PR #1, opened 2015-07-22
Initial commit of SPI interfaces

pperalta:
Includes
 - Dependencies on Receptor API and Spring XD modules
 - .gitignore updates for Eclipse and IntelliJ
 - Scaffolding for Lattice/Receptor implementations

Reviewers - **please double check my work with regard to the Maven dependencies**. I tried to place the dependency on XD modules in the top level POM but could not get it to work. Maven only fetched the dependencies when I placed them in the child-level POMs.

### DIFF
--- a/spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-lattice/src/main/java/org/springframework/cloud/data/module/deployer/lattice/ReceptorModuleDeployer.java
+++ b/spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-lattice/src/main/java/org/springframework/cloud/data/module/deployer/lattice/ReceptorModuleDeployer.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2014 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.data.module.deployer.lattice;
+
+import java.util.Map;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.springframework.cloud.data.module.ModuleStatus;
+import org.springframework.cloud.data.module.deployer.ModuleDeployer;
+import org.springframework.xd.module.ModuleDescriptor;
+
+/**
+ * @author Patrick Peralta
+ */
+public class ReceptorModuleDeployer implements ModuleDeployer {
+	private static final Logger logger = LoggerFactory.getLogger(ReceptorModuleDeployer.class);
+
+
+	@Override
+	public void deploy(ModuleDescriptor descriptor) {
+		throw new UnsupportedOperationException("todo");
+	}
+
+	@Override
+	public void undeploy(ModuleDescriptor.Key key) {
+		throw new UnsupportedOperationException("todo");
+	}
+
+	@Override
+	public ModuleStatus status(ModuleDescriptor.Key descriptor) {
+		throw new UnsupportedOperationException("todo");
+	}
+
+	@Override
+	public Map<ModuleDescriptor.Key, ModuleStatus> allStatuses() {
+		throw new UnsupportedOperationException("todo");
+	}
+}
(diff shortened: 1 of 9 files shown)


spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-spi/src/main/java/org/springframework/cloud/data/module/ModuleStatus.java
```diff
@@ -0,0 +1,196 @@
...
+		/**
+		 * All modules have been deployed.
+		 */
+		deployed,
+
+		/**
+		 * In the case of multiple modules, some have successfully deployed.
+		 * This state does not apply for individual modules.
+		 */
+		incomplete,
```
ilayaperumalg: Given this is per module status, I am trying to understand why do we need `incomplete` status here and under what scenario we would update the ModuleStatus to `incomplete` at the **ModuleDeployer** level. Based on what I see, the deployment happens per module descriptor.

ilayaperumalg: I see there is ModuleInstanceStatus and a ModuleDescriptor would have multiple modules deployments. Now I see the multiple module scenario.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-spi/src/main/java/org/springframework/cloud/data/module/deployer/ModuleDeployer.java
```diff
@@ -0,0 +1,76 @@
...
+	 * may perform this operation asynchronously; therefore
+	 * a successful deployment may not be assumed upon return.
+	 * To determine the status of a deployment, invoke
+	 * {@link #status(org.springframework.xd.module.ModuleDescriptor.Key)}.
+	 *
+	 * @param descriptor descriptor for module to be deployed
+	 *
+	 * @throws IllegalStateException if the module has already been deployed
+	 */
+	void deploy(ModuleDescriptor descriptor);
```
ilayaperumalg: Is the idea here to use `ModuleDescriptor's` children (that the XD's composed modules use now) and iterate over them for individual deployments?

markfisher: The number of individual instance deployments will be determined by the `module.count` property or whatever it's replaced with in the `spring-cloud-stream` model (where that count is likely going to be a first class concept, not "just" a deployment property). We will probably need to add that as a field on `ModuleDescriptor` itself. For example, in the Lattice/Receptor implementation, a  `ModuleDescriptor` will map to the `DesiredLRPCreateRequest` (whereas multiple ActualLRPs may be instantiated based on the provided value of `instances`). We may need to reconsider the presence of "children" on `ModuleDescriptor` depending on how composite `spring-cloud-stream` modules bubble up at this level.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-lattice/src/main/java/org/springframework/cloud/data/module/deployer/lattice/ReceptorModuleDeployer.java
```diff
@@ -0,0 +1,54 @@
...
+		throw new UnsupportedOperationException("todo");
+	}
+
+	@Override
+	public ModuleStatus status(ModuleDescriptor.Key descriptor) {
+		throw new UnsupportedOperationException("todo");
+	}
+
+	@Override
+	public Map<ModuleDescriptor.Key, ModuleStatus> allStatuses() {
```
markfisher: I wonder if 'status' would be okay as this method name as well. If you want *all* status you call the no-arg `status` method; if you want a particular module's status, you call the `status` method that accepts a `ModuleDescriptor.Key`.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-lattice/src/main/java/org/springframework/cloud/data/module/deployer/lattice/ReceptorModuleInstanceStatus.java
```diff
@@ -0,0 +1,84 @@
+/*
+ * Copyright 2014 the original author or authors.
```
markfisher: hey it's 2015 ;)

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-spi/pom.xml
```diff
@@ -6,6 +6,13 @@
 	<packaging>jar</packaging>
 	<name>spring-cloud-data-module-deployer-spi</name>
 	<description>SPI for spring-cloud-stream module deployer implementations</description>
+	<dependencies>
+		<dependency>
+			<groupId>org.springframework.xd</groupId>
+			<artifactId>spring-xd-module</artifactId>
+			<version>${spring-xd-module.version}</version>
```
markfisher: you should be able to leave out the version here (and other child poms) as long as you define the whole thing (as here, including groupId, artifactId, and version) in the parent pom

markfisher:
I will merge and make the changes for copyrights and POMs. We can discuss that `status` method afterwards.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-lattice/src/main/java/org/springframework/cloud/data/module/deployer/lattice/ReceptorModuleDeployer.java
pperalta: +1 on the `status` method rename.

PR #2, opened 2015-07-23
Introduced ModuleKey

pperalta:
Based on the conversation here: https://goo.gl/HTcLuA

Introduced a new key class to uniquely identify module descriptors. This class also contains methods to convert to/from string UID representations.

### DIFF
--- a/spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-spi/src/main/java/org/springframework/cloud/data/module/ModuleKey.java
+++ b/spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-spi/src/main/java/org/springframework/cloud/data/module/ModuleKey.java
@@ -1,0 +1,161 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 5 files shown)


spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-spi/src/main/java/org/springframework/cloud/data/module/ModuleKey.java
```diff
@@ -0,0 +1,161 @@
...
+	/**
+	 * Parse the given string and return a {@code ModuleKey} based
+	 * on the string contents.
+	 *
+	 * @param uid uid containing the fields for a {@code ModuleKey}
+	 * @return a new {@code ModuleKey} based on the provided string
+	 * @throws NullPointerException if a null uid is provided
+	 * @throws IllegalArgumentException if an invalid uid is provided
+	 */
+	public static ModuleKey fromUID(String uid) {
```
markfisher: I wonder if this might get confusing if users expect UUID format. Perhaps this method could just be called parse? The corresponding toUID() could just be this class' implementation of toString().

pperalta: I suppose using `parse` and `toString` would eliminate the use of the "uid" term. I'll add this as a comment in https://github.com/spring-cloud/spring-cloud-data/pull/4.

pperalta:
Closed in favor of https://github.com/spring-cloud/spring-cloud-data/pull/4.

PR #3, opened 2015-07-24
Added ModuleRegistry API 

dturanski:
Initial API and ModuleDeployment properties (as is) and ModuleDescriptor (needs some work).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,12 +1,25 @@
+/application.yml
+/application.properties
+asciidoctor.css
+*~
+.#*
+*#
 target/
-
 # Eclipse artifacts, including WTP generated manifests
 .classpath
 .project
 .settings/
 spring-*/src/main/java/META-INF/MANIFEST.MF
-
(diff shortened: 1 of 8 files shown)


markfisher:
I think we should have one commit that includes just the ModuleRegistry and ModuleDefinition first. The ModuleDescriptor and ModuleDeploymentProperties might belong in a separate package or even sub-project (possibly in spring-cloud-stream since ModuleLauncher also needs that info, but we might want to avoid sharing any classes), and they can probably both be simplified for the new model.

markfisher:
I squashed, polished, and merged everything except for ModuleDescriptor and ModuleDeploymentProperties: https://github.com/spring-cloud/spring-cloud-data/commit/40551d76e7bbe7d4c5a190d0f0494a55382b1744

I also pushed a branch to my fork with those 2 excluded files:
https://github.com/markfisher/spring-cloud-data/commit/dc1addb746e2de29c6a3dfd1b340b1926ec65378

PR #4, opened 2015-07-27
Initial commit of the domain model

pperalta:


spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleKey.java
```diff
@@ -0,0 +1,140 @@
...
+	/**
+	 * Return a string containing the key fields separated by
+	 * periods. This string may be used as a key in a database
+	 * to uniquely identify a module.
+	 *
+	 * @return string representation of this key
+	 */
+	public String toUID() {
+		return String.format("%s.%s", this.group, this.label);
+	}
```
pperalta: In https://github.com/spring-cloud/spring-cloud-data/pull/2 @markfisher said:

> I wonder if this might get confusing if users expect UUID format. Perhaps this method could just be called parse? The corresponding toUID() could just be this class' implementation of toString().

My response:

> I suppose using parse and toString would eliminate the use of the "uid" term.

pperalta:
Per our white boarding session today:
photo jul 27 3 31 26 pm

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleDefinition.java
```diff
@@ -0,0 +1,417 @@
...
+				? Collections.<String, String>emptyMap()
+				: Collections.unmodifiableMap(new HashMap<String, String>(parameters));
+	}
+
+	/**
+	 * Return the name of this module.
+	 *
+	 * @return module name
+	 */
+	public String getModuleName() {
```
markfisher: maybe just getName() ?

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleDefinition.java
```diff
@@ -0,0 +1,417 @@
...
+	}
+
+	/**
+	 * Return symbolic name of a module. This may be explicitly specified in
+	 * the DSL string, which is required if the stream contains another module
+	 * with the same name. Otherwise, it will be the same as the module name.
+	 *
+	 * @return module label
+	 */
+	public String getModuleLabel() {
```
markfisher: maybe just getLabel() ?

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleDefinition.java
```diff
@@ -0,0 +1,417 @@
...
+		}
+
+		/**
+		 * Create a {@code Builder} object pre-populated with the configuration
+		 * for the provided {@link ModuleDefinition}.
+		 *
+		 * @param definition module descriptor
+		 * @return pre-populated builder object
+		 */
+		public static Builder fromModuleDefinition(ModuleDefinition definition) {
```
markfisher: do we still need this?... seems that we now should only support a fully populated ModuleDefinition, whereas this method I believe was originally used for the case where depending on its context and prior usage a definition may not have contained all of the properties as needed in another context (e.g. declaration vs. deployment).

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleDeploymentRequest.java
```diff
@@ -0,0 +1,108 @@
...
+
+	/**
+	 * Number of module instances to launch.
+	 */
+	private final int count;
+
+	/**
+	 * Map of deployment properties for this module.
+	 */
+	private final Map<String, String> deploymentProperties;
```
markfisher: we may want to bring over the ModuleDeploymentProperties object (not as part of this PR, but something to discuss)

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleKey.java
markfisher: I think in this case since the class' sole purpose is to be a key, it would be ok to use toString() for that.

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleKey.java
```diff
@@ -0,0 +1,140 @@
...
+	 */
+	private final String label;
+
+	/**
+	 * Construct a {@code ModuleKey}.
+	 *
+	 * @param group name of group (stream/job) this module belongs to
+	 * @param label label to uniquely identify this module in its group
+	 */
+	public ModuleKey(String group, String label) {
```
markfisher: Do we know yet what use-cases would create a ModuleKey other than via *fromModuleDefinition()?* I'm wondering if this constructor could be private... if the fromUID() method is the only other expected way to create a key. Or alternatively, maybe we do not need the fromUID() method yet?

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleCoordinates.java
```diff
@@ -39,11 +38,11 @@
 
 	private String version;
 
-	protected ModuleDefinition() {
+	protected ModuleCoordinates() {
 		// For (subclass) JSON deserialization only
 	}
 
-	protected ModuleDefinition(String name, String artifactId, String groupId, String version) {
+	protected ModuleCoordinates(String name, String artifactId, String groupId, String version) {
```
dturanski: why protected?

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleDefinition.java
```diff
@@ -0,0 +1,417 @@
...
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.springframework.core.style.ToStringCreator;
+import org.springframework.util.Assert;
+
+/**
+ * Representation of a module in the context of a defined stream.
```
dturanski: or job

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleDefinition.java
```diff
@@ -0,0 +1,417 @@
...
+
+	/**
+	 * Parameters for module. This is specific to the type of module - for
+	 * instance an http module would include a port number as a parameter.
+	 */
+	private final Map<String, String> parameters;
+
+
+	/**
+	 * Construct a {@code ModuleDescriptor}. This constructor is private; use
```
dturanski: ModuleDefinition?

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleDefinition.java
```diff
@@ -0,0 +1,417 @@
...
+	public Map<String, String> getBindings() {
+		return this.bindings;
+	}
+
+	/**
+	 * Return parameters for module. This is specific to the type of module - for
+	 * instance an http module would include a port number as a parameter.
+	 *
+	 * @return read-only map of module parameters
+	 */
```
dturanski: Should we call these options?  That is term we generally use in the docs and ModuleOptionsMetadata, etc.

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleDefinition.java
markfisher: Actually, this is exclusive to "stream modules". We will keep the job model separate going forward.

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleDefinition.java
markfisher: We may want to consider de-emphasizing "options" as a term now that they are *just* boot properties on the `spring-cloud-stream` side. I would actually prefer "properties" here, except that it might get confusing with *deployment properties* that are combined with the `ModuleDefinition` when creating a `ModuleDeploymentRequest`.

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleKey.java
pperalta: Sounds good; I'll make these changes.

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleDefinition.java
pperalta: +1

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleDefinition.java
pperalta: +1

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleDefinition.java
pperalta: +1, removing

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleKey.java
pperalta: Yeah I think it will always either be from `ModuleDefinition` or parsing of an existing key. We can make the constructor private.

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleCoordinates.java
pperalta: Good question - I think you introduced `ModuleDefinition` into this project. ;) It does look like this should be `public`. I'll modify it.

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleDefinition.java
pperalta: +1

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleCoordinates.java
```diff
@@ -0,0 +1,158 @@
...
+ * <p>
+ * To create a new instance, either use {@link Builder} to set the individual fields:
+ * <pre>
+ * new ModuleCoordinates.Builder()
+ *     .setGroupId("org.springframework")
+ *     .setArtifactId("spring-core")
+ *     .setVersion("5.0.0")
+ *     .build()
+ * </pre>
+ * ...or use {@link #parse(String)} to parse the coordinates as a colin delimited string:
```
markfisher: typo: colin

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleCoordinates.java
```diff
@@ -0,0 +1,158 @@
...
+
+	/**
+	 * Construct a {@code ModuleCoordinates} object.
+	 *
+	 * @param groupId     group ID for artifact
+	 * @param artifactId  artifact ID
+	 * @param version     artifact version
+	 */
+	private ModuleCoordinates(String groupId, String artifactId, String version) {
+		Assert.hasLength(artifactId, "'artifactId' cannot be blank");
```
markfisher: nitpick but the order of the asserts could match the ctor args

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleCoordinates.java
```diff
@@ -0,0 +1,158 @@
...
+		Assert.hasLength(groupId, "'groupId' cannot be blank");
+		Assert.hasLength(version, "'version' cannot be blank");
+		this.artifactId = artifactId;
```
markfisher: same goes for the assignment order

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleCoordinates.java
```diff
@@ -0,0 +1,158 @@
...
+		if (this == o) {
+			return true;
+		}
+		if (!(o instanceof ModuleCoordinates)) {
+			return false;
+		}
+
+		ModuleCoordinates that = (ModuleCoordinates) o;
+
+		return this.artifactId.equals(that.artifactId) &&
```
markfisher: group, artifact, version order would be nice here too, for consistency

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleCoordinates.java
```diff
@@ -0,0 +1,158 @@
...
+				this.groupId.equals(that.groupId) &&
+				this.version.equals(that.version);
+	}
+
+	@Override
+	public int hashCode() {
+		int result = artifactId.hashCode();
```
markfisher: ditto

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleDefinition.java
dturanski: configurationProperties?

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleDefinition.java
```diff
@@ -0,0 +1,398 @@
...
+	private final String name;
+
+	/**
+	 * Symbolic name of a module. This may be generated to a default
+	 * value or specified in the DSL string.
+	 */
+	private final String moduleLabel;
+
+	/**
+	 * Name of deployable unit this module instance belongs to (such as a stream or job).
```
dturanski: do we want to remove 'job' from these comments if this class doesn't apply to jobs.

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleDefinition.java
```diff
@@ -0,0 +1,398 @@
...
+	private final Map<String, String> parameters;
+
+
+	/**
+	 * Construct a {@code ModuleDefinition}. This constructor is private; use
+	 * {@link ModuleDefinition.Builder} to create a new instance.
+	 *
+	 * @param name name of module
+	 * @param moduleLabel label used for module in stream/job definition
+	 * @param group group this module belongs to (stream/job)
```
dturanski: ditto

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleDefinition.java
```diff
@@ -0,0 +1,398 @@
...
+	 *
+	 * @return module label
+	 */
+	public String getLabel() {
+		return moduleLabel;
+	}
+
+	/**
+	 * Return name of deployable unit this module instance belongs to
+	 * (such as a stream or job).
```
dturanski: ditto

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleCoordinates.java
dturanski: I'm not crazy about the Builder pattern here since all fields are required. This introduces the possibility of a runtime exception, e.g.,  ModuleCoordinates.Builder.setGroupId("foo").build();   and less code to just use the constructor.

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleDefinition.java
```diff
@@ -0,0 +1,398 @@
...
+	 *
+	 * @return group name
+	 */
+	public String getGroup() {
+		return group;
+	}
+
+	/**
+	 * Return position in stream/job definition relative to the other
```
dturanski: one more

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleDefinition.java
markfisher: +1

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleDefinition.java
```diff
@@ -0,0 +1,398 @@
...
+	/**
+	 * Name of module.
+	 */
+	private final String name;
+
+	/**
+	 * Symbolic name of a module. This may be generated to a default
+	 * value or specified in the DSL string.
+	 */
+	private final String moduleLabel;
```
markfisher: this field can be named just "label" as well to match the new-and-improved getter

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleDefinition.java
```diff
@@ -0,0 +1,398 @@
...
+	 * {@link ModuleDefinition.Builder} to create a new instance.
+	 *
+	 * @param name name of module
+	 * @param moduleLabel label used for module in stream/job definition
+	 * @param group group this module belongs to (stream/job)
+	 * @param index position of module in stream/job definition
+	 * @param bindings map of channel bindings; may not be {@code null}
+	 * @param parameters module parameters; may be {@code null}
+	 */
+	private ModuleDefinition(String name, String moduleLabel, String group,
```
markfisher: arg name also: label

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleDefinition.java
```diff
@@ -0,0 +1,398 @@
...
+	 * @return read-only map of module parameters
+	 */
+	public Map<String, String> getParameters() {
+		return this.parameters;
+	}
+
+	@Override
+	public String toString() {
+		return new ToStringCreator(this)
+				.append("moduleName", this.name)
```
markfisher: just "name" and "label" in the toString result as well

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleDeploymentRequest.java
```diff
@@ -0,0 +1,108 @@
...
+
+package org.springframework.cloud.data.core;
+
+import java.util.Collections;
+import java.util.Map;
+
+/**
+ * Representation of a module deployment request. This includes
+ * module options defined at definition time (as part of a stream
+ * definition) as well as module deployment options (such as the
```
markfisher: module deployment *properties*

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleKey.java
```diff
@@ -0,0 +1,141 @@
...
+ * used to uniquely identify a module in a database or execution environment.
+ *
+ * @author Patrick Peralta
+ */
+public class ModuleKey implements Serializable {
+	private static final long serialVersionUID = 1L;
+
+	/**
+	 * The name of the associated group or deployment unit this
+	 * module belongs to, such as a stream or job.
```
markfisher: remove "job" here also... for all of these cases, it might be ok to just use "group" and/or "stream" (deployment unit may not be a term we need anymore if this no longer applies to jobs)

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleKey.java
markfisher: sounds good +1

spring-cloud-data-core/src/test/java/org/springframework/cloud/data/core/ModuleCoordinatesTests.java
```diff
@@ -0,0 +1,38 @@
+package org.springframework.cloud.data.core;
```
markfisher: copyright

spring-cloud-data-core/src/test/java/org/springframework/cloud/data/core/ModuleDefinitionTests.java
```diff
@@ -0,0 +1,29 @@
+package org.springframework.cloud.data.core;
```
markfisher: copyright

spring-cloud-data-core/src/test/java/org/springframework/cloud/data/core/ModuleKeyTests.java
```diff
@@ -0,0 +1,28 @@
+package org.springframework.cloud.data.core;
```
markfisher: copyright

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-spi/src/main/java/org/springframework/cloud/data/module/ModuleStatus.java
```diff
@@ -91,29 +92,27 @@
...
 	/**
-	 * Return the module label for the {@code ModuleDescriptor}.
+	 * Return the module label for the module.
 	 *
-	 * @see ModuleDescriptor.Key#getModuleLabel()
-	 *
-	 * @return module label for the {@code ModuleDescriptor}
+	 * @return module label for the module
 	 */
 	public String getName() {
```
markfisher: should this be getLabel() - or alternative should we simply provide a getter for the String representation of the full key?

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-spi/src/main/java/org/springframework/cloud/data/module/deployer/ModuleDeployer.java
```diff
@@ -32,45 +34,45 @@
 public interface ModuleDeployer {
 
 	/**
-	 * Deploy the given {@code ModuleDescriptor}. Implementations
+	 * Execute the given {@code ModuleDeploymentRequest}. Implementations
```
markfisher: maybe "Handle" instead of "Execute" here?

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-spi/src/main/java/org/springframework/cloud/data/module/deployer/ModuleDeployer.java
```diff
@@ -32,45 +34,45 @@
...
-	 * {@link #status(org.springframework.xd.module.ModuleDescriptor.Key)}.
+	 * {@link #status(ModuleKey)}.
 	 *
-	 * @param descriptor descriptor for module to be deployed
+	 * @param request request for module to be deployed
 	 *
 	 * @throws IllegalStateException if the module has already been deployed
 	 */
-	void deploy(ModuleDescriptor descriptor);
+	void deploy(ModuleDeploymentRequest request);
```
markfisher: should we return ModuleKey here as a convenience?

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleKey.java
```diff
@@ -0,0 +1,141 @@
...
+
+/**
+ * Unique identifier for a {@link ModuleDeploymentRequest}.
+ * Methods {@link #toString()} and {@link #parse(String)} can be used to convert
+ * a key to a string and a string to a key, respectively. The id string may be
+ * used to uniquely identify a module in a database or execution environment.
+ *
+ * @author Patrick Peralta
+ */
+public class ModuleKey implements Serializable {
```
markfisher: Considering the role this plays, I'm wondering if we might want to rename this: `ModuleDeploymentId` ?

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleDefinition.java
pperalta: +1

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleDefinition.java
```diff
@@ -0,0 +1,398 @@
...
+		 * @see ModuleDefinition#bindings
+		 */
+		public Builder addBindings(Map<String, String> bindings) {
+			this.bindings.putAll(bindings);
+			return this;
+		}
+
+		/**
+		 * Return name of module. Typically this module is present under
+		 * {@code $XD_HOME/modules/[module type]}.
```
markfisher: we should remove this XD path info

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleDefinition.java
```diff
@@ -0,0 +1,398 @@
...
+	public int compareTo(ModuleDefinition o) {
+		Assert.notNull(o, "ModuleDefinition must not be null");
+		return (index < o.index) ? -1 : ((index == o.index) ? 0 : 1);
+	}
+
+
+	/**
+	 * Builder object for {@code ModuleDefinition}.
+	 * This object is mutable to allow for flexibility in specifying module
+	 * type/fields/parameters during parsing.
```
markfisher: no type

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-spi/src/main/java/org/springframework/cloud/data/module/ModuleStatus.java
pperalta: How about a getter for the full key itself and let the caller decide what they need from it?

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-spi/src/main/java/org/springframework/cloud/data/module/deployer/ModuleDeployer.java
pperalta: +1

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-spi/src/main/java/org/springframework/cloud/data/module/deployer/ModuleDeployer.java
pperalta: Not sure that it's really necessary. I like returning `void` here because it helps to reenforce the notion that this is an asynchronous invocation.

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleKey.java
pperalta: +1

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-spi/src/main/java/org/springframework/cloud/data/module/deployer/ModuleDeployer.java
markfisher: the JavaDoc could state something like: "returns a ModuleDeploymentId that can be used to check on the status since the deployment is triggered asynchronously"

I would view it like a POST operation returning a URL in the location header

pperalta: On second thought, it would be useful to have that if the caller wants to invoke `status` after issuing a deployment.

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleDefinition.java
```diff
@@ -0,0 +1,397 @@
...
+	}
+
+	/**
+	 * Return map of channel bindings for this module (such as input
+	 * and/or output channels).
+	 *
+	 * @return map of channel bindings for module
+	 */
+	public Map<String, String> getBindings() {
+		return this.bindings;
```
markfisher: should this return an unmodifiable map? in any case, don't need "this" here - inconsistent with other getters in this class ;)

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleDefinition.java
```diff
@@ -0,0 +1,397 @@
...
+	}
+
+	/**
+	 * Return parameters for module. This is specific to the type of module - for
+	 * instance an http module would include a port number as a parameter.
+	 *
+	 * @return read-only map of module parameters
+	 */
+	public Map<String, String> getParameters() {
+		return this.parameters;
```
markfisher: same comment as for getBindings() (on both accounts)

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleDefinition.java
```diff
@@ -0,0 +1,397 @@
...
+	}
+
+	@Override
+	public String toString() {
+		return new ToStringCreator(this)
+				.append("name", this.name)
```
markfisher: inconsistent use of 'this' here

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleDefinition.java
pperalta: The map already is unmodifiable - take a look in the constructor.

Regarding `this` - would it be better if I *add* it to where it is missing? We're supposed to according to the coding standard:

https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-Code-Style

> A field of a class should always be referenced using `this`.

markfisher: I generally feel like in the case of a getter it's unnecessary noise (a setter is different though, since the param name usually matches).

markfisher:
squashed and merged as https://github.com/spring-cloud/spring-cloud-data/commit/fc6299073e694370dd598cc9788a2aa1d7352eef

PR #5, opened 2015-07-28
Initial stream definition domain object

pperalta:


spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/StreamDefinition.java
```diff
@@ -16,11 +16,213 @@
...
+		}
+	}
+
+
+	/**
+	 * Iterator that prevents mutation of its backing data structure.
+	 *
+	 * @param <T> the type of elements returned by this iterator
+	 */
+	private static class ReadOnlyIterator<T> implements Iterator<T> {
```
pperalta: Is this way overkill? I really want to lock these objects down and prevent any kind of mutation. Too bad the JDK doesn't have a read only `Iterator` wrapper. I could not find one in the core Spring framework either.

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/StreamDefinition.java
```diff
@@ -16,11 +16,213 @@
...
+
+	/**
+	 * Construct a {@code StreamDefinition}.
+	 *
+	 * @param name       name of stream
+	 * @param definition DSL definition for stream
+	 * @param modules    ordered list of modules for stream
+	 */
+	private StreamDefinition(String name, String definition, List<ModuleDefinition> modules) {
+		this.name = name;
```
pperalta: Constructor is missing validation; will add.

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/StreamDefinition.java
```diff
@@ -16,11 +16,217 @@
...
+	 */
+	public Iterator<ModuleDefinition> getDeploymentOrderIterator() {
+		return new ReadOnlyIterator<>(modules.descendingIterator());
+	}
+
+	@Override
+	public String toString() {
+		return new ToStringCreator(this)
+				.append("name", this.name)
+				.append(this.definition)
```
markfisher: is this intentional that the definition has no key? (I assume so since the class IS *Definition).

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/StreamDefinition.java
```diff
@@ -16,11 +16,217 @@
...
-	private String name;
+	/**
+	 * Name of stream.
+	 */
+	private final String name;
 
+	/**
+	 * DSL definition for stream.
+	 */
+	private final String definition;
```
markfisher: might be a good time to change this field (and the accessor methods) to something like dslText. I'm not crazy about that... could be something better, but I would be happy to never again see: `definition.getDefinition()` :)

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/StreamDefinition.java
markfisher: Spring's `CollectionUtils

pperalta: Should we consider adding this to `CollectionUtils`? Guava also has this concept: 

http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/collect/Iterators.html#unmodifiableIterator(java.util.Iterator)

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/StreamDefinition.java
pperalta: We use the stream name as its key, right? Or did you mean something else?

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/StreamDefinition.java
pperalta: I'll think about this one...not crazy about `dslText` either...

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/StreamDefinition.java
markfisher: This is the current format:

`[StreamDefinition@3f9261de name = 'ticktock''time | log']`

markfisher:
Here's a crazy idea: could we remove the 'definition' field altogether and have the toString return the DSL representation which it generates from the list of ModuleDefinitions? Working with this object a bit, the amount of duplication (and thus chances for invalid, conflicting data) worries me. In other words, it's possible to call the constructor passing a "definition" that does not match the list of ModuleDefinitions. I feel like this object could be simplified to the point of just having the ModuleDefinition list and a name. The parser generates that object from the DSL text.

markfisher:
Alternative to my last idea: pass the DSL text to the constructor and NOT the List of ModuleDefinitions. Instead the StreamDefinition class itself ends up being the parser. As discussed, we want the parsing at this level to be as simple as possible anyways, with no additional dependencies, and only applying syntactic validation.

markfisher:
To clarify: the StreamParser could still be a separate class since it might have other uses (e.g. supporting the REST endpoint used for auto-completion in the shell), but if we can manage to have a dependency-free and stateless parser, then StreamDefinition itself could have a static StreamParser instance.

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/StreamDefinition.java
pperalta: Oh I see - yeah it would look better with a `definition = '...'` format; I'll add that.

pperalta:
Based on the comments regarding the parser above, I'll close this PR and work in the parser along with the domain model in a future PR.

PR #6, opened 2015-07-29
Add extension and classifier to ModuleCoordinates

dturanski:


spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleCoordinates.java
```diff
@@ -55,23 +66,39 @@
...
+	 * @param version artifact version
 	 */
-	private ModuleCoordinates(String groupId, String artifactId, String version) {
-		Assert.hasLength(groupId, "'groupId' cannot be blank");
-		Assert.hasLength(artifactId, "'artifactId' cannot be blank");
-		Assert.hasLength(version, "'version' cannot be blank");
+	private ModuleCoordinates(String groupId, String artifactId, String extension, String classifier, String version) {
+		Assert.hasText(groupId, "'groupId' cannot be blank");
+		Assert.hasText(artifactId, "'artifactId' cannot be blank");
+		Assert.hasText(version, "'version' cannot be blank");
```
markfisher: I would flip extension and version here so they match the order of args.

markfisher:
LGTM; I can address that one comment while merging.

PR #11, opened 2015-08-02
initial commit for spring-cloud-data-rest

markfisher:
The purpose of this PR is to provide the initial foundation for the spring-cloud-data REST API, including the way the StreamController will interact with the StreamDefinitionRepository, ModuleRegistry, and ModuleDeployer.

This includes (temporary) stub implementations of the ModuleRegistry and StreamDefinitionRepository. The ModuleRegistry interface was updated since we need to pass type information in addition to the name of a module. That type will be necessary for mapping to the proper ModuleCoordinates. It will also allow us to continue using names such as "file" in both the source and sink position of a stream definition.

The StreamController functions for an end-to-end PoC, but it will be replaced with a version that @pperalta has been working on, and at that point it will be in-sync with the upper layers of existing XD (the ResourceAssemblers, etc) - so that it will ultimately support the Shell and UI (at least as far as stream operations are concerned).

To run the "AdminApplication", build from the spring-cloud-data root directory via `mvn clean package`, and then execute: `java -jar spring-cloud-data-rest/target/spring-cloud-data-rest-1.0.0.BUILD-SNAPSHOT.jar`.

At this point, only the `time` source and `log` sink are available.

To create the 'ticktock' stream:

```
$ curl -X POST -H "Content-Type: text/plain" -d "time | log" http://localhost:9393/streams/ticktock
```

To list all streams:

```
$ curl http://localhost:9393/streams
```

To deploy the 'ticktock' stream:

```
$ curl -X PUT http://localhost:9393/streams/ticktock
```

If successful you should see output similar to the following in the AdminApplication console:

```
2015-08-01 23:59:58.244  INFO 32845 --- [hannel-adapter1] sink.LogSink     : Received: 2015-08-01 23:59:58
2015-08-02 00:00:03.245  INFO 32845 --- [hannel-adapter1] sink.LogSink     : Received: 2015-08-02 00:00:03
2015-08-02 00:00:08.247  INFO 32845 --- [hannel-adapter1] sink.LogSink     : Received: 2015-08-02 00:00:08
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -25,6 +25,7 @@
 		<spring-xd.version>1.0.2.RELEASE</spring-xd.version>
 	</properties>
 	<modules>
+		<module>spring-cloud-data-rest</module>
 		<module>spring-cloud-data-module-registry</module>
 		<module>spring-cloud-data-module-deployers</module>
 		<module>spring-cloud-data-core</module>
--- a/spring-cloud-data-module-registry/src/main/java/org/springframework/cloud/data/module/registry/StubModuleRegistry.java
+++ b/spring-cloud-data-module-registry/src/main/java/org/springframework/cloud/data/module/registry/StubModuleRegistry.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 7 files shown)


dturanski:
LGTM

PR #13, opened 2015-08-03
initial commit for profile-driven ModuleDeployer

markfisher:
Currently this will create a ReceptorModuleDeployer bean if the "cloud" profile is active,
a LocalModuleDeployer, otherwise. We will later enhance this to be more fine-grained,
including support for a CloudFoundry ModuleDeployer implementation.

### DIFF
--- a/spring-cloud-data-rest/src/main/java/org/springframework/cloud/data/rest/AdminApplication.java
+++ b/spring-cloud-data-rest/src/main/java/org/springframework/cloud/data/rest/AdminApplication.java
@@ -18,13 +18,14 @@ package org.springframework.cloud.data.rest;
 
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
-import org.springframework.context.annotation.ComponentScan;
+import org.springframework.cloud.data.rest.config.AdminConfiguration;
+import org.springframework.context.annotation.Import;
 
 /**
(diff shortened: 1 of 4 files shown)


spring-cloud-data-rest/src/main/java/org/springframework/cloud/data/rest/AdminApplication.java
```diff
@@ -18,13 +18,14 @@
...
 import org.springframework.boot.autoconfigure.SpringBootApplication;
-import org.springframework.context.annotation.ComponentScan;
+import org.springframework.cloud.data.rest.config.AdminConfiguration;
+import org.springframework.context.annotation.Import;
 
 /**
  * @author Mark Fisher
  */
 @SpringBootApplication
-@ComponentScan(basePackages = "org.springframework.cloud.data.rest.controller")
```
ilayaperumalg: Can the `ComponentScan` help scanning more controllers in future? That way, we don't have to instantiate the `controller` beans inside AdminConfiguration. Instead we can `autowire` the module deployer into the controllers with component scan enabled in this class.

markfisher: I think for now we should keep it as simple as possible. We can add componentscan back on the AdminConfiguration class if/when needed.

PR #18, opened 2015-08-04
Polishing

pperalta:
* Added JavaDoc
* Renamed `StubStreamDefinitionRepository` to `InMemoryStreamDefinitionRepository`
* Updated parameter names in `InMemoryStreamDefinitionRepository`

### DIFF
--- a/spring-cloud-data-rest/src/main/java/org/springframework/cloud/data/rest/controller/StreamController.java
+++ b/spring-cloud-data-rest/src/main/java/org/springframework/cloud/data/rest/controller/StreamController.java
@@ -16,17 +16,11 @@
 
 package org.springframework.cloud.data.rest.controller;
 
-
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.cloud.data.core.StreamDefinition;
 import org.springframework.cloud.data.module.deployer.ModuleDeployer;
 import org.springframework.cloud.data.module.registry.ModuleRegistry;
-import org.springframework.cloud.data.module.registry.StubModuleRegistry;
 import org.springframework.cloud.data.repository.StreamDefinitionRepository;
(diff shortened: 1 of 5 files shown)


spring-cloud-data-rest/src/main/java/org/springframework/cloud/data/rest/resource/StreamDefinitionResource.java
```diff
@@ -24,29 +24,37 @@
...
- * todo: document that this is not thread safe
+ * A HATEOAS representation of a {@link StreamDefinition}.
+ * This class also includes a description of the
+ * {@link #status stream status}.
+ * <p>
+ * Note: this implementation is not thread safe.
  *
  * @author Patrick Peralta
  */
 @XmlRootElement
```
ilayaperumalg: Since we assume the controller always produces `application\json`, we may not need this annotation anymore.

spring-cloud-data-rest/src/main/java/org/springframework/cloud/data/rest/controller/StreamController.java
```diff
@@ -41,33 +35,66 @@
...
+	 * @param registry    module registry this controller will use to look up modules
+	 * @param deployer    the deployer this controller will use to deploy stream modules
 	 */
 	@Autowired
-	public StreamController(ModuleDeployer moduleDeployer) {
-		Assert.notNull(moduleDeployer, "moduleDeployer must not be null");
-		this.deployer = moduleDeployer;
+	public StreamController(StreamDefinitionRepository repository, ModuleRegistry registry,
+			ModuleDeployer deployer) {
+		Assert.notNull(repository, "repository must not be null");
```
markfisher: null check for the registry?

markfisher:
merged and addressed comments with this commit: https://github.com/spring-cloud/spring-cloud-data/commit/6c9a286a296a335c98b020ea99e9da1aea849725

PR #20, opened 2015-08-05
Simple deploy/undeploy/status via Lattice

pperalta:
Simple deploy/undeploy and status functionality via Lattice.

Sample output:
```
xd:>stream create tt --definition "time|log"
Created new stream 'tt'
xd:>stream list
  Stream Name  Stream Definition  Status
  -----------  -----------------  -------
  tt           time|log           unknown

xd:>stream deploy tt
Deployed stream 'tt'
xd:>stream list
  Stream Name  Stream Definition  Status
  -----------  -----------------  ---------
  tt           time|log           deploying

xd:>stream list
  Stream Name  Stream Definition  Status
  -----------  -----------------  --------
  tt           time|log           deployed

xd:>stream undeploy tt
Un-deployed stream 'tt'
xd:>stream list
  Stream Name  Stream Definition  Status
  -----------  -----------------  -------
  tt           time|log           unknown
```

Known issues:
* Status of `unknown` instead of `undeployed`
* Need pluggable guid generation for SPI implementations - and a better name than "guid"
* Consideration of a pluggable stream state calculator similar to Spring XD 1.x

### DIFF
--- a/spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-lattice/src/main/java/org/springframework/cloud/data/module/deployer/lattice/ReceptorModuleDeployer.java
+++ b/spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-lattice/src/main/java/org/springframework/cloud/data/module/deployer/lattice/ReceptorModuleDeployer.java
@@ -16,30 +16,105 @@
 
 package org.springframework.cloud.data.module.deployer.lattice;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
+
+import org.cloudfoundry.receptor.client.ReceptorClient;
+import org.cloudfoundry.receptor.commands.ActualLRPResponse;
+import org.cloudfoundry.receptor.commands.DesiredLRPCreateRequest;
+import org.cloudfoundry.receptor.support.EnvironmentVariable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import org.springframework.cloud.data.core.ModuleDeploymentId;
 import org.springframework.cloud.data.core.ModuleDeploymentRequest;
 import org.springframework.cloud.data.module.ModuleStatus;
 import org.springframework.cloud.data.module.deployer.ModuleDeployer;
+import org.springframework.util.StringUtils;
 
 /**
  * @author Patrick Peralta
  */
 public class ReceptorModuleDeployer implements ModuleDeployer {
+	private static final Logger logger = LoggerFactory.getLogger(ReceptorModuleDeployer.class);
+
+	public static final String DOCKER_PATH = "docker:///springcloud/stream-module-launcher";
+
+	public static final String BASE_ADDRESS = "192.168.11.11.xip.io";
+
+	private final ReceptorClient receptorClient = new ReceptorClient();
+
 	@Override
 	public ModuleDeploymentId deploy(ModuleDeploymentRequest request) {
-		throw new UnsupportedOperationException();
+		ModuleDeploymentId id = ModuleDeploymentId.fromModuleDefinition(request.getDefinition());
+		String guid = guid(id);
+
+		DesiredLRPCreateRequest lrp = new DesiredLRPCreateRequest();
+		lrp.setProcessGuid(guid);
+		lrp.setRootfs(DOCKER_PATH);
+		lrp.runAction().setPath("java");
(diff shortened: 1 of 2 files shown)


spring-cloud-data-rest/src/main/java/org/springframework/cloud/data/rest/controller/StreamController.java
```diff
@@ -211,6 +221,38 @@ private void undeployStream(StreamDefinition stream) {
...
+		if (states.contains(ModuleStatus.State.failed)) {
+			return ModuleStatus.State.failed.toString();
+		}
+		else if (states.contains(ModuleStatus.State.incomplete)) {
+			return ModuleStatus.State.incomplete.toString();
+		}
+		else if (states.contains(ModuleStatus.State.deploying)) {
+			return ModuleStatus.State.deploying.toString();
+		}
+		else if (states.contains(ModuleStatus.State.deployed)) {
```
ilayaperumalg: If one of the states is `unknown` but the one other has `deployed` then we would return `deployed` here. May be we can make sure all the states are `deployed` when returning `deployed`?

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-lattice/src/main/java/org/springframework/cloud/data/module/deployer/lattice/ReceptorModuleDeployer.java
```diff
@@ -16,30 +16,105 @@
...
+		lrp.runAction().addArg("/module-launcher.jar");
+
+		List<EnvironmentVariable> environmentVariables = new ArrayList<EnvironmentVariable>();
+		Collections.addAll(environmentVariables, lrp.getEnv());
+		environmentVariables.add(new EnvironmentVariable("MODULES", request.getCoordinates().toString()));
+		environmentVariables.add(new EnvironmentVariable("SPRING_PROFILES_ACTIVE", "cloud"));
+
+		lrp.setEnv(environmentVariables.toArray(new EnvironmentVariable[environmentVariables.size()]));
+
+		lrp.setPorts(new int[] {8080, 9000});
```
ilayaperumalg: Is the significance of creating multiple ports for the `LRP` to accommodate multiple modules inside the same container? If so, can we have the `number` of ports equals to the number of modules being deployed inside the diego container?

spring-cloud-data-rest/src/main/java/org/springframework/cloud/data/rest/controller/StreamController.java
```diff
@@ -211,6 +221,38 @@ private void undeployStream(StreamDefinition stream) {
 		}
 	}
 
+	private String calculateStreamState(String name) {
```
markfisher: Could this work here? (for lines 225-234):

```
Set<ModuleStatus.State> states = new HashSet<>();
StreamDefinition stream = repository.findOne(name);
for (ModuleDefinition module : stream.getModuleDefinitions()) {
	ModuleStatus status = deployer.status(ModuleDeploymentId.fromModuleDefinition(module)));
	states.add(status.getState());
}
```

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-lattice/src/main/java/org/springframework/cloud/data/module/deployer/lattice/ReceptorModuleDeployer.java
markfisher: we will need to change the way we handle port mappings for sure - this was taken from the xolpoc-admin originally, and there the idea was to support 8080 (the app itself) as well as 9000 because it would be used by an "http" source module by default. Obviously we need to ensure that the ports that are mapped are the ones being used

spring-cloud-data-rest/src/main/java/org/springframework/cloud/data/rest/controller/StreamController.java
markfisher: makes sense we can just add a check for states.size == 1 in this if condition

markfisher:
LGTM; will address the comments above (except for port mappings which will require a new story) and merge

markfisher:
merged as https://github.com/spring-cloud/spring-cloud-data/commit/e729ad8e40e83325b661bc53c75c2e39ad8f60b9

also added this commit: https://github.com/spring-cloud/spring-cloud-data/commit/846345d7c829b47178b22a73366ffe64f1d512c5

PR #21, opened 2015-08-06
profile-driven ModuleDeployer beans

markfisher:
if the "cloud" and "lattice" profiles are both active, the LatticeModuleDeployer bean is instantiated

if the "cloud" profile is active but "lattice" is not, the CloudFoundryModuleDeployer bean is instantiated

if the "cloud" profile is not active, the LocalModuleDeployer bean is instantiated

also added a stub implementation of CloudFoundryModuleDeployer


ilayaperumalg:
LGTM, merging.

PR #23, opened 2015-08-06
initial local module status

markfisher:
simply returns "deployed" if it has deployed a module, but "unknown" otherwise

will evolve once we decide how we'll interact with running modules (and once we have `undeploy()` implemented)


ilayaperumalg:
LGTM, merging.

PR #25, opened 2015-08-07
switch IllegalState and IllegalArg for Assertions

markfisher:
swapped these two exception types based on the following rationale:
- IllegalArgumentException means that the input is *invalid* and thus we can't even continue with the assertion
- IllegalStateException means the assertion itself failed (the data does not conform to the required state)

also added comprehensive unit tests


ilayaperumalg:
LGTM, merging.

PR #27, opened 2015-08-08
Add admin config file

ilayaperumalg:
- Set default properties in `application.yml`
   - This will also enable injecting overriding properties via `args`


markfisher:
LGTM; merged

PR #28, opened 2015-08-10
XD-3354 Setup Shell integration tests

ilayaperumalg:
- Setup Shell integration tests infrastructure
   - Moved the setup from XD to here with minimal things needed to
get started

### DIFF
--- a/spring-cloud-data-shell/src/test/java/org/springframework/cloud/data/shell/AbstractShellIntegrationTest.java
+++ b/spring-cloud-data-shell/src/test/java/org/springframework/cloud/data/shell/AbstractShellIntegrationTest.java
@@ -28,8 +28,10 @@ import org.slf4j.LoggerFactory;
 import org.springframework.cloud.data.rest.AdminApplication;
 import org.springframework.cloud.data.shell.command.StreamCommandTemplate;
 import org.springframework.shell.Bootstrap;
+import org.springframework.shell.core.CommandResult;
 import org.springframework.shell.core.JLineShellComponent;
 import org.springframework.util.AlternativeJdkIdGenerator;
+import org.springframework.util.Assert;
 import org.springframework.util.IdGenerator;
 import org.springframework.util.SocketUtils;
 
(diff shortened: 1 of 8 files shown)


spring-cloud-data-shell/src/test/java/org/springframework/cloud/data/shell/command/StreamCommandTests.java
```diff
@@ -0,0 +1,23 @@
+package org.springframework.cloud.data.shell.command;
```
markfisher: copyright header

markfisher:
we should take this opportunity to favor composition over inheritance, specifically I believe the AbstractCommandTemplate base class is not worth maintaining since it only provides one method and getShell().executeCommand() is already available.

spring-cloud-data-shell/src/test/java/org/springframework/cloud/data/shell/AbstractShellIntegrationTest.java
```diff
@@ -0,0 +1,138 @@
...
+
+	@BeforeClass
+	public static synchronized void startUp() throws InterruptedException, IOException {
+		if (application == null) {
+			application = new AdminApplication();
+			application.main(new String[] {
+					String.format("--server.port=%s", adminPort),
+					"--spring.main.show_banner=false"});
+		}
+		shell = new Bootstrap(new String[] {"--port", String.valueOf(adminPort)}).getJLineShellComponent();
```
markfisher: It's worth adding a delegating stop() in the CloudDataShell, thus no need to have both the shell and cloudDataShell fields? (could just have the CloudDataShell assigned to a 'shell' field instead)

ilayaperumalg: Yeah, good idea. trying it.

markfisher:
rebased, squashed, and merged as https://github.com/spring-cloud/spring-cloud-data/commit/41b40079c85bed5881de31f1f67cdba3cf0c802d

PR #30, opened 2015-08-10
XD-3385 Added CloudConfiguration via @Import for "cloud" profiles

trisberg:
- this allows us to handle "lattice" and "!lattice" profiles only for "cloud"

### DIFF
--- a/spring-cloud-data-rest/src/main/java/org/springframework/cloud/data/rest/config/AdminConfiguration.java
+++ b/spring-cloud-data-rest/src/main/java/org/springframework/cloud/data/rest/config/AdminConfiguration.java
@@ -85,13 +85,6 @@ public class AdminConfiguration {
 		}
 	}
 
-//	@Configuration
-//	@Profile("cloud")
-//	@Import(CloudConfiguration.class)
-//	protected static class CloudConfig {
-//		//CloudConfiguration class is only used for {"cloud"} and {"cloud", "lattice"} profiles
-//	}
-
 	@Bean
 	public WebMvcConfigurer configurer() {
 		return new WebMvcConfigurerAdapter() {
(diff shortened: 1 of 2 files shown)


spring-cloud-data-rest/src/main/java/org/springframework/cloud/data/rest/config/AdminConfiguration.java
```diff
@@ -87,27 +86,9 @@ public ModuleDeployer moduleDeployer(ModuleLauncher moduleLauncher) {
 
 	@Configuration
 	@Profile("cloud")
+	@Import(CloudConfiguration.class)
```
markfisher: would this work if the `@Import` were at class level in AdminApplication, and then `@Profile` were at class level for CloudConfiguration?

trisberg: looks like it would, I'll re-work this

trisberg:
replaced the earlier commit with a new one - 8ee05ff

spring-cloud-data-rest/src/main/java/org/springframework/cloud/data/rest/config/CloudConfiguration.java
```diff
@@ -0,0 +1,51 @@
...
+import org.springframework.context.annotation.Profile;
+
+
+/**
+ * Configuration for cloud profiles (cloud, lattice)
+ *
+ * @author Thomas Risberg
+ */
+@Profile("cloud")
+public class CloudConfiguration {
```
jvalkeal: So either one of these beans, `ReceptorModuleDeployer` or `CloudFoundryModuleDeployer` is always created, right? Then we will get into same dance with expecting other configs to work in a correct order to override these? I assume this is how `LocalModuleDeployer` would then get created. I don't really see no reliable way getting my yarn deployer configured via this profile structure.

jvalkeal: Ah there's `@Profile("cloud")` above but still this feels brittle.

markfisher: I think it's better (at least more reliable) than what we had before, so I will merge it. We can still consider other options.

markfisher:
squashed and merged as https://github.com/spring-cloud/spring-cloud-data/commit/8ef07a1a55edd0c42235432f06b18cd0d6de55c5

PR #31, opened 2015-08-10
Refactor Stream/Task parsing

pperalta:
This is part of https://jira.spring.io/browse/XD-3377. Introduced `Tokens` data structure for holding tokens and the current token position used by parsers.

`TaskDslParser` and `StreamDslParser` were minimally modified; there is potential for further refactoring.

### DIFF
--- a/spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/dsl/Tokens.java
+++ b/spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/dsl/Tokens.java
@@ -1,0 +1,244 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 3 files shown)


spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/dsl/Tokens.java
```diff
@@ -0,0 +1,244 @@
...
+		Token t = tokenStream.get(position + distance);
+		return t.kind == desiredTokenKind;
+	}
+
+	/**
+	 * Return {@code true} if there are more tokens to process.
+	 *
+	 * @return {@code true} if there are more tokens to process
+	 */
+	protected boolean more() {
```
markfisher: maybe change to `hasNext()`? (or at least `hasMore()` but the latter is more consistent with an Iterator)

pperalta: Pushed

markfisher:
This looks good to me. I'd like to get @aclement to review the general approach. Andy: the context is that we've migrated the syntactic parsing from XD so that it has no dependencies and can be used directly in the domain model. Also, we are drawing a clearer line between Stream and Task (i.e. Job), but we want to have reuse at the level of the parser (e.g. for an individual "module" and its key value pairs). The idea of introducing a `Tokens` datatype is to favor composition over inheritance for that lower level code.

aclement:
Looks good to me. And glad to see there is no semantic checking in there, that was a mistake with the original form of the parser, it was too aware of information beyond the syntax. So if the two parsers do share a common definition of a 'module' the next step is to refactor `eatModule()` into something re-usable across both that operates on a `Tokens` object?

(Incidentally if you ever need it Spring Flo for XD uses a JavaScript version of the parser...)

markfisher:
rebased, squashed, and merged as https://github.com/spring-cloud/spring-cloud-data/commit/4d9ae06086e39187039ff157a37010ff34fc72ad

PR #32, opened 2015-08-11
Spring Data Docker deploy matches SCS

cppwfs:
* added -exec extension to spring cloud data jar for consistency with SCS
* updated the Dockerfile so that it will search for -exec.jar instead of the hardcoded snapshot jar
* moved Dockerfile into the code.
* updated data-admin pom to include docker plugin

### DIFF
--- a/spring-cloud-data-rest/pom.xml
+++ b/spring-cloud-data-rest/pom.xml
@@ -17,6 +17,8 @@
 	<properties>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<start-class>org.springframework.cloud.data.rest.AdminApplication</start-class>
+		<docker.image.prefix>springcloud</docker.image.prefix>
+		<docker.image.name>data-admin</docker.image.name>
 	</properties>
 	<dependencies>
 		<dependency>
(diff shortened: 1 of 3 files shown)


spring-cloud-data-rest/pom.xml
```diff
@@ -80,6 +82,25 @@
 			<plugin>
 				<groupId>org.springframework.boot</groupId>
 				<artifactId>spring-boot-maven-plugin</artifactId>
+				<configuration>
+					<classifier>exec</classifier>
```
markfisher: not so sure we need -exec here - @dsyer could you weigh in?

markfisher: let's remove -exec for now

dsyer: It was there for the module samples originally because I wondered if it might be useful to have the module as a thin jar for composition. I don't know what relevance that has here.

markfisher: Agreed we will need the thin JAR in the module case. Here (and probably the spring-cloud-stream-module-launcher as well) it would be better to have *the* JAR be the bootified one. It does make this trickier to avoid hardcoding a version in here (while the wildcard-match on *-exec made that possible), but I don't think that's enough of a reason to force a qualifier on users who simply want to launch the AdminApplication as a boot app.

markfisher:
rebased and merged as https://github.com/spring-cloud/spring-cloud-data/commit/84b275ea15a7f40e0f049e985d090743b2bce744

PR #36, opened 2015-08-12
XD-3384 Add base features for yarn deployer

jvalkeal:
- New spring-cloud-data-yarn app which has client, appmaster and container.
- New YarnModuleDeployer working with existing yarn app
- Changes to rest app to enable yarn deployer if `yarn` profile is activated.

Some notes about this PR:

- Some instructions in main README
- This is very basic POC to demonstrate that we can actually deploy something into yarn.
- I'm adding tests in next PR's
- Yarn app based on SHDP yarn container groups/clusters. Every group is sharing same settings and thus controls instances of a module. This group can to started, stopped and ramped up/down.
- I'm crafting some SHDP jira's order to make deployer code faster and more easier to use.
- No properties passed into modules.
- Should somehow get notification from ModuleLauncher when module errors/exist so that we can exit yarn container.
- Should think how base yarn control app(yarn app instance and its appmaster) can be deployed into yarn with streams.
- We can potentially have multiple yarn apps and from shell we should somehow control which one to talk to.

### DIFF
--- a/spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-yarn/src/main/java/org/springframework/cloud/data/module/deployer/yarn/YarnModuleDeployer.java
+++ b/spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-yarn/src/main/java/org/springframework/cloud/data/module/deployer/yarn/YarnModuleDeployer.java
@@ -1,0 +1,213 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.data.module.deployer.yarn;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Properties;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.cloud.data.core.ModuleCoordinates;
+import org.springframework.cloud.data.core.ModuleDefinition;
+import org.springframework.cloud.data.core.ModuleDeploymentId;
+import org.springframework.cloud.data.core.ModuleDeploymentRequest;
+import org.springframework.cloud.data.module.ModuleStatus;
+import org.springframework.cloud.data.module.deployer.ModuleDeployer;
+import org.springframework.util.StringUtils;
+import org.springframework.yarn.boot.app.YarnContainerClusterApplication;
+import org.springframework.yarn.boot.app.YarnInfoApplication;
+
+/**
+ * {@link ModuleDeployer} which communicates to XD's yarn app running
+ * on a hadoop cluster waiting for deployment requests. This app is
+ * utilising spring yarn's container grouping functionality to create a
+ * new group per module type which then allows all modules to share same
+ * settings and group itself can controlled, i.e. ramp up/down or shutdown/destroy
+ * whole group.
+ *
+ * @author Janne Valkealahti
+ */
+public class YarnModuleDeployer implements ModuleDeployer {
+
+	private static final Logger logger = LoggerFactory.getLogger(YarnModuleDeployer.class);
+	private static final String PREFIX = "spring.yarn.internal.ContainerClusterApplication.";
+
+	public YarnModuleDeployer() {
+	}
+
+	@Override
+	public ModuleDeploymentId deploy(ModuleDeploymentRequest request) {
+		int count = request.getCount();
+		ModuleCoordinates coordinates = request.getCoordinates();
+		ModuleDefinition definition = request.getDefinition();
+		logger.info("deploying request for definition: " + definition);
+
+		ModuleDeploymentId moduleDeploymentId = new ModuleDeploymentId(definition.getGroup(), definition.getLabel());
+		String clusterId = sanitizeClusterId(convertFromModuleDeploymentId(moduleDeploymentId));
+
+		String yarnApplicationId = findRunningXdYarnApp();
+		logger.info("Using application id " + yarnApplicationId);
+		String module = coordinates.toString();
+		logger.info("deploying module: " + module);
+
+		Map<String, String> definitionParameters = definition.getParameters();
+		Map<String, String> deploymentProperties = request.getDeploymentProperties();
+		logger.info("definitionParameters: " + definitionParameters);
+		logger.info("deploymentProperties: " + deploymentProperties);
+
+		// Using same app instance yarn boot cli is using to
+		// communicate with an app running on yarn via its boot actuator
+		YarnContainerClusterApplication app = new YarnContainerClusterApplication();
+		Properties appProperties = new Properties();
+		appProperties.setProperty(PREFIX + "operation", "CLUSTERCREATE");
+		appProperties.setProperty(PREFIX + "applicationId", yarnApplicationId);
+		appProperties.setProperty(PREFIX + "clusterId", clusterId);
+		appProperties.setProperty(PREFIX + "clusterDef", "module-template");
+		appProperties.setProperty(PREFIX + "projectionType", "default");
+		appProperties.setProperty(PREFIX + "projectionData.any", Integer.toString(count));
+		appProperties.setProperty(PREFIX + "extraProperties.containerModules", module);
+		app.appProperties(appProperties);
+		String output = app.run();
+		logger.info("Output from YarnContainerClusterApplication run for CLUSTERCREATE: " + output);
+
+		app = new YarnContainerClusterApplication();
+		appProperties = new Properties();
+		appProperties.setProperty(PREFIX + "operation", "CLUSTERSTART");
+		appProperties.setProperty(PREFIX + "applicationId", yarnApplicationId);
+		appProperties.setProperty(PREFIX + "clusterId", clusterId);
+		app.appProperties(appProperties);
+		output = app.run();
+		logger.info("Output from YarnContainerClusterApplication run for CLUSTERSTART: " + output);
+
+		return moduleDeploymentId;
+	}
+
+	@Override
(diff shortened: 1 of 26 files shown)


spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-yarn/src/main/java/org/springframework/cloud/data/module/deployer/yarn/YarnModuleDeployer.java
```diff
@@ -0,0 +1,213 @@
...
+import org.springframework.cloud.data.core.ModuleDeploymentId;
+import org.springframework.cloud.data.core.ModuleDeploymentRequest;
+import org.springframework.cloud.data.module.ModuleStatus;
+import org.springframework.cloud.data.module.deployer.ModuleDeployer;
+import org.springframework.util.StringUtils;
+import org.springframework.yarn.boot.app.YarnContainerClusterApplication;
+import org.springframework.yarn.boot.app.YarnInfoApplication;
+
+/**
+ * {@link ModuleDeployer} which communicates to XD's yarn app running
```
pperalta: We should consider using "Spring Cloud Stream/Task app" instead of "XD". In general the term "XD" should not be used in this project.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-yarn/src/main/java/org/springframework/cloud/data/module/deployer/yarn/YarnModuleDeployer.java
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	@Override
+	public ModuleDeploymentId deploy(ModuleDeploymentRequest request) {
+		int count = request.getCount();
+		ModuleCoordinates coordinates = request.getCoordinates();
+		ModuleDefinition definition = request.getDefinition();
+		logger.info("deploying request for definition: " + definition);
+
+		ModuleDeploymentId moduleDeploymentId = new ModuleDeploymentId(definition.getGroup(), definition.getLabel());
```
pperalta: Here you can use ``ModuleDeploymentId.fromDefinition(definition)``

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-yarn/src/main/java/org/springframework/cloud/data/module/deployer/yarn/YarnModuleDeployer.java
```diff
@@ -0,0 +1,213 @@
...
+		String[] lines = info.split("\\r?\\n");
+		if (lines.length > 2) {
+			for (int i = 2; i < lines.length; i++) {
+				ids.add(lines[i].trim());
+			}
+		}
+		return ids;
+	}
+
+	private static String convertFromModuleDeploymentId(ModuleDeploymentId moduleDeploymentId) {
```
pperalta: Not for this PR, but I'm wondering if we should define some kind of interface in the SPI to create runtime identifiers from `ModuleDeploymentId`...

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-yarn/src/main/java/org/springframework/cloud/data/module/deployer/yarn/YarnModuleDeployer.java
```diff
@@ -0,0 +1,213 @@
...
+		Map<String, String> definitionParameters = definition.getParameters();
+		Map<String, String> deploymentProperties = request.getDeploymentProperties();
+		logger.info("definitionParameters: " + definitionParameters);
+		logger.info("deploymentProperties: " + deploymentProperties);
+
+		// Using same app instance yarn boot cli is using to
+		// communicate with an app running on yarn via its boot actuator
+		YarnContainerClusterApplication app = new YarnContainerClusterApplication();
+		Properties appProperties = new Properties();
+		appProperties.setProperty(PREFIX + "operation", "CLUSTERCREATE");
```
pperalta: Is a cluster created for each module?

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-yarn/src/main/java/org/springframework/cloud/data/module/deployer/yarn/YarnModuleDeployer.java
```diff
@@ -0,0 +1,213 @@
...
+		appProperties.setProperty(PREFIX + "applicationId", yarnApplicationId);
+		appProperties.setProperty(PREFIX + "clusterId", clusterId);
+		appProperties.setProperty(PREFIX + "clusterDef", "module-template");
+		appProperties.setProperty(PREFIX + "projectionType", "default");
+		appProperties.setProperty(PREFIX + "projectionData.any", Integer.toString(count));
+		appProperties.setProperty(PREFIX + "extraProperties.containerModules", module);
+		app.appProperties(appProperties);
+		String output = app.run();
```
pperalta: When executing this, it appeared that `app.run()` blocks until the application is running in the cluster - is this correct? If so, this is an interesting difference between Diego and Yarn - the former will accept a request to launch a process but it won't block until the process is running.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-yarn/src/main/java/org/springframework/cloud/data/module/deployer/yarn/YarnModuleDeployer.java
```diff
@@ -0,0 +1,213 @@
...
+		appProperties.setProperty(PREFIX + "applicationId", yarnApplicationId);
+		appProperties.setProperty(PREFIX + "verbose", "false");
+		appProperties.setProperty(PREFIX + "clusterId", clusterId);
+		app.appProperties(appProperties);
+		String info = app.run();
+		logger.info("Output from YarnContainerClusterApplication run for CLUSTERINFO: " + info);
+
+		boolean deployed = false;
+		String[] lines = info.trim().split("\\r?\\n");
+		if (lines.length == 3 && lines[2].contains("RUNNING")) {
```
pperalta: Are there any other possible statuses? For instance, if an exception is thrown will the output say ERROR or something like that? For now it looks like the only possible statuses are DEPLOYED and UNKNOWN. If Yarn blocks on every invocation until the application is running, it appears that a single module will never be in a DEPLOYING state.

spring-cloud-data-yarn/spring-cloud-data-yarn-appmaster/src/main/java/org/springframework/cloud/data/yarn/appmaster/XdAppmaster.java
```diff
@@ -0,0 +1,56 @@
...
+import org.springframework.yarn.am.cluster.ContainerCluster;
+import org.springframework.yarn.am.cluster.ManagedContainerClusterAppmaster;
+
+/**
+ * Custom yarn appmaster tweaking container launch settings.
+ *
+ * @author Janne Valkealahti
+ *
+ */
+public class XdAppmaster extends ManagedContainerClusterAppmaster {
```
pperalta: Maybe `SCDAppMaster instead of `XdAppMaster`?

spring-cloud-data-yarn/spring-cloud-data-yarn-client/src/main/resources/application.yml
```diff
@@ -0,0 +1,20 @@
...
+        resourceManagerHost: localhost
+    yarn:
+        appType: XD
+        appName: spring-cloud-data-yarn-app
+        applicationBaseDir: /app/
+        applicationDir: /app/spring-cloud-data-yarn-app/
+        client:
+            clientClass: org.springframework.yarn.client.DefaultApplicationYarnClient
+            files:
+              - "file:spring-cloud-data-yarn/spring-cloud-data-yarn-appmaster/target/spring-cloud-data-yarn-appmaster-1.0.0.BUILD-SNAPSHOT.jar"
```
pperalta: Can/should this version be parameterized?

spring-cloud-data-yarn/spring-cloud-data-yarn-appmaster/src/main/java/org/springframework/cloud/data/yarn/appmaster/XdAppmaster.java
markfisher: FWIW for things like `XDTemplate`, we changed to `CloudDataTemplate`.

jvalkeal:
all the comments in(above) this PR are really good. Some of those (which can be renamed during a merge and patrick agreed to do are ok). Other commends are very valid but cannot really be addressed in this PR.

Like can we parametrize version of `spring-cloud-data-yarn-appmaster-1.0.0.BUILD-SNAPSHOT.jar`, yes but this really need a proper product around this library. In a way that current XD can choose hadoop distro, so we need to do a dance around it.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-yarn/src/main/java/org/springframework/cloud/data/module/deployer/yarn/YarnModuleDeployer.java
jvalkeal: Indeed this will be handled in next PR's. I have other shdp fixes lined up for this(I don't want to do too many workarounds in this because we can enhance it in a source where trouble originates).

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-yarn/src/main/java/org/springframework/cloud/data/module/deployer/yarn/YarnModuleDeployer.java
jvalkeal: It's just a little slow for now, I'm planning to make this execute faster so that shell can return more quickly. It's slow because of some workaround I had to do for interacting with appmaster. This will be fixed in next PR.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-yarn/src/main/java/org/springframework/cloud/data/module/deployer/yarn/YarnModuleDeployer.java
jvalkeal: Yes, cluster/group is for each module which is what diego(in a sense would do) so that we can ramp up/down modules, or stop/start/destroy all similar modules.

pperalta:
Overall LGTM. Squashed, rebased, merged as e239134.

PR #37, opened 2015-08-12
XD-3315: Add support for counter sink (REST API & Shell)

ericbottard:


spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/CounterOperations.java
```diff
@@ -0,0 +1,43 @@
...
+	 */
+	CounterResource retrieve(String name);
+
+	/**
+	 * Retrieve basic information (i.e. names) for existing counters.
+	 */
+	PagedResources<MetricResource> list(/* TODO */);
+
+	/**
+	 * Delete the counter with given name
```
markfisher: "...with *the* given name." <- ends with period for consistency

spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/CloudDataTemplate.java
```diff
@@ -55,10 +60,16 @@ public CloudDataTemplate(URI baseURI) {
 		resources.put("streams/definitions", new UriTemplate(resourceSupport.getLink("streams").getHref() + "/definitions"));
 		resources.put("streams/deployments", new UriTemplate(resourceSupport.getLink("streams").getHref() + "/deployments"));
 		this.streamTemplate = new StreamTemplate(restTemplate, resources);
+		this.counterTemplate = new CounterTemplate(restTemplate, resourceSupport);
```
markfisher: We need to decide which approach is best here so we can be consistent. I like that the StreamTemplate is currently handed only what it needs and that it's simply a Map or UriTemplates at that point.

spring-cloud-data-rest-resource/src/main/java/org/springframework/cloud/data/rest/resource/MetricResource.java
```diff
@@ -0,0 +1,58 @@
...
+import org.springframework.hateoas.PagedResources;
+import org.springframework.hateoas.ResourceSupport;
+
+/**
+ * Base class for REST representations of named metrics. Can be used when a just a shallow representation
+ * of a metric is expected.
+ *
+ * @author Eric Bottard
+ */
+public class MetricResource extends ResourceSupport {
```
markfisher: We should avoid premature inheritance. Only a single property is shared, and it's very generic ("name"). So this seems like overkill. I've been purposefully refactoring other such cases as we migrate.

spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/CloudDataTemplate.java
ericbottard: Yes, I wanted to address consistency once we had this discussion.
I think in any case it's better to deal with Links rather than just UriTemplate. The problem with pre-filtering though, is that CloudDataTemplate has knowledge of all the relations, and who is interested in what

spring-cloud-data-rest/src/main/java/org/springframework/cloud/data/rest/controller/StreamController.java
```diff
@@ -257,7 +257,7 @@ else if (moduleStates.contains(ModuleStatus.State.deployed) && moduleStates.size
 	 * {@link org.springframework.hateoas.ResourceAssembler} implementation
 	 * that converts {@link StreamDefinition}s to {@link StreamDefinitionResource}s.
 	 */
-	class Assembler extends ResourceAssemblerSupport<StreamDefinition, StreamDefinitionResource> {
+	 class Assembler extends ResourceAssemblerSupport<StreamDefinition, StreamDefinitionResource> {
```
markfisher: looks like an added space here

spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/CloudDataTemplate.java
markfisher: Can we reduce that knowledge on the CloudDataTemplate side to be based on prefixes, such that it hands out all links with a given prefix to a specific template? That might be a nice compromise. That said, since CloudDataController is the "master" controller, I think it's better to err on the side of *it* having more knowledge than to pass all of the links to the specific controllers, thus giving them visibility to each other's links.

spring-cloud-data-rest-resource/src/main/java/org/springframework/cloud/data/rest/resource/MetricResource.java
ericbottard: I knew about that. The "problem" is that metrics have shallow and deep representation support. So not doing that inheritance scheme would lead to 5*2 assemblers, etc.

markfisher: Is the shallow resource/assembler really necessary?

ericbottard: it is to retain compatibility with XD1.

spring-cloud-data-rest/src/main/java/org/springframework/cloud/data/rest/config/AdminConfiguration.java
```diff
@@ -60,6 +64,11 @@
 public class AdminConfiguration {
 
 	@Bean
+	public MetricRepository metricRepository(RedisConnectionFactory redisConnectionFactory) {
```
ilayaperumalg: I think we may need some conditions like `@ConditionalOnMissingBean(MetricRepository.class)` here.

spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/CounterOperations.java
```diff
@@ -0,0 +1,43 @@
...
+
+	/**
+	 * Retrieve information about the given named counter.
+	 */
+	CounterResource retrieve(String name);
+
+	/**
+	 * Retrieve basic information (i.e. names) for existing counters.
+	 */
+	PagedResources<MetricResource> list(/* TODO */);
```
ilayaperumalg: I believe this is `CounterResource` here?

spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/CounterTemplate.java
```diff
@@ -0,0 +1,59 @@
...
+
+import org.springframework.cloud.data.rest.resource.CounterResource;
+import org.springframework.cloud.data.rest.resource.MetricResource;
+import org.springframework.hateoas.PagedResources;
+import org.springframework.hateoas.ResourceSupport;
+import org.springframework.hateoas.UriTemplate;
+import org.springframework.web.client.RestTemplate;
+
+/**
+ * Created by ericbottard on 12/08/15.
```
ilayaperumalg: `@author` tag :-)

spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/CounterTemplate.java
```diff
@@ -0,0 +1,59 @@
...
+		links = resources;
+	}
+
+	@Override
+	public CounterResource retrieve(String name) {
+		return restTemplate.getForObject(links.getLink(COUNTER_RELATION).expand(name).getHref(), CounterResource.class);
+	}
+
+	@Override
+	public PagedResources<MetricResource> list() {
```
ilayaperumalg: What's the rational behind returning paged `MetricResource` instead of paged `CounterResource`?

spring-cloud-data-rest-resource/src/main/java/org/springframework/cloud/data/rest/resource/MetricResource.java
markfisher: It looks like using the current approach would lead to 5*2 assemblers anyways since the CounterController defines 2 ResourceAssemblers (one shallow, one deep) and those are tied to the type of a "counter", thus won't also support other metrics types (e.g. field-value-counter). Is that correct?

spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/CounterOperations.java
markfisher: no, this is the case where a shallow resource is returned - just names

spring-cloud-data-rest/src/main/java/org/springframework/cloud/data/rest/controller/CounterController.java
```diff
@@ -0,0 +1,168 @@
...
+
+/**
+ * Allows interaction with Counters.
+ *
+ * @author Eric Bottard
+ */
+@RestController
+@RequestMapping("/metrics/counters")
+@ExposesResourceFor(CounterResource.class)
+public class CounterController {
```
markfisher: a lot of warnings on this class (rawtypes, unchecked), as well as unused import (ResponseBody)

spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/CounterOperations.java
ericbottard: What Mark said. This has to be MetricResource (at least if we want to keep compatibility with the previous behavior)

spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/CounterTemplate.java
ericbottard: More than author tag, no javadoc :(

spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/CounterTemplate.java
ericbottard: See comment above, this is the v1 behavior. This really starts to make sense when dealing with heavier metrics (live AggregateCounters), but for consistency, this is like this on counters as well

spring-cloud-data-rest/src/main/java/org/springframework/cloud/data/rest/controller/CounterController.java
```diff
@@ -0,0 +1,168 @@
...
+		Page<Metric<Double>> page = new PageImpl<>(filterCounters(metrics));
+		ResourceAssembler<Metric<Double>, ? extends MetricResource> assemblerToUse =
+				detailed ? counterResourceAssembler : shallowResourceAssembler;
+		return pagedAssembler.toResource(page, assemblerToUse);
+	}
+
+	/**
+	 * Retrieve information about a specific counter.
+	 */
+	@RequestMapping(value = "/{name}", method = RequestMethod.GET, produces = MediaType.APPLICATION_JSON_VALUE)
```
ilayaperumalg: I think the attributes `produces` and `method` are redundant here.

spring-cloud-data-rest/src/main/java/org/springframework/cloud/data/rest/controller/CounterController.java
ericbottard: That may certainly be the case. Where did you see the warnings? It'd be interesting to remove the current output from the maven compilation/test (which is really noisy) and see those. I need to configure IDEA properly (which is really a pain _per project_)

spring-cloud-data-rest/src/main/java/org/springframework/cloud/data/rest/controller/CounterController.java
ericbottard: Not sure about the `produces` one, but I'd leave the `method` for clarity

ilayaperumalg: Since we assume the controllers always produce `application/json` I thought we can remove `produces` attribute.

spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/CloudDataTemplate.java
ericbottard: In any case, the `resources` variable above is a field of the `CloudDataTemplate`, so I was merely trying to use a container which seemed more appropriate (collection of links). Also, we'd need to address that concatenation thing done for Streams (eg by splitting definitions and deployments concerns at the controller level).

Can we use jdk8 in this code? stream filter would come handy here...

spring-cloud-data-rest-resource/src/main/java/org/springframework/cloud/data/rest/resource/MetricResource.java
ericbottard: Yes, I put it there, seeing how we started to squash class hierarchies. But I think it'll be needed anyway (the shallow assembler can move to a top level class and be reused)

spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/CounterTemplate.java
```diff
@@ -0,0 +1,58 @@
...
+import org.springframework.web.client.RestTemplate;
+
+/**
+ * Implementation for {@link CounterOperations} that interacts with the Spring Cloud Data REST API.
+ *
+ * @author Eric Bottard
+ */
+public class CounterTemplate implements CounterOperations {
+
+	public static final String COUNTERS_COLLECTION_RELATION = "counters";
```
markfisher: do these need to be public? If so, we should agree on a consistent naming scheme

spring-cloud-data-rest-resource/src/main/java/org/springframework/cloud/data/rest/resource/MetricResource.java
markfisher: ok

spring-cloud-data-rest/src/main/java/org/springframework/cloud/data/rest/controller/CounterController.java
ericbottard: It'd be better to be consistent for sure. I think you're right it's better to let the MessageConverter advertise what we support (you caught me, this was copy/pasted...)

spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/CounterTemplate.java
ericbottard: Nope, can be private (but will need to be shared with the other template if we follow your idea)

markfisher: we can discuss it later (subsequent to this PR being merged), but my idea (which is not fully formed) would favor a naming convention over sharing constants

spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/CloudDataTemplate.java
ericbottard: Created https://jira.spring.io/browse/XD-3391

spring-cloud-data-rest/src/main/java/org/springframework/cloud/data/rest/controller/CounterController.java
ericbottard: Created https://github.com/spring-cloud/spring-cloud-data/pull/43

ericbottard:
Has been merged as https://github.com/spring-cloud/spring-cloud-data/commit/6e4fe9e89d4777f5abac73d267fcdd441e2202fc

PR #41, opened 2015-08-13
Fix admin-ui redirect

ericbottard:


spring-cloud-data-rest/src/main/java/org/springframework/cloud/data/AdminApplication.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package org.springframework.cloud.data.rest;
```
markfisher: We do not want to move the AdminApplication up. It belongs under 'rest' (the root of spring-cloud-data-rest). Why is this necessary?

spring-cloud-data-rest/src/main/java/org/springframework/cloud/data/ui/UiController.java
```diff
@@ -0,0 +1,35 @@
...
+
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.RequestMapping;
+
+/**
+ * A simple controller to trigger a redirect to the index page of the admin ui (which comes packaged as a dependency).
+ *
+ * @author Eric Bottard
+ */
+// Note how this controller is NOT a RestController
```
markfisher: should include an explanation of *why* (and does this belong in actual JavaDoc?)

spring-cloud-data-rest/src/main/java/org/springframework/cloud/data/AdminApplication.java
ericbottard: Because the "UI" is not REST _per se_.

markfisher: It is simply a redirect to the REST API right? The more important thing here is that the "-rest" suffix is mapping to the ".rest" package as the "namespace" of this project.

ericbottard: Hmm, not sure what you're saying. The "ui" part is not a REST-API. It's a regular `.html` app, that is _consuming_ the API.
I understand what you're saying wrt namespacing. IMO, there are two solutions:
* go back to having the `ui` serving part inside `.rest` and live with it
* create an extra project, which is the real "admin", made of 2 parts: the REST endpoints and the ui

markfisher: Almost, but I wasn't thinking of creating another project. Rather we could rename `spring-cloud-data-rest` to `spring-cloud-data-admin`. Then the top-level package for all code in the project would be: `org.springframework.cloud.data.admin`. The following sub-packages would still exist: config, controller, repository. Only `AdminAppplication` would reside directly under the `.admin` top level package. The `UiController` would sit alongside the "REST" Controllers within the `controller` subpackage (REST being considered an implementation detail of all the other Controllers).

ericbottard:
Reworked by renaming spring-cloud-data-rest to spring-cloud-data-admin as discussed. Also removed some redundant `@Import` and `@ComponentScan`.

spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/repository/package-info.java
```diff
@@ -0,0 +1,4 @@
+/**
+ * Spring Cloud Data REST repository classes.
```
markfisher: s/REST/Admin/

spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/controller/package-info.java
```diff
@@ -0,0 +1,4 @@
+/**
+ * Spring Cloud Data REST & UI Controllers.
```
markfisher: "Spring Cloud Data *Admin* REST & UI Controllers." (or we can probably leave out "REST & UI").

spring-cloud-data-admin/src/test/java/org/springframework/cloud/data/admin/controller/TaskControllerTests.java
```diff
@@ -29,10 +29,12 @@
 import org.junit.runner.RunWith;
 
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.SpringApplicationConfiguration;
+import org.springframework.cloud.data.admin.AdminApplication;
 import org.springframework.cloud.data.core.TaskDefinition;
```
markfisher: looks like these imports did not get re-arranged in alphabetical order

spring-cloud-data-admin/src/test/java/org/springframework/cloud/data/admin/repository/InMemoryTaskDefinitionRepositoryTests.java
```diff
@@ -29,6 +29,8 @@
 import org.junit.Before;
 import org.junit.Test;
 
+import org.springframework.cloud.data.admin.repository.DuplicateTaskException;
```
markfisher: these 2 imports should not be necessary (same package)

spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/config/CloudConfiguration.java
```diff
@@ -14,21 +14,23 @@
...
 
 /**
- * Configuration for cloud profiles (cloud, lattice)
+ * Configuration for cloud profiles (cloud, lattice).
  *
  * @author Mark Fisher
  * @author Thomas Risberg
  */
 @Profile("cloud")
+@Configuration
```
markfisher: minor thing but should we have `@Profile` on the line after `@Configuration` since this class is primarily Configuration but has its applicability narrowed by the Profile specified? Also, I believe that would be consistent with other cases within the project.

markfisher:
Affected READMEs (top-level and spring-cloud-data-admin) should be updated as part of the PR.

markfisher:
will address comments while merging

PR #42, opened 2015-08-13
Remove unncessary Table support

ericbottard:
Spring Shell has those classes already.

Yet, rewriting the s-c-d shell may be a good opportunity to switch to a much richer Table support I have yet to include in Spring Shell.


markfisher:
Nice! Merging.

PR #45, opened 2015-08-13
Add lattice-core to admin REST

ilayaperumalg:
- This will enable Admin set lattice, cloud profiles when running on lattice

### DIFF
--- a/spring-cloud-data-admin/pom.xml
+++ b/spring-cloud-data-admin/pom.xml
@@ -81,6 +81,18 @@
 			<artifactId>spring-plugin-core</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-lattice-connector</artifactId>
+			<version>${spring-cloud-lattice.version}</version>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-spring-service-connector</artifactId>
+			<version>${spring-cloud-lattice.version}</version>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter-test</artifactId>
 			<scope>test</scope>
--- a/spring-cloud-data-admin/pom.xml
+++ b/spring-cloud-data-admin/pom.xml
@@ -80,16 +80,16 @@
 			<groupId>org.springframework.plugin</groupId>
 			<artifactId>spring-plugin-core</artifactId>
 		</dependency>
+		<!-- Lattice connector dependency to create services info from lattice -->
 		<dependency>
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-lattice-connector</artifactId>
-			<version>${spring-cloud-lattice.version}</version>
 			<optional>true</optional>
 		</dependency>
+		<!-- dependency to connect to detected cloud services -->
 		<dependency>
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-spring-service-connector</artifactId>
-			<version>${spring-cloud-lattice.version}</version>
 			<optional>true</optional>
 		</dependency>
 		<dependency>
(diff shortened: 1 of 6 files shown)


spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-lattice/pom.xml
```diff
@@ -27,6 +27,7 @@
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-lattice-core</artifactId>
 			<version>${spring-cloud-lattice.version}</version>
+			<scope>runtime</scope>
```
markfisher: Why is 'runtime' scope needed? I've been running the admin on Lattice and the "lattice" and "cloud" profiles have activated, before this PR.

ericbottard: Because compile > runtime.
But I agree that if our code does not depend on that jat at compile time, it would be better to put this in runtime scope (the same way binders should be runtime scoped as well, btw)

ericbottard: Scratch that, chances are we can't go "runtime" because `@Configuration` requires compile-time visibility

ilayaperumalg: I intend to add the lattice dependency that could bring in the `RedisConnectionFactory` for `AdminConfiguration` so that the `RedisMetricRepository` gets the cloud redis connectionfactory. But I understand that's what is required here. Perhaps, I should add `spring-cloud-lattice-connector` and `spring-cloud-spring-service-connector` to admin dependencies to get them working?

ilayaperumalg:
@markfisher I have updated the PR to include lattice cloud/service connector dependencies. Does this make sense?

markfisher:
Do we also need the profile-driven configuration for the Redis ConnectionFactory that would be used under the RedisMetricRepository?

ilayaperumalg:
I believe that will be taken care by the [spring-service-connector] (https://github.com/spring-cloud/spring-cloud-connectors/blob/master/spring-cloud-spring-service-connector/src/main/java/org/springframework/cloud/config/java/CloudServiceConnectionFactory.java) dependency if running on cloud profiles:

markfisher:
well in this case we are calling that service connector directly: https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream-binders/spring-cloud-stream-binder-redis/src/main/java/org/springframework/cloud/stream/binder/redis/config/RedisServiceAutoConfiguration.java#L54

ilayaperumalg:
yeah, but in the above case, we don't use `spring-service-connector` but `lattice-connector` [only] (https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream-binders/spring-cloud-stream-binder-redis/pom.xml#L42)

ilayaperumalg:
I believe we can get away with that cloud configuration for redis binder once we have `spring-service-connector` dependency in place.

spring-cloud-data-admin/pom.xml
```diff
@@ -81,6 +81,18 @@
...
 		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-lattice-connector</artifactId>
+			<version>${spring-cloud-lattice.version}</version>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-spring-service-connector</artifactId>
+			<version>${spring-cloud-lattice.version}</version>
```
markfisher: this should be the spring-cloud version

ilayaperumalg:
@markfisher fixed the dependency version for `spring-cloud-spring-service-connector`. Also, cleaned up some of the unused dependencies and added comments.

markfisher:
LGTM; merging

markfisher:
rebased, squashed, and merged as https://github.com/spring-cloud/spring-cloud-data/commit/ba2c12ae6564be08ef237f359bf5f5ee24a59379

PR #48, opened 2015-08-14
removed StreamDefinitionParser

markfisher:


spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/StreamDefinition.java
```diff
@@ -126,6 +147,57 @@ public String toString() {
 
 
 	/**
+	 * Build a list of ModuleDefinitions out of a parsed StreamNode.
+	 *
+	 * @param name the name of the definition unit
+	 * @param stream the AST construct representing the definition
+	 */
+	private static List<ModuleDefinition> buildModuleDefinitions(String name, StreamNode stream) {
```
pperalta: There's a lot going on in here for a private static method. Perhaps this should go into a builder of some sort?

markfisher: Agreed. I will move into a ModuleDefintionBuilder in the dsl package. At first I was trying to keep it as similar to TaskDefinition as possible, but the TaskNode is a much simpler class, and nothing like this method was needed for task parsing. By the way this doesn't need to be static, but it is just used once and has no state other than its params, and is called from the constructor (so for those reasons I opted for static). That said, the builder in the dsl package would be cleaner.

pperalta:
LGTM; squashed and merged as b443ad3.

PR #51, opened 2015-08-19
XD-3384 Fix admin server yarn profile

jvalkeal:
- I'm not sure if previous merge for XD-3384 was ok, but is broken now.
- AdminApplication/AdminConfiguration were changed to rely on
  component scan which did work with yarn profile because YarnConfiguration
  did have @Configuration and its bean 'yarnConfiguration' was overriden
  from spring-yarn autoconfiguration. This didn't cause trouble when YarnConfiguration
  were imported because bean name were then `org.springframework.cloud.data.admin.config.YarnConfiguration`.
  Component scan is creating different bean names, 'yarnConfiguration' in this case.
- Disabling spring yarn auto-config, classes are needed in future PR's so
  not removing deps.
- Add @Configuration to YarnConfiguration for scan to work.
- Add tests for default/yarn profiles. Couldn't add tests for
  cloud/lattice profiles because of error
  "org.springframework.cloud.CloudException: No suitable cloud connector found"

### DIFF
--- a/spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/AdminApplication.java
+++ b/spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/AdminApplication.java
@@ -17,12 +17,18 @@
 package org.springframework.cloud.data.admin;
 
 import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.yarn.boot.YarnAppmasterAutoConfiguration;
+import org.springframework.yarn.boot.YarnClientAutoConfiguration;
+import org.springframework.yarn.boot.YarnContainerAutoConfiguration;
 
 /**
  * @author Mark Fisher
  */
 @SpringBootApplication
+@EnableAutoConfiguration(exclude = { YarnClientAutoConfiguration.class, YarnAppmasterAutoConfiguration.class,
+		YarnContainerAutoConfiguration.class })
 public class AdminApplication {
 
 	public static void main(String[] args) throws InterruptedException {
--- a/spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/config/YarnConfiguration.java
+++ b/spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/config/YarnConfiguration.java
@@ -19,6 +19,7 @@ package org.springframework.cloud.data.admin.config;
 import org.springframework.cloud.data.module.deployer.ModuleDeployer;
 import org.springframework.cloud.data.module.deployer.yarn.YarnModuleDeployer;
 import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Profile;
 
 /**
@@ -28,6 +29,7 @@ import org.springframework.context.annotation.Profile;
  * @author Janne Valkealahti
  *
  */
+@Configuration
 @Profile("yarn")
 public class YarnConfiguration {
 
--- a/spring-cloud-data-admin/src/test/java/org/springframework/cloud/data/admin/AdminApplicationTests.java
+++ b/spring-cloud-data-admin/src/test/java/org/springframework/cloud/data/admin/AdminApplicationTests.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.data.admin;
+
+import static org.hamcrest.Matchers.instanceOf;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertThat;
+
+import org.junit.Test;
+import org.springframework.boot.SpringApplication;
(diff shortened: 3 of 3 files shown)


ilayaperumalg:
@jvalkeal The error "org.springframework.cloud.CloudException: No suitable cloud connector found" appears when we enable `cloud/lattice` profiles but the current environment isn't lattice. To work around this issue, we can add `PROCESS_GUID` property to system environment. This is all it requires for the `lattice-connector` to proceed.

jvalkeal:
@ilayaperumalg I don't think it's that because I tried all possible combinations.

Error is:
```
org.springframework.cloud.CloudException: No suitable cloud connector found
	at org.springframework.cloud.CloudFactory.getCloud(CloudFactory.java:55) ~[spring-cloud-core-1.2.0.RELEASE.jar:na]
	at org.springframework.cloud.config.CloudScanHelper.initializeCloud(CloudScanHelper.java:85) ~[spring-cloud-spring-service-connector-1.2.0.RELEASE.jar:na]
	at org.springframework.cloud.config.CloudScanHelper.registerServiceBeans(CloudScanHelper.java:55) ~[spring-cloud-spring-service-connector-1.2.0.RELEASE.jar:na]
	at org.springframework.cloud.config.java.ServiceScanConfiguration.registerBeanDefinitions(ServiceScanConfiguration.java:22) ~[spring-cloud-spring-service-connector-1.2.0.RELEASE.jar:na]
	at org.springframework.cloud.config.java.CloudScanConfiguration.registerBeanDefinitions(CloudScanConfiguration.java:22) ~[spring-cloud-spring-service-connector-1.2.0.RELEASE.jar:na]
```

for test

```
@Test
public void testCloudLattice() {
//	System.setProperty("PROCESS_GUID", "12345");
	SpringApplication app = new SpringApplication(AdminApplication.class);
	ConfigurableApplicationContext context = app.run(new String[] { "--spring.profiles.active=cloud,lattice" });
	assertThat(context.containsBean("moduleDeployer"), is(true));
	assertThat(context.getBean("moduleDeployer"), instanceOf(ReceptorModuleDeployer.class));
	context.close();
}
```

I event tried to remove `@CloudScan` from `CloudConfiguration` to see if it makes difference because it imports `CloudScanConfiguration` causing this error. `ReceptorModuleDeployer` is not created unless both `cloud` and `lattice` profiles are active.

But looking boot's `CloudAutoConfiguration` it imports `CloudScanConfiguration` if `cloud` profile is active and we get to same error anyway.

jvalkeal:
Ah, `LatticeConnector` is explicitly checking *env variable*, not system props or anything from spring context environment, meaning it's not registered.

```
	@Override
	public boolean isInMatchingCloud() {
		return environment.getEnvValue("PROCESS_GUID") != null;
	}
```

and because it's almost impossible(without nasty hacks) to set env variable from unit tests, I believe this cannot be tested.

If I do this manually I don't get this error but app dies:
```
$ export PROCESS_GUID=12345
$ java -Dspring.profiles.active=cloud,lattice -jar spring-cloud-data-admin/target/spring-cloud-data-admin-1.0.0.BUILD-SNAPSHOT.jar
...
2015-08-20 08:55:31.654 ERROR 3486 --- [           main] o.s.boot.SpringApplication               : Application startup failed
java.net.ConnectException: Connection timed out
...
Wrapped by: org.springframework.web.client.ResourceAccessException: I/O error on GET request for "http://receptor.192.168.11.11.xip.io/v1/desired_lrps":Connection timed out; nested exception is java.net.ConnectException: Connection timed out
...
2015-08-20 08:55:31.657  INFO 3486 --- [       Thread-2] ationConfigEmbeddedWebApplicationContext : Closing org.springframework.boot.context.embedded.AnnotationConfigEmbeddedWebApplicationContext@75830: startup date [Thu Aug 20 08:53:21 BST 2015]; root of context hierarchy
...
```

It's a bit bad that connection error is causing app start to fail because on cloud you're expected to have network troubles even if services are actually running.

ilayaperumalg:
@jvalkeal Does the above `connection timeout` error happens even if you have **lattice** running in your environment?

jvalkeal:
@ilayaperumalg I wouldn't know because I don't have lattice my system ;)

ilayaperumalg:
@jvalkeal ;) When you are running with `cloud,lattice` profiles enabled I believe it is expected to have **lattice** running; We use similar setup locally to test deploying modules into lattice while running admin on local (with PROCESS_GUID set as env prop).

jvalkeal:
@ilayaperumalg sure, fair enough. I just didn't expect whole app to fail/die if lattice is not available. Anyway this PR wasn't about lattice ;)

spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/AdminApplication.java
```diff
@@ -17,12 +17,18 @@
...
 import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.yarn.boot.YarnAppmasterAutoConfiguration;
+import org.springframework.yarn.boot.YarnClientAutoConfiguration;
+import org.springframework.yarn.boot.YarnContainerAutoConfiguration;
 
 /**
  * @author Mark Fisher
  */
 @SpringBootApplication
+@EnableAutoConfiguration(exclude = { YarnClientAutoConfiguration.class, YarnAppmasterAutoConfiguration.class,
```
ilayaperumalg: Do these configuration classes have some kind of conditions to enable auto configuration themselves (like `@ConditionOnProperty` on a profile? I am afraid we will start having a bunch of excludes this way. Is there other option to disable auto configuration of these?

Also, what happens when you run the admin on Yarn? should these auto-configurations need to be excluded there as well?

jvalkeal: There are boot auto-configuration classes in spring-yarn. `YarnAppmasterAutoConfiguration` and `YarnContainerAutoConfiguration` are expected to kick in when fat-jar is ran either as yarn `application master` or `container`. I excluded `YarnClientAutoConfiguration` to fix bean name clash which originated from component scan(read desc of this PR). I could have just changed `YarnConfiguration` class name which would have fixed bean name clash but felt it would have been a bit wrong. So ended up excluding all because those are not really needed in web context.

"admin on yarn", good point it would break because of same component scan issue. Component scan is evil and imho we should avoid it. Think we should have a broader discussion about it.

jvalkeal:
closing as pr is already broken due to recent changes and is going to be obsolete with my next pr which contains these same fixes.

PR #52, opened 2015-08-19
bindings are now just parameters

markfisher:


spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleDefinition.java
```diff
@@ -76,17 +71,13 @@
 	 * @param bindings map of channel bindings; may not be {@code null}
```
ilayaperumalg: javadoc needs to be removed for bindings.

markfisher: thanks! updating

mminella:
Squashed and merged as 4a39810af07af32bdc6b38e5073664909aacd773

PR #56, opened 2015-08-21
XD-2919: Add Redis based StreamDefinitionRepository

ericbottard:


spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/repository/RedisStreamDefinitionRepository.java
```diff
@@ -0,0 +1,196 @@
...
+
+	private final BoundHashOperations<String, String, String> hashOperations;
+
+	private final BoundZSetOperations<String, String> zSetOperations;
+
+	private final String hashKey;
+
+	public RedisStreamDefinitionRepository(String hashKey, RedisConnectionFactory redisConnectionFactory) {
+		this.hashKey = hashKey;
+		RedisTemplate<String, String> redisTemplate = new RedisTemplate<>();
```
markfisher: why not use StringRedisTemplate and avoid the need for all the explicit serializer config?

spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/repository/RedisStreamDefinitionRepository.java
```diff
@@ -0,0 +1,196 @@
...
+	public Iterable<StreamDefinition> findAll() {
+		return findAll(new Sort("name"));
+	}
+
+	@Override
+	public Iterable<StreamDefinition> findAll(Iterable<String> strings) {
+		Collection<String> keys;
+		if (strings instanceof Collection) {
+			keys = (Collection<String>) strings;
+		} else {
```
markfisher: newline after brace

spring-cloud-data-admin/src/test/java/org/springframework/cloud/data/admin/repository/RedisStreamDefinitionRepositoryTests.java
```diff
@@ -0,0 +1,176 @@
...
+
+	@Test
+	public void testUdate() {
+		repository.save(new StreamDefinition("a", "time | log"));
+		repository.save(new StreamDefinition("a", "time | file"));
+		assertThat(repository.findOne("a"), hasProperty("dslText", is("time | file")));
+	}
+
+
+		@After
```
markfisher: funky indent

spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/repository/RedisStreamDefinitionRepository.java
```diff
@@ -0,0 +1,196 @@
...
+ */
+public class RedisStreamDefinitionRepository implements StreamDefinitionRepository {
+
+	private final BoundHashOperations<String, String, String> hashOperations;
+
+	private final BoundZSetOperations<String, String> zSetOperations;
+
+	private final String hashKey;
+
+	public RedisStreamDefinitionRepository(String hashKey, RedisConnectionFactory redisConnectionFactory) {
```
pperalta: Can we have some JavaDoc for this constructor?

spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/repository/RedisStreamDefinitionRepository.java
pperalta: +1

spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/repository/RedisStreamDefinitionRepository.java
```diff
@@ -0,0 +1,196 @@
...
+ * Also stores the list of all names in a sorted set (with constant zero score), as this
+ * guarantees lexicographic order.
+ *
+ * @author Eric Bottard
+ */
+public class RedisStreamDefinitionRepository implements StreamDefinitionRepository {
+
+	private final BoundHashOperations<String, String, String> hashOperations;
+
+	private final BoundZSetOperations<String, String> zSetOperations;
```
pperalta: The advantage of a sorted set is that it allows for paging of streams without having to load the entire key set in memory. However, the disadvantage is that mutating operations (add, remove) require multiple calls to Redis which may lead to inconsistencies if the connection to Redis is interrupted in the middle of an operation.
```
public <S extends StreamDefinition> S save(S entity) {
	zSetOperations.add(entity.getName(), 0);
	hashOperations.put(entity.getName(), entity.getDslText());
	return entity;
}
```
For instance, if the connection to Redis is dropped after adding the stream name to the set but before adding it to the hash, the stream name will be in the set even though it isn't in the hash.

Perhaps we can simply use hash operations and load the key set in memory when sorting is required?

spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/repository/RedisStreamDefinitionRepository.java
```diff
@@ -0,0 +1,196 @@
...
+	public void deleteAll() {
+		hashOperations.getOperations().delete(hashKey);
+		zSetOperations.getOperations().delete(hashKey + ".names");
+	}
+
+	/**
+	 * Return a list of StreamDefinitions, made by mapping a set of non null names and possibly null dsl texts.
+	 * In case of null dsl text, a null element is added to the result list.
+	 */
+	private List<StreamDefinition> zipToStreamDefinitions(Collection<String> names, Collection<String> definitions) {
```
pperalta: I don't understand the naming of this method - what does the term "zip" mean in this context?

pperalta: Having a `Collection` of stream names and a `Collection` of definitions isn't a strong enough contract to guarantee that the respective iterators will return matching names and definitions (for instance, these could be backed by a `HashSet`). Perhaps these should be defined as `List` instead?

spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/repository/RedisStreamDefinitionRepository.java
markfisher: a "multi"[1] redis command can be used to make multiple operations atomic, but I would opt for the simplicity of just using a hash

[1] http://docs.spring.io/spring-data/redis/docs/current/reference/html/#tx

spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/repository/RedisStreamDefinitionRepository.java
markfisher: this signature could change to (List, List) since the caller can pass a List for both 'keys' and 'dslTexts' (just remove the `if (entities instanceof Collection)` block or change Collection to List for that too).

spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/repository/RedisStreamDefinitionRepository.java
ericbottard: Well, it wouldn't be _that_ simple, as pagination is done by index, while sorted hash typically work by key. So would have to _e.g._ re-insert in a sorted hash, iterate and drop the first N, etc. Note sure it reads better.

spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/repository/RedisStreamDefinitionRepository.java
ericbottard: @pperalta zip means taking to collections and return a collection of pairs (or result of function application on two elements), similar to what your jacket zipper does. see http://stackoverflow.com/questions/1115563/what-is-zip-functional-programming or https://www.haskell.org/hoogle/?hoogle=zip

pperalta: Thanks for the description @ericbottard, I learned something new today! :)

pperalta:
Rebased, squashed, merged as c9d9d533bcb74a089fe98c97e4898e5d6984241f with a few tweaks.

PR #58, opened 2015-08-21
XD-3419: Redis module registry

pperalta:
* Implementation of module registry backed by Redis
* Test configurations modified to use StubModuleRegistry so that tests don't require Redis
* Tweaks and doc added to AbstractShellIntegrationTest

**Reviewers:**
I am seeking opinions on whether each module should be simple key/value pair in Redis (as implemented), or if three hashes keyed by module type would be preferable.

### DIFF
--- a/spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/config/AdminConfiguration.java
+++ b/spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/config/AdminConfiguration.java
@@ -29,7 +29,7 @@ import org.springframework.cloud.data.admin.repository.TaskDefinitionRepository;
 import org.springframework.cloud.data.module.deployer.ModuleDeployer;
 import org.springframework.cloud.data.module.deployer.local.LocalModuleDeployer;
 import org.springframework.cloud.data.module.registry.ModuleRegistry;
-import org.springframework.cloud.data.module.registry.StubModuleRegistry;
+import org.springframework.cloud.data.module.registry.RedisModuleRegistry;
 import org.springframework.cloud.stream.module.launcher.ModuleLauncher;
 import org.springframework.cloud.stream.module.launcher.ModuleLauncherConfiguration;
 import org.springframework.context.annotation.Bean;
@@ -76,8 +76,8 @@ public class AdminConfiguration {
 	}
 
 	@Bean
-	public ModuleRegistry moduleRegistry() {
-		return new StubModuleRegistry();
+	public ModuleRegistry moduleRegistry(RedisConnectionFactory redisConnectionFactory) {
+		return new RedisModuleRegistry(redisConnectionFactory);
 	}
 
 	@Configuration
--- a/spring-cloud-data-module-registry/src/main/java/org/springframework/cloud/data/module/registry/RedisModuleRegistry.java
+++ b/spring-cloud-data-module-registry/src/main/java/org/springframework/cloud/data/module/registry/RedisModuleRegistry.java
@@ -1,0 +1,117 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.data.module.registry;
+
+import org.springframework.cloud.data.core.ModuleCoordinates;
+import org.springframework.data.redis.connection.RedisConnectionFactory;
+import org.springframework.data.redis.core.RedisOperations;
+import org.springframework.data.redis.core.StringRedisTemplate;
+
+/**
+ * {@link ModuleRegistry} implementation backed by Redis.
+ *
+ * @author Patrick Peralta
+ */
+public class RedisModuleRegistry implements ModuleRegistry {
+
+	/**
+	 * Prefix for keys used for storing module coordinates.
+	 */
+	public static final String KEY_PREFIX = "spring.cloud.module";
+
+	/**
+	 * Group ID for default modules.
+	 */
+	private static final String GROUP_ID = "org.springframework.cloud.stream.module";
+
+	/**
+	 * Version number for default modules.
+	 */
+	private static final String VERSION = "1.0.0.BUILD-SNAPSHOT";
+
+	/**
+	 * Redis operations template.
+	 */
+	private final RedisOperations<String, String> redisOperations;
+
+
+	/**
+	 * Construct a {@code RedisModuleRegistry} with the provided
+	 * {@link RedisConnectionFactory}.
+	 *
+	 * @param redisConnectionFactory connection factory for Redis
+	 */
+	public RedisModuleRegistry(RedisConnectionFactory redisConnectionFactory) {
+		redisOperations = new StringRedisTemplate(redisConnectionFactory);
+		populateDefaults();
+	}
+
+	/**
+	 * Populate the registry with default module coordinates; will
+	 * not overwrite existing values.
+	 */
+	private void populateDefaults() {
+		populateDefault("http", "source");
+		populateDefault("time", "source");
+		populateDefault("filter", "processor");
+		populateDefault("groovy-filter", "processor");
+		populateDefault("groovy-transform", "processor");
+		populateDefault("transform", "processor");
+		populateDefault("counter", "sink");
+		populateDefault("log", "sink");
+	}
+
+	/**
+	 * Populate the registry with default values for the provided
+	 * module name and type; will not overwrite existing values.
+	 *
+	 * @param name module name
+	 * @param type module type
+	 */
+	private void populateDefault(String name, String type) {
+		redisOperations.boundValueOps(keyFor(name, type))
+				.setIfAbsent(defaultCoordinatesFor(name + '-' + type).toString());
+	}
+
+	/**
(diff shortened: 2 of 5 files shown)


markfisher:
Three hashes (one per module type) seems better to me. However, id depends how the `ModuleRegistry` needs to evolve (with additional methods) to support the REST API. The bare minimum would include:

* module register
* module unregister
* module list
* module info

In XD 1.0, `module list` is supported by a simple "findAll", and `module info` is supported by a "findByNameAndType". Currently `ModuleRegistry` has only the latter. It will clearly need a findAll that returns a Map (with keys being "type:name") or Map-of-Maps (with top-level key being "type" and 2nd-level being "name"), with the values being `ModuleCoordinates` (I prefer Map-of-Maps). That could be supported by either approach, single hash or hash-per-type, but to me hash-by-type feels more elegant. We should also consider how job module coordinates will be stored. I assume that would just be one more hash (if we go with hash-per-type) cc @mminella

spring-cloud-data-shell/src/test/java/org/springframework/cloud/data/shell/AbstractShellIntegrationTest.java
```diff
@@ -36,27 +44,53 @@
...
+	/**
+	 * Generator used to create random stream names.
+	 */
 	private final IdGenerator idGenerator = new AlternativeJdkIdGenerator();
 
 	/**
-	 * This property can be set to false when running test suite.
+	 * System property indicating whether the test infrastructure should
+	 * be shut down after all tests are executed. If running in a test
+	 * suite, this system property should be set to {@code true} to allow
```
markfisher: should this say "false" instead of "true"?

spring-cloud-data-module-registry/src/main/java/org/springframework/cloud/data/module/registry/RedisModuleRegistry.java
```diff
@@ -0,0 +1,117 @@
...
+
+	/**
+	 * Construct a {@code RedisModuleRegistry} with the provided
+	 * {@link RedisConnectionFactory}.
+	 *
+	 * @param redisConnectionFactory connection factory for Redis
+	 */
+	public RedisModuleRegistry(RedisConnectionFactory redisConnectionFactory) {
+		redisOperations = new StringRedisTemplate(redisConnectionFactory);
+		populateDefaults();
```
markfisher: I feel like we should avoid executing Redis operations in the constructor. Perhaps we could make `populateDefaults` a `@PostConstruct` method instead.

markfisher:
here is a commit on a XD-3419 branch on my fork that reflects my suggestions above:
https://github.com/markfisher/spring-cloud-data/commit/e31a75c1a96970737ad4dc19d3728a1f77c96205

spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/config/AdminConfiguration.java
```diff
@@ -76,8 +76,8 @@ public TaskDefinitionRepository taskDefinitionRepository() {
 	}
 
 	@Bean
-	public ModuleRegistry moduleRegistry() {
-		return new StubModuleRegistry();
+	public ModuleRegistry moduleRegistry(RedisConnectionFactory redisConnectionFactory) {
+		return new RedisModuleRegistry(redisConnectionFactory);
```
ericbottard: This would make redis a hard requirement, do we want that unconditionally ?

spring-cloud-data-module-registry/src/main/java/org/springframework/cloud/data/module/registry/RedisModuleRegistry.java
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	/**
+	 * Populate the registry with default module coordinates; will
+	 * not overwrite existing values.
+	 */
+	private void populateDefaults() {
```
ericbottard: Would be better to consider a default if no mapping is found (that way, no need to hardcode modules we _do_ have everytime that list changes)

spring-cloud-data-module-registry/src/main/java/org/springframework/cloud/data/module/registry/RedisModuleRegistry.java
```diff
@@ -0,0 +1,117 @@
...
+
+	/**
+	 * Prefix for keys used for storing module coordinates.
+	 */
+	public static final String KEY_PREFIX = "spring.cloud.module";
+
+	/**
+	 * Group ID for default modules.
+	 */
+	private static final String GROUP_ID = "org.springframework.cloud.stream.module";
```
ericbottard: Should be named `DEFAULT_GROUP_ID` IMO (ditto for version)

spring-cloud-data-module-registry/src/main/java/org/springframework/cloud/data/module/registry/RedisModuleRegistry.java
```diff
@@ -0,0 +1,117 @@
...
+
+	/**
+	 * Version number for default modules.
+	 */
+	private static final String VERSION = "1.0.0.BUILD-SNAPSHOT";
```
ericbottard: If we keep this logic in the long run and s-c-data version is kept in synch with s-c-s-modules versions, would be nice to have this reflect the real version. Maybe we have found a good candidate for maven resource filtering in `application.yml` after all (pro-tip: use `@` as the delimiter)

spring-cloud-data-shell/src/test/java/org/springframework/cloud/data/shell/AbstractShellIntegrationTest.java
```diff
@@ -36,27 +44,53 @@
 import org.springframework.util.SocketUtils;
 
 /**
- * Base class for all the Shell integration test classes.
+ * Base class for shell integration tests. This class sets up and tears down
+ * the infrastructure required for executing shell tests - in particular, the
+ * {@link AdminApplication} server.
+ * <p>
+ * Extensions of this class may obtain an instance of {@link StreamCommandTemplate}
```
ericbottard: Not sure we want to mention this, or if we do, phrase it as an example, as many more are to follow (_e.g._ task(), counter(), etc...)

spring-cloud-data-module-registry/src/main/java/org/springframework/cloud/data/module/registry/RedisModuleRegistry.java
markfisher: I'm actually thinking we should break out everything related to populating defaults to a separate class (lifecycle-triggered) that exists solely for that purpose and has the ModuleRegistry as a dependency. That way it would be decoupled from any particular ModuleRegistry implementation, and it could be enabled/disabled, profile-driven, etc.

markfisher: @pperalta @ericbottard please have a look at my branch which now has 2 commits on top of this PR: https://github.com/markfisher/spring-cloud-data/tree/XD-3419

This is the one where I factored out the `ModuleRegistryPopulator` role:
https://github.com/markfisher/spring-cloud-data/commit/a4e5f538ba1fd08e34d0696e47a9958fe48fc800

spring-cloud-data-module-registry/src/main/java/org/springframework/cloud/data/module/registry/RedisModuleRegistry.java
markfisher: +1 and I changed those keys in my commit of the `ModuleRegistryPopulator` (see above)

spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/config/AdminConfiguration.java
markfisher: At some point, we will probably add a "store" property that could apply to the `ModuleRegistry` and the `StreamDefinitionRepository` (or separate properties for each).

spring-cloud-data-shell/src/test/java/org/springframework/cloud/data/shell/AbstractShellIntegrationTest.java
markfisher: +1

spring-cloud-data-shell/src/test/java/org/springframework/cloud/data/shell/AbstractShellIntegrationTest.java
pperalta: +1

spring-cloud-data-module-registry/src/main/java/org/springframework/cloud/data/module/registry/RedisModuleRegistry.java
pperalta: Changes look good. Only thing I found missing was the JavaDoc suggestions @ericbottard made for `AbstractShellIntegrationTest` below.

spring-cloud-data-shell/src/test/java/org/springframework/cloud/data/shell/AbstractShellIntegrationTest.java
markfisher: fixed

spring-cloud-data-shell/src/test/java/org/springframework/cloud/data/shell/AbstractShellIntegrationTest.java
markfisher: changed

spring-cloud-data-module-registry/src/main/java/org/springframework/cloud/data/module/registry/RedisModuleRegistry.java
markfisher: we will also need version info for the shell banner

markfisher:
rebased, polished, squashed and merged as: https://github.com/spring-cloud/spring-cloud-data/commit/be89bc617599f1015fcfec947372819c8881e05c

PR #59, opened 2015-08-21
Remove CloudScan from admin cloud config

ilayaperumalg:
- Remove `@CloudScan` from CloudConfiguration and add explicit configuration
for `RedisConnectionFactory` bean
 - This will avoid creating any other service beans that might as well be created
during cloud scan

### DIFF
--- a/spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/config/CloudConfiguration.java
+++ b/spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/config/CloudConfiguration.java
@@ -16,13 +16,18 @@
 
 package org.springframework.cloud.data.admin.config;
 
-import org.springframework.cloud.config.java.CloudScan;
+import org.springframework.boot.autoconfigure.AutoConfigureBefore;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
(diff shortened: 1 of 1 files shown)


spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/config/CloudConfiguration.java
```diff
@@ -33,9 +38,23 @@
...
+	@AutoConfigureBefore(RedisAutoConfiguration.class)
+	protected static class RedisConfig {
+
+		@Bean
+		public Cloud cloud() {
+			return new CloudFactory().getCloud();
+		}
+
+		@Bean
+		@ConditionalOnMissingBean(RedisConnectionFactory.class)
```
markfisher: since this is now using `@AutoConfigureBefore`, does it still need the `@ConditionalOnMissingBean` as well?

ilayaperumalg: No, we don't need it here; will remove it.

markfisher:
LGTM; merging

markfisher:
squashed and merged as https://github.com/spring-cloud/spring-cloud-data/commit/2e621d8444d8d08b34505e5563055d1de85ae8d8

PR #60, opened 2015-08-24
CloudFoundryModuleDeployer implementation

Zteve:
This commit creates the first implementation of the ModuleDeployer interface
for Cloud Foundry. It supports status(ModuleDeploymentId) and status() methods
only.

See the README.md file for environment settings required.

Note: a small change was made to `CloudConfiguration.java` in `org.springframework.cloud.data.admin.config` so that the `CloudFoundryModuleDeployer` can be correctly wired in.


ericbottard:
Will need to use TABs, not spaces for all files. This should immediately reduce diff noise and ease the PR.
If using IDEA, please see https://github.com/spring-projects/spring-framework/blob/master/src/idea/spring-framework.xml

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/pom.xml
```diff
@@ -1,20 +1,62 @@
...
+            <groupId>org.springframework.security.oauth</groupId>
+            <artifactId>spring-security-oauth2</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+        </dependency>
+    </dependencies>
+
+    <repositories>
```
ericbottard: These two repos should be removed (and build with `mvn -s .settings.xml` until we add maven wrapper support)

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/AbstractPaginatedResponse.java
```diff
@@ -0,0 +1,161 @@
...
+
+package org.springframework.cloud.data.module.deployer.cloudfoundry;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.List;
+
+abstract class AbstractPaginatedResponse {
+
+    private volatile Integer        totalResults;
```
ericbottard: Can these really be `null`? If not, `int` would be better

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/AbstractPaginatedResponse.java
```diff
@@ -0,0 +1,161 @@
...
+
+        public void setMetadata(ListSpacesResponse.Resource.Metadata metadata) {
+            this.metadata = metadata;
+        }
+
+        static final class Metadata {
+
+            private volatile String id;
+            private volatile String url;
+            private volatile String createdAt;
```
ericbottard: While not the meat of this PR, using some form of Date object would be better (this relies on custom Jackson serializer)

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/AbstractPaginatedResponse.java
```diff
@@ -0,0 +1,161 @@
...
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.data.module.deployer.cloudfoundry;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.List;
+
+abstract class AbstractPaginatedResponse {
```
ericbottard: All classes should have your author tag(s)

ericbottard: Basic javadoc for each type is welcome too

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/ApplicationInstanceStatus.java
```diff
@@ -0,0 +1,177 @@
...
+        this.statistics = statistics;
+    }
+
+    static final class Statistics {
+
+        private volatile Usage        usage;
+        private volatile String       name;
+        private volatile List<String> uris;
+        private volatile String       host;
+        private volatile Integer      port;
```
ericbottard: use primitive `int` as appropriate

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/ApplicationInstanceStatus.java
```diff
@@ -0,0 +1,177 @@
...
+ */
+
+package org.springframework.cloud.data.module.deployer.cloudfoundry;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.List;
+
+/**
+ * Each instance of a running application corresponds to a module instance. The status and properties of an instance
```
ericbottard: The first sentence should be inverted (module => app), if not removed (as this ties this - somewhat generic - class to module semantics. Granted, it is currently in this project/package, but may be deemed reusable in the future)

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/ApplicationInstanceStatus.java
```diff
@@ -0,0 +1,177 @@
...
+ * are recorded in an {@code ApplicationInstanceStatus} object.
+ */
+final class ApplicationInstanceStatus {
+
+    private volatile String     state;
```
ericbottard: Can this be made an enum?

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/ApplicationInstanceStatus.java
```diff
@@ -0,0 +1,177 @@
...
+            return usage;
+        }
+
+        public void setUsage(Usage usage) {
+            this.usage = usage;
+        }
+
+        static final class Usage {
+
+            private volatile Integer disk;
```
ericbottard: Primitives as appropriate

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/ApplicationStatus.java
```diff
@@ -0,0 +1,42 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.data.module.deployer.cloudfoundry;
+
+import java.util.HashMap;
+import java.util.Map;
+
+final class ApplicationStatus extends HashMap<String, ApplicationInstanceStatus> {
```
ericbottard: Using inheritance here feels wrong.
Should either be replaced with
* dealing with just a dumb Map
* returning a wrapper (having `getId()`) around a plain Map at usage point (_e.g._ `GetApplicationsStatusResponse`)
* customizing Jackson ser/deser to a custom type (this `ApplicationStatus`) which would use delegation rather than inheritance

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/CloudFoundryClient.java
```diff
@@ -0,0 +1,29 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.data.module.deployer.cloudfoundry;
+
+interface CloudFoundryClient {
```
ericbottard: Javadoc highly needed here, and for each method

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/CloudFoundryModuleDeploymentIdConverter.java
```diff
@@ -0,0 +1,47 @@
...
+    private static final String PREFIX = "spring_cloud_data_";
+
+    String toApplicationName(ModuleDeploymentId moduleDeploymentId) {
+        return PREFIX + moduleDeploymentId.toString();
+    }
+
+    ModuleDeploymentId toModuleDeploymentId(String applicationName) {
+        try {
+            return ModuleDeploymentId.parse(moduleIdString(applicationName));
+        } catch (NullPointerException e) {
```
ericbottard: Should not rely on NPE and IAE here. Pull up the logic in `moduleIdString()`

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/GetApplicationStatisticsResponse.java
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.data.module.deployer.cloudfoundry;
+
+import java.util.HashMap;
+
+final class GetApplicationStatisticsResponse extends HashMap<String, ApplicationInstanceStatus> {
```
ericbottard: See comment above about inheriting from HashMap

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/ModuleDeployerConfiguration.java
```diff
@@ -0,0 +1,58 @@
...
+import org.springframework.security.oauth2.client.OAuth2RestTemplate;
+import org.springframework.security.oauth2.client.resource.OAuth2ProtectedResourceDetails;
+import org.springframework.security.oauth2.client.token.DefaultAccessTokenRequest;
+import org.springframework.security.oauth2.client.token.grant.password.ResourceOwnerPasswordResourceDetails;
+
+/**
+ * This class is present to make sure that Spring Authentication protocols for Cloud Foundry controller API access are
+ * correctly wired.
+ */
+@ComponentScan
```
ericbottard: `@ComponentScan` scans the current package for other Spring beans. Unless I'm mistaken, there is nothing else here. Can be removed.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/ModuleDeployerConfiguration.java
```diff
@@ -0,0 +1,58 @@
...
+@Configuration
+@EnableAutoConfiguration(exclude = OAuth2AutoConfiguration.class)
```
ericbottard: Although chances are that Auto Configuration is already enabled where we make use of this class, it should not trigger it itself (as this class is pretty low level and and auto conf can enable a whole world of automatic behaviors that the consumer of this class does not want / expect).
I'm not sure what this particular addition brings to the table here, but equivalent behaviour should be achieved without AutoConf

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/ModuleDeployerConfiguration.java
```diff
@@ -0,0 +1,58 @@
...
+class ModuleDeployerConfiguration {
+
+    @Bean
+    public OAuth2RestTemplate oauth2RestTemplate(OAuth2ClientContext oauth2ClientContext,
+                                                 OAuth2ProtectedResourceDetails details) {
+        return new OAuth2RestTemplate(details, oauth2ClientContext);
+    }
+
+    @Bean
+    @ConfigurationProperties("security.oauth2.client")
```
ericbottard: Maybe this sould be namespaced to indicate it relates to CF module deployer configuration

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/ModuleDeployerConfiguration.java
ericbottard: My bad, I see there is a `@Component` on `StandardCloudFoundryClient` at least. Nevertheless, this project globally uses explicit bean creation rather than component scanning, so I guess it is better to convert.
To clarify, remove `@ComponentScan` here, `@Component` on `StandardCloudFoundryClient` and add 
```
@Bean
public CloudFoundryClient cloudFoundryClient() {
    return new StandardCloudFoundryClient();
}
```
here. Rinse/repeat for all other beans, as appropriate

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/resources/application.yml
```diff
@@ -0,0 +1,20 @@
+---
+cloudfoundry:
```
ericbottard: This will typically not be read once this code is inside a bigger project at runtime (only one `application.yml` file is read on the classpath, the first one that is found, which is non-deterministic).
Defaults, *if we want them* should be set otherwise (_e.g._ as field initializers)

ericbottard: Also, see my above comment related to putting all those keys in an appropriate namesapce yet to be discussed. cc @markfisher

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/resources/application.yml
```diff
@@ -0,0 +1,20 @@
...
+  api:
+    endpoint: https://api.${host}.springapps.io
```
ericbottard: How is this `${host}` placeholder expected to be resolved?

ericbottard:
On a general note, I think we should revisit the whole `XXXRequest` (/  `XXXResponse`) naming scheme, which feels very much non-REST IMHO. While I understand their appeal in case of future extension, most of them have a unique component in them anyway (_e.g._ a String id).

We should consider moving changing (taking the more rich `ListApplicationsRequest` as an example):
```
CloudFoundryClient {
? listApplications(ListApplicationsRequest request) {...}
}
```

to
```
CloudFoundryClient {
? listApplications(String name, String spaceId) {...}
}
```

ericbottard:
Moreover, in other Spring projects, the naming scheme would typically be
```
CloudFoundryClient -> CloudFoundryOperations
StandardCloudFoundryClient -> CloudFoundryTemplate
```

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/AbstractPaginatedResponse.java
```diff
@@ -0,0 +1,170 @@
...
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.data.module.deployer.cloudfoundry;
+
+import java.util.List;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+abstract class AbstractPaginatedResponse {
```
mminella: Is there a reason Spring Data's pagination facilities can't be use for this?

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/ApplicationStatus.java
mminella: +1 leaning towards bullet 2

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/CloudFoundryClient.java
```diff
@@ -0,0 +1,31 @@
+/*
```
mminella: Just curious...does it make sense to break this out into another project like the Receptor client is?  I'd imagine that this would be useful to more than just spring-cloud-data.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/CloudFoundryModuleDeployer.java
```diff
@@ -16,32 +16,73 @@
...
+		throw new UnsupportedOperationException("to be implemented");
+		// return null;
 	}
 
 	@Override
-	public void undeploy(ModuleDeploymentId id) {
-		throw new UnsupportedOperationException();
+	public Map<ModuleDeploymentId, ModuleStatus> status() {
+		GetApplicationsStatusResponse response = this.resource.getApplicationsStatus(
+				new GetApplicationsStatusRequest());
```
mminella: Does this return all of the application statuses for the org?  Do we plan on paring it down a bit (or is there no way with the exposed API)?

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/CloudFoundryModuleDeploymentIdConverter.java
```diff
@@ -0,0 +1,49 @@
...
+
+	String toApplicationName(ModuleDeploymentId moduleDeploymentId) {
+		return PREFIX + moduleDeploymentId.toString();
+	}
+
+	ModuleDeploymentId toModuleDeploymentId(String applicationName) {
+		try {
+			return ModuleDeploymentId.parse(moduleIdString(applicationName));
+		}
+		catch (NullPointerException e) {
```
mminella: Is there a null check we can add somewhere instead of checking for an NPE?  Using Exceptions to control logic flow isn't recommended.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/DeleteApplicationResourceResponse.java
```diff
@@ -0,0 +1,42 @@
...
+
+package org.springframework.cloud.data.module.deployer.cloudfoundry;
+
+final class DeleteApplicationResourceResponse {
+
+	private volatile boolean deleted;
+
+	private volatile boolean found;
+
+	Boolean getDeleted() {
```
mminella: Why `Boolean` instead of `boolean`?

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/DeleteApplicationResourceResponse.java
```diff
@@ -0,0 +1,42 @@
...
+		return deleted;
+	}
+
+	DeleteApplicationResourceResponse withDeleted(boolean deleted) {
+		this.deleted = deleted;
+		return this;
+	}
+
+	Boolean getFound() {
```
mminella: Why `Boolean` instead of `boolean`?

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/DeleteApplicationResponse.java
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.data.module.deployer.cloudfoundry;
+
+final class DeleteApplicationResponse {
+
+	private volatile boolean deleted;
+
+	Boolean getDeleted() {
```
mminella: Why `Boolean` instead of `boolean`?

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/GetApplicationsStatusResponse.java
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.data.module.deployer.cloudfoundry;
+
+import java.util.HashMap;
+
+final class GetApplicationsStatusResponse extends HashMap<String, ApplicationStatus> {
```
mminella: Another example of inheriting from `HashMap`.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/pom.xml
```diff
@@ -1,20 +1,62 @@
...
-	<dependencies>
-		<dependency>
-			<groupId>org.springframework.cloud</groupId>
-			<artifactId>spring-cloud-data-module-deployer-spi</artifactId>
-		</dependency>
-	</dependencies>
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>spring-cloud-data-module-deployer-cloudfoundry</artifactId>
```
ilayaperumalg: Just a thought: how about having the artifactId (and this project name) as `spring-cloud-data-module-deployer-cf` which could help shortening the name a bit ;)

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/AbstractPaginatedResponse.java
```diff
@@ -0,0 +1,170 @@
...
+
+	private volatile Integer totalResults;
```
pperalta: I noticed in this class and many others the liberal usage of the `volatile` keyword. Is it required? Do we anticipate multiple threads reading and mutating individual instances of this object?

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/ApplicationStatus.java
pperalta: +1, `HashMap` should not be extended here.

Zteve:
Re: Tabs; am using spring-framework.xml from now on. Thanks.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/pom.xml
ericbottard: We don't have the habit of shortening names, so why do this here? Moreover, "cloudfoundry" is the exact part of the name that is important in this name. I think "cf" may not be obvious for outside contributors and users of the module, so I'd -1

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/AbstractPaginatedResponse.java
Zteve: Yes, please: how do we use them? (As you can see this implementation fails if there is more than one page response.)

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/AbstractPaginatedResponse.java
Zteve: I was encouraged to use `volatile` when other pieces of hidden code update/read the field. In general it does no harm to assume that for any field.  In my opinion this oughtn't to be necessary for spring/jackson accessed fields as there ought to be correct synchronisation observed in those implementations. But just in case, Ben Hale got me to put those in there.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/CloudFoundryClient.java
Zteve: These cloudfoundry REST client classes (and the Request/Response pairs) are essentially copied from the `cf-java-client` code (soon to be in v3). Eventually we ought to use that code directly, but at the moment it is hard to use it without forcing us to be a Web Application (I think). Also `cf-java-client` looks much too heavyweight for our purposes.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/CloudFoundryClient.java
Zteve: Agreed.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/CloudFoundryModuleDeploymentIdConverter.java
Zteve: Yes; `ModuleDeploymentId.parse()` checks its parameter for null and throws this; we can do it first, but we would still need to check for the syntax error.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/DeleteApplicationResourceResponse.java
Zteve: slip of the caps lock—thanks

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/DeleteApplicationResourceResponse.java
Zteve: yup

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/DeleteApplicationResponse.java
Zteve: yup

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/GetApplicationStatisticsResponse.java
Zteve: In this case the response *is* a map, and I have no information about the keys, so I can't make it a field in the response object. What do you suggest?  I would like to say `GetApplicationStatisticsResponse = HashMap<...>`, but that's not allowed in Java.  `extends` is all I have left (and it works).

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/GetApplicationsStatusResponse.java
Zteve: In this case I can avoid inheritance.  It makes the processing of the response more complicated unless I add a `withApplication` verb to do a put on the map subfield, and a `getApplications()` for getting the map out. I can do that. It makes things more consistent.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/ModuleDeployerConfiguration.java
Zteve: I can't get this scheme to work (that is, my test application fails to find a ModuleDeployer bean) so I need my hand held here.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/ModuleDeployerConfiguration.java
Zteve: Again, omitting this and the OAuth2 stuff doesn't work. What can we replace it with?

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/resources/application.yml
Zteve: I am expecting these settings to be in the manifest of the application/service that is deployed on CF because they need to be supplied from outside of the app somehow.  Most of the oauth2 parameters are set this way, so we need to set them for the entire CF app.  I put them in my own application.yml so that it is clear what is needed. The names will change, doubtless, although the oauth2 stuff cannot (I don't think).

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/resources/application.yml
Zteve: HOST is currently an environment variable just for convenience in setting the URIs. It is not used directly in the code.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/AbstractPaginatedResponse.java
Zteve: I disagree; the properties we return for a module instance (where this value surfaces) are all Strings, and I don't want any checking of this information nor conversion/reconversion from/to Strings. Besides I don't use any date-like properties of these fields. They might not even be dates, for all I know.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/AbstractPaginatedResponse.java
Zteve: Agreed.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/AbstractPaginatedResponse.java
Zteve: @author tags going in.  Javadoc will take a bit longer, but I will make a start on it. There are only a few places it is really necessary: all of the public interfaces, for example, and some of the internal classes that aren't obvious.  Most of the Request/Response pairs are self-explanatory, for example (deliberately mirroring the REST API doc).

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/AbstractPaginatedResponse.java
pperalta: > I was encouraged to use volatile when other pieces of hidden code update/read the field.

Volatile is not meant to be used when hidden code (such as serialization) is mutating fields - it is meant to be used to enforce _visibility_ of mutations across threads.

> In general it does no harm to assume that for any field.

In practice there is low overhead for mutating volatile fields - the cost isn't zero but it's pretty low:

http://stackoverflow.com/questions/4633866/is-volatile-expensive

The issue for me is that when I see `volatile` it is sending a message to me that the class was **explicitly designed** to be thread safe. 

For instance, if we take a look at `ApplicationInstanceStatus.Statistics`:

```
private volatile Usage usage;
private volatile String name;
private volatile List<String> uris;
private volatile String host;
private volatile Integer port;
private volatile Integer uptime;
private volatile Integer memoryQuota;
private volatile Integer diskQuota;
private volatile Integer fdsQuota;
```

Do we have a use case in mind where a thread will be updating a field (let's say `port`) while another thread updates another field (let's say `uptime`)? Would an instance of this class be considered "usable" if its fields are still being populated? Or is it expected that once all of the fields are populated, _that_ is when this object should be used?

I suspect it is the latter - which means having all of the fields `volatile` isn't useful if we expect that _all_ of the fields will be populated before the object is consumed (i.e. serialized as part of a request, rendered on a web page, etc).

I expect usage of instances of this class to be as follows (please correct me if I'm wrong):

1. Request is made to the admin server to obtain statistics
2. Admin server makes a request to CF to obtain statistics on the same thread that received the initial request (same thread as 1)
3. Admin server creates a new instance of `Statistics` and populates it (same thread as 1)
4. Admin server serializes the instance and sends it as part of the response (same thread as 1)

Given this usage pattern, the object will never be exposed to multiple threads.

If for some reason in the future it were exposed to multiple threads, it would be up to those threads to properly synchronize their usage of this object to make sure that one thread isn't performing a partial read (i.e. only some of the fields are populated) while another thread is updating the fields. Having the fields `volatile` doesn't help with that.

> In my opinion this oughtn't to be necessary for spring/jackson accessed fields as there ought to be correct synchronisation observed in those implementations. 

Exactly!!

> But just in case, Ben Hale got me to put those in there.

Here is where I disagree - IMO we should avoid doing things "just in case" and instead we should be _explicit_ about expected usage of our code.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/AbstractPaginatedResponse.java
mminella: See http://docs.spring.io/spring-data/rest/docs/2.4.x/reference/html/#paging-and-sorting  For the record, I'm less concerned about the implementation of how pages are constructed and more about the way they are represented.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/CloudFoundryClient.java
mminella: So is the intent that this code goes away at some point?  If so, can we get a story that outlines that in Jira and includes what we're waiting on?

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/CloudFoundryModuleDeploymentIdConverter.java
mminella: The syntax exception I'm fine with.  It's catching an NPE that feels dirty to me.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/TestApplication.java
```diff
@@ -0,0 +1,100 @@
...
+//
+//        ms = moduleDeployer.status(new ModuleDeploymentId("b", "Steve"));
+//        System.out.println("\nmoduleDeployer.status(MDI(\"b\",\"Steve\")) ==> " + ms.getState());
+//        System.out.println("\ninstance(s) ==> " + iPrettyPrint(ms.getInstances()));
+//
+//		CloudFoundryClient client = applicationContext.getBean(CloudFoundryClient.class);
+
+//		client.uploadBits(new UploadBitsRequest()
+//				.withId("b2fe9e68-7124-4489-98d0-6ea47be10a14")
+//				.withResource(new FileSystemResource("/Users/spowell/Downloads/java-main-application-1.0.0.BUILD-SNAPSHOT.jar")));
```
mminella: Can we get this commented code out?

ericbottard:
Guys, Steve and I are still heavily working on this. I marked the PR as not ready for general review for now, we'll let you know when we're ready for the big day :)

markfisher:
I think that means you should close the PR and resubmit when you're ready. I think the feedback so far has been helpful in any case.

ericbottard:
PR is a nice way for us to work with asynch feedback. And yes, feedback has definitely been useful. I just don't want you to waste time on things we obviously know (main() method, etc).

markfisher:
You can comment on any commit via Github if you're just interacting with each other at this point, but maybe this is the right middle-ground so people can keep up with the progress as mere spectators until the review ban is lifted.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/CloudFoundryModuleDeployer.java
Zteve: It returns all the application statuses for the *space*. There might be a way of paring it down if I can filter on a name pattern rather than just a name. But the api documentation is so bad I don't know how.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/AbstractPaginatedResponse.java
Zteve: I am more concerned with consuming pages of responses easily, rather than generating them. Is that easy to do?

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/CloudFoundryModuleDeploymentIdConverter.java
Zteve: Done

ericbottard:
Re-opening this PR to general discussion. This should be considered ready for merge attempt, please comment as appropriate

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/AbstractPaginatedResponse.java
```diff
@@ -0,0 +1,82 @@
...
+package org.springframework.cloud.data.module.deployer.cloudfoundry;
+
+import java.util.List;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * @author Steve Powell
+ */
+abstract class AbstractPaginatedResponse<R> {
```
mminella: Is there a reason this doesn't implement Spring Data's `Page` interface (http://docs.spring.io/autorepo/docs/spring-data-commons/current/api/org/springframework/data/domain/Page.html)?  And for that matter, why don't we just use their `PageImpl`?

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/CloudFoundryModuleDeployer.java
```diff
@@ -16,32 +16,89 @@
...
 import org.springframework.cloud.data.module.ModuleStatus;
 import org.springframework.cloud.data.module.deployer.ModuleDeployer;
 
+/**
+ * A {@link ModuleDeployer} which deploys modules as applications running in a space in CloudFoundry.
+ *
+ * @author Paul Harris
+ * @author Steve Powell
+ */
 public class CloudFoundryModuleDeployer implements ModuleDeployer {
```
mminella: I think (by looking at this) we should be naming this `AppModuleDeployer` (or something similar).  I'm assuming that we'll have two `ModuleDeployer` implementations similar to how we do with Lattice (one for processes and one for tasks) and the package/module indicates that it's for CF.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/ListServiceBindingsResponse.java
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.data.module.deployer.cloudfoundry;
+
+/**
+ * @author Eric Bottard
+ */
+class ListServiceBindingsResponse extends AbstractPaginatedResponse<ResourceResponse<ServiceBindingEntity>> {
```
mminella: Making `AbstractPaginatedResponse` abstract is creating quite a bit of cruft here in requiring the extension for all these use cases *just* to configure the generics.  Again, that should be replaced with the Spring Data components instead IMHO.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/CloudControllerRestClient.java
```diff
@@ -0,0 +1,170 @@
...
+ */
+
+package org.springframework.cloud.data.module.deployer.cloudfoundry;
+
+/**
+ * Interface to cloud controller functions which wrap the REST interface required by the {@link CloudFoundryModuleDeployer}.
+ *
+ * @author Steve Powell
+ */
+interface CloudControllerRestClient {
```
mminella: We probably should remove the "Rest" from the name here.  Does the underlying API technology matter for this interface?

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/StandardCloudControllerRestClient.java
```diff
@@ -0,0 +1,233 @@
...
+import org.springframework.util.MultiValueMap;
+import org.springframework.util.StringUtils;
+import org.springframework.web.util.UriComponentsBuilder;
+
+/**
+ * REST client specialised for module deployment requirements.
+ *
+ * @author Steve Powell
+ */
+class StandardCloudControllerRestClient implements CloudControllerRestClient {
```
mminella: I _think_ most "default" implementations of thing within the Spring world are named either "Simple*" or "Default*".  So in this case, `SimpleCloudControllerRestClient` would make sense.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/StandardCloudFoundryApplicationOperations.java
```diff
@@ -0,0 +1,233 @@
...
+
+import org.springframework.web.client.RestClientException;
+
+/**
+ * Implementation of high-level operations on applications, limited to those
+ * operations required by the {@link CloudFoundryModuleDeployer}.
+ *
+ * @author Steve Powell
+ */
+class StandardCloudFoundryApplicationOperations implements CloudFoundryApplicationOperations {
```
mminella: Same point here about the use of "Standard*"

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/AbstractPaginatedResponse.java
ericbottard: bear in mind, we play the role of the _client_ here, not the server. this is what we _get_ from CF, not something that we expose (and would benefit from spring data), unless I'm missing something

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/CloudFoundryModuleDeployer.java
ericbottard: +1

we just caught up with the process/task duality of things

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/ListServiceBindingsResponse.java
ericbottard: Again, I'm not sure I understand how this is supposed to be used. *IF* we keep this implementation (which I'd rather not do, because our goal in life is not to maintain yet another CF REST binding), then I'd definitely look into it (post S1, I guess :smile: )

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/CloudControllerRestClient.java
ericbottard: agreed

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/StandardCloudControllerRestClient.java
ericbottard: +1 for `Default` (the other client is definitely not "simple")

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/StandardCloudFoundryApplicationOperations.java
jvalkeal: I think spring convention is to use `Operations` for an interface and then its implementation is `Template`. i.e. `RestOperations` vs. `RestTemplate`.

markfisher: +1 this could be `CloudFoundryApplicationTemplate implements CloudFoundryApplicationOperations`

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/StandardCloudControllerRestClient.java
markfisher: can we go with: `CloudControllerRestTemplate implements CloudControllerRestOperations` ?

ericbottard:
Rebased, addressed review comments, added workaround for CF ENV VARS, integrated into the bigger picture of admin app, harmonized the way configurations are created

ericbottard:
Rebased on master (not using `application.yml`), tested local deploy/undeploy as well as CF deploy/undeploy

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleDeploymentRequest.java
```diff
@@ -107,6 +107,7 @@ public int getCount() {
 	 * @see #deploymentProperties
 	 */
 	public Map<String, String> getDeploymentProperties() {
+
```
pperalta: Extra line break

spring-cloud-data-admin/src/main/resources/META-INF/spring-cloud-data-admin/admin.properties
```diff
@@ -11,3 +11,18 @@ spring.yarn.client.files.0=file:${clouddata.yarn.app.appmaster.path:.}/spring-cl
 spring.yarn.client.files.1=file:${clouddata.yarn.app.container.path:.}/spring-cloud-data-yarn-container-@project.version@.jar
 spring.yarn.client.launchcontext.archiveFile=spring-cloud-data-yarn-appmaster-@project.version@.jar
 spring.yarn.client.resource.memory=1g
+
+# CF Deployer configuration
```
ilayaperumalg: All these properties need to have have `=` instead of `:` as the assignment operator. With the latest PR, this will change anyway.

ericbottard: : works just fine in properties file. Agreed for consistency though

ilayaperumalg:
Is it fine to have `spring-cloud-data-admin` running locally with the cloud config setting to target CF? If so, are there any steps (like setting environment variables etc.,) before starting the admin?
I have the admin running locally connecting to the PEZ CF environment. But the deployed modules couldn't bind to the service (for message channel binding). I see this:
```
OUT at org.springframework.boot.SpringApplication.run(SpringApplication.java:961) [spring-boot-1.3.0.BUILD-SNAPSHOT.jar!/:1.3.0.BUILD-SNAPSHOT]
Fri Sep 04 2015 16:18:43 GMT-0700 (PDT) [App/0] ERR ... 6 more
Fri Sep 04 2015 16:18:43 GMT-0700 (PDT) [App/0] ERR Caused by: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'org.springframework.cloud.stream.binder.redis.config.RedisMessageChannelBinderConfiguration': Injection of autowired dependencies failed; nested exception is org.springframework.beans.factory.BeanCreationException: Could not autowire field: private org.springframework.data.redis.connection.RedisConnectionFactory org.springframework.cloud.stream.binder.redis.config.RedisMessageChannelBinderConfiguration.redisConnectionFactory; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'redisConnectionFactory' defined in class path resource [org/springframework/cloud/stream/binder/redis/config/RedisServiceAutoConfiguration$CloudConfig.class]: Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.springframework.data.redis.connection.RedisConnectionFactory]: Factory method 'redisConnectionFactory' threw exception; nested exception is org.springframework.cloud.CloudException: No unique service matching interface org.springframework.data.redis.connection.RedisConnectionFactory found. Expected 1, found 0
Fri Sep 04 2015 16:18:43 GMT-0700 (PDT) [App/0] ERR at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:543)
Fri Sep 04 2015 16:18:43 GMT-0700 (PDT) [App/0] ERR at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:230)
Fri Sep 04 2015 16:18:43 GMT-0700 (PDT) [App/0] ERR at org.springframework.beans.factory.support.ConstructorResolver.instantiateUsingFactoryMethod(ConstructorResolver.java:368)
Fri Sep 04 2015 16:18:43 GMT-0700 (PDT) [App/0] ERR at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.instantiateUsingFactoryMethod(AbstractAutowireCapableBeanFactory.java:1123)

```

ilayaperumalg:
If the same stream is tried for re-deployment (after deleting the app from CF), I get this:
```
org.springframework.web.client.HttpClientErrorException: 400 Bad Request
	at org.springframework.web.client.DefaultResponseErrorHandler.handleError(DefaultResponseErrorHandler.java:91) ~[spring-web-4.2.1.RELEASE.jar:4.2.1.RELEASE]
	at org.springframework.security.oauth2.client.http.OAuth2ErrorHandler.handleError(OAuth2ErrorHandler.java:165) ~[spring-security-oauth2-2.0.7.RELEASE.jar:na]
	at org.springframework.web.client.RestTemplate.handleResponse(RestTemplate.java:641) ~[spring-web-4.2.1.RELEASE.jar:4.2.1.RELEASE]
	at org.springframework.web.client.RestTemplate.doExecute(RestTemplate.java:597) ~[spring-web-4.2.1.RELEASE.jar:4.2.1.RELEASE]
	at org.springframework.security.oauth2.client.OAuth2RestTemplate.doExecute(OAuth2RestTemplate.java:128) ~[spring-security-oauth2-2.0.7.RELEASE.jar:na]
	at org.springframework.web.client.RestTemplate.execute(RestTemplate.java:572) ~[spring-web-4.2.1.RELEASE.jar:4.2.1.RELEASE]
	at org.springframework.web.client.RestTemplate.postForObject(RestTemplate.java:375) ~[spring-web-4.2.1.RELEASE.jar:4.2.1.RELEASE]
	at org.springframework.cloud.data.module.deployer.cloudfoundry.CloudControllerTemplate.createRoute(CloudControllerTemplate.java:71) ~[classes/:na]
	at org.springframework.cloud.data.module.deployer.cloudfoundry.CloudFoundryApplicationTemplate.pushBindAndStartApplication(CloudFoundryApplicationTemplate.java:177) ~[classes/:na]
	at org.springframework.cloud.data.module.deployer.cloudfoundry.ApplicationModuleDeployer.deploy(ApplicationModuleDeployer.java:58) ~[classes/:na]
	at org.springframework.cloud.data.admin.controller.StreamController.deployStream(StreamController.java:241) ~[classes/:na]
	at org.springframework.cloud.data.admin.controller.StreamController.save(StreamController.java:147) ~[classes/:na]
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[na:1.8.0_20]
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[na:1.8.0_20]
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[na:1.8.0_20]
	at java.lang.reflect.Method.invoke(Method.java:483) ~[na:1.8.0_20]
```

ilayaperumalg:
After I manually `bind` the CF redis service to the deployed modules and `restaged` the apps, the channel binding occurs and stream is happy. I am not sure if this issue is only on `PEZ` environment.

spring-cloud-data-admin/src/main/resources/META-INF/spring-cloud-data-admin/admin.properties
```diff
@@ -11,3 +11,18 @@ spring.yarn.client.files.0=file:${clouddata.yarn.app.appmaster.path:.}/spring-cl
...
+#cloudfoundry.apiEndpoint:
+#cloudfoundry.organization:
+#cloudfoundry.space:
+#cloudfoundry.domain:
```
ilayaperumalg: We need to document `cloudfoundry.services`

ilayaperumalg:
ok, After setting `cloudfoundry.services` to use my redis service name, the modules are deployed fine with automating binding to the service. By default, it [points] (https://github.com/spring-cloud/spring-cloud-data/pull/60/files#diff-36d3c7ff6d2cdf6adfaa353c48c43419R38) to `redis`.

ilayaperumalg:
Once the automatic binding happens with the appropriate service name setting, the `400 bad gateway` error doesn't occur.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/README.adoc
```diff
@@ -0,0 +1,25 @@
+SPI implementation for deploying Spring Cloud Stream modules to http://cloudfoundry.org[Cloud Foundry]
+
+The external dependencies are environment variables declared (but not all
+defined) in `resources/application.yml`:
```
ilayaperumalg: No more application.yml now, will fix while merging.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/CloudControllerOperations.java
```diff
@@ -0,0 +1,170 @@
...
+
+/**
+ * Interface to cloud controller functions which wrap the REST interface required by the {@link ApplicationModuleDeployer}.
+ *
+ * @author Steve Powell
+ */
+interface CloudControllerOperations {
+
+	/**
+	 * Check that a given route exists.
```
ilayaperumalg: Is this Javadoc correct? I believe this can say `lists all the routes`.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/CloudControllerTemplate.java
```diff
@@ -0,0 +1,235 @@
...
+
+	CloudControllerTemplate(URI endpoint, ExtendedOAuth2RestOperations restOperations) {
+		this.endpoint = endpoint;
+		this.restOperations = restOperations;
+	}
+
+	@Override
+	public ListRoutesResponse listRoutes(ListRoutesRequest request) {
+		URI uri = UriComponentsBuilder.fromUri(this.endpoint)
+				.pathSegment("v2", "routes")
```
ilayaperumalg: We can possibly create a constant for `v2` as we know we are using CloudController v2 REST API. And, may be we can have this constant at `CloudControllerOperations` ?

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/CloudControllerTemplate.java
```diff
@@ -0,0 +1,235 @@
...
+
+		this.restOperations.delete(uri);
+		return new RouteMappingResponse();
+	}
+
+	@Override
+	public UploadBitsResponse uploadBits(UploadBitsRequest request) {
+		URI uri = UriComponentsBuilder.fromUri(this.endpoint)
+				.pathSegment("v2", "apps", request.getId(), "bits")
+				.queryParam("async", false)
```
ilayaperumalg: REST API specifies `async` default value is `false`. Hence, this may not be needed.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/CloudFoundryModuleDeploymentConverter.java
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * Converts between Module information and Application information.
+ * Sole owner of correspondence between module ids and application names; also builds environment for the module launcher application.
+ *
+ * @author Steve Powell
+ */
+class CloudFoundryModuleDeploymentConverter {
+
+	//TODO: validate the prefixing strategy
+	private static final String APPLICATION_PREFIX = "spring_cloud_data_";
```
ilayaperumalg: @markfisher Possible candidate for renaming when you rename the repo.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/ModuleStatusBuilder.java
```diff
@@ -0,0 +1,98 @@
...
+		ApplicationInstanceStatus.Statistics statistics = applicationInstanceStatus.getStatistics();
+
+		CloudFoundryModuleInstanceStatus status = new CloudFoundryModuleInstanceStatus()
+				.withId(id)
+				.withState(convertState(applicationInstanceStatus.getState()));
+
+		if (statistics != null) {
+			status.withAttribute("usage.time", statistics.getUsage().getTime())
+					.withAttribute("usage.cpu", statistics.getUsage().getCpu().toString())
+					.withAttribute("usage.time", statistics.getUsage().getDisk().toString())
```
ilayaperumalg: I believe this attribute needs to change to `usage.disk`.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/ModuleStatusBuilder.java
```diff
@@ -0,0 +1,98 @@
...
+					.withAttribute("usage.time", statistics.getUsage().getMemory().toString())
```
ilayaperumalg: This attribute needs to be `usage.memory`

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/data/module/deployer/cloudfoundry/CloudControllerTemplate.java
ilayaperumalg: I will leave this `as-is` for better readability.

spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/config/CloudConfiguration.java
```diff
@@ -57,7 +56,6 @@ RedisConnectionFactory redisConnectionFactory(Cloud cloud) {
 	@Profile("lattice")
 	protected static class LatticeConfig {
 
-		@Bean
```
ilayaperumalg: I think this is just a slip. This is a required bean when running on `lattice`.

ilayaperumalg: will fix while merging.

ilayaperumalg:
LGTM; Squashed, rebased and merged f506701182928239c81b7575a255e0b73a9035d6 along with polishing/fixing 4417c049522b53f8d43d9e06351f1392805d4833, 12e2afd3c4de467ecadc12da7001fe06c351e6b1

PR #61, opened 2015-08-24
Add `spring-cloud-lattice` into dependencyManagement

ilayaperumalg:
- This way the `spring-cloud-lattice-core`, `spring-cloud-lattice-connector` and
`receptor-client` versions can be management centrally


ilayaperumalg:
Until `spring-cloud-lattice` is added `spring-cloud-starters`, we can have this fix that manages all the `spring-cloud-lattice` dependencies centrally.

PR #62, opened 2015-08-25
Use spring-cloud-starters dependencies

ilayaperumalg:
- This will use the version defined from `spring-cloud-starters` for `spring-cloud-stream` and `spring-cloud-lattice`. Both are set to use `1.0.0.BUILD-SNAPSHOT`


markfisher:
rebased and merged as https://github.com/spring-cloud/spring-cloud-data/commit/341d15bdb57bf2f60314a6ca1c4c74d41e996f3e

PR #63, opened 2015-08-25
Use mocks in shell tests

ilayaperumalg:
- Create mock for `ModuleDeployer` and verify the interactions/return value on that mock


pperalta:
After some discussion with @markfisher we've decided to close this PR for now. Instead of requiring a running application, the shell tests should only mock the layer that they directly rely upon (i.e. `StreamOperations`, etc).

See this document for more details: https://docs.google.com/document/d/18uNqRAgVGO0BHdvDsVg3X78gDBeqXA_LN_C6jJ0YpKw/edit#

PR #64, opened 2015-08-25
XD-3430

markfisher:


spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/controller/StreamController.java
```diff
@@ -200,18 +205,28 @@ public void deploy(@PathVariable("name") String name, @RequestParam(required = f
...
 		Iterator<ModuleDefinition> iterator = stream.getDeploymentOrderIterator();
 		for (int i = 0; iterator.hasNext(); i++) {
 			ModuleDefinition module = iterator.next();
 			String type = (i == 0) ? "sink"
 					: (iterator.hasNext() ? "processor" : "source");
 			ModuleCoordinates coordinates = this.registry.findByNameAndType(module.getName(), type);
-			// todo: pass deployment properties
-			this.deployer.deploy(new ModuleDeploymentRequest(module, coordinates));
+			Map<String, String> moduleDeploymentProperties = new HashMap<>();
+			String modulePrefix = String.format("module.%s.", module.getLabel());
```
ilayaperumalg: I hope the prefix `module.*.` that applies the property to all modules isn't the scope of this PR.

markfisher: good point! we can add that here as well... should update this PR

spring-cloud-data-core/src/main/java/org/springframework/cloud/data/core/ModuleDeploymentRequest.java
```diff
@@ -57,13 +57,15 @@
...
-			int count, Map<String, String> deploymentProperties) {
+			Map<String, String> deploymentProperties) {
 		this.definition = definition;
 		this.coordinates = coordinates;
-		this.count = count;
 		this.deploymentProperties = deploymentProperties == null
 				? Collections.<String, String>emptyMap()
 				: Collections.unmodifiableMap(deploymentProperties);
+		this.count = this.deploymentProperties.containsKey("count")
+				? Integer.parseInt(this.deploymentProperties.get("count"))
```
pperalta: Should `"count"` be a constant somewhere as part of the SPI?

ilayaperumalg:
LGTM, merging. 

TODO:// to specify `count` property as constance in SPI (possibly in ModuleDeployer interface or a common DeploymentProperties class).

ilayaperumalg:
Merged fd440a765707e917d00f80d6fbde682fa9b6e267 along with the change to support `wildcard` prefix for deployment properties: 5779a094485f5e48ed2e158595909bc33e33e80d

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-local/src/main/java/org/springframework/cloud/data/module/deployer/local/LocalModuleDeployer.java
```diff
@@ -57,10 +57,17 @@ public LocalModuleDeployer(ModuleLauncher launcher) {
 	@Override
 	public ModuleDeploymentId deploy(ModuleDeploymentRequest request) {
 		String module = request.getCoordinates().toString();
+		if (request.getCount() != 1) {
+			logger.warn(String.format("%s only supports a single instance per module; ignoring count=%d for %s",
```
ericbottard: Why is that really? I get it because of _e.g._ port conflicts (and someone convinced me that having 8080 as a default was a good idea. And I agreed :) ), but apart from that there is really nothing stopping us from having several instances in the same process, is there? I mean, the binder would bind correctly, etc

markfisher: There wouldn't be port conflicts since ports are allocated. This was primarily to avoid the refactoring that would be necessary for launching multiple instance per module here *at least for now*, given that in XD 1.0 we never ran more than one module instance in the same process. We can definitely revisit.

ericbottard: Good to know.
Note that ports are allocated **for (http) `server.port`** only. An FTP listening module wouldn't get that special treatment. Same issue with file paths, etc

markfisher: Yes, and that's one more reason not to go down the path of multiple module instances for *local* deployment yet (if at all). In a containerized/cloud deployment ports are mapped.

PR #66, opened 2015-08-27
Introduce a helper class for formatting ModuleLauncher args

ericbottard:
This is a companion PR for https://github.com/spring-cloud/spring-cloud-stream/pull/110

### DIFF
--- a/spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-lattice/src/main/java/org/springframework/cloud/data/module/deployer/lattice/LrpModuleDeployer.java
+++ b/spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-lattice/src/main/java/org/springframework/cloud/data/module/deployer/lattice/LrpModuleDeployer.java
@@ -70,8 +70,8 @@ public class LrpModuleDeployer implements ModuleDeployer {
 		Map<String, String> rawArgs = new HashMap<>();
 		rawArgs.putAll(request.getDefinition().getParameters());
 		rawArgs.putAll(request.getDeploymentProperties());
-		Map<String, String> transformed = ModuleLauncherArgumentsHelper.qualifyArgs(0, rawArgs);
-		for (Map.Entry<String, String> entry : transformed.entrySet()) {
+		Map<String, String> qualifiedArgs = ModuleLauncherArgumentsHelper.qualifyArgs(0, rawArgs);
+		for (Map.Entry<String, String> entry : qualifiedArgs.entrySet()) {
 			environmentVariables.add(new EnvironmentVariable(entry.getKey(), entry.getValue()));
 		}
 		lrp.setEnv(environmentVariables.toArray(new EnvironmentVariable[environmentVariables.size()]));
--- a/spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-lattice/src/main/java/org/springframework/cloud/data/module/deployer/lattice/LrpModuleDeployer.java
+++ b/spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-lattice/src/main/java/org/springframework/cloud/data/module/deployer/lattice/LrpModuleDeployer.java
@@ -33,6 +33,7 @@ import org.springframework.cloud.data.core.ModuleDeploymentId;
 import org.springframework.cloud.data.core.ModuleDeploymentRequest;
 import org.springframework.cloud.data.module.ModuleStatus;
 import org.springframework.cloud.data.module.deployer.ModuleDeployer;
+import org.springframework.cloud.data.module.deployer.ModuleLauncherArgumentsHelper;
 import org.springframework.util.StringUtils;
 
 /**
@@ -66,10 +67,11 @@ public class LrpModuleDeployer implements ModuleDeployer {
 		List<EnvironmentVariable> environmentVariables = new ArrayList<EnvironmentVariable>();
 		Collections.addAll(environmentVariables, lrp.getEnv());
 		environmentVariables.add(new EnvironmentVariable("MODULES", request.getCoordinates().toString()));
-		for (Map.Entry<String, String> entry : request.getDefinition().getParameters().entrySet()) {
-			environmentVariables.add(new EnvironmentVariable(entry.getKey(), entry.getValue()));
-		}
-		for (Map.Entry<String, String> entry: request.getDeploymentProperties().entrySet()) {
+		Map<String, String> rawArgs = new HashMap<>();
+		rawArgs.putAll(request.getDefinition().getParameters());
+		rawArgs.putAll(request.getDeploymentProperties());
+		Map<String, String> transformed = ModuleLauncherArgumentsHelper.qualifyArgs(0, rawArgs);
+		for (Map.Entry<String, String> entry : transformed.entrySet()) {
 			environmentVariables.add(new EnvironmentVariable(entry.getKey(), entry.getValue()));
 		}
 		lrp.setEnv(environmentVariables.toArray(new EnvironmentVariable[environmentVariables.size()]));
--- a/spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-lattice/src/main/java/org/springframework/cloud/data/module/deployer/lattice/LrpModuleDeployer.java
+++ b/spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-lattice/src/main/java/org/springframework/cloud/data/module/deployer/lattice/LrpModuleDeployer.java
@@ -33,7 +33,7 @@ import org.springframework.cloud.data.core.ModuleDeploymentId;
 import org.springframework.cloud.data.core.ModuleDeploymentRequest;
 import org.springframework.cloud.data.module.ModuleStatus;
 import org.springframework.cloud.data.module.deployer.ModuleDeployer;
-import org.springframework.cloud.data.module.deployer.ModuleLauncherArgumentsHelper;
+import org.springframework.cloud.data.module.deployer.ModuleArgumentQualifier;
 import org.springframework.util.StringUtils;
 
 /**
@@ -70,7 +70,7 @@ public class LrpModuleDeployer implements ModuleDeployer {
 		Map<String, String> rawArgs = new HashMap<>();
 		rawArgs.putAll(request.getDefinition().getParameters());
 		rawArgs.putAll(request.getDeploymentProperties());
-		Map<String, String> qualifiedArgs = ModuleLauncherArgumentsHelper.qualifyArgs(0, rawArgs);
+		Map<String, String> qualifiedArgs = ModuleArgumentQualifier.qualifyArgs(0, rawArgs);
 		for (Map.Entry<String, String> entry : qualifiedArgs.entrySet()) {
 			environmentVariables.add(new EnvironmentVariable(entry.getKey(), entry.getValue()));
 		}
--- a/spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-local/src/main/java/org/springframework/cloud/data/module/deployer/local/LocalModuleDeployer.java
+++ b/spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-local/src/main/java/org/springframework/cloud/data/module/deployer/local/LocalModuleDeployer.java
@@ -19,6 +19,7 @@ package org.springframework.cloud.data.module.deployer.local;
 import java.net.Inet4Address;
 import java.net.URL;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -30,14 +31,18 @@ import org.springframework.cloud.data.core.ModuleDeploymentId;
 import org.springframework.cloud.data.core.ModuleDeploymentRequest;
 import org.springframework.cloud.data.module.ModuleStatus;
 import org.springframework.cloud.data.module.deployer.ModuleDeployer;
+import org.springframework.cloud.stream.module.launcher.ModuleLaunchRequest;
 import org.springframework.cloud.stream.module.launcher.ModuleLauncher;
 import org.springframework.util.Assert;
 import org.springframework.util.SocketUtils;
 import org.springframework.web.client.RestTemplate;
 
 /**
+ * A ModuleDeployer that will launch a module in-process.
+ *
  * @author Mark Fisher
  * @author Marius Bogoevici
+ * @author Eric Bottard
  */
 public class LocalModuleDeployer implements ModuleDeployer {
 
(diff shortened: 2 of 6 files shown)


ericbottard:
Did not use in the YARN case as it seems currently the deployment properties and module arguments are not passed anyway (they're just [logged][1]) cc @jvalkeal.

It may actually not be necessary to do namespacing anyway, as the yarn app uses the launcher directly. TBC.

[1]: https://github.com/spring-cloud/spring-cloud-data/blob/1b9f3d2cd6a5fe1272384c300ca5bc7621483fc6/spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-yarn/src/main/java/org/springframework/cloud/data/module/deployer/yarn/YarnModuleDeployer.java#L72

ericbottard:
Also, it seems in the Lattice case that the previous namespacing strategy (`mavenCoordinates.key=value`) was not being used? cc @markfisher (see https://github.com/spring-cloud/spring-cloud-data/blob/1b9f3d2cd6a5fe1272384c300ca5bc7621483fc6/spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-lattice/src/main/java/org/springframework/cloud/data/module/deployer/lattice/LrpModuleDeployer.java#L69)

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-lattice/src/main/java/org/springframework/cloud/data/module/deployer/lattice/LrpModuleDeployer.java
```diff
@@ -66,10 +67,11 @@ public ModuleDeploymentId deploy(ModuleDeploymentRequest request) {
...
 		Collections.addAll(environmentVariables, lrp.getEnv());
 		environmentVariables.add(new EnvironmentVariable("MODULES", request.getCoordinates().toString()));
-		for (Map.Entry<String, String> entry : request.getDefinition().getParameters().entrySet()) {
-			environmentVariables.add(new EnvironmentVariable(entry.getKey(), entry.getValue()));
-		}
-		for (Map.Entry<String, String> entry: request.getDeploymentProperties().entrySet()) {
+		Map<String, String> rawArgs = new HashMap<>();
+		rawArgs.putAll(request.getDefinition().getParameters());
+		rawArgs.putAll(request.getDeploymentProperties());
+		Map<String, String> transformed = ModuleLauncherArgumentsHelper.qualifyArgs(0, rawArgs);
```
markfisher: change "transformed" to "qualifiedArgs"? (the code is easier to read when terminology is consistent, rather than multiple terms being introduced for the same thing)

markfisher:
The namespacing is not being used in `LrpModuleDeployer

ericbottard:
> The namespacing is not being used in `LrpModuleDeployer`

is that intentional? This eventually launches the same ModuleLauncher, correct?

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-spi/src/main/java/org/springframework/cloud/data/module/deployer/ModuleLauncherArgumentsHelper.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.Map;
+
+/**
+ * A helper class to generate qualified environmental properties from a
+ * {@link org.springframework.cloud.data.core.ModuleDeploymentRequest}, in a format that is expected by
+ * ModuleLauncher and ModuleLauncherProperties.
+ *
+ * @author Eric Bottard
+ */
+public class ModuleLauncherArgumentsHelper {
```
markfisher: I assume you intended to make this `final`, given the private constructor.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-spi/src/main/java/org/springframework/cloud/data/module/deployer/ModuleLauncherArgumentsHelper.java
```diff
@@ -0,0 +1,54 @@
...
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.data.module.deployer;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * A helper class to generate qualified environmental properties from a
```
markfisher: I'm not a fan of using "Helper" in the classname since it's not descriptive and often redundant. For example, could we change this to `ModuleLauncherArgumentQualifier`? (note: singular for "Argument")

On that note, should we consider adding a `getQualifiedArgs()` method to ModuleDeploymentRequest *instead* of introducing this class?

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-local/src/main/java/org/springframework/cloud/data/module/deployer/local/LocalModuleDeployer.java
```diff
@@ -77,9 +79,11 @@ public ModuleDeploymentId deploy(ModuleDeploymentRequest request) {
 		catch (Exception e) {
 			throw new IllegalStateException("failed to determine URL for module: " + module, e);
 		}
-		args.add(String.format("--%s.server.port=%d", module, port));
-		args.add(String.format("--%s.endpoints.shutdown.enabled=true", module));
-		launcher.launch(new String[] { module }, args.toArray(new String[args.size()]));
+		args.put("server.port", String.valueOf(port));
```
markfisher: these 2 args need to be qualified also

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-spi/src/main/java/org/springframework/cloud/data/module/deployer/ModuleLauncherArgumentsHelper.java
ericbottard: That would tie ModuleDeploymentRequest to the launcher IMO

markfisher: In fact, I'd change to `ModuleArgumentQualifier`. There's no reason for the SPI implementations to know that there is a component called `ModuleLauncher

ericbottard: > (note: singular for "Argument")

This deals with arguments as a Map, hence the plural, don't you agree?

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-local/src/main/java/org/springframework/cloud/data/module/deployer/local/LocalModuleDeployer.java
ericbottard: When dealing with a ModuleLaunchRequest directly, which is what this deployer does, there is no qualification.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-local/src/main/java/org/springframework/cloud/data/module/deployer/local/LocalModuleDeployer.java
```diff
@@ -77,9 +79,11 @@ public ModuleDeploymentId deploy(ModuleDeploymentRequest request) {
...
+		args.put("endpoints.shutdown.enabled", "true");
+
+		ModuleLaunchRequest moduleLaunchRequest = new ModuleLaunchRequest(module);
```
markfisher: The `args` are not even passed.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-local/src/main/java/org/springframework/cloud/data/module/deployer/local/LocalModuleDeployer.java
ericbottard: To clarify, qualification is needed from (outside world) -> ModuleLauchRequest (which is what ModuleLauncherProperties is about). Here, we're in the same process.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-spi/src/main/java/org/springframework/cloud/data/module/deployer/ModuleLauncherArgumentsHelper.java
markfisher: re: tying `ModuleDeploymentRequest` to launcher, I agree (and take back that suggestion). The request object should not need to know about the larger context in which multiple requests are being used. So, yes, let's keep it "above" that level.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-local/src/main/java/org/springframework/cloud/data/module/deployer/local/LocalModuleDeployer.java
ericbottard: _That_, is a good point. And I thought we had a potential timing bug with regard to port allocation when I tested. sigh.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-spi/src/main/java/org/springframework/cloud/data/module/deployer/ModuleLauncherArgumentsHelper.java
markfisher: re: dealing with Map. I think the Map as input and Map as output is clear enough, and what this actually does is qualify *each* argument within the input Map.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-local/src/main/java/org/springframework/cloud/data/module/deployer/local/LocalModuleDeployer.java
markfisher: Yes. The problem with `findAvailableTcpPort` is that it will always have a race condition (between the time of allocation and the time of use). That can never be completely avoided given that we are competing with other processes as well. One thing we can do to avoid a race-condition with ourselves though is to call `findAvailableTcpPort*s*(numModules)`.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-local/src/main/java/org/springframework/cloud/data/module/deployer/local/LocalModuleDeployer.java
markfisher: Sorry, I realize that now. We are constructing a dedicated launch request instance per module in this case. In fact I believe this was the original rationale for creating that class.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-local/src/main/java/org/springframework/cloud/data/module/deployer/local/LocalModuleDeployer.java
ericbottard: Yeah. Note this was a force alarm, as I forgot to pass args, duh! What we did in XD was to wrap that class with something that remembers the last ports allocated (used as the seed) through a static field. But here we should be fine, as a launch only returns when the module has started

ericbottard:
> The ModuleLaucher does not unqualify environment variables

It will now with the companion PR to this (considers the whole Environment as a source of arguments).

But more importantly, I don't understand why using ENV VARS instead of `--foo=bar` args allowed it to work in the Lrp case (or any case, just thanks to ENV VARS). If I deployed `tcp --port=1234 | rabbit --port=5678`, and those two were passed as one (_i.e._ `MODULES=tcp,rabbit` (minding the coordinates, you get the idea)) then there *was* a conflict already. No need to resort to say, 2 instances of the _same_ module.

markfisher:
RE: my comment about the `LrpModuleDeployer` and its use of env vars...

It's not actually a problem since the `ModuleLauncherProperties` itself is expecting the values in any way supported by boot (including env vars). So, the "MODULES" env var as well as the "ARGS.*" env vars are targeting the `ModuleLauncherProperties`, NOT individual modules. Individual modules will no longer have properties bound directly from env vars, but rather they will be passed as *args* by the `ModuleLauncher`.

ericbottard:
Exactly. So I think we're saying the same thing. My change *is* necessary

markfisher:
Yes, and your point about conflict is correct. Any matching module property name would be a conflict. These problems would have been apparent as soon as we started passing requests for more than one module per-process, but we hadn't done that yet (waiting on the direct-binding support).

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-local/src/main/java/org/springframework/cloud/data/module/deployer/local/LocalModuleDeployer.java
```diff
@@ -77,9 +79,12 @@ public ModuleDeploymentId deploy(ModuleDeploymentRequest request) {
...
+		moduleLaunchRequest.setArguments(args);
```
markfisher: I think we should make the `ModuleLaunchRequest` accept the arguments as a 2nd constructor arg. We can remove the setter, but keep the *add* method. That would make it less likely to accidentally forget passing args as you did here at first.

ericbottard: Good point

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-spi/src/main/java/org/springframework/cloud/data/module/deployer/ModuleArgumentQualifier.java
```diff
@@ -0,0 +1,54 @@
...
+	 * Return a qualified version of the given arguments, suitable for passing to the ModuleLauncher, in the context
+	 * of launching a module appearing at position {@literal index} in the list of modules to be launched.
+	 *
+	 * <p>No other operation is performed on the keys. Callers may want to <i>e.g.</i> turn the result to uppercase and
+	 * replace dots with underscores if the arguments are to be passed <i>via</i> environment variables.
+	 * </p>
+	 *
+	 * @param index position of the module the arguments pertain to
+	 * @param unqualified a raw map of key value pairs to be passed as arguments
+	 * @return a map of key value pairs where each ket has been transformed, in "dot" form
```
markfisher: typo: ket

markfisher:
squashed and merged as https://github.com/spring-cloud/spring-cloud-data/commit/980e4d20e9a3e3a49bf7571e40865ce95a3bbe1a

PR #67, opened 2015-08-27
XD-3423 Basic task commands available for SCD

cppwfs:


spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/TaskOperations.java
```diff
@@ -0,0 +1,51 @@
...
+
+	/**
+	 * List tasks known to the system.
+	 */
+	public PagedResources<TaskDefinitionResource> list(/* TODO */);
+
+	/**
+	 * Create a new task.
+	 */
+	public TaskDefinitionResource createTask(String name, String definition);
```
markfisher: I just verified the same inconsistency exists for `StreamOperations`, but I think this should just be "create" (none of the other operations include the word "task".

spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/TaskTemplate.java
```diff
@@ -0,0 +1,93 @@
...
+	private final UriTemplate deploymentsPath;
+	
+	private final UriTemplate defaultPath;
+
+	TaskTemplate(RestTemplate restTemplate, Map<String, UriTemplate> resources) {
+		Assert.notNull(resources, "URI Resources can't be null");
+		Assert.notNull(resources.get(DEFINITIONS_PATH), "Definitions path is required");
+		Assert.notNull(resources.get(DEPLOYMENTS_PATH), "Deployments path is required");
+		Assert.notNull(resources.get(DEFAULT_PATH), "Default path is required");
+		this.restTemplate = restTemplate;
```
markfisher: should assert the RestTemplate is not null

spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/TaskTemplate.java
```diff
@@ -0,0 +1,93 @@
...
+	private final RestTemplate restTemplate;
+
+	private final UriTemplate definitionsPath;
+
+	private final UriTemplate deploymentsPath;
+	
+	private final UriTemplate defaultPath;
+
+	TaskTemplate(RestTemplate restTemplate, Map<String, UriTemplate> resources) {
+		Assert.notNull(resources, "URI Resources can't be null");
```
markfisher: typically we use the terminology: "X must not be null"

spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/TaskTemplate.java
```diff
@@ -0,0 +1,93 @@
...
+		return restTemplate.getForObject(uriTemplate, TaskDefinitionResource.Page.class);
+	}
+
+	@Override
+	public TaskDefinitionResource createTask(String name, String definition) {
+		MultiValueMap<String, Object> values = new LinkedMultiValueMap<String, Object>();
+		values.add("name", name);
+		values.add("definition", definition);
+		TaskDefinitionResource task = restTemplate.postForObject(
+				defaultPath.expand(), values, TaskDefinitionResource.class);
```
markfisher: The `StreamTemplate` uses `definitionsPath` here (it doesn't use a `defaultPath` for anything).

spring-cloud-data-rest-resource/src/main/java/org/springframework/cloud/data/rest/resource/TaskDefinitionResource.java
```diff
@@ -30,6 +31,12 @@
 
 	private String dslText;
 
+	/**
+	 * Stream status (i.e. running, complete, etc).
+	 */
+	private String status = "";
```
markfisher: for `StreamDefinitionResource` the status is null by default instead of empty string (should both default to "unknown" instead? cc @pperalta)

spring-cloud-data-shell/src/main/java/org/springframework/cloud/data/shell/command/TaskCommands.java
```diff
@@ -0,0 +1,147 @@
...
+import org.springframework.stereotype.Component;
+
+/**
+ * Task commands.
+ *
+ * @author Glenn Renfro
+ */
+@Component
+// todo: reenable optionContext attributes
+	public class TaskCommands implements CommandMarker {
```
markfisher: class declaration should not be indented

spring-cloud-data-shell/src/main/java/org/springframework/cloud/data/shell/command/TaskCommands.java
```diff
@@ -0,0 +1,147 @@
...
+					.addValue(3, task.getStatus());
+		}
+
+		return table;
+	}
+
+	@CliCommand(value = CREATE_TASK, help = "Create a new task definition")
+	public String createTask(
+			@CliOption(mandatory = true, key = { "", "name" }, help = "the name to give to the task") String name,
+			@CliOption(mandatory = true, key = { "definition" }, help = "a task definition, using the DSL (e.g. \"http --port=9000 | hdfs\")") String dsl){
```
markfisher: the definition shown in the example is for a stream, not a task

spring-cloud-data-shell/src/main/java/org/springframework/cloud/data/shell/command/TaskCommands.java
```diff
@@ -0,0 +1,147 @@
...
+		this.taskOperations().createTask(name, dsl);
+		return String.format(
+				"Created new task '%s'", name);
+	}
+
+	@CliCommand(value = LAUNCH_TASK, help = "Launch a previously created task")
+	public String launchTask(
+			@CliOption(key = { "", "name" }, help = "the name of the launch to launch", mandatory = true) String name,
```
markfisher: "launch to launch" ... should be "task to launch" correct?

spring-cloud-data-shell/src/test/java/org/springframework/cloud/data/shell/AbstractShellIntegrationTest.java
```diff
@@ -24,14 +24,14 @@
 import org.junit.rules.TestName;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
```
markfisher: preserve this newline... updated your organize imports prefs to treat "org.springframework" as a specific grouping

spring-cloud-data-shell/src/test/java/org/springframework/cloud/data/shell/command/TaskCommandTemplate.java
```diff
@@ -0,0 +1,127 @@
+/*
+ * Copyright 2002-2015 the original author or authors.
```
markfisher: I think "2015" is sufficient here

spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/CloudDataTemplate.java
```diff
@@ -51,7 +52,12 @@
 	 */
 	private final CounterTemplate counterTemplate;
 
+	/**
+	 * REST client template for task operations.
+	 */
+	private final TaskTemplate taskTemplate;
```
mminella: Is there a reason we aren't coding to the `*Operations` interfaces here?

spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/TaskTemplate.java
```diff
@@ -0,0 +1,93 @@
...
+	@Override
+	public TaskDefinitionResource.Page list() {
+		String uriTemplate = definitionsPath.toString();
+		uriTemplate = uriTemplate + "?size=10000";
+		return restTemplate.getForObject(uriTemplate, TaskDefinitionResource.Page.class);
+	}
+
+	@Override
+	public TaskDefinitionResource createTask(String name, String definition) {
+		MultiValueMap<String, Object> values = new LinkedMultiValueMap<String, Object>();
```
mminella: I know the StreamTemplate uses the `MultiValueMap` as well...I'd be interested to understand why given the additional functionality above a regular `Map` doesn't seem to be used...

spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/TaskTemplate.java
```diff
@@ -0,0 +1,93 @@
...
+		values.add("definition", definition);
+		TaskDefinitionResource task = restTemplate.postForObject(
+				defaultPath.expand(), values, TaskDefinitionResource.class);
+		return task;
+	}
+
+	@Override
+	public void launch(String name, Map<String, String> properties) {
+		String uriTemplate = deploymentsPath.toString() + "/{name}";
+		MultiValueMap<String, Object> values = new LinkedMultiValueMap<String, Object>();
```
mminella: See previous comment about the `MultiValuedMap`...

spring-cloud-data-rest-resource/src/main/java/org/springframework/cloud/data/rest/resource/TaskDefinitionResource.java
mminella: +1 for this never being null or empty.  Even though we're not using an enum, we should still use something.  In Spring Batch, we initialize  `JobExecution#status` to `BatchStatus#STARTING` since it's the first possible state...

spring-cloud-data-shell/src/main/java/org/springframework/cloud/data/shell/command/TaskCommands.java
```diff
@@ -0,0 +1,147 @@
...
+	@Autowired
+	private UserInput userInput;
+
+	@CliAvailabilityIndicator({ LIST_TASK, CREATE_TASK, LAUNCH_TASK, STATUS_TASK, DESTROY_TASK})
+	public boolean available() {
+		return cloudDataShell.getCloudDataOperations() != null;
+	}
+
+	@CliCommand(value = LIST_TASK, help = "List created tasks")
+	public Table listTasks() {
```
mminella: I feel like the comment @markfisher made in the `TaskOperations` should be applied here.  We already know these commands are for tasks since it's on the `TaskCommands` class.  We should remove the word `task` from each of these commands.

spring-cloud-data-rest-resource/src/main/java/org/springframework/cloud/data/rest/resource/TaskDefinitionResource.java
cppwfs: So here is the question.  What is the state of a task that has not been run?  I assumed it was blank or we create a State of "Ready".  Then when run, then it would change to the appropriate state.

spring-cloud-data-shell/src/main/java/org/springframework/cloud/data/shell/command/TaskCommands.java
cppwfs: doh!

spring-cloud-data-shell/src/main/java/org/springframework/cloud/data/shell/command/TaskCommands.java
cppwfs: agreed

spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/TaskTemplate.java
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * @author Glenn Renfro
+ */
+public class TaskTemplate implements TaskOperations {
+
+	private static final String DEFINITIONS_PATH = "tasks/definitions";
+
+	private static final String DEPLOYMENTS_PATH = "tasks/deployments";
+
+	private static final String DEFAULT_PATH = "tasks/";
```
ericbottard: How is that "default"? 

The uri template /tasks/{name} should not be used. Should be /tasks/definitions/{name} (to be changed at controller side as well)

Moreover, those things are not "paths", they are "relations".
The things that are looked up in the map could be named paths, though.

spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/TaskTemplate.java
ericbottard: This is because this is http posting, which supports the equivalent of `?foo=bar&foo=wizz` (same key). And so the payload needs to be of that type to be picked up by the HttpMessageConverter (namely, FormHMC)

spring-cloud-data-shell/src/main/java/org/springframework/cloud/data/shell/command/TaskCommands.java
```diff
@@ -0,0 +1,147 @@
...
+			@CliOption(mandatory = true, key = { "definition" }, help = "a task definition, using the DSL (e.g. \"http --port=9000 | hdfs\")") String dsl){
+		this.taskOperations().createTask(name, dsl);
+		return String.format(
+				"Created new task '%s'", name);
+	}
+
+	@CliCommand(value = LAUNCH_TASK, help = "Launch a previously created task")
+	public String launchTask(
+			@CliOption(key = { "", "name" }, help = "the name of the launch to launch", mandatory = true) String name,
+			@CliOption(key = { PROPERTIES_OPTION }, help = "the properties for this launch", mandatory = false) String properties,
```
ericbottard: Maybe should read "the _deployment_ properties for this launch"

spring-cloud-data-shell/src/main/java/org/springframework/cloud/data/shell/command/TaskCommands.java
```diff
@@ -0,0 +1,147 @@
...
+		return String.format("Destroyed task '%s'", name);
+	}
+
+	@CliCommand(value = STATUS_TASK, help = "Retrieve status info on an existing task")
+	public String statusTask(
+			@CliOption(key = { "", "name" }, help = "the name of the task ", mandatory = true) String name) {
+
+		return String.format("Feature Not Available");
+	}
+	TaskOperations taskOperations() {
```
ericbottard: newline

spring-cloud-data-shell/src/test/java/org/springframework/cloud/data/shell/command/TaskCommandTemplate.java
```diff
@@ -0,0 +1,127 @@
...
+		tasks.add(taskName);
+		String createMsg = "Created";
+
+		assertEquals(createMsg + " new task '" + taskName + "'", cr.getResult());
+
+		verifyExists(taskName, actualDefinition);
+	}
+
+	/**
+	 * Destroy all tasks that were created using the 'create' method. Commonly called in a @After annotated method
```
ericbottard: dot at the end

spring-cloud-data-shell/src/test/java/org/springframework/cloud/data/shell/command/TaskCommandTemplate.java
```diff
@@ -0,0 +1,127 @@
...
+			String taskname = tasks.get(s);
+			CommandResult cr = shell.executeCommand("task destroy --name " + taskname);
+			//stateVerifier.waitForDestroy(taskname);
+			assertTrue("Failure to destroy task " + taskname + ".  CommandResult = " + cr.toString(),
+					cr.isSuccess());
+		}
+	}
+
+	/**
+	 * Destroy a specific task
```
ericbottard: dot at the end

spring-cloud-data-shell/src/test/java/org/springframework/cloud/data/shell/command/TaskCommandTemplate.java
```diff
@@ -0,0 +1,127 @@
...
+	 *
+	 * @param task The task to destroy
```
ericbottard: actually, a task _name_

spring-cloud-data-shell/src/main/java/org/springframework/cloud/data/shell/command/TaskCommands.java
markfisher: -1 we are not using the word "deploy" for tasks (just "launch")

ericbottard: my point was, "the properties for.." could mean anything. Which kind of properties?

markfisher: or a Batch job these map to the `JobParameters`...  @mminella: is there a generic term that applies at the Task level?

spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/TaskTemplate.java
cppwfs: If I'm right, according to the FormHttpMessageConverter docs,
```this converter can read and write the "application/x-www-form-urlencoded" media type as MultiValueMap<String, String> and it can also write (but not read) the "multipart/form-data" media type as MultiValueMap<String, Object>.
```

spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/CloudDataTemplate.java
cppwfs: good catch

spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/TaskTemplate.java
mminella: What does that have to do with whether that functionality is being used or not?  I'm all for using that structure when it's needed, I just am missing why it's needed here (I'll admit I'm nit picking here)...

spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/CloudDataTemplate.java
markfisher: This will be very important given the testing plan @pperalta and I just started discussing... basically we will want to facilitate using mocks of the *Operations interfaces.

spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/TaskTemplate.java
ericbottard: because said HttpMessageConverter supports() is based on that payload type

spring-cloud-data-admin/src/test/java/org/springframework/cloud/data/admin/controller/TaskControllerTests.java
```diff
@@ -32,7 +32,6 @@
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.mockito.ArgumentCaptor;
-
```
markfisher: preserve this newline

spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/TaskTemplate.java
```diff
@@ -0,0 +1,89 @@
...
+	private final UriTemplate definitionsPath;
+
+	private final UriTemplate deploymentsPath;
+	
+
+	TaskTemplate(RestTemplate restTemplate, Map<String, UriTemplate> resources) {
+		Assert.notNull(resources, "URI Resources must not be be null");
+		Assert.notNull(resources.get(DEFINITIONS_PATH), "Definitions path is required");
+		Assert.notNull(resources.get(DEPLOYMENTS_PATH), "Deployments path is required");
+		Assert.notNull(restTemplate, "Rest Template must not be null");
```
markfisher: shouldn't be a space: "RestTemplate" (the classname)

spring-cloud-data-rest-resource/src/main/java/org/springframework/cloud/data/rest/resource/TaskDefinitionResource.java
```diff
@@ -30,6 +31,12 @@
 
 	private String dslText;
 
+	/**
+	 * Stream status (i.e. running, complete, etc).
+	 */
+	private String status;
+	
+
 	public TaskDefinitionResource() {
```
markfisher: is this no-arg constructor ever used? (there is no setter for name or status, so I don't see how a valid object would be created that way)

spring-cloud-data-shell/src/main/java/org/springframework/cloud/data/shell/command/TaskCommands.java
```diff
@@ -0,0 +1,148 @@
...
+	private static final String PROPERTIES_FILE_OPTION = "propertiesFile";
+
+
+	@Autowired
+	private CloudDataShell cloudDataShell;
+
+	@Autowired
+	private UserInput userInput;
+
+	@CliAvailabilityIndicator({ LIST, CREATE, LAUNCH, STATUS, DESTROY})
```
markfisher: should be a space before closing brace

spring-cloud-data-shell/src/main/java/org/springframework/cloud/data/shell/command/TaskCommands.java
```diff
@@ -0,0 +1,148 @@
...
+		return table;
+	}
+
+	@CliCommand(value = CREATE, help = "Create a new task definition")
+	public String create(
+			@CliOption(mandatory = true, key = { "", "name" }, help = "the name to give to the task") String name,
+			@CliOption(mandatory = true, key = { "definition" }, help = "a task definition, using the DSL (e.g. \"taskName\")") String dsl){
+		this.taskOperations().create(name, dsl);
+		return String.format(
+				"Created new task '%s'", name);
```
markfisher: this should be on one line

spring-cloud-data-shell/src/main/java/org/springframework/cloud/data/shell/command/TaskCommands.java
```diff
@@ -0,0 +1,148 @@
...
+
+	@CliAvailabilityIndicator({ LIST, CREATE, LAUNCH, STATUS, DESTROY})
+	public boolean available() {
+		return cloudDataShell.getCloudDataOperations() != null;
+	}
+
+	@CliCommand(value = LIST, help = "List created tasks")
+	public Table list() {
+		final PagedResources<TaskDefinitionResource> tasks = taskOperations().list();
+
```
markfisher: remove newline

spring-cloud-data-shell/src/main/java/org/springframework/cloud/data/shell/command/TaskCommands.java
```diff
@@ -0,0 +1,148 @@
...
+				.addHeader(1, new TableHeader("Task Name"))
+				.addHeader(2, new TableHeader("Task Definition"))
+				.addHeader(3, new TableHeader("Task Status"));
+		for (TaskDefinitionResource task : tasks) {
+			table.newRow()
+					.addValue(1, task.getName())
+					.addValue(2, task.getDslText())
+					.addValue(3, (task.getStatus() == null) ? "" : task.getStatus());
+		}
+
```
markfisher: remove newline

spring-cloud-data-shell/src/main/java/org/springframework/cloud/data/shell/command/TaskCommands.java
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	@CliCommand(value = STATUS, help = "Retrieve status info on an existing task")
+	public String status(
+			@CliOption(key = { "", "name" }, help = "the name of the task ", mandatory = true) String name) {
+
+		return String.format("Feature Not Available");
+	}
+	
+	TaskOperations taskOperations() {
```
markfisher: why is this not private?

spring-cloud-data-shell/src/main/java/org/springframework/cloud/data/shell/command/TaskCommands.java
```diff
@@ -0,0 +1,148 @@
...
+	public String destroy(
+			@CliOption(key = { "", "name" }, help = "the name of the task to destroy", mandatory = true) String name) {
+		taskOperations().destroy(name);
+		return String.format("Destroyed task '%s'", name);
+	}
+
+	@CliCommand(value = STATUS, help = "Retrieve status info on an existing task")
+	public String status(
+			@CliOption(key = { "", "name" }, help = "the name of the task ", mandatory = true) String name) {
+
```
markfisher: remove newline

spring-cloud-data-shell/src/test/java/org/springframework/cloud/data/shell/AbstractShellIntegrationTest.java
```diff
@@ -147,30 +149,20 @@ protected StreamCommandTemplate stream() {
...
 	 */
-	protected String generateStreamName() {
-		return generateStreamName(null);
+	protected TaskCommandTemplate task() {
+		return new TaskCommandTemplate(cloudDataShell);
 	}
-
 	// Util methods
-	private String generateUniqueName(String name) {
+	protected String generateUniqueName(String name) {
```
markfisher: since they are now protected instead of private, these methods should have JavaDoc (similar to what was deleted for the previous `generateStreamName()` method).

spring-cloud-data-shell/src/test/java/org/springframework/cloud/data/shell/command/TaskCommandTemplate.java
```diff
@@ -0,0 +1,127 @@
...
+	 */
+	public void destroyTask(String task) {
+		CommandResult cr = shell.executeCommand("task destroy --name " + task);
+		//stateVerifier.waitForDestroy(task);
+		assertTrue("Failure to destroy task " + task + ".  CommandResult = " + cr.toString(),
+				cr.isSuccess());
+		tasks.remove(task);
+	}
+
```
markfisher: extra line here

spring-cloud-data-shell/src/test/java/org/springframework/cloud/data/shell/command/TaskCommandTests.java
```diff
@@ -0,0 +1,37 @@
...
+import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.cloud.data.shell.AbstractShellIntegrationTest;
+
+/**
+ * @author Glenn Renfro
+ */
+public class TaskCommandTests extends AbstractShellIntegrationTest {
+
```
markfisher: extra line here

cppwfs: Line removed

PR #68, opened 2015-08-28
XD-3425: Shell commands for module registry

pperalta:
New classes for module registration:
 * `ModuleRegistration`
 * `ModuleRegistrationResource`

`ModuleRegistry` modified to support `ModuleRegistration`

New functionality: module list, info, register, unregister:
```
cloud-data:>module list
      Source      Processor             Sink         Task
  ----------  --------------------  -----------  --------
      ftp         filter                counter
      http        groovy-filter         log
      time        groovy-transform      redis
                  transform

cloud-data:>module info http --type source
Information about source module 'http':

Maven coordinates:
org.springframework.cloud.stream.module:http-source:jar:1.0.0.BUILD-SNAPSHOT

  Option Name  Description  Default  Type
  -----------  -----------  -------  ----

cloud-data:>module register mainframe --type source --coordinates
com.megacorp.modules:mainframe:2.4
Successfully registered module 'source:mainframe'

cloud-data:>module list
      Source         Processor             Sink         Task
  -------------  --------------------  -----------  --------
      ftp            filter                counter
      http           groovy-filter         log
      mainframe      groovy-transform      redis
      time           transform

cloud-data:>module info mainframe:source
Information about source module 'mainframe':

Maven coordinates: com.megacorp.modules:mainframe:jar:2.4

  Option Name  Description  Default  Type
  -----------  -----------  -------  ----

cloud-data:>module unregister mainframe:source
Successfully unregistered module 'mainframe' with type source
```
Missing functionality:
 * Module options display
 * Module composition

### DIFF
--- a/spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/controller/ModuleController.java
+++ b/spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/controller/ModuleController.java
@@ -1,0 +1,160 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.data.admin.controller;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.cloud.data.core.ModuleCoordinates;
+import org.springframework.cloud.data.core.ModuleType;
+import org.springframework.cloud.data.module.registry.ModuleRegistration;
+import org.springframework.cloud.data.module.registry.ModuleRegistry;
+import org.springframework.cloud.data.rest.resource.DetailedModuleRegistrationResource;
+import org.springframework.cloud.data.rest.resource.ModuleRegistrationResource;
+import org.springframework.data.domain.PageImpl;
+import org.springframework.data.domain.Pageable;
+import org.springframework.data.web.PagedResourcesAssembler;
+import org.springframework.hateoas.ExposesResourceFor;
+import org.springframework.hateoas.PagedResources;
+import org.springframework.hateoas.mvc.ResourceAssemblerSupport;
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.ResponseBody;
+import org.springframework.web.bind.annotation.ResponseStatus;
+import org.springframework.web.bind.annotation.RestController;
+
+/**
+ * Handles all Module related interactions.
+ *
+ * @author Glenn Renfro
+ * @author Mark Fisher
+ * @author Gunnar Hillert
+ * @author Eric Bottard
+ * @author Gary Russell
+ * @author Patrick Peralta
+ */
+@RestController
+@RequestMapping("/modules")
+@ExposesResourceFor(ModuleRegistrationResource.class)
+public class ModuleController {
+
+	private final Assembler moduleAssembler = new Assembler();
+
+	private final ModuleRegistry registry;
+
+	@Autowired
+	public ModuleController(ModuleRegistry registry) {
+		this.registry = registry;
+	}
+
+	/**
+	 * List module registrations.
+	 */
+	@RequestMapping(value = "", method = RequestMethod.GET)
+	@ResponseStatus(HttpStatus.OK)
+	@ResponseBody
+	public PagedResources<? extends ModuleRegistrationResource> list(Pageable pageable,
+			PagedResourcesAssembler<ModuleRegistration> assembler,
+			@RequestParam(value = "type", required = false) ModuleType type,
+			@RequestParam(value = "detailed", defaultValue = "false") boolean detailed) {
+
+		List<ModuleRegistration> list = new ArrayList<>(registry.findAll());
+		if (type != null) {
+			for (Iterator<ModuleRegistration> iterator = list.iterator(); iterator.hasNext();) {
+				if (iterator.next().getType() != type) {
+					iterator.remove();
+				}
+			}
+		}
+
+		// modules have a "natural" sort order of type followed by name,
+		// use that instead of what "pageable" indicates as a sort order
+		Collections.sort(list);
+		return assembler.toResource(new PageImpl<>(list), moduleAssembler);
+	}
+
+	/**
+	 * Retrieve detailed information about a particular module.
+	 *
+	 * @param type  module type
+	 * @param name  module name
+	 * @return detailed module information
+	 */
+	@RequestMapping(value = "/{type}/{name}", method = RequestMethod.GET)
+	@ResponseStatus(HttpStatus.OK)
+	@ResponseBody
+	public DetailedModuleRegistrationResource info(@PathVariable("type") ModuleType type,
+			@PathVariable("name") String name) {
+		ModuleRegistration registration = registry.find(name, type);
+
+		return (registration == null ? null :
+				new DetailedModuleRegistrationResource(moduleAssembler.toResource(registration)));
+	}
+
+	/**
+	 * Register a module name and type with its Maven coordinates.
+	 *
+	 * @param type  module type
+	 * @param name  module name
+	 * @param coordinates  Maven coordinates for the module artifact
+	 * @param force if {@code true}, overwrites a pre-existing registration
+	 */
+	@RequestMapping(value = "/{type}/{name}", method = RequestMethod.POST)
+	@ResponseStatus(HttpStatus.CREATED)
+	public void register(
+			@PathVariable("type") ModuleType type,
+			@PathVariable("name") String name,
+			@RequestParam("coordinates") String coordinates,
+			@RequestParam(value = "force", defaultValue = "false") boolean force) {
+		if (!force && registry.find(name, type) != null) {
+			return;
+		}
+		registry.save(new ModuleRegistration(name, type, ModuleCoordinates.parse(coordinates)));
+	}
+
+	/**
+	 * Unregister a module name and type.
+	 *
+	 * @param type the module type
+	 * @param name the module name
+	 */
+	@RequestMapping(value = "/{type}/{name}", method = RequestMethod.DELETE)
+	@ResponseStatus(HttpStatus.OK)
+	public void unregister(@PathVariable("type") ModuleType type, @PathVariable("name") String name) {
+		registry.delete(name, type);
+	}
+
+	class Assembler extends ResourceAssemblerSupport<ModuleRegistration, ModuleRegistrationResource> {
+
+		public Assembler() {
+			super(ModuleController.class, ModuleRegistrationResource.class);
+		}
+
+		@Override
+		public ModuleRegistrationResource toResource(ModuleRegistration registration) {
+			return new ModuleRegistrationResource(registration.getName(),
+					registration.getType().name(), registration.getCoordinates().toString(), false);
+		}
+	}
+
+}
--- a/spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/controller/ModuleController.java
+++ b/spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/controller/ModuleController.java
@@ -71,7 +71,7 @@ public class ModuleController {
 	 */
 	@RequestMapping(method = RequestMethod.GET)
 	@ResponseStatus(HttpStatus.OK)
-	public PagedResources<? extends ModuleRegistrationResource> list(Pageable pageable,
+	public PagedResources<? extends ModuleRegistrationResource> list(
 			PagedResourcesAssembler<ModuleRegistration> assembler,
 			@RequestParam(value = "type", required = false) ModuleType type,
 			@RequestParam(value = "detailed", defaultValue = "false") boolean detailed) {
--- a/spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/controller/ModuleController.java
+++ b/spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/controller/ModuleController.java
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- * http://www.apache.org/licenses/LICENSE-2.0
+ *      http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
@@ -39,7 +39,6 @@ import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.RequestParam;
-import org.springframework.web.bind.annotation.ResponseBody;
 import org.springframework.web.bind.annotation.ResponseStatus;
 import org.springframework.web.bind.annotation.RestController;
 
@@ -70,9 +69,8 @@ public class ModuleController {
 	/**
 	 * List module registrations.
 	 */
-	@RequestMapping(value = "", method = RequestMethod.GET)
+	@RequestMapping(method = RequestMethod.GET)
 	@ResponseStatus(HttpStatus.OK)
-	@ResponseBody
 	public PagedResources<? extends ModuleRegistrationResource> list(Pageable pageable,
 			PagedResourcesAssembler<ModuleRegistration> assembler,
 			@RequestParam(value = "type", required = false) ModuleType type,
@@ -102,8 +100,8 @@ public class ModuleController {
 	 */
 	@RequestMapping(value = "/{type}/{name}", method = RequestMethod.GET)
 	@ResponseStatus(HttpStatus.OK)
-	@ResponseBody
-	public DetailedModuleRegistrationResource info(@PathVariable("type") ModuleType type,
+	public DetailedModuleRegistrationResource info(
+			@PathVariable("type") ModuleType type,
 			@PathVariable("name") String name) {
 		ModuleRegistration registration = registry.find(name, type);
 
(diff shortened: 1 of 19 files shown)


spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/controller/ModuleController.java
```diff
@@ -0,0 +1,160 @@
...
+	public ModuleController(ModuleRegistry registry) {
+		this.registry = registry;
+	}
+
+	/**
+	 * List module registrations.
+	 */
+	@RequestMapping(value = "", method = RequestMethod.GET)
+	@ResponseStatus(HttpStatus.OK)
+	@ResponseBody
```
ilayaperumalg: `@ResponseBody` is redundant when using `@RestController` for the class. also, we can remove `value = ""` in the RequestMapping (ok to be in there for clarity).

spring-cloud-data-shell/src/main/java/org/springframework/cloud/data/shell/command/ModuleCommands.java
```diff
@@ -0,0 +1,250 @@
...
+
+		public String name;
+
+		public QualifiedModuleName(String name, ModuleType type) {
+			this.name = name;
+			this.type = type;
+		}
+	}
+
+}
```
ilayaperumalg: newline

spring-cloud-data-rest-resource/src/main/java/org/springframework/cloud/data/rest/resource/ModuleRegistrationResource.java
```diff
@@ -0,0 +1,107 @@
...
+	protected ModuleRegistrationResource() {
+	}
+
+	/**
+	 * Construct a {@code ModuleRegistrationResource}.
+	 *
+	 * @param name module name
+	 * @param type module type
+	 * @param coordinates coordinates for module artifact
+	 * @param composed true if this is a composed module
```
ilayaperumalg: Are we going to have `composed` module at the ModuleRegistration level? If I understand it correct, it will be the responsibility of the module author to compose modules together (using aggregate builder pattern) and make it pointing to a single maven co-ordinate. For the modules with different maven co-ordinates, we can do the module composition using `direct-binding` while still have it connected to other modules via binder.

spring-cloud-data-rest-resource/src/main/java/org/springframework/cloud/data/rest/resource/ModuleRegistrationResource.java
```diff
@@ -0,0 +1,107 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
```
markfisher: all of the copyright headers should have 5 spaces before "http:" (recently a bulk update took care of most, we all need to ensure our templates are up-to-date)

spring-cloud-data-rest-resource/src/main/java/org/springframework/cloud/data/rest/resource/ModuleRegistrationResource.java
markfisher: If we add back `module compose`, I think it will merely be a way to save a shortcut, e.g.:

```
module compose  loudhttp --definition "http | transform --expression=payload.toUpperCase()"
```

so that streams can be defined with that:

```
stream create example --definition "loudhttp | log"
```

We need to decide where we handle the expansion of the composed module, but I think we should do that expansion prior to looking up registry entries. In other words, the above stream definition would still lead to a lookup of 3 modules: http, filter, log.

spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/CloudDataOperations.java
```diff
@@ -26,15 +26,21 @@
...
-	public StreamOperations streamOperations();
+	StreamOperations streamOperations();
 
 	/**
 	 * Counter related operations.
 	 */
-	public CounterOperations counterOperations();
-
+	CounterOperations counterOperations();
+
```
markfisher: trailing whitespace?

spring-cloud-data-rest-resource/src/main/java/org/springframework/cloud/data/rest/resource/ModuleRegistrationResource.java
pperalta: Agreed; I'll get rid of anything having to do with composition until (if) we add it back.

spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/ModuleOperations.java
```diff
@@ -0,0 +1,78 @@
...
+
+	/**
+	 * Register a module name and type with its Maven coordinates.
+	 *
+	 * @param name  module name
+	 * @param type  module type
+	 * @param coordinates  Maven coordinates for the module artifact
+	 * @param force if {@code true}, overwrites a pre-existing registration
+	 */
+	ModuleRegistrationResource registerModule(String name, ModuleType type,
```
markfisher: For consistency (with other methods in here as well as the methods of the ModuleController), we should drop "Module" from the `register`/`unregister` method names. Also relevant, see this change we agreed on earlier today for TaskOperations: https://github.com/cppwfs/spring-cloud-data/commit/ab4f6b5ad15b3f4123d8911ba0740bb118f9720c#diff-9a647bdc0cb461b538bff06ed9c338efL39

spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/ModuleOperations.java
```diff
@@ -0,0 +1,78 @@
...
+	ModuleRegistrationResource register(String name, ModuleType type,
+			String coordinates, boolean force);
+
+	/**
+	 * Unregister a module name and type.
+	 *
+	 * @param name  module name
+	 * @param type  module type
+	 */
+	void unregisterModule(String name, ModuleType type);
```
markfisher: should also change this from `unregisterModule` to just `unregister`

spring-cloud-data-rest-client/src/main/java/org/springframework/cloud/data/rest/client/ModuleTemplate.java
```diff
@@ -0,0 +1,97 @@
...
+
+	@Override
+	public PagedResources<ModuleRegistrationResource> list(ModuleType type) {
+		// TODO handle pagination at the client side
+		String uri = uriTemplate + "?size=10000" + ((type == null) ? "" : "&type=" + type.name());
+		return restTemplate.getForObject(uri, ModuleRegistrationResource.Page.class);
+	}
+
+	@Override
+	public void unregisterModule(String name, ModuleType moduleType) {
```
markfisher: should change to `unregister` here as well

spring-cloud-data-shell/src/main/java/org/springframework/cloud/data/shell/command/ModuleCommands.java
```diff
@@ -0,0 +1,227 @@
...
+	private CloudDataShell cloudDataShell;
+
+
+	@CliAvailabilityIndicator({LIST_MODULES, MODULE_INFO, UNREGISTER_MODULE, REGISTER_MODULE})
+	public boolean available() {
+		return cloudDataShell.getCloudDataOperations() != null;
+	}
+
+	@CliCommand(value = MODULE_INFO, help = "Get information about a module")
+	public String moduleInfo(
```
markfisher: should rename `info`

spring-cloud-data-shell/src/main/java/org/springframework/cloud/data/shell/command/ModuleCommands.java
```diff
@@ -0,0 +1,227 @@
...
+						.addValue(4, o.getType() == null ? "<unknown>" : o.getType());
+				table.getRows().add(row);
+			}
+			result.append(table.toString());
+		}
+		return result.toString();
+	}
+
+	@CliCommand(value = REGISTER_MODULE, help = "Register a new module")
+	public String registerModule(
```
markfisher: should rename `register`

spring-cloud-data-shell/src/main/java/org/springframework/cloud/data/shell/command/ModuleCommands.java
```diff
@@ -0,0 +1,227 @@
...
+					help = "force update if module already exists (only if not in use)",
+					specifiedDefaultValue = "true",
+					unspecifiedDefaultValue = "false")
+			boolean force) {
+		moduleOperations().register(name, type, coordinates, force);
+		return String.format(("Successfully registered module '%s:%s'"), type, name);
+	}
+
+	@CliCommand(value = UNREGISTER_MODULE, help = "Unregister a module")
+	public String unregisterModule(
```
markfisher: should rename `unregister`

spring-cloud-data-shell/src/main/java/org/springframework/cloud/data/shell/command/ModuleCommands.java
```diff
@@ -0,0 +1,227 @@
...
+			ModuleType type) {
+
+		QualifiedModuleName module = processArgs(name, type);
+		moduleOperations().unregisterModule(module.name, module.type);
+		return String.format(("Successfully unregistered module '%s' with type %s"),
+				module.name, module.type);
+	}
+
+	@CliCommand(value = LIST_MODULES, help = "List all modules")
+	public Table listModules() {
```
markfisher: should rename `list`

spring-cloud-data-rest-resource/src/main/java/org/springframework/cloud/data/rest/resource/DetailedModuleRegistrationResource.java
```diff
@@ -0,0 +1,198 @@
...
+ */
+public class DetailedModuleRegistrationResource extends ModuleRegistrationResource {
+
+	/**
+	 * Optional short description of the module.
+	 */
+	private String shortDescription;
+
+	/**
+	 * List of module options.
```
markfisher: do we want to maintain the term "options" at this level?... or let the fact these are mapped to boot *properties* bubble up to here?

spring-cloud-data-rest-resource/src/main/java/org/springframework/cloud/data/rest/resource/DetailedModuleRegistrationResource.java
```diff
@@ -0,0 +1,198 @@
...
+			this.description = description;
+			this.defaultValue = defaultValue;
+			this.hidden = hidden;
+		}
+
+		public String getName() {
+			return name;
+		}
+
+
```
markfisher: extra lines between several of these getters

spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/controller/ModuleController.java
```diff
@@ -0,0 +1,158 @@
...
+			for (Iterator<ModuleRegistration> iterator = list.iterator(); iterator.hasNext();) {
+				if (iterator.next().getType() != type) {
+					iterator.remove();
+				}
+			}
+		}
+
+		// modules have a "natural" sort order of type followed by name,
+		// use that instead of what "pageable" indicates as a sort order
+		Collections.sort(list);
```
ericbottard: That doesn't make sense to me. If we provide a sort order, then it should be honored. If we don't want to support that, then let's drop the pageable.

spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/controller/ModuleController.java
```diff
@@ -0,0 +1,158 @@
...
+	@ResponseStatus(HttpStatus.CREATED)
+	public void register(
+			@PathVariable("type") ModuleType type,
+			@PathVariable("name") String name,
+			@RequestParam("coordinates") String coordinates,
+			@RequestParam(value = "force", defaultValue = "false") boolean force) {
+		if (!force && registry.find(name, type) != null) {
+			return;
+		}
+		registry.save(new ModuleRegistration(name, type, ModuleCoordinates.parse(coordinates)));
```
ericbottard: Should we see if the module is currently in use?

spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/controller/StreamController.java
```diff
@@ -214,9 +216,14 @@ private void deployStream(StreamDefinition stream, Map<String, String> deploymen
 		Iterator<ModuleDefinition> iterator = stream.getDeploymentOrderIterator();
 		for (int i = 0; iterator.hasNext(); i++) {
 			ModuleDefinition module = iterator.next();
-			String type = (i == 0) ? "sink"
-					: (iterator.hasNext() ? "processor" : "source");
-			ModuleCoordinates coordinates = this.registry.findByNameAndType(module.getName(), type);
+			ModuleType type = (i == 0) ? ModuleType.sink
```
ericbottard: I thought we were going away from the type enum. What made us change our minds?

spring-cloud-data-shell/src/main/java/org/springframework/cloud/data/shell/command/ModuleCommands.java
```diff
@@ -0,0 +1,227 @@
...
+
+
+	@CliAvailabilityIndicator({LIST_MODULES, MODULE_INFO, UNREGISTER_MODULE, REGISTER_MODULE})
+	public boolean available() {
+		return cloudDataShell.getCloudDataOperations() != null;
+	}
+
+	@CliCommand(value = MODULE_INFO, help = "Get information about a module")
+	public String moduleInfo(
+			@CliOption(mandatory = true,
```
ericbottard: Splitting type and name is not TAB completion friendly. I know <type>:<name> is not ideal, but this is the only way to provide completion

spring-cloud-data-rest-resource/src/main/java/org/springframework/cloud/data/rest/resource/DetailedModuleRegistrationResource.java
pperalta: If we are calling these "properties" across the board then we should rename to "options". Are we dropping the use of "options" entirely?

spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/controller/ModuleController.java
pperalta: +1, removed

spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/controller/ModuleController.java
pperalta: This would not affect streams currently in use, just new ones. What do you think @markfisher?

spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/controller/StreamController.java
pperalta: I believe the idea is that on the SCS side there isn't strict checking of module types, but on the SCD side it will be more opinionated. 

@markfisher and I went back and forth on this one a few times. I can see the possibility of going with strings if it's possible to add a new module type without modifying SCD...

spring-cloud-data-shell/src/main/java/org/springframework/cloud/data/shell/command/ModuleCommands.java
pperalta: I added this method to handle the `:`...is there a way to make this work with autocompletion?

```java
/**
 * Return a {@link QualifiedModuleName} for the given arguments.
 * If {@code type} is {@code null}, the module type may be obtained
 * from the module name if the module name is in the format
 * {@code name:type}.
 *
 * @param name module name
 * @param type module type; may be {@code null}
 * @return {@code QualifiedModuleName} for the provided arguments
 */
private QualifiedModuleName processArgs(String name, ModuleType type) {
	if (type == null) {
		String[] split = name.split("\\:");
		if (split.length != 2) {
			throw new IllegalArgumentException(
					String.format("Expected format of 'name:type' for module name %s", name));
		}
		return new QualifiedModuleName(split[0], ModuleType.valueOf(split[1]));
	}
	else {
		return new QualifiedModuleName(name, type);
	}
}
```

spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/controller/ModuleController.java
```diff
@@ -0,0 +1,158 @@
...
+	public ModuleController(ModuleRegistry registry) {
+		this.registry = registry;
+	}
+
+	/**
+	 * List module registrations.
+	 */
+	@RequestMapping(method = RequestMethod.GET)
+	@ResponseStatus(HttpStatus.OK)
+	public PagedResources<? extends ModuleRegistrationResource> list(
```
pperalta: @ericbottard - I removed `Pagable`; do I need to change the return type here as well?

spring-cloud-data-shell/src/main/java/org/springframework/cloud/data/shell/command/ModuleCommands.java
ericbottard: Ah, I see.
So completion would work with that single "name" arg, but then why require
a --type arg?

What can't work is this: `register --type --name <TAB>` because the
completer for the name arg doesn't "see" the other args, namely --type here

spring-cloud-data-rest-resource/src/main/java/org/springframework/cloud/data/rest/resource/DetailedModuleRegistrationResource.java
markfisher: for now, it probably makes sense to keep "options"... a term for the subset of boot properties of a module that are intended to be configured via the DSL (and for which we'll be providing metadata-driven auto-completion, etc).

spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/controller/ModuleController.java
markfisher: I think it's ok for this to affect a stream that is explicitly redeployed, given that in order to overwrite an existing entry with the same key, one needs to use "force=true" anyways.

markfisher:
squashed and merged as https://github.com/spring-cloud/spring-cloud-data/commit/95b8af406cdbcba9500fd3dda88c0670df30a10f

spring-cloud-data-admin/src/main/java/org/springframework/cloud/data/admin/controller/ModuleController.java
ericbottard: Yes, I understand. Which is why it may be a problem: what if I have `time (v1) --fixedDelay | log` and I upload time v2, which doesn't have a `fixedDelay` option. If the stream gets redeployed (for whatever reason, maybe just scaled up for example), then boom

markfisher: If a stream is redeployed using a module that has been explicitly updated and that update causes a failure, then the redeploy itself will fail. That should be expected. I don't think we should couple the registry to the runtime state of streams. By requiring "force" to override any existing registry entry, we are already requiring the user to think twice about updates to modules that *might* be in use already.

PR #69, opened 2015-09-01
Add better YARN application model

jvalkeal:
Fix yarn profile
- AdminApplication/AdminConfiguration were changed to rely on component scan which did work with yarn profile because YarnConfiguration did have @Configuration and its bean 'yarnConfiguration' was overriden from spring-yarn autoconfiguration. This didn't cause trouble when YarnConfiguration were imported because bean name were then org.springframework.cloud.data.admin.config.YarnConfiguration. Component scan is creating different bean names, 'yarnConfiguration' in this case.
- Disabling spring yarn auto-config, classes are needed in future PR's so not removing deps.
- Add @Configuration to YarnConfiguration for scan to work.
- Add tests for default/yarn profiles.

XD-3400 XD-3401 XD-3402
- Update to shdp 2.3.0.M2
- Add statemachine to control deployment flow into yarn
- Support passing module props from stream into module ran by yarn
- Support automatically pushing and starting yarn app instance
- Update README to reflect these changes

### DIFF
--- a/spring-cloud-data-admin/src/test/java/org/springframework/cloud/data/admin/AdminApplicationTests.java
+++ b/spring-cloud-data-admin/src/test/java/org/springframework/cloud/data/admin/AdminApplicationTests.java
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- * http://www.apache.org/licenses/LICENSE-2.0
+ *      http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
--- a/spring-cloud-data-admin/src/test/java/org/springframework/cloud/data/admin/AdminApplicationTests.java
+++ b/spring-cloud-data-admin/src/test/java/org/springframework/cloud/data/admin/AdminApplicationTests.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.data.admin;
+
+import static org.hamcrest.Matchers.instanceOf;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertThat;
+
+import org.junit.Test;
+import org.springframework.boot.SpringApplication;
+import org.springframework.cloud.data.module.deployer.local.LocalModuleDeployer;
+import org.springframework.cloud.data.module.deployer.yarn.YarnModuleDeployer;
+import org.springframework.context.ConfigurableApplicationContext;
+
+/**
+ * Generic tests for {@link AdminApplication}.
+ *
+ * @author Janne Valkealahti
+ *
+ */
+public class AdminApplicationTests {
+
+	@Test
+	public void testDefaultProfile() {
+		SpringApplication app = new SpringApplication(AdminApplication.class);
+		ConfigurableApplicationContext context = app.run(new String[] { "--server.port=0" });
+		assertThat(context.containsBean("processModuleDeployer"), is(true));
+		assertThat(context.getBean("processModuleDeployer"), instanceOf(LocalModuleDeployer.class));
+		context.close();
+	}
+
+	@Test
+	public void testYarnProfile() {
+		SpringApplication app = new SpringApplication(AdminApplication.class);
+		ConfigurableApplicationContext context = app.run(new String[] { "--spring.profiles.active=yarn",
+				"--server.port=0" });
+		assertThat(context.containsBean("processModuleDeployer"), is(true));
+		assertThat(context.getBean("processModuleDeployer"), instanceOf(YarnModuleDeployer.class));
+		context.close();
+	}
+
+}
--- a/spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-yarn/src/main/java/org/springframework/cloud/data/module/deployer/yarn/DefaultYarnCloudAppService.java
+++ b/spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-yarn/src/main/java/org/springframework/cloud/data/module/deployer/yarn/DefaultYarnCloudAppService.java
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- * http://www.apache.org/licenses/LICENSE-2.0
+ *      http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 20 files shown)


spring-cloud-data-admin/src/test/java/org/springframework/cloud/data/admin/AdminApplicationTests.java
```diff
@@ -0,0 +1,56 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
```
pperalta: Minor nitpick - this line should be indented 5 spaces in (6 if you include the one space between the `*` and the first character of every other line.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-yarn/src/main/java/org/springframework/cloud/data/module/deployer/yarn/DefaultYarnCloudAppService.java
```diff
@@ -0,0 +1,248 @@
...
+	@Override
+	public Collection<CloudAppInfo> getApplications() {
+		ArrayList<CloudAppInfo> infos = new ArrayList<CloudAppInfo>();
+
+		YarnInfoApplication app = new YarnInfoApplication();
+		Properties appProperties = new Properties();
+		appProperties.setProperty(PREFIX_YIA + "operation", "PUSHED");
+		app.appProperties(appProperties);
+		String info = app.run();
+		if (logger.isDebugEnabled()) {
```
pperalta: With SLF4J you can avoid the `if` by doing:

```java
logger.debug("Full response for SUBMITTED: {}", info);
```
In the case where you want to avoid a `toString` invocation, SLF4J will evaluate the log level before constructing the string.

See section **Better yet, use parameterized messages** here:

http://www.slf4j.org/faq.html#logging_performance

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-yarn/src/main/java/org/springframework/cloud/data/module/deployer/yarn/YarnCloudAppService.java
```diff
@@ -0,0 +1,155 @@
...
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.data.module.deployer.yarn;
+
+import java.util.Collection;
+import java.util.Map;
+
+/**
+ * Interface used to glue a state machine and yarn application logic together.
```
pperalta: Is a state machine specifically required to implement this interface?

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-yarn/src/main/java/org/springframework/cloud/data/module/deployer/yarn/YarnCloudAppService.java
```diff
@@ -0,0 +1,155 @@
...
+	 */
+	void pushApplication(String appVersion);
+
+	/**
+	 * Submit new application instance into yarn.
+	 *
+	 * @param appVersion the app version
+	 * @return the application id
+	 */
+	String submitApplication(String appVersion);
```
pperalta: Should the JavaDoc (either at the individual method or at the interface level) indicate the difference between "submit" and "push"? Should it describe general expected usage/lifecycle?

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-yarn/src/main/java/org/springframework/cloud/data/module/deployer/yarn/YarnCloudAppStateMachine.java
```diff
@@ -0,0 +1,437 @@
...
+			String applicationId = yarnCloudAppService.submitApplication(appVersion);
+			context.getExtendedState().getVariables().put(VAR_APPLICATION_ID, applicationId);
+
+			for (int i = 0; i < 60; i++) {
+				if (isRunning()) {
+					return;
+				}
+				try {
+					Thread.sleep(1000);
+				} catch (InterruptedException e) {
```
pperalta: Swallowing the `InterruptedException` without resetting the interrupt flag will cause this loop to run to completion. Is this the intended behavior? Would it make sense to reset the flag and throw a `RuntimeException`?

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-yarn/src/test/java/org/springframework/cloud/data/module/deployer/yarn/YarnCloudAppStateMachineTests.java
```diff
@@ -0,0 +1,357 @@
...
+		}
+
+	}
+
+	private static class TestYarnCloudAppService implements YarnCloudAppService {
+
+		String app = null;
+		String instance = null;
+
+		volatile CountDownLatch getApplicationsLatch = new CountDownLatch(1);
```
pperalta: I don't see these being mutated anywhere - would `final` be more appropriate here?

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-yarn/src/main/java/org/springframework/cloud/data/module/deployer/yarn/YarnCloudAppService.java
```diff
@@ -0,0 +1,155 @@
...
+	 * @param clusterId the cluster id
+	 */
+	void stopCluster(String yarnApplicationId, String clusterId);
+
+	/**
+	 * Gets the clusters states.
+	 *
+	 * @return the clusters states
+	 */
+	Map<String, String> getClustersStates();
```
pperalta: Should the JavaDoc indicate what strings are returned? For instance, is the key of the map the application id?

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-yarn/src/main/java/org/springframework/cloud/data/module/deployer/yarn/YarnCloudAppService.java
jvalkeal: In this case yes because this is kinda internal interface. Didn't want to use implementation directly to ease testing. It's wired in `YarnConfiguration`.

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-yarn/src/main/java/org/springframework/cloud/data/module/deployer/yarn/YarnCloudAppStateMachine.java
jvalkeal: good point but we don't want to throw here. we'll eventually handle this in an error state(which is not yet implemented).

spring-cloud-data-module-deployers/spring-cloud-data-module-deployer-yarn/src/test/java/org/springframework/cloud/data/module/deployer/yarn/YarnCloudAppStateMachineTests.java
jvalkeal: `app` and `instance` are both changed in tests.

pperalta:
Looks good overall - and the experience with deploying a stream is much better! I have a few small suggestions that I pushed to https://github.com/pperalta/spring-cloud-data/commit/69011a500b3e0e2da5640b2e5aef06eb0a730510. @jvalkeal if you agree I'll include these when I merge.

jvalkeal:
@pperalta yeah, looks ok!

pperalta:
Rebased, squashed, merged as 4632f1df763aa345f7e5dd0f25caf94fbe64d566

PR #72, opened 2015-09-03
local deployer uses server.port if provided

markfisher:
If the 'server.port' for a module is provided via deployment properties,
the LocalModuleDeployer will now attempt to use that port rather than allocating
one. This allows the use of an explicitly provided port for the 'http' source, e.g.:
- stream deploy example --properties "module.http.server.port=9000"


pperalta:
Merged as 51677dc2d3bce368a0eed3d3458fea88f1c67a8a

PR #75, opened 2015-09-03
Remove application.yml from admin

ilayaperumalg:
- This will enable other non-admin contexts (especially the module contexts in case of local module deployer) not to use the
application.yml properties


spring-cloud-data-admin/src/main/resources/bootstrap.yml
```diff
@@ -1,9 +1,33 @@
+server:
```
ericbottard: I don't think `bootstrap.yml` is meant to contain props that are actually applicative properties. Its aim is to _bootstrap_, _i.e._ to tell where very early things, like the config server, are.

ericbottard:
> This will enable other non-admin contexts (especially the module contexts in case of local module deployer) not to use the application.yml properties

not quite sure I understand what you mean, but in anycase, the module context should not "see" the admin context anyway (ClassLoader wise)

markfisher:
Is it worth considering a `@PropertySource` for these properties instead? Agreed, the *real* solution will avoid any classpath visibility, but this is causing a critical bug at the moment.

ericbottard:
merged as 73ef50f

PR #76, opened 2015-09-04
Yarn profile based config

ilayaperumalg:
- Move yarn specific config from admin.properties to `bootstrap-yarn.yml` so that these config will only be loaded when `yarn` profile is active.


ericbottard:
Maybe we'll regret not using `spring.config.name` instead of 73ef50ff7014198c64c7f8734b1d3db1da116598 , we'd get that behavior for free :)

This stuff still does not belong in `bootstrap.yml`.

jvalkeal:
@ericbottard @trisberg  ups, just merged this because it fixed broken config.

PR #80, opened 2015-09-07
Interpret more application states

twoseat:
Currently only STARTED is recognized and interpreted as module state
`deployed`. This commit recognizes all the application instance states
we are aware of, and interprets them as various module instance states.

It also attempts to obtain application instance status irrespective of
the reported state of the application, and ignores the exception that
might be thrown.

### DIFF
--- a/spring-cloud-dataflow-module-deployers/spring-cloud-dataflow-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/dataflow/module/deployer/cloudfoundry/ModuleStatusBuilder.java
+++ b/spring-cloud-dataflow-module-deployers/spring-cloud-dataflow-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/dataflow/module/deployer/cloudfoundry/ModuleStatusBuilder.java
@@ -56,9 +56,15 @@ class ModuleStatusBuilder {
 		return this;
 	}
 
-	// TODO: Find out what the application state candidates are
 	private static ModuleStatus.State convertState(String state) {
 		switch (state) {
+			case "STARTED":
+			case "STARTING":
+			case "STOPPED":
+				return ModuleStatus.State.deploying;
+			case "CRASHED":
+			case "CRASHING":
+				return ModuleStatus.State.failed;
 			case "RUNNING":
 				return ModuleStatus.State.deployed;
 			default:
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-module-deployers/spring-cloud-dataflow-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/dataflow/module/deployer/cloudfoundry/ModuleStatusBuilder.java
```diff
@@ -56,9 +56,15 @@ ModuleStatusBuilder withId(ModuleDeploymentId id) {
 		return this;
 	}
 
-	// TODO: Find out what the application state candidates are
 	private static ModuleStatus.State convertState(String state) {
 		switch (state) {
+			case "STARTED":
+			case "STARTING":
+			case "STOPPED":
```
markfisher: We need to add "undeployed" to the enum first, but would that be an appropriate mapping for "STOPPED"?

twoseat: An app can be `STOPPED` because it has been created but not yet started (we can't create it directly as `STARTED`, so every app will be in this state for a while when it's first deployed), or because it is in the process of being destroyed (which is a short-lived state). So `undeployed` would be technically correct, as it describes either of those states, but `deploying` is probably more informative as `STOPPED` will generally only be encountered during app startup (i.e. module `deploying`).

I'd leave it as is, but if `undeployed` better fits your mental model then it's a reasonable change.

markfisher: Thanks Paul. Based on that description, I would say "deploying" in indeed the right mapping.

At some point we might want to consider the option of a "stream undeploy" triggering apps to STOP instead of always being a DELETE. We would probably need to bubble up the distinction in some way (STOP would be preferable when you will later want to re-deploy with minimal overhead, but DELETE is required I assume if you want to change any deployment properties/env vars when re-deploying). AFAIK, we don't have the same distinction on Lattice where *stream deploy* maps to CREATE and *stream undeploy* maps to REMOVE.

twoseat: I'd have to learn more about Lattice's approach, but it may not be worth the extra complexity it would expose. The biggest expense in a delete+deploy is the upload, and CF will cache that if the bits haven't changed (though we're not yet doing the upload in the right way to take advantage of that). We could certainly save several seconds a time by being able to `restart` rather than `redeploy` though, so worth looking into.

markfisher: Good point on the caching; that will be nice when we're taking advantage of it. In any case, this is a topic to explore another day. For now STOPPED -> "deploying" makes perfect sense. Thanks for the detail.

PR #83, opened 2015-09-09
LRP request addRoute -> addHttpRoute

ilayaperumalg:


ilayaperumalg:
This change is due with this [PR] (https://github.com/cloudfoundry-incubator/receptor-client/pull/14) change.

PR #93, opened 2015-09-17
Fixes passing complex arguments.

ericbottard:
What I thought was originally a quotes (' and ") problem was actually a bash escape problem (especially parentheses for example). So now properly bash-escape args that we pass.

Thx @nebhale !


markfisher:
closing this as it's now being addressed via the changes here: https://github.com/spring-cloud/spring-cloud-dataflow/commit/5ca04a4fbe3c5a27f864ecddbf42aee38a35c023#diff-5f11938b65d189a3e2a72a6d25a08cb7R313

PR #94, opened 2015-09-22
Rewrite of the CF Deployer using cf-java-client.

ericbottard:
This brings several immediate benefits:
- less code
- higher abstraction level, so some things are taken care of for us (e.g.
route creation/removal as part of application lifecycle)
- incremental upload, something pretty complicated to implement in itself


Zteve:
The PR work done by https://github.com/ericbottard was augmented, and squashed into a single commit https://github.com/spring-cloud/spring-cloud-dataflow/commit/5ca04a4fbe3c5a27f864ecddbf42aee38a35c023.

The fix to XD-3505 (contributed by https://github.com/spring-cloud/spring-cloud-dataflow/commits/master?author=twoseat) is also included as commit https://github.com/spring-cloud/spring-cloud-dataflow/commit/a67dee286ee68d773064ef8ddf7623245e01e599.

PR #96, opened 2015-09-25
Fix parser handling of dots in option names

ericbottard:
Subtle bug related to method overloading (evil) and the fact that
`x ? a : b ` resolves to the common super type (more evil)


markfisher:
merged as https://github.com/spring-cloud/spring-cloud-dataflow/commit/6a0ef1fb3fc99497ab891d9a79017e0c5a05444b

added newline before "else"

did not include the .gitignore change

PR #98, opened 2015-09-28
Add type conversion support

ilayaperumalg:
- Support inputType/outputType for the module (same support that Spring XD has)
 - Set the module parameters as s-c-s binding properties based on the inputType/outputType specified in DSL


ilayaperumalg:
This PR depends on https://github.com/spring-cloud/spring-cloud-stream/pull/142

PR #100, opened 2015-10-06
XD-3595, XD-3596

markfisher:
Added StreamControllerTests

StreamController now throws a DuplicateStreamException if a stream
already exists with the same name as provided in a save request.

### DIFF
--- a/spring-cloud-dataflow-admin/src/main/java/org/springframework/cloud/dataflow/admin/controller/StreamController.java
+++ b/spring-cloud-dataflow-admin/src/main/java/org/springframework/cloud/dataflow/admin/controller/StreamController.java
@@ -28,13 +28,14 @@ import org.slf4j.LoggerFactory;
 
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.cloud.dataflow.admin.repository.DuplicateStreamException;
 import org.springframework.cloud.dataflow.admin.repository.StreamDefinitionRepository;
+import org.springframework.cloud.dataflow.core.BindingProperties;
 import org.springframework.cloud.dataflow.core.ModuleCoordinates;
 import org.springframework.cloud.dataflow.core.ModuleDefinition;
 import org.springframework.cloud.dataflow.core.ModuleDeploymentId;
 import org.springframework.cloud.dataflow.core.ModuleDeploymentRequest;
 import org.springframework.cloud.dataflow.core.ModuleType;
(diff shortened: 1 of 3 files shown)


spring-cloud-dataflow-admin/src/main/java/org/springframework/cloud/dataflow/admin/controller/StreamController.java
```diff
@@ -145,6 +146,11 @@ public void save(@RequestParam("name") String name,
 			boolean deploy) {
 
 		StreamDefinition stream = new StreamDefinition(name, dsl);
+		if (this.repository.exists(name)) {
```
pperalta: Minor nitpick - seems to make sense to only construct the `StreamDefinition` after checking the repository to see if it exists. I can change this on merge.

markfisher: +1

spring-cloud-dataflow-admin/src/test/java/org/springframework/cloud/dataflow/admin/controller/StreamControllerTests.java
```diff
@@ -0,0 +1,201 @@
...
+	public void testDestroyStreamNotFound() throws Exception {
+		mockMvc.perform(
+				delete("/streams/definitions/myStream").accept(MediaType.APPLICATION_JSON)).andDo(print())
+				.andExpect(status().is5xxServerError());
+		assertEquals(0, repository.count());
+	}
+
+	@Test
+	@SuppressWarnings({"unchecked", "rawtypes"})
+	public void testDeploy() throws Exception {
```
pperalta: `StreamController` works with `ModuleDeployer`, not `ModuleLauncher`. It seems to me that this test is assuming too much knowledge about the local module deployer. Instead of mocking `ModuleLauncher` should it mock `ModuleDeployer` and assert that the correct `ModuleDeploymentRequest`s are being created?

markfisher: The mock ModuleLauncher is what we get (autowired) from the existing TestDependencies configuration class. We can discuss with @mminella if perhaps we can instead mock the deployer and then make changes to both Stream and Task controller tests. If so, I think that should be done after merging this PR as it would involve refactoring both.

pperalta:
LGTM; merged as c74086e19541863d1dd2d3a8efa7a84ecc9cf5ae

PR #101, opened 2015-10-06
XD-3594

markfisher:
* ensures the proper module type is recognized when named channels are used in either the source or sink position
* added tests to StreamControllerTests for single and multiple module streams with both source and sink channels

### DIFF
--- a/spring-cloud-dataflow-admin/src/main/java/org/springframework/cloud/dataflow/admin/controller/StreamController.java
+++ b/spring-cloud-dataflow-admin/src/main/java/org/springframework/cloud/dataflow/admin/controller/StreamController.java
@@ -227,8 +227,7 @@ public class StreamController {
 		boolean isDownStreamModulePartitioned = false;
 		for (int i = 0; iterator.hasNext(); i++) {
 			ModuleDefinition currentModule = iterator.next();
-			ModuleType type = (i == 0) ? ModuleType.sink
-					: (iterator.hasNext() ? ModuleType.processor : ModuleType.source);
+			ModuleType type = determineModuleType(currentModule, i, iterator.hasNext());
 			ModuleRegistration registration = this.registry.find(currentModule.getName(), type);
 			if (registration == null) {
 				throw new IllegalArgumentException(String.format(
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-admin/src/main/java/org/springframework/cloud/dataflow/admin/controller/StreamController.java
```diff
@@ -252,6 +251,37 @@ private void deployStream(StreamDefinition stream, Map<String, String> cumulated
 		}
 	}
 
+	private ModuleType determineModuleType(ModuleDefinition moduleDefinition, int index, boolean hasNext) {
```
ericbottard: Can this go in StreamDefinition itself. I've been duplicating this in TAB completion code?

ericbottard: I was about to file another comment when I realized `index` is in "reverse" order. May want to rename / add javadoc to clarify

markfisher: We are trying to avoid "type" being part of a `ModuleDefinition` since it's really only relevant at the time of deployment, and because we will want to support more flexible topologies (than linear pipelines) in the future and keeping ModuleDefinition simpler will likely facilitate that. On the code-completion side, I would expect that you could rely on the info in the `ModuleRegistration` instead?

Regarding the name of the `hasNext` parameter, I agree we could rename to something clearer. I considered a few names and then went along with the `hasNext` since it matches the method on the `Iterator` where the boolean originates. That said, as we cross a method boundary, it would be better to clarify the ordering. Perhaps we could change it to `hasMoreUpstream` or something. Suggestions?

ericbottard: > because we will want to support more flexible topologies

I was expecting that answer, fine by me.

> you could rely on the info in the ModuleRegistration

The problem is exactly that to get hold of a `ModuleRegistration`, I need a type :)


`isFirst` is good, and would shift the weirdness at the point where it is computed, rather than where it is used.

markfisher: I will change to `isFirst`. Let's continue the discussion about auto-completion offline. I would imagine the logic is a bit different anyways - considering, for example, you could have a processor OR sink following a pipe.

PR #105, opened 2015-10-07
clarify determineModuleType() method

markfisher:


ericbottard:
LGTM, but I think we could
1. move this out of the Controller to a util method in the same package as StreamDefinition
2. add tests that use `new StreamDefinition(name, dsl)` that cover all cases (as some logic here has a corresponding logic in `ModuleDefinitionBuilder`

As a side note, while working on completion, I was surprised to see that parameters were added behind a user's back (`foo | bar --x=y` doesn't have an empty parameters map for `foo`, for example)

markfisher:
I consider any `Utils` method a last-resort and would definitely not do that until there's another use-case.

RE: #2 more tests are always good, and can be done separately from this PR. Not sure what you have in mind as "corresponding logic".

Binding properties being added automatically to connect modules is one of the most obvious benefits of using dataflow over manually connecting spring-cloud-stream modules. Is that what you consider "behind a user's back"?

ericbottard:
> I consider any Utils method a last-resort and would definitely not do that until there's another use-case.

Me too. But this does not belong in the Controller IMO, and given the second point below, this was a way to make it visible

> RE point 2: more tests are always good, and can be done separately from this PR. Not sure what you have in mind as "corresponding logic".

I meant _e.g._ https://github.com/spring-cloud/spring-cloud-dataflow/blob/4d8b0d8cb6a83d38c53cf0b6319918f1209443f9/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/ModuleDefinitionBuilder.java#L85-L92

> Binding properties being added automatically to connect modules is one of the most obvious benefits of using dataflow over manually connecting spring-cloud-stream modules. Is that what you consider "behind a user's back"?

Absolutely agree wrt benefits. What I was referring to was the addition as a side effect of something that (at least in my mind, but this can be argued) was a parsing operation.

markfisher:
As long as the method is only used in the context of the controller it *does* belong there, as a private method. Let's merge this one for now, and then separately continue the discussion of whether this logic really does have another use (it's easy to move now that it's encapsulated in a single method).

To your last point, `ModuleDefinitionBuilder` is not in the scope of the parser; it's the component that translates from a parsed stream to the domain objects for that stream's modules. Part of that is determining how they are connected.

PR #108, opened 2015-10-12
XD-335, XD-3582: Options metadata and DSL completion

ericbottard:
Not yet fully useful because of the huge number of options returned, I will work on a POC for server-side progressive results surfacing as a separate PR.
If not effective, will need to limit discoverability of options somehow, in which case the code in this PR is still 100% valid.

Requires https://github.com/spring-cloud/spring-cloud-stream/pull/150


spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/ExpansionStrategy.java
```diff
@@ -0,0 +1,40 @@
...
+
+import org.springframework.cloud.dataflow.core.StreamDefinition;
+
+/**
+ * Used to enhance a well formed stream definition by adding yet more text to it (<i>e.g.</i> adding more options to a
+ * module).
+ * 
+ * @author Eric Bottard
+ */
+public interface ExpansionStrategy {
```
pperalta: This interface requires two invocations - one to determine if the strategy applies, another to mutate a list passed to it.

A few observations about this interface:

1. It requires the caller to invoke `shouldTrigger` to determine if the strategy applies to the parse result. What happens if the caller does not invoke this method but invokes `addProposals` instead?

2. The `addProposals` method returns `void` and mutates a collection passed to it. It assumes that the collection being passed to it is mutable.

Could we combine these aspects into a single method for this interface?

```
/**
* For a given stream DSL text and {@link StreamDefinition}, return a list
* of {@link CompletionProposal} that apply.
* 
* @param text DSL text for the stream
* @param streamDefinition stream definition
* @param detailLevel integer representing the amount of detail to include
* in the generated {@code CompletionProposal}s
* @return a list of {@code CompletionProposal}; if no proposals apply
* an empty list is returned
*/
List<CompletionProposal> generateProposals(String text, StreamDefinition streamDefinition, int detailLevel);
```
This way, callers can maintain their own `List` of proposals, and the invocation to `shouldTrigger` won't be required because an empty `List` will be returned if the completion should not be triggered.

spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/resource/CompletionProposalsResource.java
```diff
@@ -0,0 +1,76 @@
...
+
+	public void addProposal(String text, String explanation) {
+		proposals.add(new Proposal(text, explanation));
+	}
+
+	public List<Proposal> getProposals() {
+		return proposals;
+	}
+
+	public void setProposals(List<Proposal> proposals) {
```
pperalta: If `setProposals` is invoked after `addProposal`, the previously added `Proposal` will be lost. Is this setter required? We could also consider an `addProposals` method that will add the provided proposals to the existing list.

spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/CompletionUtils.java
```diff
@@ -0,0 +1,98 @@
...
+			} else {
+				return new ModuleType[] {ModuleType.processor, ModuleType.sink};
+			}
+		} // MUST be source
+		else {
+			return new ModuleType[] {ModuleType.source};
+		}
+	}
+
+	static Coordinates adapt(ModuleCoordinates coordinates) {
```
pperalta: Instead of having this method, should we create a 4 param constructor for `Coordinates` with "exec" as the default classifier?

Way **off topic** general question not related to this PR (and probably not worth having the discussion here) - why do we have both `ModuleCoordinates` and `Coordinates`? It appears that `ModuleCoordinates` does everything that `Coordinates` does + more.

spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/CompletionUtils.java
```diff
@@ -0,0 +1,98 @@
...
+		return new Coordinates(coordinates.getGroupId(), coordinates.getArtifactId(), coordinates.getExtension(), "exec", coordinates.getVersion());
+	}
+
+	/**
+	 * Given a candidate module name, maybe prefix it with an auto-generated label if its use would clash with
+	 * an already existing definition.
+	 */
+	static String maybeQualifyWithLabel(String moduleName, StreamDefinition streamDefinition) {
```
pperalta: `ModuleDefinition` already has a `label` field - why do we need this one?

spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/CompletionUtils.java
```diff
@@ -0,0 +1,98 @@
...
+	static final Set<String> IMPLICIT_PARAMETER_NAMES = new HashSet<>();
+	static {
+		IMPLICIT_PARAMETER_NAMES.add(BindingProperties.INPUT_BINDING_KEY);
+		IMPLICIT_PARAMETER_NAMES.add(BindingProperties.OUTPUT_BINDING_KEY);
+	}
+
+	/**
+	 * Return the type(s) of a given module definition, in the context of code completion.
+	 */
+	static ModuleType[] inferType(ModuleDefinition moduleDefinition) {
```
pperalta: Would it be possible to combine this with `StreamController.determineModuleType`, or are they incompatible uses? Why does this return an array; how is it possible that a module is of multiple types?

spring-cloud-dataflow-admin/src/main/java/org/springframework/cloud/dataflow/admin/controller/CompletionController.java
```diff
@@ -0,0 +1,95 @@
...
+
+	@Autowired
+	public CompletionController(StreamCompletionProvider completionProvider) {
+		this.completionProvider = completionProvider;
+	}
+
+	/**
+	 * Return a list of possible completions given a prefix string that the user has started typing.
+	 *  @param start the amount of text written so far
+	 * @param detailLevel the level of detail the user wants in completions
```
pperalta: Is there an accepted range of values for `detailLevel`? This should be specified here and in other places where it is accepted as a parameter.

spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/StacktraceFingerprintingRecoveryStrategy.java
```diff
@@ -0,0 +1,136 @@
...
+					throw exception;
+				}
+			}
+		}
+	}
+
+	/**
+	 * Extract the top frames (until the call to the {@link StreamDefinition} constructor appears) of the given exception.
+	 */
+	private void computeFingerprint(E exception) {
```
pperalta: Since this method is keeping track of the generated fingerprint, should it be `addFingerprint` instead?

spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/ExpansionStrategy.java
ericbottard: You're right about the two invocation thing (which is not uncommon in other designs). This can be avoided without changing the method signature though.

And indeed, I think I'll keep the "mutate the list" approach, because I will need the return value (currently `null`) to indicate whether or not a Strategy should short-circuit the whole invocation chain. I discovered it will be needed for cases like ` ... | log --level=INF<TAB>` which complete with a well known value (from an enum) and so we *know* for sure at the semantic level that `INF` is not sufficient. Yet, at the pure lexical parsing level, `--level=INF` was wellformed, and so other strategies get invoked, resulting in propositions like `--level=INF --name=` (in addition to `--level=INFO` which is the only one we want, at that time). 

Hope that makes sense.

spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/resource/CompletionProposalsResource.java
ericbottard: Agreed. Quick automatic addition when `getProposals()` (which is needed for jackson serialization) was added I guess.

spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/CompletionUtils.java
ericbottard: To answer your first question, I think I'll need to take my shot at the second:
We have the two classes because we did not want leakage from the launcher project to dataflow.

And so, even if we added a 4 param constructor, this would still seem tedious to me and stubborn as I am, I would still add a helper method. What we would need is a constructor `new Coordinates(ModuleCoordinates c)` which is a no go, given the answer above.

spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/CompletionUtils.java
ericbottard: This is a helper method to decide whether or not we need to introduce a label in a completion proposal.
Given the following input
```
http | filter --expression=true | fil<TAB>
```

we don't generate `http | filter --expression=true | filter` but `http | filter --expression=true | filter2: filter` instead.

spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/CompletionUtils.java
ericbottard: I had this discussion with @markfisher too.
Those methods are actually quite different. Why the one you mention operates in the context of a definition that we know (or at least assume) is 100% well formed (going from source-ish to sink-ish), this one -- although it receives an input that is well formed token-wise -- may receive an incomplete definition (we're here for code completion after all).

Hence, 

```
http | filter --expression=true | transform
```

is something we could get here, although `transform` is not a sink.

This method returns several values exactly for that reason: we're going to try to see if `transform` is a sink, or a processor

spring-cloud-dataflow-admin/src/main/java/org/springframework/cloud/dataflow/admin/controller/CompletionController.java
ericbottard: There is no strict contract, other than `[1, ..]`. Strategies are to interpret this as they wish, with the open secret that this actually refers to the successive number of invocations of `<TAB>`

spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/StacktraceFingerprintingRecoveryStrategy.java
ericbottard: I would expect a method named `addFoo()` to take a Foo as an argument. This is mostly about "computing" it, also storing it, you're right. Any verb that captures those two ideas I'll buy, but not convinced by `add`. Sorry `¯\_(ツ)_/¯`

spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/ExpansionStrategy.java
ericbottard: So, to clarify, I'll get rid of the the `shouldTrigger()` call

pperalta: So would the contract read like this?

```
/**
 * For a given stream DSL text and {@link StreamDefinition},
 * <ul>
 *   <li>
 *     Generate {@link CompletionProposal}s that apply (if any)
 *     and add them to the provided {@code collector} list
 *   </li>
 *   <li>
 *     Return {@code true} if other strategies should be applied for
 *     the stream DSL text; {@code false} if this should be the last
 *     strategy applied
 *   </li>
 * </ul>
 *
 * @param text DSL text for the stream
 * @param streamDefinition stream definition
 * @param detailLevel integer representing the amount of detail to include
 * in the generated {@code CompletionProposal}s
 * @return {@code true} if other strategies should be applied for
 * the stream DSL text; {@code false} if this should be the last
 * strategy applied
 */
boolean addProposals(String text, StreamDefinition streamDefinition, int detailLevel, 
		List<CompletionProposal> collector);
```

spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/CompletionUtils.java
pperalta: I understand the desire to avoid dependencies between these projects. A few other things to consider:

1. The method name didn't help me understand what the method is actually doing. Perhaps we could name it `fromModuleCoordinates` or something along those lines?
2. What if the provided `coordinates` object already has a classifier? Is it our desire to ignore it and force it to be "exec" under all cases?

spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/CompletionUtils.java
pperalta: I see, that makes sense. Can we add some JavaDoc to indicate the format that we're expecting the module name to be in, and what the method returns if there is no need to auto generate a label?

spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/ExpansionStrategy.java
ericbottard: yes, except I would invert the meaning of the returned boolean

spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/CompletionUtils.java
pperalta: Makes sense, the uses are completely different. Could the method name and/or JavaDoc reflect that this is returning *potential* types for a module definition? When I read the method signature and existing JavaDoc, it implies that the module being passed can be of multiple types which isn't accurate.

spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/CompletionUtils.java
ericbottard: 1) np, will change
2) good point, did not even realize ModuleCoordinates had a classifier itself

spring-cloud-dataflow-admin/src/main/java/org/springframework/cloud/dataflow/admin/controller/CompletionController.java
pperalta: Can this be indicated in the JavaDoc?

spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/CompletionUtils.java
ericbottard: I think adding an example to the javadoc would make it clearer. Will do

spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/CompletionUtils.java
ericbottard: Will at least change the name, and try to enhance the javadoc

spring-cloud-dataflow-admin/src/main/java/org/springframework/cloud/dataflow/admin/controller/CompletionController.java
ericbottard: will do

spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/StacktraceFingerprintingRecoveryStrategy.java
pperalta: It does makes sense that you'd expect an "add..." method to accept the object that it is adding as an argument. At the same time, when I see a method called "compute" that returns void, it isn't clear to me what it is doing with the result that comes from the computation. IMO the fact that invoking this method mutates an internal data structure is not something that should be hidden. Perhaps `addFingerprintForException`? I'm open to other options.

ericbottard: `addFingerprintForException` sounds good to me

spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/CompletionUtils.java
markfisher: the method name should definitely imply *plural* return (e.g. determinePotentialType*s*)

spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/RecoveryStrategy.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.List;
+
+/**
+ * Used to provide completions on ill-formed stream definitions, after an initial (failed) parse.
+ * 
+ * @param <E> the kind of exception that is intercepted during parsing
+ * 
+ * @author Eric Bottard
+ */
+public interface RecoveryStrategy<E extends Exception> {
```
pperalta: Should we make the same modifications here as we did in `ExpansionStrategy`?

ericbottard: No, as we need to decide via stacktrace fingerprinting whether or not to trigger.
Moreover, I'm not that happy with the change we did in ExpansionStrategy. For some implementations it's ok, for others there is quite a cognitive burden at the start of the method to decide not to do anything

pperalta: Can you elaborate more on this:

> for others there is quite a cognitive burden at the start of the method to decide not to do anything

Is there an example you have in mind? I certainly don't want to make the code harder to understand! I had assumed that the `addProposals` implementations would simply invoke the logic that was previously in `shouldTrigger`.

ericbottard: they do, but when `addProposals()` is 15 lines long and 10 of those lines is for deciding whether or not to "trigger", it becomes less readable.

`PipeIntoOtherModulesExpansionStrategy` is the one I'm thinking about. not a big deal.

pperalta:
Merged as be5cd0b74dcb93566563c946cad8509e16c56117

PR #111, opened 2015-10-16
XD-3593: artifact registry

ericbottard:
~~This PR builds on top of https://github.com/spring-cloud/spring-cloud-dataflow/pull/112/~~
The meat of the functionality is 61516b4

This adds a new kind of "ModuleType" (renamed ArtifactType in the 3rd commit) named "library".
At deployment time, the `includes` parameter (if present) is scanned and anything that does not look like maven coordinates is tried against a prior registration of kind `library`.

This is the basic implementation. A more interesting one would be to scan for any property of time "MavenCoordinates" and apply the same logic. This would allow stuff like `jdbc --driver=mysql`, provided that the `jdbc` module declared its `driver` property as type `MavenCoordinates`. This needs better exposition of module metadata than what we have currently now though. cc @markfisher


markfisher:
@ericbottard could you please update this PR so that it no longer depends on #112? I think we want to revisit the refactoring in #112 in isolation, but I'd rather not hold this one up any longer than necessary.

ericbottard:
Made this PR independent of #112

markfisher:
@ericbottard I do not see any new commits

ericbottard:
No new commits, rather *less* commits.

Or are you referring to the ability to support arbitrary properties of type `ModuleCoordinates`? I'd like to think about it some more, and would make a different PR anyway

markfisher:
ok - got it... so you simply removed what was the middle commit?

ericbottard:
The _initial_ commit(s) actually, but yes

spring-cloud-dataflow-admin/src/main/java/org/springframework/cloud/dataflow/admin/config/AdminConfiguration.java
```diff
@@ -69,13 +69,13 @@ public TaskDefinitionRepository taskDefinitionRepository() {
 	}
 
 	@Bean
-	public ModuleRegistry moduleRegistry(RedisConnectionFactory redisConnectionFactory) {
-		return new RedisModuleRegistry(redisConnectionFactory);
+	public ArtifactRegistry moduleRegistry(RedisConnectionFactory redisConnectionFactory) {
```
markfisher: the method name should also change to "artifactRegistry"

spring-cloud-dataflow-admin/src/main/java/org/springframework/cloud/dataflow/admin/config/ModuleRegistryPopulator.java
```diff
@@ -53,18 +53,18 @@
...
-	 * Construct a {@code ModuleRegistryPopulator} with the provided {@link ModuleRegistry}.
+	 * Construct a {@code ModuleRegistryPopulator} with the provided {@link ArtifactRegistry}.
 	 *
-	 * @param moduleRegistry the {@link ModuleRegistry} to populate.
+	 * @param artifactRegistry the {@link ArtifactRegistry} to populate.
 	 */
-	public ModuleRegistryPopulator(ModuleRegistry moduleRegistry) {
-		Assert.notNull(moduleRegistry, "ModuleRegistry must not be null");
-		this.moduleRegistry = moduleRegistry;
+	public ModuleRegistryPopulator(ArtifactRegistry artifactRegistry) {
```
markfisher: This class name should also change.

spring-cloud-dataflow-admin/src/main/java/org/springframework/cloud/dataflow/admin/controller/TaskController.java
```diff
@@ -66,13 +66,13 @@
 	/**
 	 * The module registry this controller will use to look up modules.
```
markfisher: should update javadoc: "artifact registry"

spring-cloud-dataflow-admin/src/test/java/org/springframework/cloud/dataflow/admin/configuration/TestDependencies.java
```diff
@@ -35,8 +35,8 @@
 public class TestDependencies extends WebMvcConfigurationSupport {
 
 	@Bean
-	public ModuleRegistry moduleRegistry() {
-		return new InMemoryModuleRegistry();
+	public ArtifactRegistry moduleRegistry() {
```
markfisher: update bean method name

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/ArtifactType.java
```diff
@@ -17,10 +17,34 @@
...
+
+	/**
+	 * A module type to execute a short-lived process.
+	 */
+	task,
+
+	/**
+	 * A supporting library to a module, either as a single (jar) maven artifact, or as a bom.
+	 */
+	library,
```
markfisher: unnecessary comma after library

spring-cloud-dataflow-module-registry/src/main/java/org/springframework/cloud/dataflow/module/registry/ArtifactRegistration.java
```diff
@@ -17,40 +17,39 @@
...
+	 * Artifact type.
 	 */
-	private final ModuleType type;
+	private final ArtifactType type;
 
 	/**
-	 * Maven coordinates for module artifact.
+	 * Maven coordinates for the artifact.
 	 */
 	private final ModuleCoordinates coordinates;
```
markfisher: we should also rename `ModuleCoordinates` right?

spring-cloud-dataflow-module-registry/src/main/java/org/springframework/cloud/dataflow/module/registry/ArtifactRegistration.java
```diff
@@ -17,40 +17,39 @@
...
 
 	/**
-	 * Construct a {@code ModuleRegistration} object.
+	 * Construct a {@code ArtifactRegistration} object.
 	 *
 	 * @param name module name
```
markfisher: update javadoc "module" -> "artifact"

spring-cloud-dataflow-module-registry/src/main/java/org/springframework/cloud/dataflow/module/registry/ArtifactRegistry.java
```diff
@@ -0,0 +1,71 @@
...
+
+package org.springframework.cloud.dataflow.module.registry;
+
+
+import java.util.List;
+
+import org.springframework.cloud.dataflow.core.ArtifactType;
+
+/**
+ * {@code ArtifactRegistry} is used to manage module registrations, as well
```
markfisher: update javadoc: "module" -> "artifact"

spring-cloud-dataflow-module-registry/src/main/java/org/springframework/cloud/dataflow/module/registry/ArtifactRegistry.java
```diff
@@ -0,0 +1,71 @@
...
+public interface ArtifactRegistry {
+
+	/**
+	 * Look up the registration for an artifact by name and type. If
+	 * a registration does not exist, {@code null} is returned.
+	 *
+	 * @param name the artifact name
+	 * @param type the artifact type
+	 *
+	 * @return registration for a module/library, or {@code null} if not found
```
markfisher: just replace "module/library" with "artifact" in this class... it's meant to be completely generic at this level

spring-cloud-dataflow-module-registry/src/main/java/org/springframework/cloud/dataflow/module/registry/RedisArtifactRegistry.java
```diff
@@ -21,18 +21,18 @@
...
+ * {@link ArtifactRegistry} implementation backed by Redis.
  *
  * @author Patrick Peralta
  * @author Mark Fisher
  */
-public class RedisModuleRegistry implements ModuleRegistry {
+public class RedisArtifactRegistry implements ArtifactRegistry {
 
 	/**
 	 * Prefix for keys used for storing module coordinates.
```
markfisher: we should update the value of the redis key as well

spring-cloud-dataflow-module-registry/src/main/java/org/springframework/cloud/dataflow/module/registry/package-info.java
```diff
@@ -1,5 +1,5 @@
 /**
- * Root package of the ModuleRegistry support.
+ * Root package of the ArtifactRegistry support.
  */
 
 package org.springframework.cloud.dataflow.module.registry;
```
markfisher: we should update the package as well as project name

markfisher:
How will we register/list libraries (e.g. via the shell)? I assume we will add `LibraryCommands` parallel to `ModuleCommands` etc., and allow those to share the common `ArtifactRegistry` under the covers.

ericbottard:
> How will we register/list libraries (e.g. via the shell)? I assume we will add LibraryCommands parallel to ModuleCommands etc., and allow those to share the common ArtifactRegistry under the covers.

We can either change the current `module register` command to read `artifact register`, or add an alias to it (both would share the same help/description) or create 2 separate commands. I'd vote for option 1 or 2, personally

ericbottard:
To clarify, the alias/2nd command would read `library register`

ericbottard:
OK I rebased this following the merge of be5cd0b74dcb93566563c946cad8509e16c56117, renaming a few more occurrences (that was fun).

In addition to the shell command discussion, I wonder if we want to rename `moduleResolver` -> `artifactResolver`

ericbottard:
> How will we register/list libraries (e.g. via the shell)? I assume we will add LibraryCommands parallel to ModuleCommands etc., and allow those to share the common ArtifactRegistry under the covers.

As a side note, the `module list` command already automatically adds a "Library" column.

markfisher:
I noticed that... after these changes, `module` commands are actually *too* generic. That was what originally led to my question about new `library` commands. At the user level, those concepts should remain distinct. The fact that under the covers they share a generic *artifact* registry is an implementation detail that should not leak out at that level.

markfisher:
I think we need a separate command altogether. For example, when using `library register` there should be no `--type` option. For that same reason, I don't think the 2 commands can share help/description.

ericbottard:
OK, so what do you think about this:
* `module register` and `library register` being actually the same command, aliased
* `module list` showing only "modules"
* `artifact list` showing only libraries
* `module info` should show maven coordinates, in addition to configuration properties
* `artifact info` should show maven coordinates only

ericbottard:
Damn, our posts crossed. 2 separate commands it is then

ericbottard:
I'd make that a separate PR though, this one is already big

spring-cloud-dataflow-admin/src/main/java/org/springframework/cloud/dataflow/admin/controller/StreamController.java
```diff
@@ -80,7 +81,7 @@
 	/**
 	 * The module registry this controller will use to look up modules.
```
markfisher: this javadoc needs to be updated: "module" -> "artifact" and "look up modules" -> "look up modules and libraries"

markfisher:
I only had one more comment (minor javadoc) after a re-review. But I noticed that you did not change the package or project name yet either. I think it's probably clearer to do that after merging this one though. Agree?

ericbottard:
> I only had one more comment (minor javadoc) after a re-review. But I noticed that you did not change the package or project name yet either. I think it's probably clearer to do that after merging this one though. Agree?

Actually, I have. Not pushed yet

markfisher:
squashed and merged as https://github.com/spring-cloud/spring-cloud-dataflow/commit/5cb81c49a240304be14bcf5d724cfd36df403d39

PR #112, opened 2015-10-16
Factor out partition logic from StreamController.deploy() flow

ericbottard:


spring-cloud-dataflow-admin/src/main/java/org/springframework/cloud/dataflow/admin/controller/StreamController.java
```diff
@@ -222,34 +222,77 @@ private void deployStream(StreamDefinition stream, Map<String, String> cumulated
...
+
+		private ModulePartitionInfo(int instances, boolean isPartitioned) {
+			this.instances = instances;
+			this.isPartitioned = isPartitioned;
+		}
+
+		/**
+		 * Return an info suitable for the initial module, <i>i.e.</i> the first one to be deployed.
+		 */
+		private static ModulePartitionInfo initial() {
```
markfisher: I think "initial" also violates the clarity we're trying to achieve with regard to order (sounds like the FIRST module in a stream to me... as opposed to first to be deployed).

Also, we need to consider how we'll handle a situation where a stream definition ends with a named channel and yet partitioning is expected.

spring-cloud-dataflow-admin/src/main/java/org/springframework/cloud/dataflow/admin/controller/StreamController.java
```diff
@@ -222,34 +222,77 @@ private void deployStream(StreamDefinition stream, Map<String, String> cumulated
...
 			cumulatedDeploymentProperties = Collections.emptyMap();
 		}
-		Iterator<ModuleDefinition> iterator = stream.getDeploymentOrderIterator();
-		int nextModuleCount = 0;
-		boolean isDownStreamModulePartitioned = false;
-		for (int i = 0; iterator.hasNext(); i++) {
+
+		ModulePartitionInfo downstreamInfo = ModulePartitionInfo.initial();
+
+		for (StreamDefinition.DeploymentOrderIterator iterator = stream.getDeploymentOrderIterator();
```
markfisher: I think a `while` loop would be clearer here.

spring-cloud-dataflow-admin/src/main/java/org/springframework/cloud/dataflow/admin/controller/StreamController.java
```diff
@@ -222,34 +222,77 @@ private void deployStream(StreamDefinition stream, Map<String, String> cumulated
...
-			boolean isLast = (i == 0);
-			ModuleType type = determineModuleType(currentModule, isFirst, isLast);
-			ModuleRegistration registration = this.registry.find(currentModule.getName(), type);
-			if (registration == null) {
-				throw new IllegalArgumentException(String.format(
-						"Module %s of type %s not found in registry", currentModule.getName(), type));
-			}
-			ModuleCoordinates coordinates = registration.getCoordinates();
+
+			ModuleCoordinates coordinates = getModuleCoordinates(iterator, currentModule);
```
markfisher: should not pass the `iterator` itself... perhaps separate the `determineType` method, e.g.:

```
ModuleType type = determineType(currentModule, iterator.hasMoreUpstream(), iterator.hasMoreDownstream());
ModuleCoordinates coordinates = getModuleCoordinates(currentModule, type);
```

ericbottard:
@markfisher I did some additional changes, hopefully clarifying even further.

I guess the next step, especially to reduce the number of args to that method I introduced, would be to do 2 passes, the first one computing a `[LinkedHash]Map<ModuleDefinition, moduleDeploymentProperties>`

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/StreamDefinition.java
```diff
@@ -127,25 +127,21 @@ public String toString() {
...
+	 * to upstream (source) module.
+	 * Also prevents mutation of its backing data structure.
 	 */
-	private static class ReadOnlyIterator<T> implements Iterator<T> {
-		private final Iterator<T> wrapped;
-
-		public ReadOnlyIterator(Iterator<T> wrapped) {
-			this.wrapped = wrapped;
-		}
+	public class DeploymentOrderIterator implements Iterator<ModuleDefinition> {
```
markfisher: should this be static?, taking modules as a constructor arg

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/StreamDefinition.java
```diff
@@ -153,6 +149,28 @@ public void remove() {
...
+		public boolean hasMoreUpstream() {
+			return hasNext();
+		}
+
+		/**
+		 * Return the next module to deploy after the one currently pointed at, if any.
+		 * This does not advance this iterator.
+		 */
+		public ModuleDefinition peekUpstream() {
+			return i >= 0 ? modules.get(i) : null;
```
markfisher: nit: could call `hasMoreUpstream()` instead of `i >= 0` here (possibly this could evolve, and the `hasMoreUpstream()` method would always be accurate)

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/StreamDefinition.java
ericbottard: Or non static and referring to the outer field. Amounts to the same, really. What's the problem with non-static inner class?

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/StreamDefinition.java
ericbottard: Correct

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/StreamDefinition.java
```diff
@@ -153,6 +149,28 @@ public void remove() {
 			throw new UnsupportedOperationException();
 		}
 
+		/**
+		 * Return whether or not this iterator is past the very first module to deploy.
+		 */
+		public boolean hasMoreDownstream() {
```
ericbottard: I thought about calling this `hadMoreDownstream()` (had vs has), WDYT?

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/StreamDefinition.java
```diff
@@ -153,6 +149,28 @@ public void remove() {
...
+			return i < modules.size() - 2;
```
markfisher: I don't understand how this could work. If a stream has 2 modules e.g. "time | log" and `next()` has not yet been called, on this iterator, `i` would have a value of 1. In that case this method would return `false` which is *correct* since nothing has been passed yet. Then, however, if `next()` is called, `i` would have a value of 0, and this would still return `false`.

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/StreamDefinition.java
markfisher: Before these changes, the iterator instance actually delegated to a wrapped iterator, and that way the state was consistent with that wrapped iterator, even if there were more than one instance of the `ReadOnlyIterator`. With these changes, there could be more than one instance of this iterator, and each would have its own state for the position. Is that something we want to allow?

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/StreamDefinition.java
ericbottard: First of all, maybe we should add a check that this is called only once `next()` has been called at least once.

This is really about the position that is currently pointed to (ie the result of the last call to next). In your example, once next has been called once (yielding `log`), then there were no modules beyond that one. Remember those method names were for iterator consistency, but this is really `!isLast()`

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/StreamDefinition.java
ericbottard: Before, the wrapped iterator was always a new iterator anyway, and the caller could not interact with it directly (hence was forced to advance the wrapped iterator thru `ReadOnlyIterator `). This is the same here, or am I missing something?

markfisher: In general, I prefer avoiding non-static inner classes because of the tendency to increase coupling (given the ability of the inner class to access fields of the outer class). If the state that needs to be used from the outer class is simple enough to pass (one collection in this case), it's safer to do so and make the inner class static. That also facilitates passing an immutable view of the shared state, especially when a collection.

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/StreamDefinition.java
markfisher: It seems wrong to have a method which is only applicable after `next()` has been called. I think that's a red flag that we aren't expressing the right thing.

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/StreamDefinition.java
```diff
@@ -127,25 +127,21 @@ public String toString() {
 
 
 	/**
-	 * Iterator that prevents mutation of its backing data structure.
-	 *
-	 * @param <T> the type of elements returned by this iterator
+	 * Iterator that walks through ModuleDefinitions in deployment order, that is, from downstream (sink, if present)
+	 * to upstream (source) module.
```
markfisher: This javadoc would not be accurate in the case of named channels (at least not on the "source" end). For sink it does say "if present", but perhaps we should clarify the case where it would not be?

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/StreamDefinition.java
pperalta: In general I agree with passing along whatever state a new instance of a class needs into its constructor. Static inner classes also help avoid exposure of any of the parent class that isn't required.

However in this circumstance, the list is immutable and the iterator prevents mutation - so I’m OK with this non-static inner class having direct access to the list of modules. There is indeed a tighter coupling between the iterator and its parent stream definition, but I think that is appropriate here.

markfisher: Actually, I see no reason for `DeploymentOrderIterator` to be public in the first place (nor the corresponding `getter`). Can we make it private? (and non-static is fine)

ericbottard: It was made public to give access to its additional methods (like `hasMoreDownstream()`).
Since we've switched from enhanced for loop to while loop, I'd even make it **not** an iterator.

Actually, I feel like we're running in circles here, trying to shoehorn pretty complex logic (am I first?, am I last? are there named channels?) in the iterator pattern. And having vocabulary issues (the order being reversed, words like "first" and "last" become ambiguous) does not help.

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/StreamDefinition.java
```diff
@@ -127,25 +127,21 @@ public String toString() {
 
 
 	/**
-	 * Iterator that prevents mutation of its backing data structure.
-	 *
-	 * @param <T> the type of elements returned by this iterator
+	 * Iterator that walks through ModuleDefinitions in deployment order, that is, from downstream (sink, if present)
```
markfisher: change "walks through" -> "traverses"... no need to anthropomorphize ;)

spring-cloud-dataflow-admin/src/main/java/org/springframework/cloud/dataflow/admin/controller/StreamController.java
```diff
@@ -304,27 +354,22 @@ else if (isFirst) {
...
-		Iterator<ModuleDefinition> iterator = stream.getDeploymentOrderIterator();
-		while (iterator.hasNext()) {
-			ModuleDefinition module = iterator.next();
-			if (module.equals(currentModule) && iterator.hasNext()) {
-				ModuleDefinition prevModule = iterator.next();
-				Map<String, String> moduleDeploymentProperties = getModuleDeploymentProperties(prevModule, cumulatedDeploymentProperties);
-				return moduleDeploymentProperties.containsKey(BindingProperties.PARTITION_KEY_EXPRESSION) ||
-						moduleDeploymentProperties.containsKey(BindingProperties.PARTITION_KEY_EXTRACTOR_CLASS);
-			}
+	private boolean moduleHasPartitionInfo(ModuleDefinition module, Map<String, String> cumulatedDeploymentProperties) {
```
markfisher: for consistency with `isPartitionedConsumer(..)`, can we call this `isPartitionedProducer(..)`?

spring-cloud-dataflow-admin/src/main/java/org/springframework/cloud/dataflow/admin/controller/StreamController.java
```diff
@@ -222,34 +221,85 @@ private void deployStream(StreamDefinition stream, Map<String, String> cumulated
...
+		private final boolean isPartitioned;
+
+		private ModulePartitionInfo(int instances, boolean isPartitioned) {
+			this.instances = instances;
+			this.isPartitioned = isPartitioned;
+		}
+
+		/**
+		 * Return info suitable for handling partitioning for the last downstream module. If it is a regular sink, then
+		 * no partitioning happens. If it is a module that redirects to a named channel, then the number of partitions
```
markfisher: change "redirects" to "publishes"

spring-cloud-dataflow-admin/src/main/java/org/springframework/cloud/dataflow/admin/controller/StreamController.java
```diff
@@ -222,34 +221,85 @@ private void deployStream(StreamDefinition stream, Map<String, String> cumulated
...
+	 * to partitioning.
+	 *
+	 * @param cumulatedDeploymentProperties the whole set of stream deployment properties
+	 * @param downstreamInfo information about the downstream module in the stream
+	 * @param currentModule the module whose deployment properties we want to alter
+	 * @param upstreamModule the next module after {@literal currentModule} in deployment order, or null if none
+	 * @param moduleDeploymentProperties the deployment properties for the current module
+	 * @return information about the current module, to be used as 'previous' in the next invocation
+	 */
+	private ModulePartitionInfo postProcessPartitioningProperties(Map<String, String> cumulatedDeploymentProperties, ModulePartitionInfo downstreamInfo, ModuleDefinition currentModule, ModuleDefinition upstreamModule,  Map<String, String> moduleDeploymentProperties) {
```
markfisher: this method does too much - not only (potentially) updating the properties, but also returning a `ModulePartitionInfo` instance

ericbottard:
Closing for now, will revisit later.
One path to consider is to explicitly do 2 passes for deployment, the first one computing effective module deployment properties (as those are used throughout). Also, having the iteration logic (whatever its form, `Iterator` may be too decoupled after all) point at 2 modules at once seems to be a requirement: the actual module being deployed, and the one upstream.

PR #115, opened 2015-10-20
Update spring-cloud-starter version to snapshot

ilayaperumalg:
- This will let the spring-cloud-dataflow use the snapshot versions of spring-cloud-starters.
More importantly `spring-cloud-stream` when the module launcher is used.


markfisher:
This change has already been applied via https://github.com/spring-cloud/spring-cloud-dataflow/commit/be5cd0b74dcb93566563c946cad8509e16c56117#diff-600376dffeb79835ede4a0b285078036

PR #117, opened 2015-10-20
Don't show banner for local module deployer

ilayaperumalg:
- The spring-cloud-dataflow banner gets displayed when the local module deployer deploys the modules. We need to disable showing the dataflow banner for the modules.


markfisher:
LGTM; merged

PR #118, opened 2015-10-21
XD-3601 Set jmx domain name as group + label

ilayaperumalg:
- Set jmx domain name as `group + label` from the module deployment request
   - This will help avoid any domain name clash

### DIFF
--- a/spring-cloud-dataflow-module-deployers/spring-cloud-dataflow-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/dataflow/module/deployer/cloudfoundry/ApplicationModuleDeployer.java
+++ b/spring-cloud-dataflow-module-deployers/spring-cloud-dataflow-module-deployer-cloudfoundry/src/main/java/org/springframework/cloud/dataflow/module/deployer/cloudfoundry/ApplicationModuleDeployer.java
@@ -245,11 +245,11 @@ class ApplicationModuleDeployer implements ModuleDeployer {
 		args.put("modules", request.getCoordinates().toString());
 		args.putAll(ModuleArgumentQualifier.qualifyArgs(0, request.getDefinition().getParameters()));
 		args.putAll(ModuleArgumentQualifier.qualifyArgs(0, request.getDeploymentProperties()));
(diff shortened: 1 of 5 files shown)


ilayaperumalg:
This change depends on https://github.com/spring-cloud/spring-cloud-stream/pull/160

ericbottard:
Should we update the other deployers as well then? even though less critical

ilayaperumalg:
Sure, pushed the change to update all the deployers except YARN deployer. It looks like setting JMX domain name for the module deployment is more intrusive.

markfisher:
Shouldn't we be using the UPPERCASE_WITH_UNDERSCORE format for all but the `LocalModuleDeployer`?

ericbottard:
> Shouldn't we be using the UPPERCASE_WITH_UNDERSCORE format for all but the LocalModuleDeployer

I'd vote for the contrary, using `lowercase.dotted` as the canonical form, and having deployers selectively turn that into UPPERCASE_WITH_UNDERSCORE as needed. Moreover, there may be an issue with keys like `foo.barWizz` that turn to FOO_BAR_WIZZ (but you can't be sure on the other way)

ilayaperumalg:
Having small letters as the default is a good idea. But for this change, we are concerned about only `spring.jmx.default-domain` property.  I also think that for the deployers excluding `local` deployer, we need to having `_` instead of `.` and having smaller case should be fine. checking that.

ericbottard:
Taking the CF deployer as an example, it does take care of that (currently it uses a magical ENV var to pass all program arguments, but even if it did not, it's the responsibility of the deployer to convert _all_ of its parameters to the form it expects). So no need to pre-emptively pass underscores IMO

ilayaperumalg:
yeah, at least we need `underscore` for lattice environment. Verifying if that works and will update the PR.

ilayaperumalg:
That looks ok on `lattice`. Updated the PR.

ericbottard:
Dumb me deleted Ilaya's branch, and re-pushing does not seem to make Github happy.
Will recreate a PR (from Ilaya's repo if I can)

PR #119, opened 2015-10-21
Use 'exec' as the default qualifier when registering modules

ericbottard:
Will need to `FLUSHDB` your redis to remove all keys.

### DIFF
--- a/spring-cloud-dataflow-admin/src/main/java/org/springframework/cloud/dataflow/admin/config/ArtifactRegistryPopulator.java
+++ b/spring-cloud-dataflow-admin/src/main/java/org/springframework/cloud/dataflow/admin/config/ArtifactRegistryPopulator.java
@@ -53,6 +53,16 @@ public class ArtifactRegistryPopulator {
 	private static final String DEFAULT_VERSION = "1.0.0.BUILD-SNAPSHOT";
 
 	/**
+	 * Default classifier for default modules.
+	 */
+	private static final String DEFAULT_CLASSIFIER = "exec";
+
+	/**
+	 * Default extension for default modules.
+	 */
+	private static final String DEFAULT_EXTENSION = "jar";
+
+	/**
 	 * The {@link ArtifactRegistry} to populate.
 	 */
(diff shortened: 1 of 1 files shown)


spring-cloud-dataflow-admin/src/main/java/org/springframework/cloud/dataflow/admin/config/ArtifactRegistryPopulator.java
```diff
@@ -115,8 +125,8 @@ private void populateDefault(String name, ArtifactType type) {
 	 * @return default coordinates for the provided module
 	 */
 	private ArtifactCoordinates defaultTaskCoordinatesFor(String moduleName) {
-		return ArtifactCoordinates.parse(String.format("%s:%s:%s",
-				DEFAULT_TASK_GROUP_ID, moduleName, DEFAULT_VERSION));
+		return ArtifactCoordinates.parse(String.format("%s:%s:%s:%s:%s",
+				DEFAULT_TASK_GROUP_ID, moduleName, DEFAULT_EXTENSION, DEFAULT_CLASSIFIER, DEFAULT_VERSION));
```
mbogoevici: Should we use 'exec' for all kinds of artifacts - or is this affecting only modules ?

ericbottard: this is only affecting OOTB `stream` and `task` registrations.

Newly introduced libraries would typically be hand-registered, for now.

markfisher:
with this change can we remove the use of `DEFAULT_MODULE_CLASSIFIER` in the `ModuleLauncher`? (would be good, since currently it's impossible to use a module that does not use "exec") cc @mbogoevici

mbogoevici:
@markfisher we probably should - that assumption belongs at a higher level anyway

markfisher:
Agreed. And as long as we are populating the registry with those classifiers, I think we should remove the assumption at the `ModuleLauncher` level.

markfisher:
LGTM; merging

PR #120, opened 2015-10-21
Fix closed set value code completion

ericbottard:
When trying to complete `http --use.ssl=tr<TAB>`, we know that `true`
is the only possible value. `tr` is not a possible value, and hence no
other completion strategy shall be used until `true` has been typed in full.

### DIFF
--- a/spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/DefaultValueHintProvider.java
+++ b/spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/DefaultValueHintProvider.java
@@ -20,16 +20,27 @@ import java.util.List;
 
 import org.springframework.boot.configurationmetadata.ConfigurationMetadataProperty;
 import org.springframework.boot.configurationmetadata.ValueHint;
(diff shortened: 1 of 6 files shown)


spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/DefaultValueHintProvider.java
```diff
@@ -20,16 +20,27 @@
...
 
 	@Override
 	public List<ValueHint> guessValueHints(ConfigurationMetadataProperty property, ClassLoader classLoader) {
 		return property.getValueHints();
 	}
+
+	@Override
+	public boolean isExclusive(ConfigurationMetadataProperty property) {
+		for (ValueProvider valueProvider : property.getValueProviders()) {
+			if ("any".equals(valueProvider.getName())) {
```
ericbottard: see http://docs.spring.io/spring-boot/docs/current-SNAPSHOT/reference/htmlsingle/#_any

pperalta:
Merged as c2e11508aae16889d5bb0c8fd27aaef9f6b620f2

PR #121, opened 2015-10-21
Make code more readable

alexVengrovsk:


spring-cloud-dataflow-admin/src/main/java/org/springframework/cloud/dataflow/admin/completion/TapOnChannelExpansionStrategy.java
```diff
@@ -47,13 +47,12 @@ public boolean shouldTrigger(String dslStart, Exception exception) {
...
-		String streamAndModule = dsl.substring("tap:stream:".length());
-		int firstDot = streamAndModule.indexOf('.');
-		String streamName = streamAndModule;
+		String streamName = dsl.substring("tap:stream:".length());
 		String moduleName = "";
-		if (firstDot > 0) {
-			streamName = streamAndModule.substring(0, firstDot);
-			moduleName = streamAndModule.substring(firstDot + 1);
+		if (streamName.contains(".")) {
+			String[] splited = streamName.split("\\.")
```
markfisher: Thanks for the Pull Request. I would probably change "splited" (mis-spelled and/or not a common use of the adjective, which would normally be just "split" itself) to either "splits" or "tokens".

markfisher: If you're ok with that, I can make the change while merging. Otherwise LGTM.

alexVengrovsk: Yes, I am ok with your decision about change while merging.

markfisher:
LGTM; lost the original commit (because the fork's master was updated?) but committed these same changes via https://github.com/spring-cloud/spring-cloud-dataflow/commit/da40c9ef99173a2fb28588b1e09741c69c499201

PR #122, opened 2015-10-21
JavaDoc and various tweaks for ValueHintProvider

pperalta:


spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/ValueHintProvider.java
```diff
@@ -22,19 +22,34 @@
...
  * @author Eric Bottard
  */
 public interface ValueHintProvider {
 
-	List<ValueHint> guessValueHints(ConfigurationMetadataProperty property, ClassLoader classLoader);
+	/**
+	 * For a given property, return a list of {@link ValueHint} that may apply.
+	 *
+	 * @param property     property for which to generate value hints
+	 * @param classLoader  class loader for the artifact/module that this
```
pperalta: Tried to explain what this `ClassLoader` is for; hopefully this makes sense and the terminology is correct in this context.

spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/BooleanValueHintProvider.java
```diff
@@ -42,12 +42,10 @@
...
 	@Override
-	public List<ValueHint> guessValueHints(ConfigurationMetadataProperty property, ClassLoader classLoader) {
-		if ("java.lang.Boolean".equals(property.getType())) {
-			return BOOLEANS;
-		} else {
-			return Collections.emptyList();
-		}
+	public List<ValueHint> generateValueHints(ConfigurationMetadataProperty property, ClassLoader classLoader) {
+		return "java.lang.Boolean".equals(property.getType())
+				? Collections.unmodifiableList(BOOLEANS)
```
ericbottard: Can move the unmodifiable wrapping in the initializer. Will do on merge

ericbottard:
Merged with minor tweak as 143d5c5

spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/BooleanValueHintProvider.java
pperalta: +1

Issue #130, opened 2015-10-25
incorrect module deployment lookup - admin attempts to find a source instead of a processor


whummer:
When deploying the following pipe:
```
queue:in > groovy-transform --script=.../myscript.groovy --variables='...' > queue:out
```

I get an admin exception:
```
2015-10-24 23:01:42.967 ERROR 5 --- [io-9393-exec-18] o.s.c.d.a.c.RestControllerAdvice         : Caught exception while handling a request
java.lang.IllegalArgumentException: Module groovy-transform of type source not found in registry
        at org.springframework.cloud.dataflow.admin.controller.StreamController.deployStream(StreamController.java:236) ~[admin.jar!/:1.0.0.BUILD-SNAPSHOT]
        at org.springframework.cloud.dataflow.admin.controller.StreamController.deploy(StreamController.java:219) ~[admin.jar!/:1.0.0.BUILD-SNAPSHOT]
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[na:1.8.0_60]
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[na:1.8.0_60]
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[na:1.8.0_60]
        at java.lang.reflect.Method.invoke(Method.java:497) ~[na:1.8.0_60]
        at org.springframework.web.method.support.InvocableHandlerMethod.doInvoke(InvocableHandlerMethod.java:221) ~[spring-web-4.2.1.RELEASE.jar!/:4.2.1.RELEASE]
```

The exception indicates that the admin is looking for a "source" module, while it should in fact be looking for a "processor" module (because groovy-transform receives input data from the named channel `queue:in`).

Part of this could be an issue with the code in https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-admin/src/main/java/org/springframework/cloud/dataflow/admin/controller/StreamController.java#L231
```
			boolean isFirst = !iterator.hasNext();
			boolean isLast = (i == 0);
```
Shouldn't that be the other way around?

ericbottard:
I created https://github.com/spring-cloud/spring-cloud-dataflow/pull/131 for this.
Might be as simple as this (good for our later `deploy()` simplification @markfisher )

whummer:
[...]

Issue #128, opened 2015-10-23
Support source channel to sink channel mapping with no module (bridge)


whummer:
When deploying the following flow:
```
queue:simple1-out > queue:simple2-in
```

SCDF Admin fails with:
```
2015-10-23 09:22:02.235 ERROR 5 --- [nio-9393-exec-6] o.s.c.d.a.c.RestControllerAdvice         : Caught exception while handling a request
java.lang.IllegalArgumentException: Module bridge of type source not found in registry
        at org.springframework.cloud.dataflow.admin.controller.StreamController.deployStream(StreamController.java:235) ~[admin.jar!/:1.0.0.BUILD-SNAPSHOT]
        at org.springframework.cloud.dataflow.admin.controller.StreamController.save(StreamController.java:156) ~[admin.jar!/:1.0.0.BUILD-SNAPSHOT]
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[na:1.8.0_60]
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[na:1.8.0_60]
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[na:1.8.0_60]
        at java.lang.reflect.Method.invoke(Method.java:497) ~[na:1.8.0_60]
        at org.springframework.web.method.support.InvocableHandlerMethod.doInvoke(InvocableHandlerMethod.java:221) ~[spring-web-4.2.1.RELEASE.jar!/:4.2.1.RELEASE]
        at org.springframework.web.method.support.InvocableHandlerMethod.invokeForRequest(InvocableHandlerMethod.java:137) ~[spring-web-4.2.1.RELEASE.jar!/:4.2.1.RELEASE]
        at org.springframework.web.servlet.mvc.method.annotation.ServletInvocableHandlerMethod.invokeAndHandle(ServletInvocableHandlerMethod.java:111) ~[spring-webmvc-4.2.1.RELEASE.jar!/:4.2.1.RELEASE]
        at org.springframework.web.servlet.mvc.method.annotation.RequestMappingHandlerAdapter.invokeHandlerMethod(RequestMappingHandlerAdapter.java:806) ~[spring-webmvc-4.2.1.RELEASE.jar!/:4.2.1.RELEASE]
[...]

PR #131, opened 2015-10-26
Fix and simplify module type determination upon deployment

ericbottard:


spring-cloud-dataflow-admin/src/main/java/org/springframework/cloud/dataflow/admin/controller/StreamController.java
```diff
@@ -286,37 +284,19 @@ private ModuleDefinition postProcessLibraryProperties(ModuleDefinition module) {
...
-				// stream begins with a named channel in the source position
-				type = ArtifactType.processor;
-			}
-			else { 
-				type = ArtifactType.source;
-			}
-		}
-		else {
-			type = ArtifactType.processor;
+	private ArtifactType determineModuleType(ModuleDefinition moduleDefinition) {
```
ilayaperumalg: Can this method be changed to `determineArtifactType` ?

ericbottard: I don't think that would make sense, as it only applies to modules (and stream modules, for that matter)

ilayaperumalg: ok, my reasoning was that it determines and returns `ArtifactType` and I believe there is no module type now ;)

spring-cloud-dataflow-admin/src/main/java/org/springframework/cloud/dataflow/admin/controller/StreamController.java
```diff
@@ -286,37 +284,19 @@ private ModuleDefinition postProcessLibraryProperties(ModuleDefinition module) {
...
+		// Parser has already taken care of source/sink named channels, etc
+		boolean hasOutput = moduleDefinition.getParameters().containsKey(BindingProperties.OUTPUT_BINDING_KEY);
+		boolean hasInput = moduleDefinition.getParameters().containsKey(BindingProperties.INPUT_BINDING_KEY);
+		if (hasInput && hasOutput) {
+			return ArtifactType.processor;
+		} else if (hasInput) {
```
markfisher: newlines after closing braces please

spring-cloud-dataflow-admin/src/main/java/org/springframework/cloud/dataflow/admin/controller/StreamController.java
markfisher: The "module" types (source, processor, and sink) are still relevant as a logical type within a stream, and they are a subset of artifact types (which currently includes "task" and "library" as well). I agree with Eric that this method should be clear that it only returns module types.

markfisher:
As part of this PR, please add a unit test to cover the case that led to the issue being reported (e.g. a stream definition such as: "queue:x > someProcessor > queue:y"). The issue is https://github.com/spring-cloud/spring-cloud-dataflow/issues/130

whummer:
@markfisher the issue you are referring to is #128. The problem there is that the `bridge` module does not exist. Not sure if #128 is related to #130 (although it would be great to have both solved :) ). thanks

ericbottard:
Added test, fixed formatting

markfisher:
LGTM; merging

markfisher:
squashed and merged as https://github.com/spring-cloud/spring-cloud-dataflow/commit/4238a95ac842627a6dc15ed9d7163164af6c93b7

PR #134, opened 2015-11-03
[Do not merge yet] XD-3672: Mesos/Marathon Deployer implementation

ericbottard:

PR #132, opened 2015-10-29
[Do not merge yet] Add a Kubernetes Deployer implementation

ericbottard:
This builds on top of https://github.com/spring-cloud/spring-cloud-dataflow/pull/113 where a lot of discussion has already happened.

My commit decouples the previous implementation from hardcoded redis requirement by introducing Spring Cloud Connector for Kubernetes.

Here are the things I think that should still be addressed, not necessarily as part of this PR:
* this deployer, as the Marathon one, does not support `module.index`. Maybe fail deployment in case partitioning is requested
* The `kubernetesClient` object is currently created with the simple constructor, using configuration that is drawn from ENV_VARS and system props (like documented [here](https://github.com/fabric8io/kubernetes-client)). We should rather use the constructor that takes an explicit `Config` and use boot `@ConfigurationProperties`, for consistency
* The original PR by @frosenberg added an IT test, but this requires a running kubernetes cluster. Execution of that test should be made optional thanks to a JUnit rule (test is currently disabled)

Apart from that, I think this is now in a state that can sustain further review, and is a good candidate for merge

PR #135, opened 2015-11-04
[Do not merge yet] XD-3616 Pass module resolver props via dataflow admin

ilayaperumalg:
- This PR adds the ability to pass `module-resolver` properties from `spring-cloud-dataflow-admin`.
The module resolver properties would have a prefix `moduleResolver.` in the deployment properties when provided during `deploy` operation.
For instance, a stream deployment command could override the module resolver properties like this:
  `stream deploy <stream> --properties "module.time.moduleResolver.remoteRepositories=repo1,module.time.moduleResolver.localRepository=/tmp/m2,module.log.moduleResolver.offline=true"`

Since the `moduleResolver` configuration is setup at the dataflow-admin `startup` itself, we can not rely to simple property override for the `moduleResolver` properties.
 - There were few road blocks while addressing this in few different approaches (especially in local deployer):
   - Spawning separate process for `ModuleLauncherApplication` from `LocalModuleDeployer` needed to have its own classpath and process management (for deploying/un-deploy/destroy)
   - The oracle-tools runtime library seemed to be a better option; but the classpath management within the dataflow-admin -> module-launcher -> oracle-tools gets complicated.

The current approach explicitly sets the module resolver properties via `ModuleResolver` implementation.

This resolves spring-cloud/spring-cloud-dataflow#161


ilayaperumalg:
This PR depends on https://github.com/spring-cloud/spring-cloud-stream/pull/195

ericbottard:
I don't understand this approach. Those configuration properties are not module related IMO, they're really a way to always pass the same props to the module launcher. I would have expected something in the line of https://github.com/spring-cloud/spring-cloud-dataflow/pull/134/files#diff-67a657831e9e3bfc5b7db7dedf282f35R45 and https://github.com/spring-cloud/spring-cloud-dataflow/pull/132/files#diff-896933cba3771b5af8750c3c7f4d8b24R86, only streamlined somehow.

What am I missing?

ilayaperumalg:
First of all, this PR addresses the way to pass `ModuleResolver` properties only as any other module **launcher** properties can be provided as part of `DeploymentProperties` already. For instance, `module.<label>.<anyLauncherProperty>`. 

> Those configuration properties are not module related IMO, they're really a way to always pass the same props to the module launcher

I believe they are module related as one can choose module resolver properties for each module that get deployed. Also, we need a way to pass these module resolver properties via dataflow-admin and passing them as deployment properties with prefix `module.<label>.moduleResolver.*` is one possible way.

ericbottard:
Those props are not only for the moduleResolver I agree, but for that particular case, why would anyone want to use different repositories per module?

ilayaperumalg:
It is just an option ;) Most probably one would end up doing `module.*.moduleResolver.<property>, to override the `ModuleResolver` properties.

ilayaperumalg:
Updated the PR to support passing module launcher/resolver properties at startup level which can be overridden by the properties passed via `deployment` properties. Also, dropped the support to pass these properties for `local` deployer.

spring-cloud-dataflow-module-deployers/spring-cloud-dataflow-module-deployer-spi/src/main/java/org/springframework/cloud/dataflow/module/deployer/ModuleDeployer.java
```diff
@@ -38,6 +38,8 @@
 
 	public static final String JMX_DEFAULT_DOMAIN_KEY  = "spring.jmx.default-domain";
 
+	public static final String MODULE_LAUNCHER_PROPERTY_PREFIX = "launcher.";
```
ericbottard: Add some javadoc to explain how this is being used (e.g. with a concrete example)

spring-cloud-dataflow-module-deployers/spring-cloud-dataflow-module-deployer-local/src/main/java/org/springframework/cloud/dataflow/module/deployer/local/LocalModuleDeployer.java
```diff
@@ -52,6 +55,9 @@
 
 	private final RestTemplate restTemplate = new RestTemplate();
 
+	@Autowired
+	ModuleResolver moduleResolver;
```
ericbottard: Changes to this class seem to be from the previous attempt, as nothing is being done with the resolver, am I correct?

spring-cloud-dataflow-module-deployers/spring-cloud-dataflow-module-deployer-lattice/src/main/java/org/springframework/cloud/dataflow/module/deployer/lattice/TaskModuleDeployer.java
```diff
@@ -76,7 +81,20 @@ public ModuleDeploymentId deploy(ModuleDeploymentRequest request) {
 		for (Map.Entry<String, String> entry : request.getDefinition().getParameters().entrySet()) {
 			environmentVariables.add(new EnvironmentVariable(entry.getKey(), entry.getValue()));
 		}
-
+		// Module launcher properties from Deployment properties
```
ericbottard: All these duplicated handling may be moved to the SPI layer, to some helper method.
Maybe by turning `ModuleDeployer` into an abstract class + injecting the properties.

markfisher: -1 for turning `ModuleDeployer` into an abstract class

ericbottard: This would also make sure that we have consistent handling of need-to-be-qualified properties and don't-need-to-be-qualified props, IMO

ericbottard: > -1 for turning ModuleDeployer into an abstract class

Or *introducing* an abstract base class `implements ModuleDeployer`

markfisher: It's important that the SPI remain interface based. We should reconsider the problem if inheritance seems like the solution.

pperalta: I agree with @markfisher here - I do not believe that turning `ModuleDeployer` into an abstract class is a good idea. If there is a common way of translating deployment properties into module launcher properties, that functionality should exist in its own class, and the SPI implementations can use that to perform the translation.

ilayaperumalg: Does anyone disagree on *not* to support passing module resolver/launcher properties via deployment properties?
It might be ok just to have the support to set the module resolver/launcher properties for the spring-cloud-dataflow-adming at the startup level.

ilayaperumalg:
Updated the PR after admin/deployer SPI splits. Now that `ModuleLauncherProperties` is moved to `module-deployer-spi`. Subsequent PR is [submitted](https://github.com/spring-cloud/spring-cloud-dataflow-admin-lattice/pull/3) for `lattice`. Once we agree on the approach, I will add the PRs for other deployers as well.

markfisher:
The `ModuleDeploymentRequest` already has deployment properties:

```
public Map<String, String> getDeploymentProperties() {
    return deploymentProperties;
}
```

I'm thinking we could add `AdminConfigurationProperties` that includes something like `defaultDeploymentProperties`. Those would be overridable, but the full set of deployment properties should be passed along in the appropriate way for any deployment environment anyways (e.g. env vars for a cloud or container based runtime).

ericbottard:
Please note that `deploymentProperties` above apply to a **module** (hence they are [qualified](https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-module-deployer-spi/src/main/java/org/springframework/cloud/dataflow/module/deployer/ModuleArgumentQualifier.java#L47)), while **resolver** properties are really props that target the whole Launcher (hence, **not** qualified)

ilayaperumalg:
I think we need to decide/confirm if we really want to support passing module resolver/launcher properties via module deployment properties. If we are ok with having an option to specify module resolver/launcher properties **per** `deployer` at startup, then concerns around setting via deployment properties **per** `module` can be neglected.

markfisher:
My two concerns are: 1) to an end user the fact we have a "launcher" for a module should be completely opaque and 2) we hope to remove the launcher altogether, likely pushing more responsibility to the admin side. If the main driver for this is simply the Maven repository location, I would prefer that to at least feel like an Admin configuration property from the user's perspective, and I don't think it needs to be configurable on a per-deployment basis (it could be dynamically refreshed, or have a new repo added to the list, via cloud config - but it really should map 1:1 with the Admin). Interested in any feedback from @mbogoevici as well, since he's probably done the most thinking about a post-launcher model.

mbogoevici:
I don't think that we have a specific need to supply/override resolver settings per deployment, it just so happens that with the launcher model, they need to be included with each module launch. I agree with the general direction that @markfisher was suggesting, in that, regardless of the mechanism used to pass them on, the user should set them only at the admin level.

ilayaperumalg:
In that case, the PR at its current state addresses `launcher/resolver` properties to be considered as admin level property.

mbogoevici:
If we passed such properties through the `ModuleDeploymentRequest`, even if as part of a separate set of properties, then we would establish a firm contract around the property names, e.g. `remoteRepositories` between the administrator and the deployer, and it's not very clear to me how that would be better than just injecting an object that contains some pre-parsed administrator configuration into the deployers (which is what the current PR does) - and then the individual implementations can choose to act on these properties any way they see fit. To note, we've already established some implicit contracts around the `includes` and `excludes` which imply the underlying use of a module launcher.

markfisher:
I would prefer the admin level configuration properties be explicit, e.g. just a single property for the repository locations.

mbogoevici:
I don't think we should have a separate set of properties for the module launcher though - just a generic `AdminProperties` with `repositoryLocations` and other relevant stuff.

ilayaperumalg:
ok, we can explicitly set these properties as separate **admin** configuration properties:

```
localRepository
RemoteRepositories
offline
```

ericbottard:
I would expect to be able to pass free form properties to the launcher. I know we don't want to couple scs with scd, but duplicating that Properties class gives us the worst of two worlds IMO: coupling without safety. So I'd rather go the complete "un-couple" way and allow anything to be passed to the launcher.

markfisher:
My point is that "launcher" as it exists today is going away. We should not be thinking about passing "free form" properties to it. One of the ideas we're considering is that the repository config is purely Admin-side in the near future.

ericbottard:
Also, I understand we don't want to disclose too much that there is a ModuleLauncher at play, but these are not really "AdminProperties" per se. For example, if running the admin on a laptop and targeting a behind-corporate-firewall PCF install, one has to understand that e.g. the remoteRepositories configuration applies from the point of view of the PCF install...

mbogoevici:
I don't really like the notion of freeform properties, even if I've been initially tempted by it, for the reason that producing them requires some knowledge about how they are going to be interpreted downstream, which one way or another couples the admin server to the module launcher's behaviour (even if it's something minimal like a `remoteRepositories` property). I think that they are `AdminProperties` in the sense that they configure a special aspect of executing the modules, i.e. for a given SCDF installation, this is the repository where artifacts are resolved from. I'm not sure how the firewall comes into discuss.

ilayaperumalg:
Updated the PR to use admin configuration properties for the module resolver properties. Those properties would still be parsed as `String` values as they are command-line args for the module launcher. The associated change is [made] (https://github.com/spring-cloud/spring-cloud-dataflow-admin-cloudfoundry/pull/5) in CF deployer repo.

ericbottard:
There are a couple of things I don't understand with the current state of this PR (1d6d86f70686808e97c5e254034a4975da7faf5a):
* It seems for the local case, the properties are not used at all (or are they because they have the exact same name as the ones for the launcher, and as such are visible? This is confusing at best)
* I think we could still have rich (typed) properties on the `@ConfigurationProperties` class (that's the point of that annotation after all) even though they are converted back to Strings.
* That conversion should rather happen in the class itself for example, rather than via a static method in a separate class.

ilayaperumalg:
From the beginning, we excluded the `local` case for this PR. 
We can definitely add `typed` properties for the configuration properties (like what we have in module launcher). My reasoning was not to force any validations (like File etc.,) at the `admin` side as these properties are going to be for the target deployer environment.

ericbottard:
Ah, I see what you mean. I would still vote in favor of typed props though, especially regarding the remote repositories. Their syntax is likely to become more structured, for example when we support authentication, etc. WDYT ?

ilayaperumalg:
Yeah,  +1 for adding more validation against `remoteRepositories` once we have a syntax for authentication etc.,

ericbottard:
Reworked a bit and merged as 0b75488

ilayaperumalg:
This resolves https://github.com/spring-cloud/spring-cloud-dataflow/issues/161

PR #137, opened 2015-11-05
Added contributing section

jadekler:


ericbottard:
Thanks for this useful addition. However, I believe it would be better suited in a file named `CONTRIBUTING.md` (see https://github.com/blog/1184-contributing-guidelines)

jadekler:
@ericbottard Per conv with @dsyer, going to re-PR as src/main/asciidoc to generate a .adoc (same as most of the other spring-cloud projects). Let me know if you'd like me to do so differently.

PR #144, opened 2015-11-12
Use Assert instead of if-check

alexVengrovsk:
[Review on Reviewable](https://reviewable.io/reviews/spring-cloud/spring-cloud-dataflow/144)


markfisher:
cherry picked and merged as https://github.com/spring-cloud/spring-cloud-dataflow/commit/66b7902c923f330301515ff35271684da3363db7

Issue #148, opened 2015-11-14
Add support to build Admin with individual SPI deployers


sabbyanandan:
As a s-c-d developer, I'd like to break the build lifecycle to bundle SPI deployers optionally, so I don't have to build *admin* with all the deployer variations as one whole thing.

**Acceptance:**
* *admin* project build includes option to build by SPI deployer
* *Local* deployer is included by default

PR #168, opened 2015-11-17
split Admin SPI from implementations

markfisher:
modified spring-cloud-dataflow-admin to be a starter

all SPI implementations other than local will now be in their own repos:
* https://github.com/spring-cloud/spring-cloud-dataflow-admin-cloudfoundry
* https://github.com/spring-cloud/spring-cloud-dataflow-admin-lattice
* https://github.com/spring-cloud/spring-cloud-dataflow-admin-yarn

for now the local SPI and the shell remain in this repo

also removed spring-cloud-dataflow-yarn (will move to the yarn admin repo)

This resolves spring-cloud/spring-cloud-dataflow#148


spring-cloud-dataflow-yarn/spring-cloud-dataflow-yarn-container/pom.xml
```diff
@@ -1,72 +0,0 @@
...
-		<version>1.0.0.BUILD-SNAPSHOT</version>
-	</parent>
-
-	<dependencies>
-		<dependency>
-			<groupId>org.springframework.data</groupId>
-			<artifactId>spring-yarn-boot</artifactId>
-			<version>2.3.0.M2</version>
-			<exclusions>
-<!--
```
pperalta: Shall I remove this on merge?

spring-cloud-dataflow-admin-local/pom.xml
```diff
@@ -0,0 +1,71 @@
...
+		<plugins>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-resources-plugin</artifactId>
+			</plugin>
+			<plugin>
+				<groupId>org.springframework.boot</groupId>
+				<artifactId>spring-boot-maven-plugin</artifactId>
+			</plugin>
+			<!--
```
pperalta: Should this be removed on merge?

spring-cloud-dataflow-yarn/spring-cloud-dataflow-yarn-container/pom.xml
markfisher: this file is deleted altogether

spring-cloud-dataflow-admin-local/pom.xml
markfisher: we might want to add Docker support later, but we have an example in the lattice admin now, so this could go away until we're ready to add it

spring-cloud-dataflow-yarn/spring-cloud-dataflow-yarn-container/pom.xml
pperalta: Opps...my bad 😛

spring-cloud-dataflow-admin-local/pom.xml
pperalta: OK, if it is something that is likely to come back then we can keep it for now; just wanted to make sure it wasn't left over by mistake.

PR #169, opened 2015-11-18
Fix module-deployers-spi parent

ilayaperumalg:
- use `dataflow-parent` instead of `deployers-parent` which no longer exists.


markfisher:
LGTM; merged

PR #138, opened 2015-11-06
[Do not merge yet] Add a simple out-of-process local deployer

ericbottard:
Discussed this a bit with Ilaya.
This turned out to be pretty easy to implement.

Say goodbye to CL issues, and hello to partitioning support..

```
dataflow:>stream create foo --definition "http | log"
Created new stream 'foo'
dataflow:>stream deploy foo --properties "module.http.partitionKeyExpression=payload.length(), module.log.count=3"
Deployed stream 'foo'
dataflow:>runtime modules
╔═══════════════════════╤═══════════╤════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════╗
║Module Id / Instance Id│Unit Status│                                                Nb of Instances / Attributes                                                ║
╟───────────────────────┼───────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╢
║foo.http               │ deployed  │                                                             1                                                              ║
╟┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┼┈┈┈┈┈┈┈┈┈┈┈┼┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈╢
║                       │           │working.dir = /var/folders/ty/nnp8nms96w76mxhdpmbdgcx00000gn/T/foo.http-0-1029941279105341630                               ║
║foo.http-0             │ deployed  │     stdout = /var/folders/ty/nnp8nms96w76mxhdpmbdgcx00000gn/T/foo.http-0-1029941279105341630/stdout_7418152489030773686.log║
║                       │           │     stderr = /var/folders/ty/nnp8nms96w76mxhdpmbdgcx00000gn/T/foo.http-0-1029941279105341630/stderr_6712135003765518816.log║
║                       │           │        url = http://192.168.0.24:26527                                                                                     ║
[...]

PR #172, opened 2015-11-19
Add Out-Of-Process local ModuleDeployer

ericbottard:
This replaces https://github.com/spring-cloud/spring-cloud-dataflow/pull/138

### DIFF
--- a/spring-cloud-dataflow-admin-local/src/main/java/org/springframework/cloud/dataflow/admin/spi/local/LocalConfiguration.java
+++ b/spring-cloud-dataflow-admin-local/src/main/java/org/springframework/cloud/dataflow/admin/spi/local/LocalConfiguration.java
@@ -16,6 +16,8 @@
 
 package org.springframework.cloud.dataflow.admin.spi.local;
 
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.cloud.dataflow.module.deployer.ModuleDeployer;
 import org.springframework.cloud.stream.module.launcher.ModuleLauncher;
 import org.springframework.cloud.stream.module.launcher.ModuleLauncherConfiguration;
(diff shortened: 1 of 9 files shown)


spring-cloud-dataflow-admin-local/src/main/java/org/springframework/cloud/dataflow/admin/spi/local/LocalConfiguration.java
```diff
@@ -25,21 +27,42 @@
...
 @Configuration
-@Import(ModuleLauncherConfiguration.class)
 public class LocalConfiguration {
 
-	@Bean
-	public ModuleDeployer processModuleDeployer(ModuleLauncher moduleLauncher) {
-		return new LocalModuleDeployer(moduleLauncher);
+	static final String LOCAL_DEPLOYER_PREFIX = "deployer.local";
+
+	@ConditionalOnProperty(prefix = LocalConfiguration.LOCAL_DEPLOYER_PREFIX, name = "out-of-process", havingValue = "false", matchIfMissing = true)
```
mbogoevici: I think that the default should be out-of-process, and allow in-proces to be used as an option for debugging, etc.

ericbottard:
Updated to make out of process the default

spring-cloud-dataflow-admin-local/src/main/java/org/springframework/cloud/dataflow/admin/spi/local/OutOfProcessModuleDeployer.java
```diff
@@ -0,0 +1,233 @@
...
+			}
+			running.remove(id);
+		}
+	}
+
+	@Override
+	public ModuleStatus status(ModuleDeploymentId id) {
+		List<Instance> instances = running.get(id);
+		if (instances == null) {
+			return null;
```
mbogoevici: We shouldn't return `null` here, probably `unknown` until https://github.com/spring-cloud/spring-cloud-dataflow/pull/194 is merged, when it would be `undeployed`. In particular, this results in an NPE for streams that are created but not deployed.

spring-cloud-dataflow-admin-local/src/main/java/org/springframework/cloud/dataflow/admin/spi/local/OutOfProcessModuleDeployer.java
```diff
@@ -0,0 +1,233 @@
...
+			args.put(SERVER_PORT_KEY, String.valueOf(port));
+
+			ProcessBuilder builder = new ProcessBuilder(properties.getJavaCmd(), "-jar", moduleLauncherPath);
+			builder.environment().clear();
+			builder.environment().putAll(args);
+
+			try {
+				Path workDir = Files.createTempDirectory(properties.getWorkingDirectoriesRoot(), moduleDeploymentId.toString() + "-" + i + "-");
+				if (properties.isDeleteFilesOnExit()) {
+					workDir.toFile().deleteOnExit();
```
mbogoevici: This requires registering files for deletion too: see below.

As an improvement - I think that we can create a single directory per module deployment (rather than per instance) and create separate files per instance id inside that directory - thus being able to do something like `tail -f stdout_*` to aggregate the output of multiple modules.

spring-cloud-dataflow-admin-local/src/main/java/org/springframework/cloud/dataflow/admin/spi/local/OutOfProcessModuleDeployer.java
```diff
@@ -0,0 +1,233 @@
...
+
+		private final File stderr;
+
+		private final URL moduleUrl;
+
+		private Instance(ModuleDeploymentId moduleDeploymentId, int instanceNumber, ProcessBuilder builder, Path workDir) throws IOException {
+			this.moduleDeploymentId = moduleDeploymentId;
+			this.instanceNumber = instanceNumber;
+			builder.directory(workDir.toFile());
+			builder.redirectOutput(this.stdout = Files.createTempFile(workDir, "stdout_", ".log").toFile());
```
mbogoevici: Need to register the file to `deleteOnExit()` or else the directory will not be deleted either.

ericbottard:
Addressed review comments

mbogoevici:
LGTM - another thought while testing would be to provide an option to shutdown the child processes when the admin exits (although I can't see right now a reason why we wouldn't want that all the time, since a newly arrived admin cannot reassert control over the launched processes anyway). Will create a separate issue.

mbogoevici:
Rebased, merged w/https://github.com/spring-cloud/spring-cloud-dataflow/commit/be29384a8536541ec16e489dd3eb7d1a0d2d1cee

Comments to be addressed separately

ericbottard:
The child processes are already killed when admin exits.
Le 3 déc. 2015 7:30 PM, "Marius Bogoevici" <notifications@github.com> a
écrit :

mbogoevici:
That is only true when the process exits as a result of SIGINT (Ctrl-C) and the entire process group is killed by the OS, but not as a result of SIGTERM (kill). This can be easily reproduced by doing a `kill PID` on the admin server. We should actually rely on the destruction of `OutOfProcessDeployer` to destroy the child processes (optionally, perhaps).

mbogoevici:
But relying on the shutdown hook is currently blocked by https://github.com/spring-cloud/spring-cloud-commons/issues/62.

mbogoevici:
Actually, the child processes don't exit even when doing `kill -s SIGINT pid` - it really happens only on a Ctrl-C.

ericbottard:
> That is only true when the process exits as a result of SIGINT (Ctrl-C) 
Ah, did not realize that. Seems you're on top of it though :-)

PR #175, opened 2015-11-20
Apply new Spring Shell tables consistently

ericbottard:


spring-cloud-dataflow-shell/src/main/java/org/springframework/cloud/dataflow/shell/command/ModuleCommands.java
```diff
@@ -84,38 +88,37 @@ public String info(
...
-					.addHeader(3, new TableHeader("Default"))
-					.addHeader(4, new TableHeader("Type"));
-			for (ConfigurationMetadataProperty o : options) {
-				final TableRow row = new TableRow();
-				row.addValue(1, o.getId())
-						.addValue(2, o.getDescription() == null ? "<unknown>" : o.getDescription())
-						.addValue(3, prettyPrintDefaultValue(o))
-						.addValue(4, o.getType() == null ? "<unknown>" : o.getType());
-				table.getRows().add(row);
+			TableModelBuilder<Object> modelBuidler = new TableModelBuilder<>();
```
markfisher: typo: "Buidler"

markpollack:
```
dataflow:>stream list
╔═══════════╤═════════════════╤════════╗
║Stream Name│Stream Definition│ Status ║
╠═══════════╪═════════════════╪════════╣
║ticktock   │time | log       │deployed║
╚═══════════╧═════════════════╧════════╝
```

nice!

markpollack:
Why does the command for module info reverse the arguments.  In XD it was

```
xd:>module info --name source:http
```
but now it is

```
dataflow:>module info --name http:source
```

markpollack:
The table doesn't seem to expand to fill all available width.  I remember seeing that is was possible in the initial work.  For example - [screenshot](https://gist.githubusercontent.com/markpollack/13ae32e4c95591c0c9c1/raw/9e3bfda84176da03caaf60af1ce25d12fe24fe66/dataflow-shell.png)

ericbottard:
The current implementations don't **expand** to fill available width. They try to expand to fit content, and if they can't, they can shrink to accomodate content.
Forcing to fill width should be easy by setting the desired width to Integer.MAX_VALUE (while retaining the minimum acceptable width to its correct, current value).
Not sure we want to apply this to all tables though

markpollack:
It seems like in the screen shot above the first column could have been just one or two columns wider to avoid wrapping just 1-2 letters in the next row.  Should we create another issue to look into this?

ericbottard:
> It seems like in the screen shot above the first column could have been just one or two columns wider to avoid wrapping just 1-2 letters in the next row. Should we create another issue to look into this?

I don't see wrapping happening above. Are you referring to 
```
dataflow:>module list
╔══════════════╤════════════════╤══════════╤════╤═══════╗
║    source    │   processor    │   sink   │task│library║
╠══════════════╪════════════════╪══════════╪════╪═══════╣
║ftp           │groovy-filter   │counter   │    │       ║
║http          │groovy-transform│file      │    │       ║
║load-generator│noop            │ftp       │    │       ║
║sftp          │transform       │gemfire   │    │       ║
║time          │                │hdfs      │    │       ║
║twitterstream │                │jdbc      │    │       ║
║              │                │log       │    │       ║
║              │                │redis     │    │       ║
║              │                │throughput│    │       ║
║              │                │websocket │    │       ║
║              │                │          │    │       ║
╚══════════════╧════════════════╧══════════╧════╧═══════╝
```

?

ericbottard:
Scratch that. Missed your attached screenshot, will take a look

ericbottard:
Yep, seems like a bug. If you could narrow how to reproduce, that would be cool. May need to output the perceived terminal width. Or maybe this is a rounding error (thought about this, but never witnessed it)

sabbyanandan:
I absolutely love the new grid layout!

```
╔══════════════╤════════════════╤══════════╤════╤═══════╗
║    source    │   processor    │   sink   │task│library║
╠══════════════╪════════════════╪══════════╪════╪═══════╣
║ftp           │groovy-filter   │counter   │    │       ║
║http          │groovy-transform│file      │    │       ║
║load-generator│noop            │ftp       │    │       ║
║sftp          │transform       │gemfire   │    │       ║
║time          │                │hdfs      │    │       ║
║twitterstream │                │jdbc      │    │       ║
║              │                │log       │    │       ║
║              │                │redis     │    │       ║
║              │                │throughput│    │       ║
║              │                │websocket │    │       ║
║              │                │          │    │       ║
╚══════════════╧════════════════╧══════════╧════╧═══════╝
```

markpollack:
Find me on skype and we can screenshare, that will be better to help debug together.

ericbottard:
Found why the first column wraps: I hard coded a max size of 30 for all columns, because some of them would otherwise report a non-breakable minimum width of ~70, which would make the table illegible in many cases.

Waiting for a change on `module info` to ask for a `--prefix` option, which would alleviate the problem somehow.

Issue #152, opened 2015-11-14
Add consistent support for "undeployed" state across the deployers


sabbyanandan:
As a developer, I'd like to add support for *undeployed* status consistently across all the deployers, so I can present the correct status instead of the current *unknown*. This is applicable for existing streams without any deployment context associated with it.

PR #194, opened 2015-11-24
XD-3680: Consistent Stream.undeployed state

ericbottard:
fixes spring-cloud/spring-cloud-dataflow#152 

This is a first step of things discussed with @markfisher and @pperalta.
The task `complete` enum value is not taken care of in this PR. Primarily looking for feedback on: 
* javadoc and API contract expressed in the new top level enum. 
* logic implemented in `reduce()`

### DIFF
--- a/spring-cloud-dataflow-module-deployer-spi/src/main/java/org/springframework/cloud/dataflow/module/DeploymentState.java
+++ b/spring-cloud-dataflow-module-deployer-spi/src/main/java/org/springframework/cloud/dataflow/module/DeploymentState.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.dataflow.module;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Deployment states for modules and streams. Unless indicated, these states
+ * may represent the state of<ul>
+ * <li>an entire stream</li>
+ * <li>the global state of a deployed module as part of a stream</li>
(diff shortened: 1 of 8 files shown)


spring-cloud-dataflow-module-deployer-spi/src/main/java/org/springframework/cloud/dataflow/module/DeploymentState.java
```diff
@@ -0,0 +1,100 @@
...
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Deployment states for modules and streams. Unless indicated, these states
+ * may represent the state of<ul>
+ * <li>an entire stream</li>
+ * <li>the global state of a deployed module as part of a stream</li>
+ * <li>the state of a particular instance of a module, in cases where {@literal module.count > 1}</li>
```
pperalta: For this JavaDoc, was the intent for module.count to be in fixed width font or the same font as the rest of the text?

ericbottard: Fixed. Did i screw up?

pperalta: Here's the difference between `{@literal}` and `{@code}`:

https://blogs.oracle.com/darcy/entry/javadoc_tip_code_and_literal

I'll modify this on merge.

spring-cloud-dataflow-module-deployer-spi/src/main/java/org/springframework/cloud/dataflow/module/deployer/ModuleDeployer.java
```diff
@@ -69,7 +69,7 @@
 	 *
 	 * @param id id for the module this status is for
 	 *
-	 * @return module deployment status
+	 * @return module deployment status or {@literal null} if this deployer knows nothing about the given deployment
```
pperalta: Is this accurate? I think we're supposed to return `unknown` for unknown deployments.

mbogoevici: I don't think we should return `null` either, also see https://github.com/spring-cloud/spring-cloud-dataflow/pull/172#discussion_r46330735

ericbottard: yes. This was an earlier approach. Changed my mind when we started discussing.

pperalta:
Due to rebasing, will close this PR and replace with https://github.com/spring-cloud/spring-cloud-dataflow/pull/205.

Issue #187, opened 2015-11-23
TAB completion is impractical, with 100s of options


sabbyanandan:
As a user, I'm trying to use `module info` command to lookup supported options for each module; however, I'm seeing 100s of options listed instead. Most majority of them are irrelevant to the module itself and are Boot specific overrides. 

**Acceptance:**
When I lookup `module info http:source`, I should only see `http` module specific overrides such as:
```
  Option Name            Description                                                                                                                                       Default                                                              Type
  ---------------------  ------------------------------------------------------------------------------------------------------------------------------------------------  -------------------------------------------------------------------  ---------------------------------
  https                  true for https://                                                                                                                                 false                                                                boolean
  maxContentLength       the maximum allowed content length                                                                                                                1048576                                                              int
  messageConverterClass  the name of a custom MessageConverter class, to convert HttpRequest to Message; must have a constructor with a 'MessageBuilderFactory' parameter  org.springframework.integration.x.http.NettyInboundMessageConverter  java.lang.String
  port                   the port to listen to                                                                                                                             9000                                                                 int
  sslPropertiesLocation  location (resource) of properties containing the location of the pkcs12 keyStore and pass phrase                                                  classpath:httpSSL.properties                                         java.lang.String
  outputType             how this module should emit messages it produces                                                                                                  <none>                                                               org.springframework.util.MimeType

PR #199, opened 2015-11-26
Add incremental TAB completion of configuration properties

ericbottard:
Instead of showing the 100s of configuration properties that are
currently visible though boot metadata, stop at each dot.

This resolves spring-cloud/spring-cloud-dataflow#187.

### DIFF
--- a/spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/AddModuleOptionsExpansionStrategy.java
+++ b/spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/AddModuleOptionsExpansionStrategy.java
@@ -16,25 +16,26 @@
 
 package org.springframework.cloud.dataflow.completion;
 
-import static org.springframework.cloud.dataflow.completion.CompletionProposal.*;
+import static org.springframework.cloud.dataflow.completion.CompletionProposal.expanding;
 
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
+import org.springframework.boot.configurationmetadata.ConfigurationMetadataGroup;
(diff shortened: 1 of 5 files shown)


ghillert:
Hey Eric, 

Not sure how nasty it would be, but could you provide some tests for this?

ericbottard:
Yep, intented to add some but got carried away. Would be cool to validate the behavior before that, if you can though (ie manual tests to see how it feels)

ghillert:
@ericbottard Cannot build after rebase with the latest from `master`. Hitting compilation errors in `spring-cloud-dataflow-completion`.

ericbottard:
Can you paste the error ? Need https://github.com/spring-cloud/spring-cloud-stream/pull/210

ghillert:
By the way do any of the XD-documented `http-source` properties work? Looks like things like `--port` don't show up. Otherwise, property completion works. Once you add some tests we could merge.

ericbottard:
The option is not `port` but `server.port` (boot option)

ghillert:
@ericbottard Shouldn't we keep certain basic properties backwards compatible?

ericbottard:
Not for me to decide :) My understanding is that we embrace boot. cc @markfisher 
I personally don't think this is a big deal, especially if we have good discoverability of those options

markfisher:
In general, we should maintain backwards compatibility and not rename options arbitrarily. However, the `--port` option is a special case. In fact it's handled universally for any module now via the `server.port` boot property (configurable as a deployment property rather than a "module option").

ghillert:
Ok - cool. @ericbottard So once you have some tests in place, I can merge.

ericbottard:
Added some tests that cover new behavior

ghillert:
Merged with PR f584c2f

Issue #176, opened 2015-11-20
REST: Disable HAL to provide better backwards compatibility with XD API


ghillert:

ghillert:
Required by: https://github.com/spring-cloud/spring-cloud-dataflow-ui/pull/9

garyrussell:
Do we really want to do this? I recently fixed the shell to work with HAL: 

https://github.com/spring-cloud/spring-cloud-dataflow/pull/143

ghillert:
@garyrussell @sabbyanandan @ilayaperumalg  Good for pointing this out. As the shell is already refactored to work with this, I would be in favor of closing this PR and rather refactor the UI to work with the new REST API. My concern was originally the substantial breaking change in the REST API but since there all kinds of other changes, that might be ok then.

PR #204, opened 2015-12-02
issue-176 REST: Disable HAL

ghillert:
This resolves spring-cloud/spring-cloud-dataflow#176


garyrussell:
See my comment on #176 before merging.

PR #205, opened 2015-12-03
XD-3680: Consistent state calculation

pperalta:
Replacement for https://github.com/spring-cloud/spring-cloud-dataflow/pull/194 by @ericbottard.

This includes the following changes:
* Rebased with merge conflicts resolved (based on top of master)
* State `incomplete` renamed to `partial`.
* Stream and module state calculation split

After discussion with @markfisher we decided to split the state calculation. There are some subtle differences between stream and module state calculation. For instance, if module states include `deployed` and `failed`, the resulting state is `partial`. However if a stream includes modules in `deployed` and `failed` states, the stream should be `failed` since it cannot function with a failed module.

This resolves spring-cloud/spring-cloud-dataflow#152


ericbottard:
This LGTM, all valid points. Will merge

ericbottard:
Merged as 3fb02d0

ericbottard:
Also changed https://github.com/spring-cloud/spring-cloud-dataflow-admin-cloudfoundry/commit/d67304117d0b0a8b02d4ece11f407178154bfebe

Issue #207, opened 2015-12-04
Out of process deployer should terminate deployed modules when admin server shuts down


mbogoevici:
Currently, this happens if the admin process is sent `Ctrl-C` via shell or a `kill -pid` (both of which terminate the process group). Generally speaking, it would be desirable for the deployer to shut down the  deploys modules when it exits gracefully (i.e. no `SIGKILL`)

PR #208, opened 2015-12-04
Out of process deployer improvements: module shutdown and directories

mbogoevici:
This resolves spring-cloud/spring-cloud-dataflow#207.

- handling the shutdown of the children processes in the shutdown hook of the admin;
- reworking the paths for the out of process modules - a single temporary root is created for the admin, and module log files are created underneath

### DIFF
--- a/spring-cloud-dataflow-admin-local/src/main/java/org/springframework/cloud/dataflow/admin/spi/local/OutOfProcessModuleDeployer.java
+++ b/spring-cloud-dataflow-admin-local/src/main/java/org/springframework/cloud/dataflow/admin/spi/local/OutOfProcessModuleDeployer.java
@@ -97,14 +97,14 @@ public class OutOfProcessModuleDeployer implements ModuleDeployer {
 				workDir.toFile().deleteOnExit();
 			}
 			for (int i = 0; i < request.getCount(); i++) {
-				int port = SocketUtils.findAvailableTcpPort(DEFAULT_SERVER_PORT);
-				args.put(SERVER_PORT_KEY, String.valueOf(port));
+				int managementPort = SocketUtils.findAvailableTcpPort(DEFAULT_MANAGEMENT_PORT);
+				args.put(MANAGEMENT_PORT_KEY, String.valueOf(managementPort));
 
 				ProcessBuilder builder = new ProcessBuilder(properties.getJavaCmd(), "-jar", moduleLauncherPath);
 				builder.environment().clear();
 				builder.environment().putAll(args);
 
-				Instance instance = new Instance(moduleDeploymentId, i, builder, workDir);
+				Instance instance = new Instance(moduleDeploymentId, i, builder, workDir, managementPort);
 				processes.add(instance);
 				if (properties.isDeleteFilesOnExit()) {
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-admin-local/src/main/java/org/springframework/cloud/dataflow/admin/spi/local/OutOfProcessModuleDeployer.java
```diff
@@ -192,8 +210,13 @@ private Instance(ModuleDeploymentId moduleDeploymentId, int instanceNumber, Proc
 			this.moduleDeploymentId = moduleDeploymentId;
 			this.instanceNumber = instanceNumber;
 			builder.directory(workDir.toFile());
-			builder.redirectOutput(this.stdout = Files.createTempFile(workDir, "stdout_" + instanceNumber + "_", ".log").toFile());
-			builder.redirectError(this.stderr = Files.createTempFile(workDir, "stderr_" + instanceNumber + "_", ".log").toFile());
+			String workDirPath = workDir.toFile().getAbsolutePath();
+			this.stdout =
+					Files.createFile(FileSystems.getDefault().getPath(workDirPath, "stdout_" + instanceNumber + ".log")).toFile();
```
ericbottard: I the root directory is global for all deployments, then file name needs to include stream name as well

ericbottard: Also, what's with all the `getAbsolutePath()` plus jumping through all those `FileSystems.getDefault()` hoops etc?

mbogoevici: In the end, it results in something like this: `$TEMP/spring-cloud-data-flow-8372153501625223078/ticktock.log/stdout_0.log`

mbogoevici: As for the API usage, it's not trivial to create a non-system specific Path relative to another Path, so a String absolute path has to be used. Anyway, simplified this a bit.

ericbottard:
With the port changes, it is not possible to start several instances of one module.
Actually, it is not even possible to start something like "time | log" with both count=1 apparently, as the second module tries to use the same default `server.port=8080`.

Also, albeit the single directory seems like a good idea, there _could_ be an issue if modules create / expect files in their working directory

mbogoevici:
The last commit was a terrible idea then. We should replicate the logic of InProcessModuleDeployer and allocate a random port unless explicitly by the user. The only limitation would be that one wouldn't be able to do multicount when explicitly setting the port `http --server.port=9090`, but this is as far as we can go, IMO.

ericbottard:
I agree. I'll give it a spin dropping the last commit

mbogoevici:
Cool, thanks!

ericbottard:
Merged as 83440f9 (squashing all commits but the last one)

Issue #213, opened 2015-12-07
Shell does not show first module in Module list 


cppwfs:
Steps to reproduce
1) Execute Module List
2) look for file source or timestamp task and you will note that they are missing.

PR #214, opened 2015-12-07
GH-213 Fixes bug where shell omits first row of module list.

cppwfs:
Steps to reproduce
1) Execute Module List
2) look for file source or timestamp task and you will note that they are missing.

This resolves spring-cloud/spring-cloud-dataflow#213


ghillert:
I know I am pesty, but could you add a test for that :-)

ericbottard:
I will when I merge. My bug, my test :)

ericbottard:
Merged as 06d7bfd with an additional test

Issue #203, opened 2015-11-30
Cannot load admin-ui 


sabbyanandan:
As a user, I'm trying to access the `admin-ui` running locally at http://localhost:9393; however, I'm getting the following error in `yarn-admin` application console.

```
2015-11-28 19:50:24.920 ERROR 53145 --- [nio-9393-exec-1] o.s.c.d.a.c.RestControllerAdvice         : Caught exception while handling a request

java.lang.IllegalStateException: required key [security.basic.enabled] not found
	at org.springframework.core.env.AbstractPropertyResolver.getRequiredProperty(AbstractPropertyResolver.java:160) ~[spring-core-4.2.3.RELEASE.jar!/:4.2.3.RELEASE]
	at org.springframework.core.env.AbstractEnvironment.getRequiredProperty(AbstractEnvironment.java:561) ~[spring-core-4.2.3.RELEASE.jar!/:4.2.3.RELEASE]
	at org.springframework.cloud.dataflow.admin.controller.SecurityController.getSecurityInfo(SecurityController.java:46) ~[spring-cloud-dataflow-admin-starter-1.0.0.BUILD-SNAPSHOT.jar!/:1.0.0.BUILD-SNAPSHOT]
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[na:1.8.0_31]
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[na:1.8.0_31]
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[na:1.8.0_31]
	at java.lang.reflect.Method.invoke(Method.java:483) ~[na:1.8.0_31]
	at org.springframework.web.method.support.InvocableHandlerMethod.doInvoke(InvocableHandlerMethod.java:222) ~[spring-web-4.2.3.RELEASE.jar!/:4.2.3.RELEASE]
	at org.springframework.web.method.support.InvocableHandlerMethod.invokeForRequest(InvocableHandlerMethod.java:137) ~[spring-web-4.2.3.RELEASE.jar!/:4.2.3.RELEASE]
	at org.springframework.web.servlet.mvc.method.annotation.ServletInvocableHandlerMethod.invokeAndHandle(ServletInvocableHandlerMethod.java:110) ~[spring-webmvc-4.2.3.RELEASE.jar!/:4.2.3.RELEASE]
[...]

PR #216, opened 2015-12-08
GH-203 Fix SecurityInfo REST endpoint

ilayaperumalg:
- Use `SecurityProperties` instead of infering from environment
 - Thanks to @jvalkeal for the suggestion

This resolves spring-cloud/spring-cloud-dataflow/issues/203

### DIFF
--- a/spring-cloud-dataflow-admin-starter/src/main/java/org/springframework/cloud/dataflow/admin/controller/SecurityController.java
+++ b/spring-cloud-dataflow-admin-starter/src/main/java/org/springframework/cloud/dataflow/admin/controller/SecurityController.java
@@ -35,8 +35,12 @@ import org.springframework.web.bind.annotation.RestController;
 @EnableConfigurationProperties(SecurityProperties.class)
 public class SecurityController {
 
+	private final SecurityProperties securityProperties;
+
 	@Autowired
-	SecurityProperties securityProperties;
+	public SecurityController(SecurityProperties securityProperties) {
+		this.securityProperties = securityProperties;
+	}
 
 	/**
 	 * Return security information.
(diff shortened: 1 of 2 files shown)


ilayaperumalg:
This fixes https://github.com/spring-cloud/spring-cloud-dataflow/issues/203

spring-cloud-dataflow-admin-starter/src/main/java/org/springframework/cloud/dataflow/admin/controller/SecurityController.java
```diff
@@ -31,10 +32,11 @@
  */
 @RestController
 @RequestMapping("/security/info")
+@EnableConfigurationProperties(SecurityProperties.class)
 public class SecurityController {
 
 	@Autowired
-	private Environment environment;
+	SecurityProperties securityProperties;
```
markfisher: why is this not private? also, can we go ahead and switch away from field injection while we're making changes here?

ilayaperumalg: Sure, this needs to be private. I can also change it to use constructor level injection. Earlier I followed the typical approach to autowire the configuration properties via field injection.

jvalkeal:
This should take away that previous error. `security.basic.enabled` is `true` by default so I don't know if application actually works if that is not disabled on admin(you'd need to find generated random password from logs). Local does disable it in its yml config. But that is another topic so I think this looks good. I anyway have `servers.yml` file which user can use to configure custom settings.

jvalkeal:
Merged

mbogoevici:
A bit late into the game, but we could add a `@PropertySource("adminDefaults.properties")` with defaults to the admin server to override Boot's own defaults. which can include disabling security. So that we don't have to do this for each admin server implementation.

ilayaperumalg:
@mbogoevici I don't think we do something to override the existing `boot` properties here. All we do here is to expose whatever the settings for security property.

PR #228, opened 2015-12-11
Module info command to input type:name

### LINKED ISSUE #217: DSL: Retain original syntax to fetch 'module info'
As a user, I'd want to use XD DSL such as the `module info source:http` in order to query all the options supported by `http` source module. This is currently changed to `module info http:source`, which breaks backward compatibility.

**Acceptance:**
Issuing `module info source:http` (in Shell) results with the options associated with `http` source module

ilayaperumalg:
- This resolves #217
 - Port `QualifiedModuleNameConverter` from XD to here
 - This will make sure the syntax `type:name` is used for the module info command


ilayaperumalg:
This was merged as 8738511f1df317adda37baef62eb988447633be0

PR #232, opened 2015-12-14
Move repository entries to .settings.xml

### LINKED ISSUE #230: Move repo definitions to settings.xml
As a developer, I'd like to cleanup parent `pom.xml` by moving the repo definitions to `.settings.xml` file.

While we are at it, we could cleanup plugin configurations for the ref docs. The parent `pom.xml` doesn't have `surefire` and `javadoc` configurations either.

ilayaperumalg:
- Also cleanup unused entries for `docs`

This resolves #230


.settings.xml
```diff
@@ -1,60 +1,70 @@
...
+	<servers>
+		<server>
+			<id>repo.spring.io</id>
+			<username>${env.CI_DEPLOY_USERNAME}</username>
+			<password>${env.CI_DEPLOY_PASSWORD}</password>
+		</server>
+	</servers>
+	<profiles>
+		<profile>
+			<id>spring</id>
```
markpollack: Add this boilerplate text 

```
      <!--
          N.B. this profile is only here to support users and IDEs that do not use Maven 3.3. 
          It isn't needed on the command line if you use the wrapper script (mvnw) or if you use 
          a native Maven with the right version. Eclipse users should points their Maven tooling to
          this settings file, or copy the profile into their ~/.m2/settings.xml.
      -->
```

Also, i don't think the `</servers>` section is needed but I haven't checked.

ilayaperumalg: Yeah, I don't think the `servers` section is needed. I see that the CI build plan doesn't use it. Addressed the above.

markpollack:
Closed via 93b8eb34c3aa0ca35efb92988d76a2a38ee3941b

Make another PR if you think there is additional cruft to remove from some poms

PR #263, opened 2015-12-18
Make sure closed set value provider runs last

ericbottard:
Fixes spring-cloud/spring-cloud-dataflow#262

### DIFF
--- a/spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/StreamCompletionProvider.java
+++ b/spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/StreamCompletionProvider.java
@@ -17,10 +17,14 @@
 package org.springframework.cloud.dataflow.completion;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
+
+import javax.annotation.PostConstruct;
 
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.cloud.dataflow.core.StreamDefinition;
+import org.springframework.core.annotation.AnnotationAwareOrderComparator;
 
 /**
  * Provides code completion on a (maybe ill-formed) stream definition.
@@ -34,6 +38,15 @@ public class StreamCompletionProvider {
 
 	@Autowired
 	private List<? extends ExpansionStrategy> completionExpansionStrategies = new ArrayList<>();
+
+	@PostConstruct
+	public void sortStrategies() {
+		// Here we mean HIGHEST_PRECEDENCE to win over others, so must be *last*
+		AnnotationAwareOrderComparator.sort(completionExpansionStrategies);
+		Collections.reverse(completionExpansionStrategies);
+		AnnotationAwareOrderComparator.sort(completionRecoveryStrategies);
+		Collections.reverse(completionRecoveryStrategies);
+	}
 
 	/*
 	 * Attempt to parse the text the user has already typed in. This either succeeds,
--- a/spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/ConfigurationPropertyValueHintExpansionStrategy.java
+++ b/spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/ConfigurationPropertyValueHintExpansionStrategy.java
@@ -16,7 +16,7 @@
 
 package org.springframework.cloud.dataflow.completion;
 
-import static org.springframework.cloud.dataflow.completion.CompletionProposal.*;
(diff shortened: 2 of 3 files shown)


spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/StreamCompletionProvider.java
```diff
@@ -35,6 +39,15 @@
 	@Autowired
 	private List<? extends ExpansionStrategy> completionExpansionStrategies = new ArrayList<>();
 
+	@PostConstruct
+	public void sortStrategies() {
+		// Here we mean HIGHEST_PRECEDENCE to win over others, so must be *last*
+		AnnotationAwareOrderComparator.sort(completionExpansionStrategies);
```
markfisher: This seems like an abuse of the `@Ordered` annotation to avoid the real problem which is a result of using field injection for a collection. We need to remove all field injection eventually, so I'd rather see this change to constructor injection and explicit configuration of the desired order within the list.

markfisher: ...or possibly the addition of a `CompositeExpansionStrategy` - either way having the list order within configuration code

ericbottard: I'm not sure I understand how field injection is the culprit here. While I agree with the move to constructor injection, what's wrong with `@Ordered`? Are you saying `StreamCompletionProvider` should know about the desired order?

markfisher: True, the culprit is not field-injection per se but autowiring (field-injection having no other option), *and* specifically autowiring of a list of otherwise arbitrarily defined beans in the context. `@Ordered` is specifically designed for lifecycle operations (it's not used for order of things like converters, etc). The intention with `@Ordered` is that its precedence order applies for the startup process and is reversed for the shutdown process, and it's also common to have multiple instances within a "tier" (all beans having the same order are processed within the same tier, where order is once again arbitrary). To me, the red flag with this particular PR is that you set `HIGHEST_PRECEDENCE` (which makes sense semantically) but then sort in reverse, whereas the intended usage scenarios treat `HIGHEST_PRECEDENCE` as *first* in the list and `LOWEST_PRECEDENCE` as last. That is why I used the word "abuse".

It seems as though some central component needs to know the order here, rather than each instance itself having an order value. That would provide a strictly ordered list with, for example, exactly one instance in the final position since it avoids the possibility of "tiers" (what happens when another instance is added and also has `HIGHEST_PRECEDENCE` for its `order` property?). It could be `StreamCompletionProvider` itself, but if that doesn't feel like its responsibility, then the next option would be a `CompositeExpansionStrategy` that manages a list internally but is injected as the sole `ExpansionStrategy` instance for the `StreamCompletionProvider` (a common pattern within `spring-cloud-stream` and `spring-cloud-dataflow` already). Or, if it feels like the order should be a purely configuration-side responsibility, we could switch to *non*-autowired constructor injection and explicitly provide the list of beans within a `@Configuration` class.

ericbottard: Gotcha (and agreed))

ericbottard: Hmmm, there is a caveat though: we need post-creation injection because of
a circular dependency b/w SCProvider and ExpandOneDashToTwoDashesRS

markfisher: *circular dependency* sounds like another red flag... any way to refactor to avoid it?

ericbottard: Maybe. I guess '-<tab>' and '--<tab>' could be handled by the same
completer. Will look into it (DSL parser may fail differently)

markfisher:
These 2 tests seem to be failing more frequently as of late. So, I have temporarily ignored them with this commit: https://github.com/spring-cloud/spring-cloud-dataflow/commit/d4e9caaf2dfac941f1a86d7f9a44d0209ba69722

Un-ignoring those methods should be part of the scope of this PR, once we also address the refactoring discussed above.

ericbottard:
Closing in favor of https://github.com/spring-cloud/spring-cloud-dataflow/pull/277

PR #265, opened 2015-12-18
Fix passing producer/consumer properties

### LINKED ISSUE #264: producer/consumer properties aren't passed via DSL
For the following stream, the producer/consumer properties need to be set appropriately when the message channel is bound.

```
dataflow:>stream create a1 --definition "time | log"
Created new stream 'a1'
dataflow:>stream deploy a1 --properties "module.time.producer.partitionKeyExpression=payload,module.log.count=2,module.log.partitioned=true,module.log.consumer.concurrency=2"
Deployed stream 'a1'
dataflow:>
```

The above DSL to set the producer/consumer properties don't set any of the module deployment properties and subsequently `spring.cloud.stream.producer/consumerProperties`.

ilayaperumalg:
- When the DSL has `module.<label>.producer/consumer.` properties, then they need to be set as `spring.cloud.stream.producer/consumerProperties`.
 - The partition properties also need to be updated per binding case.
 - This fixes #264

### DIFF
--- a/spring-cloud-dataflow-admin-starter/src/main/java/org/springframework/cloud/dataflow/admin/controller/StreamDeploymentController.java
+++ b/spring-cloud-dataflow-admin-starter/src/main/java/org/springframework/cloud/dataflow/admin/controller/StreamDeploymentController.java
@@ -57,6 +57,9 @@ import org.springframework.web.bind.annotation.RestController;
  * Controller for deployment operations on {@link StreamDefinition}.
  *
  * @author Eric Bottard
+ * @author Mark Fisher
+ * @author Patrick Peralta
+ * @author Ilayaperumal Gopinathan
  */
 @RestController
 @RequestMapping("/streams/deployments")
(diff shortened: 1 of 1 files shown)


spring-cloud-dataflow-admin-starter/src/main/java/org/springframework/cloud/dataflow/admin/controller/StreamDeploymentController.java
```diff
@@ -82,6 +85,18 @@
 
 	private static final String DEFAULT_PARTITION_KEY_EXPRESSION = "payload";
 
+	private static final String CHANNEL_BINDING_PRODUCER_PROPERTIES_PREFIX = "spring.cloud.stream.producerProperties.";
+
+	private static final String CHANNEL_BINDING_CONSUMER_PROPERTIES_PREFIX = "spring.cloud.stream.consumerProperties.";
+
+	private static final String PARTITION_KEY = CHANNEL_BINDING_PRODUCER_PROPERTIES_PREFIX + BindingProperties.PARTITION_KEY_EXPRESSION;
```
markfisher: should be `PARTITION_KEY_EXPRESSION`

markfisher: will change while merging - otherwise LGTM

markfisher:
merged along with https://github.com/spring-cloud/spring-cloud-dataflow/commit/4662cc7d5eaf7deaf5117ed64b2cdbd0498ee9bf

PR #268, opened 2015-12-21
Pass producer/consumer properties as per-binding properties

### LINKED ISSUE #209: Enable trackHistory as a deployment property
Related to the feature reported [here] (https://github.com/spring-cloud/spring-cloud-stream/issues/196), we would like to have a `trackHistory` as a deployment property that can be set via DSL.

ilayaperumalg:
- This will make `trackHistory` property as a per-binding property
This is in contrast to how XD DSL supports `trackHistory`. In XD, this property is set as `module.<label>.trackHistory`.
But, we want this to be a `per-binding` property in SCS. Hence, syntax for passing `trackHistory` property would be
module.<label>.producer/consumer.trackHistory
 - This PR change also enables any producer/consumer property can be passed as output/input binding property.

- Add test

This fixes #209

### DIFF
--- a/spring-cloud-dataflow-admin-starter/src/main/java/org/springframework/cloud/dataflow/admin/controller/StreamDeploymentController.java
+++ b/spring-cloud-dataflow-admin-starter/src/main/java/org/springframework/cloud/dataflow/admin/controller/StreamDeploymentController.java
@@ -274,10 +274,20 @@ public class StreamDeploymentController {
 		for (Map.Entry<String, String> entry : streamDeploymentProperties.entrySet()) {
 			if (entry.getKey().startsWith(modulePrefix)) {
 				if (entry.getKey().startsWith(producerPropertyPrefix)) {
+					// Set per-binding properties
+					moduleDeploymentProperties.put(BindingProperties.OUTPUT_BINDING_KEY_PREFIX +
+							entry.getKey().substring(producerPropertyPrefix.length()), entry.getValue());
+					// Set ChannelBindingServiceProperties' producer properties
+					// todo: Once the producer/consumer properties are moved to per-binding properties, following step
+					// 		isn't needed: https://github.com/spring-cloud/spring-cloud-stream/issues/256
 					moduleDeploymentProperties.put(CHANNEL_BINDING_PRODUCER_PROPERTIES_PREFIX +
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-admin-starter/src/main/java/org/springframework/cloud/dataflow/admin/controller/StreamDeploymentController.java
```diff
@@ -274,10 +274,20 @@ else if (hasOutput) {
 		for (Map.Entry<String, String> entry : streamDeploymentProperties.entrySet()) {
 			if (entry.getKey().startsWith(modulePrefix)) {
 				if (entry.getKey().startsWith(producerPropertyPrefix)) {
+					// Set per-binding properties
+					moduleDeploymentProperties.put(BindingProperties.OUTPUT_BINDING_KEY_PREFIX +
+							entry.getKey().substring(producerPropertyPrefix.length()), entry.getValue());
+					// Set ChannelBindingServiceProperties' producer properties
+					// todo: Once the producer/consumer properties are moved to per-binding properties, following step
```
markfisher: The properties are already per-binding properties, e.g.:

```
private int partitionCount = 1;

private String partitionKeyExpression;

private String partitionKeyExtractorClass;

private String partitionSelectorClass;

private String partitionSelectorExpression;
```

The challenge is that they are not qualified as "producer.*" properties at the binding level, nor should they be since each binding is already corresponding to a single direction (inbound vs. outbound).

markfisher: In other words, while "consumer" and "producer" might make sense within the scope of the dataflow DSL, we need to map those to the corresponding bindings without carrying over the terms "producer" or "consumer" on the `spring-cloud-stream` side (where they would be superfluous). On the dataflow side, we do know the channel names (input/output) for which are providing the properties. It seems that we need to map from `producer.*` to the output channel binding properties and map from `consumer.*` to the input channel binding properties. The key following `prodcuer/consumer.` should match that expected in `BindingProperties` itself.

ilayaperumalg:
Closing this PR to create a new one that is built on top of https://github.com/spring-cloud/spring-cloud-stream/pull/261

PR #272, opened 2015-12-22
Passing producer/consumer props as per-binding props

### LINKED ISSUE #209: Enable trackHistory as a deployment property

ilayaperumalg:
- Hanndle producer/consumer properties to be passed as per-binding properties in SCS
 - Address `trackHistory` and partition properties.
 - In general, any producer/consumer properties will be set as output/input channel BindingProperties' config

This fixes #209

### DIFF
--- a/spring-cloud-dataflow-admin-starter/src/main/java/org/springframework/cloud/dataflow/admin/controller/StreamDeploymentController.java
+++ b/spring-cloud-dataflow-admin-starter/src/main/java/org/springframework/cloud/dataflow/admin/controller/StreamDeploymentController.java
@@ -78,22 +78,8 @@ public class StreamDeploymentController {
 	 */
 	private final ModuleDeployer deployer;
 
-	private static final String DEFAULT_PARTITION_KEY_EXPRESSION = "payload";
-
-	private static final String CHANNEL_BINDING_PRODUCER_PROPERTIES_PREFIX = "spring.cloud.stream.producerProperties.";
-
-	private static final String CHANNEL_BINDING_CONSUMER_PROPERTIES_PREFIX = "spring.cloud.stream.consumerProperties.";
-
-	private static final String PARTITION_KEY_EXPRESSION = CHANNEL_BINDING_PRODUCER_PROPERTIES_PREFIX + BindingProperties.PARTITION_KEY_EXPRESSION;
-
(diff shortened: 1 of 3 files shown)


ilayaperumalg:
Updated the PR to reflect the changes from https://github.com/spring-cloud/spring-cloud-stream/pull/262

spring-cloud-dataflow-admin-starter/src/main/java/org/springframework/cloud/dataflow/admin/controller/StreamDeploymentController.java
```diff
@@ -329,9 +315,8 @@ private boolean isPartitionedConsumer(ModuleDefinition module,
 			Map<String, String> moduleDeploymentProperties,
 			boolean upstreamModuleSupportsPartition) {
 		return upstreamModuleSupportsPartition ||
-				(module.getParameters().containsKey(BindingProperties.INPUT_BINDING_KEY) &&
```
markfisher: why is this check for input binding key no longer necessary? (or was it ever?)

ilayaperumalg: Given these `partition` properties are only applicable for the module that is a `producer`, I believe having this explicit check is unnecessary.

markfisher: you mean `consumer` right? :)

ilayaperumalg: oops, I looked at the wrong method `upstreamModuleHasPartitionInfo`at my IDE . yeah, `consumer` in this case.

markfisher:
rebased and merged as https://github.com/spring-cloud/spring-cloud-dataflow/commit/36e97d667d9ec04be4e9dcb6f8a28ac60a0eab0d

PR #277, opened 2015-12-28
Make sure closed set value provider runs last

ericbottard:
Fixes spring-cloud/spring-cloud-dataflow#262

I made the configuration order explicit, and got rid of the circular
dependency at the cost of duplicating a bit of logic found in
StreamCompletionProvider in ExpandOneDashToTwoDashesRecoveryStrategy.

### DIFF
--- a/spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/CompletionConfiguration.java
+++ b/spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/CompletionConfiguration.java
@@ -17,6 +17,9 @@
 package org.springframework.cloud.dataflow.completion;
 
 
+import java.util.Arrays;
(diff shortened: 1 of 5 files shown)


markfisher:
Something still feels a bit off given the coupling between `ExpandOneDashToTwoDashesRecoveryStrategy` and `ConfigurationPropertyNameAfterDashDashRecoveryStrategy` while they are also in the list as 2 distinct strategy instances (perhaps they should just be combined?). However, that can be a follow-up discsions. This PR is a big improvement overall. Will merge.

ericbottard:
Sadly they can't be combined because the exceptions that trigger them are very different. One is a pure tokenization exception, while the other is an AST parse exception (that can remember the last point up to which parsing was "well formed"). We need to fall back in a situation similar to the latter to be able to know e.g. which properties have already been set for the module etc.

Issue #247, opened 2015-12-16
Add field-value-counter shell commands


sabbyanandan:
As a user, I'd like to shell commands for [field-value-counter] sink, so I can interact with the counter. There's currently [no support](https://github.com/spring-cloud/spring-cloud-dataflow/tree/master/spring-cloud-dataflow-shell/src/main/java/org/springframework/cloud/dataflow/shell/command) for this at the shell `level`.

**Acceptance:**
* Following commands are available for use in the shell

```
private static final String DISPLAY_FV_COUNTER = "field-value-counter display";
private static final String LIST_FV_COUNTERS = "field-value-counter list";
private static final String DELETE_FV_COUNTER = "field-value-counter delete";
```
* Using `field-value-counter display` command includes similar options from XD such as it allows to specify the counter `name` or `size` or `pattern`
* Unit tests included to simulate command operations

PR #282, opened 2015-12-31
Add Field Value Counter shell support

ericbottard:
Depends on https://github.com/spring-cloud/spring-cloud-stream-modules/pull/149

Add REST endpoints to expose FVC functionality
Add Shell commands to display, list names of and delete FVCs

Fixes spring-cloud/spring-cloud-dataflow#247


ilayaperumalg:
@ericbottard I rebased the PR against the latest master and for some reason the tests are stuck at the `spring-cloud-dataflow-shell` project.

ilayaperumalg:
It went through after few attempts. but it was weird though as it seemed to only happened on this branch.

spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/FieldValueCounterOperations.java
```diff
@@ -0,0 +1,44 @@
...
+
+	/**
+	 * Retrieve basic information (i.e. names) for existing counters.
+	 */
+	PagedResources<MetricResource> list();
+
+	/**
+	 * Delete the counter with given name.
+	 */
+	void reset(String name);
```
ilayaperumalg: I believe this operation just needs to reset the field value counter with the given `name` not deleting it.

spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/FieldValueCounterTemplate.java
```diff
@@ -0,0 +1,59 @@
...
+import org.springframework.web.client.RestTemplate;
+
+/**
+ * Implementation for {@link FieldValueCounterOperations} that interacts with the Spring Cloud Data Flow REST API.
+ *
+ * @author Eric Bottard
+ */
+public class FieldValueCounterTemplate implements FieldValueCounterOperations {
+
+	public static final String FVC_COLLECTION_RELATION = "field-value-counters";
```
ilayaperumalg: May be we can move these constants to the interface `FieldValueCounterOperations`

spring-cloud-dataflow-shell/src/main/java/org/springframework/cloud/dataflow/shell/command/FieldValueCounterCommands.java
```diff
@@ -0,0 +1,117 @@
...
+
+	protected FieldValueCounterCommands() {
+		super("Field Value Counter");
+	}
+
+	private static final String DISPLAY_COUNTER = "field-value-counter display";
+
+	private static final String LIST_COUNTERS = "field-value-counter list";
+
+	private static final String DELETE_COUNTER = "field-value-counter reset";
```
ilayaperumalg: If we decide to go with `reset` operation, I think we can rename this operation to `reset`

spring-cloud-dataflow-shell/src/main/java/org/springframework/cloud/dataflow/shell/command/FieldValueCounterCommands.java
```diff
@@ -0,0 +1,117 @@
...
+	@CliCommand(value = DISPLAY_COUNTER, help = "Display the value of a field value counter")
+	public List<Object> display(
+			@CliOption(key = { "", "name" }, help = "the name of the field value counter to display", mandatory = true
+					/*,optionContext = "existing-counter disable-string-converter"*/) String name,
+			final @CliOption(key = "pattern", help = "the pattern used to format the values (see DecimalFormat)",
+					mandatory = false, unspecifiedDefaultValue = NumberFormatConverter.DEFAULT) NumberFormat pattern) {
+		FieldValueCounterResource counter = fvcOperations().retrieve(name);
+
+		LinkedHashMap<String, Object> header = new LinkedHashMap<>();
+		header.put("key", "Key");
```
ilayaperumalg: Can this be `Field Name` instead?

spring-cloud-dataflow-shell/src/main/java/org/springframework/cloud/dataflow/shell/command/FieldValueCounterCommands.java
```diff
@@ -0,0 +1,117 @@
...
+
+	@CliAvailabilityIndicator({ LIST_COUNTERS, DISPLAY_COUNTER, DELETE_COUNTER })
+	public boolean available() {
+		return dataFlowShell.getDataFlowOperations() != null;
+	}
+
+	@CliCommand(value = DISPLAY_COUNTER, help = "Display the value of a field value counter")
+	public List<Object> display(
+			@CliOption(key = { "", "name" }, help = "the name of the field value counter to display", mandatory = true
+					/*,optionContext = "existing-counter disable-string-converter"*/) String name,
```
ilayaperumalg: Not sure why this `optionContext` is commented here.

spring-cloud-dataflow-admin-starter/src/main/java/org/springframework/cloud/dataflow/admin/controller/FieldValueCounterController.java
```diff
@@ -0,0 +1,140 @@
...
+		}
+
+		@Override
+		protected MetricResource instantiateResource(String name) {
+			return new MetricResource(name);
+		}
+	}
+
+	/**
+	 * Knows how to assemble {@link FieldValueCounterResource}s out of counter {@link FieldValueCounter}s.
```
ilayaperumalg: I think it can be better this way:

```
Knows how to assemble {@link FieldValueCounterResource} out of  {@link FieldValueCounter}.
``

Removal of plural and extra word `counter`.

spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/FieldValueCounterOperations.java
ericbottard: But resetting is the same as deleting, this is what I tried to convey in our previous discussion

spring-cloud-dataflow-shell/src/main/java/org/springframework/cloud/dataflow/shell/command/FieldValueCounterCommands.java
ericbottard: +1

spring-cloud-dataflow-shell/src/main/java/org/springframework/cloud/dataflow/shell/command/FieldValueCounterCommands.java
ericbottard: I'm not sure, because it's quite different from the fieldName that was used when creating the counter. This is an actual value that *that* fieldName was seen with, while `Value` on the line below is the number of times it was seen with that value.

I think we could do `Key -> Value, Value -> Count`

spring-cloud-dataflow-shell/src/main/java/org/springframework/cloud/dataflow/shell/command/FieldValueCounterCommands.java
ericbottard: This is a remnant from XD, used in particular to handle completion of names. Other commands have left those as comments, but I'm happy to remove it altogether

spring-cloud-dataflow-admin-starter/src/main/java/org/springframework/cloud/dataflow/admin/controller/FieldValueCounterController.java
ericbottard: +1

spring-cloud-dataflow-shell/src/main/java/org/springframework/cloud/dataflow/shell/command/FieldValueCounterCommands.java
ilayaperumalg: oh yeah, that's right. I like `Value -> Count` as it represents field-value-counter :-)

spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/FieldValueCounterOperations.java
ilayaperumalg: I agree that it is a `delete` of the field value counter values when we do `reset`. From the end user's perspective,  `delete` operation would make the user feel like the deletion of the field value counter itself while `reset` operation would make it feel like resetting the field value counter (though it actually deletes the field values so far in the fvc). So, having the **name** `reset` in all the places would make it feel better :-)

spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/FieldValueCounterTemplate.java
ericbottard: Actually, I wouldn't do this as the concept of relation is tied to REST / hypermedia, which the `Template` embodies, but the general contract (`Operations`) doesn't mandate it.

ericbottard:
Addressed review comments (the branch builds fine for me)

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed, rebased and merged as 8c7d267

Issue #299, opened 2016-01-11
REST: Expose Version Information


ghillert:
expose version information under `/management/info`. 

Required by: https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/2

PR #300, opened 2016-01-11
gh-299 REST: Expose Version Information

ghillert:
resolves https://github.com/spring-cloud/spring-cloud-dataflow/issues/299

required by https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/2

### DIFF
--- a/spring-cloud-dataflow-admin-local/src/main/resources/admin.yml
+++ b/spring-cloud-dataflow-admin-local/src/main/resources/admin.yml
@@ -2,7 +2,11 @@ server:
   port: 9393
 management:
   contextPath: /management
-
+info:
+  app:
+    name: "@project.artifactId@"
+    description: "@project.description@"
+    version: "@project.version@"
 security:
   basic:
     enabled: false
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-admin-local/src/main/resources/admin.yml
```diff
@@ -2,7 +2,11 @@ server:
   port: 9393
 management:
   contextPath: /management
-
+info:
```
ilayaperumalg: Is there a way to define these `info.app.*` properties via admin-starter so that we don't have to explicitly specify these properties for each SPI implementation?

ghillert: Currently we don't have a "base yaml" file in the admin-starter project. Also, wouldn't that mean that the properties are resolved against the `spring-cloud-dataflow-admin-starter`, probably not something we want to happen.

ilayaperumalg: ok, In that case, I think we need to have corresponding changes in all the SPI implementations so that we don't lose track.

ghillert: Yep - Will create corresponding PRs.

ghillert:
@ilayaperumalg Added PRs for respective SPIs.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Rebased and merged as a73816b

ilayaperumalg:
The PR for Cloud foundry SPI is still missing.

ilayaperumalg:
Pushed the changes into CF Admin Repo: https://github.com/spring-cloud/spring-cloud-dataflow-admin-cloudfoundry/commit/b79ba89ef8ca2d9e12ed1ee6ff54cad4dd6de932 after verification.

PR #302, opened 2016-01-12
GH-275 Allow redeployment for the local out of process deployer

### LINKED ISSUE #275: Can't re-deploy streams when running in local mode
When I deploy a stream, and I undeploy it, when I try and deploy the same stream again, the scdf-admin throws an error.  The stack trace shows that the tmp directory for that stream already exists for that stream and that is preventing the stream from being deployed again.

mbogoevici:
Resolves #275.

* Introduces a 'module group deployment id', set for each stream/task deployment, and shared among the modules.
* Set up module output/error directories to be use the deployment id as part of their path, thus avoiding clashes at deployment time.

### DIFF
--- a/spring-cloud-dataflow-admin-local/src/main/java/org/springframework/cloud/dataflow/admin/spi/local/OutOfProcessModuleDeployer.java
+++ b/spring-cloud-dataflow-admin-local/src/main/java/org/springframework/cloud/dataflow/admin/spi/local/OutOfProcessModuleDeployer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 the original author or authors.
+ * Copyright 2015-2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -29,6 +29,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 
 import javax.annotation.PostConstruct;
@@ -54,6 +55,7 @@ import org.springframework.web.client.RestTemplate;
  * A ModuleDeployer implementation that spins off a new JVM process per
  * module instance.
  * @author Eric Bottard
+ * @author Marius Bogoevici
  */
 public class OutOfProcessModuleDeployer implements ModuleDeployer {
 
@@ -93,7 +95,16 @@ public class OutOfProcessModuleDeployer implements ModuleDeployer {
 
 
 		try {
-			Path workDir = Files.createDirectory(Paths.get(logPathRoot.toFile().getAbsolutePath(),
+			String moduleDeploymentGroupId = request.getDeploymentProperties().get(MODULE_DEPLOYMENT_GROUP_ID);
+			if (moduleDeploymentGroupId == null) {
+				moduleDeploymentGroupId = UUID.randomUUID().toString();
+			}
+			Path moduleDeploymentGroupDir = Paths.get(logPathRoot.toFile().getAbsolutePath(), moduleDeploymentGroupId);
+			if (!Files.exists(moduleDeploymentGroupDir)) {
+				Files.createDirectory(moduleDeploymentGroupDir);
+				moduleDeploymentGroupDir.toFile().deleteOnExit();
+			}
+			Path workDir = Files.createDirectory(Paths.get(moduleDeploymentGroupDir.toFile().getAbsolutePath(),
 					moduleDeploymentId.toString()));
 			if (properties.isDeleteFilesOnExit()) {
 				workDir.toFile().deleteOnExit();
--- a/spring-cloud-dataflow-admin-local/src/main/java/org/springframework/cloud/dataflow/admin/spi/local/OutOfProcessModuleDeployer.java
+++ b/spring-cloud-dataflow-admin-local/src/main/java/org/springframework/cloud/dataflow/admin/spi/local/OutOfProcessModuleDeployer.java
@@ -95,11 +95,11 @@ public class OutOfProcessModuleDeployer implements ModuleDeployer {
 
 
 		try {
-			String moduleDeploymentGroupId = request.getDeploymentProperties().get(MODULE_DEPLOYMENT_GROUP_ID);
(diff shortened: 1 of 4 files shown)


ericbottard:
I'm wondering: do we really need to introduce a new deployment property for this? Seems like the randomness could be taken care of at the OOPMD level only.

mbogoevici:
I've considered the alternatives: for strictly solving this problem, we can simply randomize the destination folder and get done with it alright.
But at the same time, I think it would be a better idea to propagate information that allows to correlate module deployments and stream/task deployment requests. The long term goal would be mainly be to able to diagnose stray processes (you can't tell right now what request spawned them without parsing their logs), trace/debug the deployment process and so on. My initial choice was an internal deployment property because those are automatically propagated to the module (and running application instance) itself. Anyway, this is open for discussion.

spring-cloud-dataflow-admin-local/src/main/java/org/springframework/cloud/dataflow/admin/spi/local/OutOfProcessModuleDeployer.java
```diff
@@ -93,7 +95,16 @@ public ModuleDeploymentId deploy(ModuleDeploymentRequest request) {
 
 
 		try {
-			Path workDir = Files.createDirectory(Paths.get(logPathRoot.toFile().getAbsolutePath(),
+			String moduleDeploymentGroupId = request.getDeploymentProperties().get(MODULE_DEPLOYMENT_GROUP_ID);
+			if (moduleDeploymentGroupId == null) {
+				moduleDeploymentGroupId = UUID.randomUUID().toString();
```
markfisher: we could use `group` + timestamp here as well, I guess?

markfisher:
What if we add a `timestamp` property to the `ModuleDeploymentRequest` object. Then rather than using a deployment property, we could simply use the `group` and `timestamp` properties from that request object in the deployer. We could even consider a level of the hierarchy for the group, e.g. `../<group>/<timestamp>/...`

spring-cloud-dataflow-admin-local/src/main/java/org/springframework/cloud/dataflow/admin/spi/local/OutOfProcessModuleDeployer.java
mbogoevici: That would work.

mbogoevici:
That could work, too, but it requires handling in each and every individual deployer implementation. What I like about deployment properties was that they're automatically propagated to the deployment. This may diverge from our current use, which is mostly for runtime configuration, but I suppose we can propagate some metadata as well. Maybe we can provide a separate Metadata section in the ModuleDeploymentRequest ?

spring-cloud-dataflow-module-deployer-spi/src/main/java/org/springframework/cloud/dataflow/module/deployer/ModuleDeployer.java
```diff
@@ -38,6 +39,8 @@
 
 	public static final String JMX_DEFAULT_DOMAIN_KEY  = "spring.jmx.default-domain";
 
+	public static final String MODULE_DEPLOYMENT_GROUP_ID = "spring_cloud_dataflow.module_deployment_group_id";
```
markfisher: should this use dots and hyphens rather than underscores? (e.g. `spring.cloud.dataflow.module-deployment-group-id`),... and is "dataflow" sufficient for the prefix?

mbogoevici: I'm fine with `dataflow.module-deployment-group-id`.

markfisher: I wonder if "group-deployment-id" would be sufficient?... not only removing "module" as a bit redundant but reversing the other 2 terms as it's really an ID for a *group* having been *deployed*.

mbogoevici: Make sense. I thought of it as the stream being a "module deployment group" as in "group of modules that got deployed". "Group deployment" works as well, I suppose, and I like "group deployment id" being similar with "module deployment id". Still, would want to keep the `dataflow.` prefix, correct?

markfisher: I do think it makes sense to keep the prefix.

mbogoevici:
Pushed, rebased.

spring-cloud-dataflow-admin-local/src/main/java/org/springframework/cloud/dataflow/admin/spi/local/OutOfProcessModuleDeployer.java
```diff
@@ -93,7 +95,16 @@ public ModuleDeploymentId deploy(ModuleDeploymentRequest request) {
 
 
 		try {
-			Path workDir = Files.createDirectory(Paths.get(logPathRoot.toFile().getAbsolutePath(),
+			String groupDeploymentId = request.getDeploymentProperties().get(GROUP_DEPLOYMENT_ID);
+			if (groupDeploymentId == null) {
+				groupDeploymentId = request.getDefinition().getGroup() + "-" + System.currentTimeMillis();
```
ericbottard: This makes groupDeploymentId calculation appear at 2 places. Can we assume it is never null, with the "only" caveat that some unit tests may have to explicitly pass one?

mbogoevici: I'd prefer not to have to pass anything for the unit tests, if possible at all. But I see what you're saying trying to do the same calculation here for a bogus group deployment id is wrong. After all we're just trying to create a unique directory if no group deployment id is provided - so perhaps we an just use a timestamp for the folder instead.

ericbottard:
Merged as 2ae69c6

PR #329, opened 2016-01-24
@EnableDataflowAdminServer annotation

joshlong:
adds support for `@EnableDataflowAdminServer`. This also required defaulting the properties configured from `admin.yml` but supporting user-defined overrides through the normal channels, Spring Boot auto-configuration, the Spring Cloud Config Server, etc.


markfisher:
need to address tabs vs spaces across the board

markfisher:
the last commit says "trying to restore spaces".... but we use tabs. Just look at the diff - a modified file should not have every single line showing as changed

joshlong:
i think i got 'em all :)

markfisher:
This PR needs to be rebased.

markfisher:
Something crazy is happening in the history on this PR. There are a series of commits that occur more than once in the 35 commits that now make up this PR. I'm closing it. Please submit a new one with a clean linear history that is rebased on the current upstream master.

Issue #327, opened 2016-01-22
Properly report not found entities as 404


ericbottard:
A lot of XXXController methods simply return silently when they fail to find a resource (often by name) on which to act.
We should create a dedicated Exception class (maybe with `name` and `kind`/`type` fields) and properly throw it, mapping it to http status 404

Issue #266, opened 2015-12-18
Properly report failure to register a module/library


ilayaperumalg:
When a user tries to register a library that already exists and without `--force` option, the shell output can state the library isn't updated or something. Currently, it says the library is successfully registered which is misleading.

PR #330, opened 2016-01-25
Properly report 'not found' and 'already exists' errors at http level.

ericbottard:
This fixes spring-cloud/spring-cloud-dataflow#327 and
fixes spring-cloud/spring-cloud-dataflow#266


markpollack:
Certainly alot of clean up in here but I think we need to have a quick discussion about the exceptions.  We used to have in XD some hierarchy, which wasn't ideal in several ways, but reducing everything to a ResourceNotFound/ResourceAlreadyExists exceptions might not be ideal either.  That said in https://github.com/spring-projects/spring-xd/blob/master/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/rest/RestControllerAdvice.java  we didnt' handle any of these exceptions differently.  

@cppwfs @mminella as 'their' exceptions got zapped in this PR and to get some general comments.

ericbottard:
I chose to report those as flat exceptions and http level exceptions because a lot of logic is happening at the level already.
The alternative, which could allow for a more fine grained hierarchy (but would we actually benefit from it? we did not really in XD) would be to have business-level exceptions, and have for example each implementation of the repositories be responsible for the checks.

spring-cloud-dataflow-admin-starter/src/main/java/org/springframework/cloud/dataflow/admin/controller/TaskController.java
```diff
@@ -125,6 +125,10 @@ public TaskController(TaskDefinitionRepository repository, ArtifactRegistry regi
 	@RequestMapping(value = "/definitions", method = RequestMethod.POST)
 	public void save(@RequestParam("name") String name,
 			@RequestParam("definition") String dsl) {
+		TaskDefinition previous = repository.findOne(name);
+		if (previous != null) {
+			throw new ResourceAlreadyExistsException(name, "task definition");
+		}
```
mminella: Why are we making the repositories dumb to support this change?  Why can't we keep the repository defensive (have them responsible for maintaining the integrity of their data) and have the Controller catch exceptions thrown and translate them appropriately for the HTTP context?

ericbottard:
I'd like to find a solution that is consistent. To me, 
1. either all exceptions should be thrown at the repository level (and we can use `RestControllerAdvice` to turn them into proper http semantics)
2. or they can be thrown at the Controller level.

I'd prefer option 1, but there are some problems with it. For one, we can't throw an exception on `findOne()` for example (as the interface javadoc states that it should return `null`). Also, while I don't like it, a lot of business logic is in the controller already.

mminella:
I'm not saying all exceptions should be thown from the repository or not.  What I'm saying is that the repository should be responsible for maintaining it's state and have the controllers translate feedback from the repositories accordingly.  I'm ok with `findOne()` returning null when not found and having that check in the controller because the fact that is an exception condition is due to the REST context it's being consumed within.  I don't believe this is a black and white thing.  I think we should look at the repository and do what's right for that component.  Then, after that, what do the controllers need to do to translate into something appropriate for REST.

ericbottard:
Will open another PR

Issue #331, opened 2016-01-25
Developer needs to remove unused task columns from Rest & shell


cppwfs:
Remove the following columns
* ExternalExecutionId
* statusCode

PR #332, opened 2016-01-25
SCDF-331 Removes Unused Task Columns

cppwfs:
* Removes the statusCode and externalExecutionId
* Adds timestamp task back to ArtifactRegistryPopulator

Resolves spring-cloud/spring-cloud-dataflow#331

### DIFF
--- a/spring-cloud-dataflow-shell/src/main/java/org/springframework/cloud/dataflow/shell/command/TaskCommands.java
+++ b/spring-cloud-dataflow-shell/src/main/java/org/springframework/cloud/dataflow/shell/command/TaskCommands.java
@@ -177,9 +177,7 @@ public class TaskCommands implements CommandMarker {
 		modelBuilder.addRow().addValue("Id ").addValue(taskExecutionResource.getExecutionId());
 		modelBuilder.addRow().addValue("Name ").addValue(taskExecutionResource.getTaskName());
 		modelBuilder.addRow().addValue("Parameters ").addValue(taskExecutionResource.getParameters());
-		modelBuilder.addRow().addValue("External Execution Id ").addValue(taskExecutionResource.getExternalExecutionID());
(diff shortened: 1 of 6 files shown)


spring-cloud-dataflow-shell/src/main/java/org/springframework/cloud/dataflow/shell/command/TaskCommands.java
```diff
@@ -177,9 +177,7 @@ public Table view(
 		modelBuilder.addRow().addValue("Id ").addValue(taskExecutionResource.getExecutionId());
 		modelBuilder.addRow().addValue("Name ").addValue(taskExecutionResource.getTaskName());
 		modelBuilder.addRow().addValue("Parameters ").addValue(taskExecutionResource.getParameters());
-		modelBuilder.addRow().addValue("External Execution Id ").addValue(taskExecutionResource.getExternalExecutionID());
```
ericbottard: For the record, since all the properties of the `taskExecutionResource` use getters, you could use a `BeanListTableModel`

cppwfs: In this case I'm not sure it fits in that we are getting detail for one specific task and not a list of them.

ericbottard: Ah sorry, I read too quickly then.

ericbottard: But so you're saying this is a 2 lines (and many columns) Table?
Why not transpose it then, it may be more readable

ericbottard: And **then** I read the code correctly. Nevermind my previous comment

PR #335, opened 2016-01-28
GH-322 Provide group information for streams

### LINKED ISSUE #322: Pass consumer group information for stream modules
Required to bring project up to date with https://github.com/spring-cloud/spring-cloud-stream/pull/276

Otherwise, each application instance will have a distinct consumer group, resulting in duplicate messages.

mbogoevici:
Resolves #322
Streams create durable, named group subscriptions for consumers, as follows:
* intra-stream consumers will use the `default` group subscription;
* consumers from named queues will use the stream name as consumer group;

Also provides a basic way to align with Spring Cloud Stream dropping `topic:` and `queue:` support by not passing them on to the bindings - support at DSL level should address this more comprehensively (see https://github.com/spring-cloud/spring-cloud-dataflow/issues/336).

### DIFF
--- a/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/BindingProperties.java
+++ b/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/BindingProperties.java
@@ -54,6 +54,16 @@ public class BindingProperties {
 	public static final String OUTPUT_TYPE_KEY = OUTPUT_BINDING_KEY_PREFIX + CONTENT_TYPE_KEY;
 
 	/**
+	 * Group property key for input.
+	 */
+	public static final String INPUT_GROUP_KEY = INPUT_BINDING_KEY_PREFIX + "group";
+
+	/**
+	 * Subscription durability for input.
+	 */
+	public static final String INPUT_DURABLE_SUBSCRIPTION_KEY = INPUT_BINDING_KEY_PREFIX + "durableSubscription";
+
+	/**
 	 * Partition properties
 	 */
 	public static final String DEFAULT_PARTITION_KEY_EXPRESSION = "payload";
--- a/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/BindingProperties.java
+++ b/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/BindingProperties.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 the original author or authors.
+ * Copyright 2015-2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,6 +19,7 @@ package org.springframework.cloud.dataflow.core;
  * Spring Cloud Stream property names mostly used for binding.
  *
  * @author Ilayaperumal Gopinathan
+ * @author Marius Bogoevici
  */
 public class BindingProperties {
 
--- a/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/ModuleDefinitionBuilder.java
+++ b/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/ModuleDefinitionBuilder.java
@@ -22,6 +22,7 @@ import java.util.LinkedList;
 import java.util.List;
 
 import org.springframework.cloud.dataflow.core.dsl.ArgumentNode;
+import org.springframework.cloud.dataflow.core.dsl.ChannelType;
 import org.springframework.cloud.dataflow.core.dsl.ModuleNode;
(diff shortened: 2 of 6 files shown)


spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/BindingProperties.java
```diff
@@ -54,6 +54,16 @@
...
 
 	/**
+	 * Group property key for input.
+	 */
+	public static final String INPUT_GROUP_KEY = INPUT_BINDING_KEY_PREFIX + "group";
+
+	/**
+	 * Subscription durability for input.
+	 */
+	public static final String INPUT_DURABLE_SUBSCRIPTION_KEY = INPUT_BINDING_KEY_PREFIX + "durableSubscription";
```
ilayaperumalg: I think this property needs to be added in `spring-cloud-stream`'s [BindingProperties] (https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingProperties.java) so that the it gets updated as `ConsumerProperties` when binding.

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/BindingProperties.java
```diff
@@ -54,6 +54,16 @@
 	public static final String OUTPUT_TYPE_KEY = OUTPUT_BINDING_KEY_PREFIX + CONTENT_TYPE_KEY;
 
 	/**
+	 * Group property key for input.
```
ilayaperumalg: 2016

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/ModuleDefinitionBuilder.java
```diff
@@ -108,4 +117,14 @@ else if (argument.getName().equalsIgnoreCase("outputType")) {
 		}
 		return moduleDefinitions;
 	}
+
+	private String removeChannelTypePrefix(String sourceChannelName) {
```
ilayaperumalg: Can we change the parameter name `sourceChannelName` to just `channelName` as it applies for both source/sink named channel names?

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/ModuleDefinitionBuilder.java
```diff
@@ -83,8 +87,9 @@ else if (argument.getName().equalsIgnoreCase("outputType")) {
 				}
 			}
 			if (m > 0) {
-				builder.setParameter(BindingProperties.INPUT_BINDING_KEY,
-						String.format("%s.%d", streamName, m - 1));
+				builder.setParameter(BindingProperties.INPUT_BINDING_KEY, String.format("%s.%d", streamName, m - 1));
+				builder.setParameter(BindingProperties.INPUT_GROUP_KEY, DEFAULT_CONSUMER_GROUP_NAME);
+				builder.setParameter(BindingProperties.INPUT_DURABLE_SUBSCRIPTION_KEY,"true");
```
ilayaperumalg: Given this specific property is used only in rabbit binder and the consumer module is going to use the same `durable` queue that the upstream module would create, setting this property explicitly here may not be needed. I believe this is needed only for the named channel bindings where we **explicitly** create a `durable` queue for the consumer.

markfisher: need a space after the comma

mbogoevici: I believe we shouldn't make assumptions about how Spring Cloud Stream behaves. Using the `default` queue is a bit of a stretch already. I'd rather be consistent.

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/BindingProperties.java
mbogoevici: Added via https://github.com/spring-cloud/spring-cloud-stream/commit/40709fbf140e0cf79dd3485fbff67e05b186bc15

mbogoevici:
Pushed changes.

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/ModuleDefinitionBuilder.java
ilayaperumalg: yeah, ok. will fix the format (space after comma) when merging.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed, rebased and merged as bff7f69

PR #337, opened 2016-01-28
Use different port H2 DB datasource URL

### LINKED ISSUE #334: Admin local will not start if Kafka is already running locally
The default H2 port is 9092, same as the default port for Kafka brokers. Unless there is a very specific need to to use port 9092 for H2, I would suggest to customize that value for our embedded DB.

ilayaperumalg:
- 9092 collides with Kafka default port

Resolves #334

### DIFF
--- a/spring-cloud-dataflow-admin-local/src/main/resources/admin.yml
+++ b/spring-cloud-dataflow-admin-local/src/main/resources/admin.yml
@@ -18,7 +18,7 @@ spring:
     config:
       uri: http://localhost:8888
   datasource:
-    url: jdbc:h2:tcp://localhost:9092/mem:dataflow
+    url: jdbc:h2:tcp://localhost:9093/mem:dataflow
     username: sa
     password:
     driverClassName: org.h2.Driver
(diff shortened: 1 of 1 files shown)


spring-cloud-dataflow-admin-local/src/main/resources/admin.yml
```diff
@@ -18,7 +18,7 @@ spring:
     config:
       uri: http://localhost:8888
   datasource:
-    url: jdbc:h2:tcp://localhost:9092/mem:dataflow
+    url: jdbc:h2:tcp://localhost:9093/mem:dataflow
```
markfisher: This is still in the range that would be commonly used for a Kafka cluster. Maybe `19092`?

ilayaperumalg: oh yeah, sure.

ilayaperumalg:
updated the PR.

markfisher:
LGTM; squashed and merged as https://github.com/spring-cloud/spring-cloud-dataflow/commit/6a07dd99f722f4b5faa2ee835c9f973f129d92b0

PR #340, opened 2016-01-28
@EnableDataflowAdminServer

joshlong:

PR #342, opened 2016-01-29
add @EnableDataflowAdminServer

markfisher:
Replaces PR #340


markfisher:
Before merging, we're likely to rename this `@EnableDataflowServer` (dropping "Admin"). That could be the first step of a larger change across the board, to replace "admin" with "server" for consistency with other spring-cloud projects. As a follow up to this PR, we need to rename the `spring-cloud-dataflow-admin-starter` subproject anyways, since that is not a proper "starter" in the first place. It could be renamed `spring-cloud-dataflow-server-core`.

markpollack:
I can make the name change on merge.

spring-cloud-dataflow-admin-starter/src/main/java/org/springframework/cloud/dataflow/admin/AdminApplication.java
```diff
@@ -16,31 +16,21 @@
...
  *
  * @author Mark Fisher
+ * @author Josh Long
  */
-@SpringBootApplication
-@EnableAutoConfiguration(exclude = {
-		OAuth2AutoConfiguration.class
-})
+@Configuration
+@EnableDataflowAdminServer
```
ilayaperumalg: I find it interesting to remove `@SpringBootApplication` on this **starter** class while the `@EnableDataflowAdminServer` doesn't have `@SpringBootApplication` either. But I see 'EnableAutoConfiguration` in `AdminConfiguration` class. Is it worth to annotate  `@EnableDataflowAdminServer` with `@SpringBootApplication` and remove `@EnableAutoConfiguration` from `AdminConfiguration` (which could make all the required annotation in one place and leaving the AdminConfiguration as just a Configuration class)?

markpollack:
Using https://github.com/spring-cloud/spring-cloud-config/tree/master/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server as a basis for comparison, `@EnableAutoConfiguration` is on the `main` class vs the configuration class.  Though `@SpringBootApplication` is basically 

``
@Configuration
@EnableAutoConfiguration
@ComponentScan
```
So AdminConfiguration has @ComponentScan so all the functionality is the same.  I guess @SpringBootApplication is more for end user apps as a starting point (like having a single java file with configuration in it.

The javadocs says 
```
This is a convenience annotation that is equivalent to declaring @Configuration,
@EnableAutoConfiguration and @ComponentScan.
```

PR #345, opened 2016-01-30
adding spring-cloud-starter-dataflow-server-local

joshlong:
changes the name of the starter to be foreword compatible


markfisher:
updated the subproject name in the parent pom, then squashed and merged as: https://github.com/spring-cloud/spring-cloud-dataflow/commit/bbe41149469532035e53841a8bbd7f2b93ab872d

PR #351, opened 2016-02-01
Spring Cloud Data Flow Dependencies

### LINKED ISSUE #348: Add a BOM
Spring Cloud Data Flow doesn't (at the moment) provide any dependency management for users building on top of its code base. The pattern adopted (recently) in the other Spring Cloud projects is recommended, where there is a separate spring-cloud-dataflow-dependencies project at the top level and the parent uses it as an import.

joshlong:
this resolves #348 and adds a `spring-cloud-dataflow-dependencies` project that could then be added to the Maven `pom.xml` when users [select the people's _beloved_ checkbox](https://github.com/spring-io/initializr/pull/186) from the [Spring Initializr](http://start.spring.io)

### DIFF
--- a/spring-cloud-dataflow-dependencies/pom.xml
+++ b/spring-cloud-dataflow-dependencies/pom.xml
@@ -1,0 +1,107 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+		 xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+	<!--
+		NB: This BOM is designed to support consumption of Spring Cloud Data Flow
+		from the <a href="http://start.spring.io">Spring Initializr</a>.
+ 	-->
+	<parent>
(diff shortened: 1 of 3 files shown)


spring-cloud-dataflow-dependencies/pom.xml
```diff
@@ -0,0 +1,103 @@
...
+    <relativePath/>
+	</parent>
+	<artifactId>spring-cloud-dataflow-dependencies</artifactId>
+	<version>1.0.0.BUILD-SNAPSHOT</version>
+	<packaging>pom</packaging>
+	<name>spring-cloud-dataflow-dependencies</name>
+	<description>Spring Cloud Data Flow Dependencies</description>
+	<properties>
+		<spring-cloud-commons.version>1.1.0.BUILD-SNAPSHOT</spring-cloud-commons.version>
+		<spring-cloud-dataflow.version>1.0.0.BUILD-SNAPSHOT</spring-cloud-dataflow.version>
```
markfisher: is this necessary? won't it always match `${project.version}`?

dsyer: I believe `${project.version}` will work here (you have to be careful though if this POM ever gets used as a parent - probably it won't be).

markfisher:
rebased, polished, and merged via https://github.com/spring-cloud/spring-cloud-dataflow/commit/288e83986158cea3bf2ab248df85cc19b0a24347

joshlong:
thanks very much for this!

PR #354, opened 2016-02-01
the packaging should be a .jar, not a .pom

joshlong:
the packaging should be a .jar, not a .pom


markfisher:
rebased and merged as https://github.com/spring-cloud/spring-cloud-dataflow/commit/7a9da33139cd29d2929d2c3d76c32389d3c808c0

Issue #274, opened 2015-12-22
Upgrading to use data-commons 1.11 breaks build


markpollack:
We should not hardcode a specific version of data commons but pick it up via the Brixton dependency.

markpollack:
Fixed as part of 69cac13b401e30507de531b00cc7ddd6e07bc32d

PR #359, opened 2016-02-02
Align dependencies with BOMs in Spring Cloud

dsyer:
To allow users to manage dependencies in application projects
for data flow and the rest of Spring Cloud independently, we
need to clean up the BOM a bit and use the Cloud BOMs individually
in the libraries in data flow.

Also fixes the wrapper configuration so Maven can be used on the
command line to do a manual deployment.


markpollack:
The build fails with the same issue as seen in https://github.com/spring-cloud/spring-cloud-dataflow/issues/274

Some newer version of spring data is being used despite picking data-common 1.9.1 explicitly.  Looking into it.

markpollack:
Added profile to main pom for repo locations and fixed test for use with latest data commons.  Squashed and merged as 69cac13

markpollack:
Also created issues in respective projects for TODO comments in the pom

PR #361, opened 2016-02-02
Disable management security by default

### LINKED ISSUE #360: Security is always enabled
Though `admin.yml` is set with `security.basic.enabled` to `false`, it looks like the admin application when started always has the security enabled.

ilayaperumalg:
- Set `management.security.enabled` to `false` to disable actuator endpoints by default

This resolves #360


markfisher:
rebased and merged as https://github.com/spring-cloud/spring-cloud-dataflow/commit/1b5d066b4e5d68e1ca5ea2f24bc9a0620dc97ff3

Issue #343, opened 2016-01-29
Refactor AdminServer to support an @EnableDataFlowServer annotation


markpollack:
This provides a means for users to compose their own Data Flow Server that will help users add their own security configuration and bundle jdbc drivers for connecting to the task repository (such as oracle's).

markpollack:
This issue comes out of Josh Long's work, which also added an `EnvironmentPostProcessor` that would set various properties via the `admin.yml` as default values for other SPI implementations

PR #363, opened 2016-02-02
Set appropriate shutdown URL for in-process deployer

### LINKED ISSUE #362: Management context path isn't needed for in-process deployer
Currently, the in-process deployer uses `managementContextPath` for the shutdown operations on the deployed modules. Since the deployed modules wouldn't use the same management context path, this is not needed and the shutdown path is formed incorrectly.

This will make `undeploy` operation fail.

ilayaperumalg:
- The modules deployed by the in-process deployer should not use `managementContextPath` from the deployer environment.
The modules are supposed to have their own environment. Currently, the inclusion of `spring-cloud-dataflow-admin-starter` classes directory in the module launcher
classloader results in invoking `AdminDefaultEnvironmentPostProcessor` which in turn sets up `defaultProperties` environment from the deployer itself.
 - This also means that the modules can not be undeployed when using in-process deployer because of the incorrect shutdown URL

This resolves #362

### DIFF
--- a/spring-cloud-dataflow-admin-local/pom.xml
+++ b/spring-cloud-dataflow-admin-local/pom.xml
@@ -35,7 +35,6 @@
 				<filtering>true</filtering>
 				<includes>
 					<include>META-INF/spring.factories</include>
-					<include>admin.yml</include>
 					<include>banner.txt</include>
 				</includes>
 			</resource>
--- a/spring-cloud-dataflow-admin-local/pom.xml
+++ b/spring-cloud-dataflow-admin-local/pom.xml
@@ -34,6 +34,7 @@
 				<directory>src/main/resources</directory>
 				<filtering>true</filtering>
 				<includes>
+					<include>admin.yml</include>
 					<include>META-INF/spring.factories</include>
 					<include>banner.txt</include>
 				</includes>
--- a/spring-cloud-dataflow-admin-local/src/test/java/org/springframework/cloud/dataflow/admin/spi/local/AdminDefaultTestEnvironmentPostProcessor.java
+++ b/spring-cloud-dataflow-admin-local/src/test/java/org/springframework/cloud/dataflow/admin/spi/local/AdminDefaultTestEnvironmentPostProcessor.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2015-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.dataflow.admin.spi.local;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.env.EnvironmentPostProcessor;
+import org.springframework.core.env.ConfigurableEnvironment;
+
+/**
+ * @author Ilayaperumal Gopinathan
+ */
+public class AdminDefaultTestEnvironmentPostProcessor implements EnvironmentPostProcessor {
+
+	@Override
+	public void postProcessEnvironment(ConfigurableEnvironment environment, SpringApplication application) {
+		if (environment.getPropertySources().contains("defaultProperties")) {
+			environment.getPropertySources().remove("defaultProperties");
+		}
+	}
+}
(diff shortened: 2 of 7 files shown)


ilayaperumalg:
This PR depends on https://github.com/spring-cloud/spring-cloud-stream/pull/303

markfisher:
I don't fully understand these changes yet, but I do see that it is undoing changes that were just added as part of https://github.com/spring-cloud/spring-cloud-dataflow/issues/343 where @joshlong was explicitly switching away from the hardcoded management context path, since the `@EnableDataFlowServer` use-case could enable customization of any of the properties within `admin.yml`.

markfisher:
sorry... I think I get it now - the shutdown url for the *admin* is not the shutdown url for the apps that need to be un-deployed (obviously we do not want to shutdown the admin itself when attempting to undeploy one of those in-process apps)

spring-cloud-dataflow-admin-local/pom.xml
```diff
@@ -35,7 +35,6 @@
 				<filtering>true</filtering>
 				<includes>
 					<include>META-INF/spring.factories</include>
-					<include>admin.yml</include>
```
markfisher: why is this being removed here?

ilayaperumalg:
Updated the PR.

spring-cloud-dataflow-admin-local/pom.xml
```diff
@@ -34,8 +34,8 @@
 				<directory>src/main/resources</directory>
 				<filtering>true</filtering>
 				<includes>
-					<include>META-INF/spring.factories</include>
 					<include>admin.yml</include>
+					<include>META-INF/spring.factories</include>
```
markfisher: as Eric commented here: https://github.com/spring-cloud/spring-cloud-dataflow/commit/40e321827b5cc9937fdc084c7ef7f1d3e6a3707f#commitcomment-15854422 - I don't think the `spring.factories` file should be included here in the first place, should it?

spring-cloud-dataflow-admin-local/src/test/java/org/springframework/cloud/dataflow/admin/spi/local/AdminDefaultTestEnvironmentPostProcessor.java
```diff
@@ -0,0 +1,34 @@
...
+package org.springframework.cloud.dataflow.admin.spi.local;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.env.EnvironmentPostProcessor;
+import org.springframework.core.env.ConfigurableEnvironment;
+
+/**
+ * @author Ilayaperumal Gopinathan
+ */
+public class AdminDefaultTestEnvironmentPostProcessor implements EnvironmentPostProcessor {
```
markfisher: What is the rationale behind this? It seems like a dangerous idea to make a significant change to how the properties work within tests since that means we are not testing the same behavior that would occur in real world usage.

spring-cloud-dataflow-admin-starter/pom.xml
```diff
@@ -125,7 +125,6 @@
 				<directory>src/main/resources</directory>
 				<filtering>true</filtering>
 				<includes>
-					<include>admin.yml</include>
 					<include>META-INF/spring.factories</include>
```
markfisher: we don't really need `spring.factories` here either do we?

spring-cloud-dataflow-admin-local/src/test/java/org/springframework/cloud/dataflow/admin/spi/local/AdminDefaultTestEnvironmentPostProcessor.java
ilayaperumalg: The problem here is that the `defaultProperties` property source is available for the `module` being deployed because the module context also gets `AdminDefaultEnvironmentPostProcessor` from `s-c-d-admin-starter` **only** in test as the test classpath also includes `s-c-d-admin-starter`. The `defaultProperties` property source shouldn't be available at first place for the modules inside the test.

There is another option to remove the `s-c-d-admin-starter` via maven sure-fire plugin `excludeClasspathComponent` but we would have to skip the `LocalConfigurationTests`.

spring-cloud-dataflow-admin-local/pom.xml
ilayaperumalg: I think we need this entry in order to include `spring.factories` inside the package `jar`.

ilayaperumalg: I verified this and it requires this entry to include it in the jar.

markfisher: yes, makes sense as there's only one resource set... if we were expecting any issues from filtering (we shouldn't in this case), then we would need to provide a non-filtering resource set as well

markfisher:
squashed and merged as bcaaf4a1e966b086f054adeade84f5b2ca1d16ec along with bbe4a21aa4d3b71631373ec2f88a7e559bdd6528

PR #368, opened 2016-02-05
Support tap/named channel DSL

### LINKED ISSUE #314: Add DSL taps support
As a user, I'd like to tap an existing stream via DSL syntax, so I can get a copy of the primary pipeline data for ad-hoc analysis. 

**example:**
```
stream create tweets --definition "twitterstream | log"
stream deploy tweets --properties "module.*.includes=org.springframework.cloud:spring-cloud-stream-binder-rabbit:1.0.0.M3,module.*.excludes=:*-redis"

stream create tweetlang  --definition "tap:stream:tweets > log"
stream deploy tweetlang --properties "module.*.includes=org.springframework.cloud:spring-cloud-stream-binder-rabbit:1.0.0.M3,module.*.excludes=:*-redis"
```

**Acceptance:**
* Decide whether we need `tap:stream..`, `topic:stream..` or `<streamName>.0`
* `tweetlang` stream taps `tweets` stream successfully
* Documentation along with samples included for reference

ilayaperumalg:
- After the binder refactoring on SCS to use pub/sub model and the consumer groups, this PR provides a simplistic approach
to support tap DSL.

  - Add consistent syntax for assigning the binding `destination` name
   - This turns out to be the module `label` instead of `index`
  - For the tap support, the DSL would look like this:

```
 stream create a1 --definition "http | transform | log" --deploy
 stream create tap-http --definition "a1.http > jdbc" --deploy
 stream create tap-transform --definition "a1.transform > counter" --deploy
```
All the named channel syntaxes such as `queue:`, `tap:` and `topic:` are removed.

The named channels that aren't part of the stream endpoints would have colon (:) prefix like this:

```
 stream create tap-a1 --definition "file > :test" --deploy
 stream create test-stream --definition ":abc > log" --deploy

```

This resolves #314


spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/StreamParser.java
```diff
@@ -154,18 +154,16 @@ private StreamNode eatStream() {
...
-
 		Tokens tokens = getTokens();
-		List<ModuleNode> moduleNodes;
+		List<ModuleNode> moduleNodes = new ArrayList<>();
 		if (bridge) {
 			// Create a bridge module to hang the source/sink channels off
 			tokens.decrementPosition(); // Rewind so we can nicely eat the sink channel
-			moduleNodes = new ArrayList<>();
-			moduleNodes.add(new ModuleNode(null, "bridge", tokens.peek().startPos,
+			moduleNodes.add(new ModuleNode(null, "noop", tokens.peek().startPos,
```
ilayaperumalg: Since `noop` processor is already available, I ended up adding it here instead of `bridge` processor. We could still add `bridge` processor in SCSM if we really need or may be rename `noop` to `bridge` and use `bridge` processor.

markfisher: +1 for renaming `noop` to `bridge`

ericbottard: The README should be amended I guess, stating that it is being used not only for performance testing (measuring the cost of binders) and named channels binding

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/StreamParser.java
```diff
@@ -145,8 +146,7 @@ private String eatStreamName() {
 	private StreamNode eatStream() {
 		String streamName = eatStreamName();
 		SourceChannelNode sourceChannelNode = eatSourceChannel();
-
-		// This construct: queue:foo > topic:bar is a source then a sink channel
+		// This construct: foo > bar is a source then a sink channel
```
markfisher: This should still have `:foo` and `:bar`, right?

Also, we should avoid referring to those as "channels" from now on, i.e. we need to realign terminology across the board so that "channel" is reserved for the actual `MessageChannel` component *within an app* that is being bound to the broker via a binding. Whenever we discuss the broker side, we should refer to it as simply "destination" - and whereas we used to use "named channel" as a way to distinguish from pipes, I believe we should avoid "named" as an adjective from now on as well, since it's redundant. Every destination is "named". The destinations that are established for pipe bindings just happen to use a naming convention of `<stream>.<label>`now.

spring-cloud-dataflow-core/src/test/java/org/springframework/cloud/dataflow/core/StreamDefinitionTests.java
```diff
@@ -143,30 +143,30 @@ public void sinkChannelNameIsAppliedToSinkModule() throws Exception {
...
-		boolean isException = false;
-		try {
-			new StreamDefinition("test", "badLog > :foo");
-		}
-		catch (Exception e) {
-			isException = true;
-		}
-		assertTrue(isException);
-	}
+//	@Test
```
markfisher: is this commented out because the Exception is not being raised as expected?

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/ChannelNode.java
```diff
@@ -60,62 +57,32 @@ public String toString() {
...
-		for (int t = 0, max = nameComponents.size(); t < max; t++) {
-			if (t > 0) {
-				s.append(":");
-			}
-			s.append(nameComponents.get(t));
-		}
-		return s.toString();
-	}
-
 	private String getIndexingComponents() {
```
markfisher: this method should be renamed `getLabelComponents()`

markfisher:
The syntax for "taps" should be no different than any other destination. Instead of `a1.http > jdbc`, the syntax should be: `:a1.http > jdbc`.

In other words, since the bindings created for pipes are just generating destinations on the broker exactly the same way as "foo" would be for `:foo > somesink`, there should be no special treatment within the DSL. We happen to follow the `<stream>.<label>` convention for those bindings, and we advertise that so that users can easily "tap" one of those pipe destinations.

That also would simplify the parser code and eliminate the special treatment of a DOT in the token (which could mean we allow DOTs in module names).

ilayaperumalg:
Pushed the changes that uses `:` prefix for all the destination names (from stream definitions as well as the external destination names).

ilayaperumalg:
With the change to rename `no-op` processor to `bridge` [here] (https://github.com/spring-cloud/spring-cloud-stream-modules/pull/180), updated the PR to use `bridge` module to connect to external destination names.

ilayaperumalg:
From @ericbottard : 

```
TapOnChannelExpansionStrategy will need to be revised as well, I believe.
BTW, I'm not sure what the impact of @ConditionalOnMissingBean for that bean was in 40e3218
```

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/ChannelNode.java
```diff
@@ -21,21 +21,18 @@
...
  * @author David Turanski
+ * @author Ilayaperumal Gopinathan
  */
 public class ChannelNode extends AstNode {
 
-	private ChannelType channelType;
+	private String nameComponent;
 
-	private List<String> nameComponents;
+	private List<String> labelComponents;
```
ericbottard: Not all channel references are related to "labels", which are a tap-only concept, are they?

ilayaperumalg:
@ericbottard yeah, good point. I will look into it.

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/ChannelNode.java
ilayaperumalg: Given we aren't using the indexes inside the stream definition, they are just the components after DOT token and there is no impact even if the named channel has `.` in it. Perhaps, we need a better name here?

ericbottard: Yes, my concern is about naming

ilayaperumalg:
@ericbottard It looks like the `TapOnChannelRecoveryStrategy` has some issues getting registered and I submitted a PR for it [here] (https://github.com/spring-cloud/spring-cloud-dataflow/pull/372). Maybe it would be better we address this completion strategy separately from this PR so that this PR would only focus on the tap DSL in particular. I will create new issue created to address this strategy post the PR #372 merge.

ericbottard:
@ilayaperumalg Sure.
Are you saying it does not get registered? I believe it is because of `@ConditionalOnMissingBean`, which may look at the return type of the method. I wanted to investigate this, having that gut feeling (Josh added it) but haven't had the time.
Let's fix it and adapt to new tap syntax in a separate PR

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/SourceChannelNode.java
```diff
@@ -48,11 +48,7 @@ public SourceChannelNode copyOf() {
 	}
 
 	public String getChannelName() {
-		return channelNode.getChannelName();
-	}
-
-	public Object getChannelType() {
-		return channelNode.getChannelType();
+			return this.channelNode.getChannelName();
```
markfisher: excessive indent

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/StreamParser.java
```diff
@@ -290,168 +286,54 @@ private SinkChannelNode eatSinkChannel() {
...
-	 *     <li>queue</li>
-	 *     <li>topic</li>
-	 *     <li>tap</li>
-	 * </ul>
+	 * A channel reference is the label component when referencing a specific
+	 * module/label in a stream definition.
+	 *
 	 * Expected format:
-	 * {@code identifier [ ':' identifier ]* [ '.' identifier ]*}
+	 * {@code [ ':' identifier ]* [ '.' identifier ]*}
```
markfisher: this doesn't look valid now (shows 2 things with ZERO-or-more)... shouldn't it be something like below?:

```{@code ':' identifier ['.' identifier]*}```

File: spring-cloud-dataflow-core/src/test/java/org/springframework/cloud/dataflow/core/dsl/StreamParserTests.java [hunk 8]
```diff
@@ -339,73 +340,53 @@ public void expressions_gh1_2() {
...
 	}
 
 	@Test
 	public void channelVariants() {
-		// Job is not a legal channel prefix
-		checkForParseError("trigger > job:foo", DSLMessage.EXPECTED_CHANNEL_PREFIX_QUEUE_TOPIC, 10, "job");
 
 		// This looks like a label and so file is treated as a sink!
-		checkForParseError("queue: bar > file", DSLMessage.NO_WHITESPACE_IN_CHANNEL_DEFINITION, 7);
+		checkForParseError("http > :test value", DSLMessage.UNEXPECTED_DATA_AFTER_STREAMDEF, 13);
```
markfisher: the comment above this is no longer related

spring-cloud-dataflow-core/src/test/java/org/springframework/cloud/dataflow/core/dsl/StreamParserTests.java
```diff
@@ -339,73 +340,53 @@ public void expressions_gh1_2() {
...
 
-		// 'queue' looks like a module all by itself so everything after is unexpected
-		checkForParseError("queue : bar > file", DSLMessage.NO_WHITESPACE_IN_CHANNEL_DEFINITION, 6);
+		checkForParseError(":boo .xx > file", DSLMessage.NO_WHITESPACE_IN_CHANNEL_DEFINITION, 5);
+		checkForParseError(":boo . xx > file", DSLMessage.NO_WHITESPACE_IN_CHANNEL_DEFINITION, 5);
+		checkForParseError(":boo. xx > file", DSLMessage.NO_WHITESPACE_IN_CHANNEL_DEFINITION, 6);
+		checkForParseError(":boo.xx. yy > file", DSLMessage.NO_WHITESPACE_IN_CHANNEL_DEFINITION, 9);
+		checkForParseError(":boo.xx .yy > file", DSLMessage.NO_WHITESPACE_IN_CHANNEL_DEFINITION, 8);
+		checkForParseError(":boo.xx . yy > file", DSLMessage.NO_WHITESPACE_IN_CHANNEL_DEFINITION, 8);
```
markfisher: looks like we should have one test with `: boo` (space between colon and identifier)

spring-cloud-dataflow-core/src/test/java/org/springframework/cloud/dataflow/core/dsl/StreamParserTests.java
```diff
@@ -339,73 +340,53 @@ public void expressions_gh1_2() {
...
+		StreamNode ast = parse(":foo.bar > file");
+		assertEquals("[(foo.bar:1>8)>(ModuleNode:file:11>15)]", ast.stringify(true));
+		assertEquals("foo.bar", ast.getSourceChannelNode().getChannelName());
 	}
 
 	@Test
 	public void sourceTapChannel() {
-		StreamNode ast = parse("tap:queue:xxy > file");
-		assertEquals("[(tap:queue:xxy:0>13)>(ModuleNode:file:16>20)]", ast.stringify(true));
+		StreamNode ast = parse(":xxy > file");
```
markfisher: this test method name is no longer relevant for this

spring-cloud-dataflow-core/src/test/java/org/springframework/cloud/dataflow/core/dsl/StreamParserTests.java
```diff
@@ -451,10 +432,10 @@ public void errorCases06() {
 
 	@Test
 	public void errorCases07() {
-		checkForParseError("foo > bar", DSLMessage.EXPECTED_CHANNEL_PREFIX_QUEUE_TOPIC, 6, "bar");
-		checkForParseError("foo >", DSLMessage.OOD, 5);
-		checkForParseError("foo > --2323", DSLMessage.EXPECTED_CHANNEL_PREFIX_QUEUE_TOPIC, 6, "--");
-		checkForParseError("foo > *", DSLMessage.UNEXPECTED_DATA, 6, "*");
+		//checkForParseError("foo > bar", DSLMessage.EXPECTED_CHANNEL_PREFIX, 6, "bar");
```
markfisher: can we not have a test that matches this commented-out one? (if not, we should delete rather than comment out, but it seems like we should ensure such a test does lead to an error still)

spring-cloud-dataflow-core/src/test/java/org/springframework/cloud/dataflow/core/dsl/StreamParserTests.java
```diff
@@ -451,10 +432,10 @@ public void errorCases06() {
...
+		checkForParseError(":foo >", DSLMessage.OOD, 6);
+		checkForParseError(":foo > --2323", DSLMessage.EXPECTED_MODULENAME, 7, "--");
```
markfisher: this is not testing the same situation as the one it modified (that was a missing prefix error)

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/ChannelNode.java
markfisher: We will need to do some follow-up renaming in general (changing "channel" to "destination" in most cases, for example), so we can include this on the list. In the *typical* case, the "labelComponents" would only be present for a "tap" (since that means the source destination is one fro a stream, thus having `<stream>.<label>` syntax). However, since someone could explicitly create a destination with a dot in it now, it's generally valid. IIUC, the expected pattern is now: `identifier[.identifier]*`, and that entire thing is actually the NAME of the destination, so I'd argue that even "nameComponent" is misnamed currently. We need to decide on the right name for 2 things:

1. the required identifier that either stands alone or precedes the first DOT
2. the list of optional identifiers that are themselves separated by dots, following the first dot

All that said, I feel as though this arbitrary distinction based on the first dot is no longer necessary. Isn't the whole thing just the name? Do we even need to know if it's a "tap" anymore?

spring-cloud-dataflow-core/src/test/java/org/springframework/cloud/dataflow/core/dsl/StreamParserTests.java
```diff
@@ -339,73 +340,53 @@ public void expressions_gh1_2() {
...
 	@Test
 	public void sourceTapChannel() {
-		StreamNode ast = parse("tap:queue:xxy > file");
-		assertEquals("[(tap:queue:xxy:0>13)>(ModuleNode:file:16>20)]", ast.stringify(true));
+		StreamNode ast = parse(":xxy > file");
+		assertEquals("[(xxy:1>4)>(ModuleNode:file:7>11)]", ast.stringify(true));
 	}
 
 	@Test
 	public void sourceTapChannel2() {
```
markfisher: following from my comment above, I believe we can delete the existing `sourceTapChannel` and rename this one to `sourceTapChannel`.

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/SourceChannelNode.java
markfisher: will fix on merge

spring-cloud-dataflow-core/src/test/java/org/springframework/cloud/dataflow/core/dsl/StreamParserTests.java
markfisher: actually that passes the parser since IIUC colon is an identifier token and then `boo` is the *next* token regardless of space between them

spring-cloud-dataflow-core/src/test/java/org/springframework/cloud/dataflow/core/dsl/StreamParserTests.java
markfisher: actually if uncommented, that one produces `DSLMessage.UNEXPECTED_DATA_AFTER_STREAMDEF` which is not real clear to me at this point

markfisher:
I have a commit on my branch that addresses many of the comments here, but just a couple open questions remain. I will post a link soon.

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/ChannelNode.java
ilayaperumalg: It is just a `name` and there is no distinction between the `tap` and the `named` destination now.

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/StreamParser.java
ilayaperumalg: Yeah, I agree.

spring-cloud-dataflow-core/src/test/java/org/springframework/cloud/dataflow/core/dsl/StreamParserTests.java
ilayaperumalg: yeah, in this case it is tricky to suggest the exact DSL error message. For this definition `foo > bar`, foo is parsed as Module Node and since there is GT token the `bar` is expected to be a named destination. 
1) We could either say, "Expected Channel Prefix" at the `bar` position 
2) but `foo` could also be a named destination. 

May be it is ok to say (1) here? If so, [this] (https://gist.github.com/ilayaperumalg/284d39af2a31dfc8a5db) can be the change we need.

markfisher: I vote for option 1

markfisher:
Here is my branch where I've added one commit (to address PR comments here) on top of the changes here, which I also squashed - from 3 commits - to a single commit: https://github.com/markfisher/spring-cloud-dataflow/commits/tap-dsl

ilayaperumalg:
@markfisher Pushed a commit on top of your changes to include the fix on proper DSL error message and updated this PR with all those commits.

markfisher:
rebased, made a name change ('error' -> 'errorMessage'), added a comment, squashed those changes and merged via the following 3 commits:
6aa417533a841ae0916f567556722cd7a8ea58b3
04f9daccb0b38cc6e665a6e406cbe1210a89635c
bc449193d98a3b29d44868ac2110142165a30ef4

spring-cloud-dataflow-core/src/test/java/org/springframework/cloud/dataflow/core/dsl/StreamParserTests.java
ilayaperumalg: Here the `colon` is parsed as a token and the DSL message `NO_WHITESPACE_IN_CHANNEL_DEFINITION` is thrown only if the parser finds a whitespace `around` the DOT inside the `identifier` that is parsed after the `colon`. Hence, the whitespace after the `:` is ignored here. I believe this is ok given it is only at the beginning after colon.

PR #370, opened 2016-02-08
Support consumer group option

### LINKED ISSUE #336: Revise the use of `queue:` and `topic:` prefixes for named channels
Since we do not support the distinction between queues and topics anymore (see https://github.com/spring-cloud/spring-cloud-stream/issues/263 and related), the prefixes do not serve any real purpose anymore. However, we might want to keep them (or at least `topic:` to be more exact) for backwards compatibility with Spring XD within the DSL.

ilayaperumalg:
This resolves #336 

This PR is based out of tap DSL support PR #368 and main change is [here](https://github.com/spring-cloud/spring-cloud-dataflow/commit/5ef67f145ea4d49950cedc945841db9df32a0aee)

### DIFF
--- a/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/ModuleDefinitionBuilder.java
+++ b/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/ModuleDefinitionBuilder.java
@@ -40,7 +40,9 @@ import org.springframework.util.Assert;
  */
 class ModuleDefinitionBuilder {
 
-	public static final String DEFAULT_CONSUMER_GROUP_NAME = "default";
+	private static final String DEFAULT_CONSUMER_GROUP_NAME = "default";
+
+	private static final String CONSUMER_GROUP_PARAMETER = "group";
 
 	private final String streamName;
 
@@ -100,9 +102,19 @@ class ModuleDefinitionBuilder {
 		}
 		SourceDestinationNode sourceDestination = streamNode.getSourceDestinationNode();
 		if (sourceDestination != null) {
-			builders.getLast().setParameter(BindingProperties.INPUT_BINDING_KEY, sourceDestination.getDestinationName());
-			builders.getLast().setParameter(BindingProperties.INPUT_GROUP_KEY, streamName);
-			builders.getLast().setParameter(BindingProperties.INPUT_DURABLE_SUBSCRIPTION_KEY, "true");
+			ModuleDefinition.Builder sourceModuleBuilder = builders.getLast();
+			sourceModuleBuilder.setParameter(BindingProperties.INPUT_BINDING_KEY, sourceDestination.getDestinationName());
+			String consumerGroupName = streamName;
+			if (sourceDestination.getArguments() != null) {
+				ArgumentNode[] argumentNodes = sourceDestination.getArguments();
+				for (ArgumentNode argument: argumentNodes) {
+					if (argument.getName().equalsIgnoreCase(CONSUMER_GROUP_PARAMETER)) {
+						consumerGroupName = argument.getValue();
+					}
+				}
+			}
+			sourceModuleBuilder.setParameter(BindingProperties.INPUT_GROUP_KEY, consumerGroupName);
+			sourceModuleBuilder.setParameter(BindingProperties.INPUT_DURABLE_SUBSCRIPTION_KEY, "true");
 		}
 		SinkDestinationNode sinkDestination = streamNode.getSinkDestinationNode();
 		if (sinkDestination != null) {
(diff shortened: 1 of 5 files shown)


ericbottard:
`TapOnChannelExpansionStrategy` will need to be revised as well, I believe.
BTW, I'm not sure what the impact of `@ConditionalOnMissingBean` for that bean was in https://github.com/spring-cloud/spring-cloud-dataflow/commit/40e321827b5cc9937fdc084c7ef7f1d3e6a3707f

ericbottard:
Sorry @ilayaperumalg, I believe my last comment was more relevant to https://github.com/spring-cloud/spring-cloud-dataflow/pull/368

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/SourceChannelNode.java
```diff
@@ -48,11 +48,11 @@ public SourceChannelNode copyOf() {
 	}
 
 	public String getChannelName() {
-		return channelNode.getChannelName();
+			return this.channelNode.getChannelName();
```
markfisher: excessive indent

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/ModuleDefinitionBuilder.java
```diff
@@ -87,29 +89,36 @@ else if (argument.getName().equalsIgnoreCase("outputType")) {
...
 		if (sourceChannel != null) {
-			String sourceChannelName = removeChannelTypePrefix(sourceChannel.getChannelName());
-			builders.getLast().setParameter(BindingProperties.INPUT_BINDING_KEY,
-					sourceChannelName);
-			builders.getLast().setParameter(BindingProperties.INPUT_GROUP_KEY, streamName);
-			builders.getLast().setParameter(BindingProperties.INPUT_DURABLE_SUBSCRIPTION_KEY, "true");
+			ModuleDefinition.Builder sourceModuleBuilder = builders.getLast();
+			sourceModuleBuilder.setParameter(BindingProperties.INPUT_BINDING_KEY, sourceChannel.getChannelName());
+			String consumerGroupName = streamName;
+			if (sourceChannel.getArguments() != null) {
```
markfisher: We might want to wait on adding this feature (the user-provided `--group` property via the DSL). I'd like to hear @mbogoevici's thoughts.

mbogoevici: I'm with supporting declaring some specific user group on the DSL. I guess it won't be done very frequently, but I suppose it's good to have it there, since it's a first class feature of the consumers. By default it would be set to the stream name as far as my reading of the code goes, correct? That's what we want, because different streams get different consumer groups then by default.

markfisher: default is the stream name yes

ilayaperumalg: Yeah, `streamName` is the default consumer group name for all the `destinations` that don't belong to pipeline bindings.

markfisher:
@ilayaperumalg can you rebase this now that PR #368 has been merged?

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/ModuleDefinitionBuilder.java
markfisher: also @mbogoevici I think this extraction of "arguments" for the destination token can evolve to support other binder properties as we've started discussing

mbogoevici: Yes @markfisher , and actually expanded to the outputs - e.g. supplying partitioning configuration for a destination.

ilayaperumalg:
@markfisher sure. I am rebasing this branch now. Meanwhile, shall we confirm if we want to use the option `consumerGroup` instead of `group` as group seems more generic IMO.

markfisher:
honestly, I like 'group' since that is exactly the property this maps to: `spring.cloud.stream.bindings.<channel>.group=foo`

ilayaperumalg:
Updated the PR after rebasing off master.

spring-cloud-dataflow-core/src/test/java/org/springframework/cloud/dataflow/core/dsl/StreamParserTests.java
```diff
@@ -468,6 +468,12 @@ public void bridge01() {
 		assertEquals("[(bar:1>4)>(ModuleNode:bridge:5>6)>(boo:8>11)]", sn.stringify(true));
 	}
 
+	@Test
+	public void testSourceChannelArgs() {
```
markfisher: hey didn't we just change all the "channel" occurrences to "destination"? ;)

ilayaperumalg: oops ;)

markfisher:
squashed and merged as https://github.com/spring-cloud/spring-cloud-dataflow/commit/b958155f91a2880dc5ddc36ca21ec62cb156ef11

PR #372, opened 2016-02-08
Register TapOnChannelExpansionStrategy appropriately

### LINKED ISSUE #371: TapOnChannelExpansionStrategy isn't added to the StreamCompletionProvider
The `TapOnChannelExpansionStrategy` is not getting added to the `StreamCompletionProvider` when the dataflow admin is configured.

ilayaperumalg:
- Since this bean is made conditional, the exact type needs to be specified as there are few other beans with the same interface type exist
 - Need to explicitly `add` this bean to the StreamCompletionProvider
 - Rename `TapOnChannelExpansionStrategy` to `TapOnChannelRecoveryStrategy`

This resolves #371

### DIFF
--- a/spring-cloud-dataflow-admin-starter/src/main/java/org/springframework/cloud/dataflow/admin/completion/TapOnChannelRecoveryStrategy.java
+++ b/spring-cloud-dataflow-admin-starter/src/main/java/org/springframework/cloud/dataflow/admin/completion/TapOnChannelRecoveryStrategy.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 the original author or authors.
+ * Copyright 2015-2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,7 +18,6 @@ package org.springframework.cloud.dataflow.admin.completion;
 
 import java.util.List;
 
-import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.cloud.dataflow.admin.repository.StreamDefinitionRepository;
 import org.springframework.cloud.dataflow.completion.CompletionProposal;
 import org.springframework.cloud.dataflow.completion.RecoveryStrategy;
@@ -33,11 +32,15 @@ import org.springframework.cloud.dataflow.core.dsl.ParseException;
  * <p>Lives in this package as it needs access to a {@link StreamDefinitionRepository}.</p>
  *
  * @author Eric Bottard
+ * @author Ilayaperumal Gopinathan
  */
-public class TapOnChannelExpansionStrategy implements RecoveryStrategy<ParseException> {
+public class TapOnChannelRecoveryStrategy implements RecoveryStrategy<ParseException> {
 
-	@Autowired
-	private StreamDefinitionRepository streamDefinitionRepository;
+	private final StreamDefinitionRepository streamDefinitionRepository;
+
+	public TapOnChannelRecoveryStrategy(StreamDefinitionRepository streamDefinitionRepository) {
+		this.streamDefinitionRepository = streamDefinitionRepository;
+	}
 
 	@Override
 	public boolean shouldTrigger(String dslStart, Exception exception) {
--- a/spring-cloud-dataflow-admin-starter/src/main/java/org/springframework/cloud/dataflow/admin/config/AdminConfiguration.java
+++ b/spring-cloud-dataflow-admin-starter/src/main/java/org/springframework/cloud/dataflow/admin/config/AdminConfiguration.java
@@ -26,6 +26,7 @@ import javax.sql.DataSource;
 import org.h2.tools.Server;
 import org.slf4j.LoggerFactory;
 
+import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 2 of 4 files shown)


spring-cloud-dataflow-admin-starter/src/main/java/org/springframework/cloud/dataflow/admin/completion/TapOnChannelRecoveryStrategy.java
```diff
@@ -33,11 +32,15 @@
  * <p>Lives in this package as it needs access to a {@link StreamDefinitionRepository}.</p>
  *
  * @author Eric Bottard
+ * @author Ilayaperumal Gopinathan
  */
-public class TapOnChannelExpansionStrategy implements RecoveryStrategy<ParseException> {
+public class TapOnChannelRecoveryStrategy implements RecoveryStrategy<ParseException> {
```
ericbottard: Doh! Thx for the fix

spring-cloud-dataflow-admin-starter/src/main/java/org/springframework/cloud/dataflow/admin/config/AdminConfiguration.java
```diff
@@ -158,9 +163,13 @@ public void configurePathMatch(PathMatchConfigurer configurer) {
 	}
 
 	@Bean
-	@ConditionalOnMissingBean
+	@ConditionalOnMissingBean(TapOnChannelRecoveryStrategy.class)
```
ericbottard: I would simply drop the `@ConditionalOnMissingBean` here.

spring-cloud-dataflow-admin-starter/src/main/java/org/springframework/cloud/dataflow/admin/config/AdminConfiguration.java
```diff
@@ -158,9 +163,13 @@ public void configurePathMatch(PathMatchConfigurer configurer) {
...
 	public RecoveryStrategy<?> tapOnChannelExpansionStrategy() {
-		return new TapOnChannelExpansionStrategy();
+		RecoveryStrategy<?> recoveryStrategy = new TapOnChannelRecoveryStrategy(streamDefinitionRepository());
+		if (this.streamCompletionProvider != null) {
```
ericbottard: Ah, so [this](https://github.com/spring-cloud/spring-cloud-dataflow/commit/1317402a4280b34c20298837ae4d4e7659f20909) is the reason for it not being registered. Good catch

spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/CompletionConfiguration.java
```diff
@@ -51,24 +52,22 @@
...
-		);
-		List<ExpansionStrategy> expansionStrategies = Arrays.asList(
-				addModuleOptionsExpansionStrategy(),
-				pipeIntoOtherModulesExpansionStrategy(),
-				unfinishedModuleNameExpansionStrategy(),
-				// Make sure this one runs last, as it may clear already computed proposals
-				// and return its own as the sole candidates
-				configurationPropertyValueHintExpansionStrategy()
-		);
+		List<RecoveryStrategy<?>> recoveryStrategies = new ArrayList<>();
```
ericbottard: Is this change so that we can mutate the list later? I'd rather leave it as before and make a copy in the constructor.

I can change this while merging

spring-cloud-dataflow-admin-starter/src/main/java/org/springframework/cloud/dataflow/admin/config/AdminConfiguration.java
ilayaperumalg: I believe it is better to have `ConditionalOnMissingBean` in all the beans defined in this admin configuration to provide an ability to extend/override the beans from this `starter` project to the project that includes it.

spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/CompletionConfiguration.java
ilayaperumalg: yeah, it is for adding the strategies later in the list. Sure, thanks for changing it!

spring-cloud-dataflow-admin-starter/src/main/java/org/springframework/cloud/dataflow/admin/config/AdminConfiguration.java
ericbottard: The other completion related beans don't have this. Also, there is always the ability to override by name, I'm not sure I understand the rationale here...

ilayaperumalg: I think it is always a safe thing to add `@ConditionalOnMissingBean` at the `starter` project to avoid any conflicts in general and adding this specific type here would make it better for the project that uses this `starter` project.

ericbottard:
Tested and merged as a5d8980 with the minor change discussed above.

PR #378, opened 2016-02-10
Rename channel -> destination in StreamParser

ilayaperumalg:
- All the references to `channel` in the StreamParser now changed to `destination`

This resolves spring-cloud/spring-cloud-dataflow#379

### DIFF
--- a/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/ModuleDefinitionBuilder.java
+++ b/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/ModuleDefinitionBuilder.java
@@ -98,15 +98,15 @@ class ModuleDefinitionBuilder {
 			}
 			builders.add(builder);
 		}
-		SourceDestinationNode sourceChannel = streamNode.getSourceDestinationNode();
-		if (sourceChannel != null) {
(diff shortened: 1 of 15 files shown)


spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/ModuleDefinitionBuilder.java
```diff
@@ -98,15 +98,15 @@ else if (argument.getName().equalsIgnoreCase("outputType")) {
 			}
 			builders.add(builder);
 		}
-		SourceChannelNode sourceChannel = streamNode.getSourceChannelNode();
+		SourceDestinationNode sourceChannel = streamNode.getSourceDestinationNode();
```
markfisher: should change the variable names too... basically a case-insensitive search for "channel" in this file should find nothing

ilayaperumalg: ah, missed those in my search. fixing it.

ilayaperumalg:
Updated the PR with few additional renaming changes.

markfisher:
rebased and merged as https://github.com/spring-cloud/spring-cloud-dataflow/commit/b336633d03726e4744a8580fb6079dd74f17056d

PR #383, opened 2016-02-10
general improvements for property keys

markfisher:
- renamed BindingProperties to BindingPropertyKeys
- dropped "_KEY" from those constant names
- created StreamPropertyKeys for the instance count (since that is a top-level property)
- moved a couple constants from BindingProperties to the StreamDeploymentController since they are only used there (also they were not KEYS)
- added more JavaDoc
- updated tests


ilayaperumalg:
LGTM, rebased and merged as 566b776

PR #389, opened 2016-02-12
Set SCDF to use Milestone vs snapshot.

cppwfs:
SCDF should be using M1 milestone and not snapshot.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -25,7 +25,7 @@
 		<java.version>1.7</java.version>
 		<spring-cloud-dataflow-ui.version>1.0.0.BUILD-SNAPSHOT</spring-cloud-dataflow-ui.version>
 		<spring-cloud-stream.version>1.0.0.BUILD-SNAPSHOT</spring-cloud-stream.version>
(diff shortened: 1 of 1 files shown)


markpollack:
Why not snapshot?

markpollack:
Because of https://github.com/spring-cloud/spring-cloud-dataflow/issues/390

cppwfs:
Correct

markfisher:
I don't think this is necessary. The latest snapshot dependency will always include the latest commits, even if those were in a milestone. The actual problem is that the `TaskDatabaseInitializer` has been replaced, and yet it is still being included within the dataflow configuration.

markfisher:
Closing since instead of this PR, we need the one that addresses #391

markfisher:
reopening so that we keep spring-cloud-dataflow on M1 of spring-cloud-task

markfisher:
By keeping spring-cloud-dataflow on the M1 of spring-cloud-task we will not be bringing in the changes to  the initializer. Then, after the next dataflow milestone, we will move back to snapshots at which point #391 will be necessary.

markfisher:
merged

PR #393, opened 2016-02-12
Add tap DSL doc

### LINKED ISSUE #386: Document DSL support for Taps
As a user, I'd like to refer to docs on `taps` with concrete examples, so I can use it as reference while building streams with taps. 

**Acceptance:**

- Perhaps the best would be to adapt and resurface [ref. guide on `taps`](http://docs.spring.io/spring-xd/docs/1.3.1.BUILD-SNAPSHOT/reference/html/#taps) from XD
- DSL semantics for tapping a stream / named-channel are documented
- Examples included

ilayaperumalg:
- also add doc for explicit destination names

This resolves spring-cloud/spring-cloud-dataflow#386

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
@@ -162,10 +162,10 @@ taps can be created at the output of `http`, `step1` and `step2`.
 To create a stream that acts as a 'tap' on another stream requires to specify the `source destination name` for the tap stream. The syntax for source destination name is:
 
 ```
-`<stream-name>.<label/app name>`
+`<stream-name>.<label/app-name>`
 ```
 To create a tap at the output of `http` in the stream above, the source destination name is `mainstream.http`
-To create a tap at the output of first transform app in the stream above, the source destination name is `mainstream.step1`
+To create a tap at the output of the first transform app in the stream above, the source destination name is `mainstream.step1`
 
 The tap stream DSL looks like this:
 
@@ -181,26 +181,30 @@ Note the colon (:) prefix before the destination names. The colon lets the parse
 [[explicit-destination-names]]
 == Connecting to explicit destination names at the broker
 
-One can connect to a specific source destination name located in the broker (Rabbit, Kafka) and connect it to a downstream set of apps.
-In this case, the DSL looks like this:
+One can connect to a specific destination name located in the broker (Rabbit, Kafka etc.,) either at the `source` or at the `sink` position.
+
+The following stream has the destination name at the `source` position:
 
 ```
-  stream create --definition ":mySource > log" --name ingest_from_broker --deploy
+  stream create --definition ":myDestination > log" --name ingest_from_broker --deploy
 ```
 
-In the above stream, the destination name `mySource` is the `source` destination name and the stream receives messages from the source destination `mySource` at the broker and connects it to the `log` app.
+This stream receives messages from the destination `myDestination` located at the broker and connects it to the `log` app.
 
-The destination can act as `sink` as well.
 
-```
-  stream create --definition "http > :mySink" --name ingest_to_broker --deploy
-```
-In the above stream, the destination name `mySink` is the `sink` destination name and the stream sends the messages from `http` app to the destination `mySink` at the broker.
-
-It is also possible to connect `source` and `sink` destinations at the broker.
+The following stream has the destination name at the `sink` position:
 
 ```
-  stream create --definition ":myOtherSource > :myOtherSink" --name source_sink_destinations --deploy
+  stream create --definition "http > :myDestination" --name ingest_to_broker --deploy
+```
+This stream sends the messages from `http` app to the destination `myDestination` located at the broker.
+
+From the above streams, notice that the `http` and `log` apps are interacting with each other via `broker` (through the destination `myDestination`) rather than having a pipe directly between `http` and `log` within a single stream.
+
+It is also possible to connect two different destinations (`source` and `sink` positions) at the broker in a stream.
+
+```
+  stream create --definition ":destination1 > :destination2" --name bridge_destinations --deploy
 ```
 
-In the above stream, both the source and sink destinations are located in the broker. The messages flow from the source destination to the sink destination via a `bridge` app that connects the source and sink destinations.
+In the above stream, both the destinations (`destination1` and `destination2`) are located in the broker. The messages flow from the source destination to the sink destination via a `bridge` app that connects them.
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
@@ -138,3 +138,48 @@ When a stream is comprised of multiple modules with the same name, they must be 
 
   stream create --definition "http | firstLabel: transform --expression=payload.toUpperCase() | secondLabel: transform --expression=payload+'!' | log" --name myStreamWithLabels --deploy
 
+[[tap-dsl]]
+== Tap DSL
+
+Taps can be created at various producer endpoints in a stream. For a stream like this:
+
+  stream create --definition "http | step1: transform --expression=payload.toUpperCase() | step2: transform --expression=payload+'!' | log" --name mainstream --deploy
+
+taps can be created at the output of `http`, `step1` and `step2`.
+
+Tap DSL requires to specify the `source destination name` for the tap stream. The syntax for source destination name is:
+
+`<stream-name>.<label/app name>`
+
+To create a tap at the output of `http` in the stream above, the source destination name is `mainstream.http`
+To create a tap at the output of first transform app in the stream above, the source destination name is `mainstream.step1`
+
+The tap stream DSL looks like this:
+
+  stream create --definition ":mainstream.http > counter" --name tap_at_http --deploy
+
+  stream create --definition ":mainstream.step1 > jdbc" --name tap_at_step1_transformer --deploy
+
+Note the colon (:) prefix before the destination names. The colon lets the parser to parse this as the destination name instead of app name.
+
+[[explicit-destination-names]]
+== Connecting to explicit destination names at the broker
+
+One can connect to specific source destination name located in the broker (Rabbit, Kafka) and connect it to the downstream.
+In this case, the DSL looks like this:
+
+  stream create --definition ":mySource > log" --name ingest_from_broker --deploy
+
+In the above stream, the destination name `mySource` is the `source` destination name and stream receives messages from the source destination `mySource` at the broker and connect it to the `log` app.
+
+The destination can act as `sink` as well.
+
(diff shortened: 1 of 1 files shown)


spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -138,3 +138,48 @@ When a stream is comprised of multiple modules with the same name, they must be
...
+To create a tap at the output of `http` in the stream above, the source destination name is `mainstream.http`
+To create a tap at the output of first transform app in the stream above, the source destination name is `mainstream.step1`
+
+The tap stream DSL looks like this:
+
+  stream create --definition ":mainstream.http > counter" --name tap_at_http --deploy
+
+  stream create --definition ":mainstream.step1 > jdbc" --name tap_at_step1_transformer --deploy
+
+Note the colon (:) prefix before the destination names. The colon lets the parser to parse this as the destination name instead of app name.
```
ericbottard: > lets the parser to parse this

extra "to"

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -138,3 +138,48 @@ When a stream is comprised of multiple modules with the same name, they must be
...
+
+[[explicit-destination-names]]
+== Connecting to explicit destination names at the broker
+
+One can connect to specific source destination name located in the broker (Rabbit, Kafka) and connect it to the downstream.
```
ericbottard: Should use either "a specific source..." or plural form

Moreover, "connect it to the downstream" seems like a malformed sentence. Maybe "connect it to **a** downstream **set of apps**"?

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -138,3 +138,48 @@ When a stream is comprised of multiple modules with the same name, they must be
...
+Tap DSL requires to specify the `source destination name` for the tap stream. The syntax for source destination name is:
+
+`<stream-name>.<label/app name>`
+
+To create a tap at the output of `http` in the stream above, the source destination name is `mainstream.http`
+To create a tap at the output of first transform app in the stream above, the source destination name is `mainstream.step1`
+
+The tap stream DSL looks like this:
+
+  stream create --definition ":mainstream.http > counter" --name tap_at_http --deploy
```
ericbottard: Please use triple backquote markers rather than automation, it's easier to act on automatically if needed

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -138,3 +138,48 @@ When a stream is comprised of multiple modules with the same name, they must be
...
+[[tap-dsl]]
+== Tap DSL
+
+Taps can be created at various producer endpoints in a stream. For a stream like this:
+
+  stream create --definition "http | step1: transform --expression=payload.toUpperCase() | step2: transform --expression=payload+'!' | log" --name mainstream --deploy
+
+taps can be created at the output of `http`, `step1` and `step2`.
+
+Tap DSL requires to specify the `source destination name` for the tap stream. The syntax for source destination name is:
```
markfisher: "Tap DSL" sounds too formal, since now we are simply referring to a destination (as with any use of destination, not just "taps") while we happen to know the naming convention that is used for the destinations created on behalf of streams. I would change this to something like: "To create a stream that acts as a 'tap' on another stream, specify..."

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -138,3 +138,48 @@ When a stream is comprised of multiple modules with the same name, they must be
...
+
+[[explicit-destination-names]]
+== Connecting to explicit destination names at the broker
+
+One can connect to specific source destination name located in the broker (Rabbit, Kafka) and connect it to the downstream.
+In this case, the DSL looks like this:
+
+  stream create --definition ":mySource > log" --name ingest_from_broker --deploy
+
+In the above stream, the destination name `mySource` is the `source` destination name and stream receives messages from the source destination `mySource` at the broker and connect it to the `log` app.
```
ericbottard: **the** stream receives
and connect**s** it to

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -138,3 +138,48 @@ When a stream is comprised of multiple modules with the same name, they must be
...
+
+  stream create --definition "http > :mySink" --name ingest_to_broker --deploy
+
+In the above stream, the destination name `mySink` is the `sink` destination name and the stream sends the messages from `http` app to the destination `mySink` at the broker.
+
+It is also possible to connect `source` and `sink` destinations at the broker.
+
+  stream create --definition ":myOtherSource > :myOtherSink" --name source_sink_destinations --deploy
+
+In the above stream, both the source and sink destinations are located in the broker and messages flow from the source destination to the sink destination.
```
ericbottard: This sentence seems to imply that the flow happens entirely at the broker level, while there are actually two hops and an implicit bridge app involved. I think we should clarify this

ilayaperumalg:
Updated the PR to reflect the suggestions.

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -135,6 +145,62 @@ If directed graphs are needed instead of the simple linear streams described abo
...
+```
+  stream create --definition "http | step1: transform --expression=payload.toUpperCase() | step2: transform --expression=payload+'!' | log" --name mainstream --deploy
+
+```
+taps can be created at the output of `http`, `step1` and `step2`.
+
+To create a stream that acts as a 'tap' on another stream requires to specify the `source destination name` for the tap stream. The syntax for source destination name is:
+
+```
+`<stream-name>.<label/app name>`
```
markfisher: should use a hyphen in `app-name` as well so there's no whitespace... could be confusing

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -135,6 +145,62 @@ If directed graphs are needed instead of the simple linear streams described abo
...
+```
+To create a tap at the output of `http` in the stream above, the source destination name is `mainstream.http`
+To create a tap at the output of first transform app in the stream above, the source destination name is `mainstream.step1`
```
markfisher: add "the" before "first transform"

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -135,6 +145,62 @@ If directed graphs are needed instead of the simple linear streams described abo
...
+Note the colon (:) prefix before the destination names. The colon lets the parser parses this as the destination name instead of app name.
+
+[[explicit-destination-names]]
+== Connecting to explicit destination names at the broker
+
+One can connect to a specific source destination name located in the broker (Rabbit, Kafka) and connect it to a downstream set of apps.
+In this case, the DSL looks like this:
+
+```
+  stream create --definition ":mySource > log" --name ingest_from_broker --deploy
```
markfisher: I think "mySource" is a confusing name here since it might imply a correlation to a `source` app in the `spring-cloud-stream` sense. Maybe just call it "myDestination" or something.

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -135,6 +145,62 @@ If directed graphs are needed instead of the simple linear streams described abo
...
+```
+
+In the above stream, the destination name `mySource` is the `source` destination name and the stream receives messages from the source destination `mySource` at the broker and connects it to the `log` app.
+
+The destination can act as `sink` as well.
```
markfisher: add "a" -> "as a `sink`"

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -135,6 +145,62 @@ If directed graphs are needed instead of the simple linear streams described abo
...
+
+```
+  stream create --definition "http > :mySink" --name ingest_to_broker --deploy
+```
+In the above stream, the destination name `mySink` is the `sink` destination name and the stream sends the messages from `http` app to the destination `mySink` at the broker.
```
markfisher: add "the":  "the `http` app"

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -135,6 +145,62 @@ If directed graphs are needed instead of the simple linear streams described abo
...
+```
+  stream create --definition ":mySource > log" --name ingest_from_broker --deploy
+```
+
+In the above stream, the destination name `mySource` is the `source` destination name and the stream receives messages from the source destination `mySource` at the broker and connects it to the `log` app.
+
+The destination can act as `sink` as well.
+
+```
+  stream create --definition "http > :mySink" --name ingest_to_broker --deploy
```
markfisher: same comment  here about not using "sink" in the name of the destination. It might be okay to use the same name here in both of these examples, destination in source position and destination in sink position (by the way, I often refer to it that way "in the source position" to clarify that it's different from the case of using an actual `source`). If we do use the same name for both, we can even mention that they are interacting with each other but *via* the broker as 2 separate streams rather than having a pipe directly between `http` and `log` within a single stream.

ilayaperumalg:
Updated the PR after addressing the latest review comments.

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -26,14 +26,16 @@ The Spring Cloud Data Flow Admin Server footnote:[The server is implemented by t
 
 New streams are created by posting stream definitions. The definitions are built from a simple DSL. For example, let's walk through what happens if we execute the following shell command:
 
+```
     dataflow:> stream create --definition "time | log" --name ticktock
```
ericbottard: Using \`\`\`, you don't need indentation. Will fix on merge

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -135,6 +145,66 @@ If directed graphs are needed instead of the simple linear streams described abo
...
+The tap stream DSL looks like this:
+
+```
+  stream create --definition ":mainstream.http > counter" --name tap_at_http --deploy
+
+  stream create --definition ":mainstream.step1 > jdbc" --name tap_at_step1_transformer --deploy
+
+```
+
+Note the colon (:) prefix before the destination names. The colon lets the parser parses this as the destination name instead of app name.
```
ericbottard: typo: lets the parser parseS, will fix

ericbottard:
Merged as 3534b0c

PR #414, opened 2016-02-17
Add additional env vars to deployed module

dsyer:
User can specify additional command line args like this in the
local deployer app:

```
deployer:
  local:
    sharedArgs:
      includes: org.springframework.cloud:spring-cloud-starter-config:1.1.0.BUILD-SNAPSHOT
      spring.application.name: ${module.label}
      spring.cloud.config.uri: http://localhost:8888
```

Then all modules are launched with "includes" on the classpath
and enough information to bootstrap the config client.

The "includes" is a feature of the module launcher, and it is a
comma separated list, so we have to be careful to merge
module and deployment specific arguments with the same name.

### DIFF
--- a/spring-cloud-dataflow-deployer-local/pom.xml
+++ b/spring-cloud-dataflow-deployer-local/pom.xml
@@ -25,6 +25,11 @@
 		</dependency>
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-configuration-processor</artifactId>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter-test</artifactId>
 			<scope>test</scope>
 		</dependency>
--- a/spring-cloud-dataflow-deployer-local/src/main/java/org/springframework/cloud/dataflow/deployer/local/OutOfProcessModuleDeployer.java
+++ b/spring-cloud-dataflow-deployer-local/src/main/java/org/springframework/cloud/dataflow/deployer/local/OutOfProcessModuleDeployer.java
@@ -26,8 +26,12 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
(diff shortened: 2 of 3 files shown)


ericbottard:
Should these shared args really apply only to the local deployer? Also, note that those args *in fine* apply to the **module**, not really to the deployer

dsyer:
They are "shared" by all modules, hence the name. It's not for the local deployed right now, but I suppose a similar feature might make sense.

ericbottard:
Also, just to clarify, know that the same outcome may be attained by using module deployment properties:

`stream deploy foo --properties "module.*.includes=xxx,module.*.spring.cloud.config.uri=yy"` *etc.*

dsyer:
It's the same outcome as far as the module is concerned. But for users it's completely different (they don't have to type that includes mumbo jumbo for every stream deployment).

dsyer:
Ping. Can we merge this or something like it? Even if we don't support includes any more (boo) we can expect that this feature might be useful (it's not related directly to includes).

markpollack:
Didn't realize this was assigned to me.  BTW, the spot to support for adding a jar to the classpath seems to be a natural fit for a boot loader.  While I am not in having transitive or excludes functionality, would seem a boot loader would be the spot for any more sophisticated implementations that change the classpath at runtime.

ericbottard:
Irrespective of whether we merge this as-is or not, could we consider adding this functionality to _any_ deployer (_i.e._ not make the keys deployer.**local**.sharedArgs but just deployer.sharedArgs)

markfisher:
If at all possible, we should defer this feature until we switch to the generic deployer and figure out the `ModuleLauncher

markpollack:
Sorry for such a long delay.  I created https://github.com/spring-cloud/spring-cloud-deployer/issues/70 and also https://github.com/spring-cloud/spring-cloud-dataflow/issues/661 and https://github.com/spring-cloud/spring-cloud-dataflow/issues/662 to add env-var support now that we are based on AppDeployer.   

It isn't clear to me if there is a Boot Launcher implementation that would provide the same functionality as the ModuleLauncher for including other jars at runtime.  If there is a Boot feature for this, we can expose it in dataflow/deployers.

Issue #417, opened 2016-02-24
Use a prefix for AdminProperties


ericbottard:
This seems to be a generally sound idea, and is even more important when running tests via maven, which has a hidden `localRepository` that conflicts with ours.

PR #416, opened 2016-02-19
Renaming `admin` -> `server`

ilayaperumalg:
- Doc updates
 - Change `admin.yml` to `dataflow-server.yml`
 - Change package name `admin` to `server`
 - Change shell command `admin` to `dataflow`

Resolves spring-cloud/spring-cloud-dataflow#352

PR #418, opened 2016-02-24
Use a prefix for DataFlowServerProperties

ericbottard:
Fixes spring-cloud/spring-cloud-dataflow#417


markfisher:
`AdminProperties` is about to be renamed when this is merged: https://github.com/spring-cloud/spring-cloud-dataflow/pull/416

Perhaps we should wait? Also, as you might be anticipating based on the prefix name you chose, the properties that are currently defined might move to the `spring-cloud-deployer` layer once we make that shift.

ericbottard:
Yes, I know. And to add to the confusion, the current set of 3 properties are actually used by the _launcher_ on the launched module side.
I wouldn't mind waiting if this wasn't needed by https://github.com/spring-cloud/spring-cloud-dataflow-admin-cloudfoundry/pull/23 (which for some reason did not appear in the timeline of _this_ PR)

ericbottard:
Rebased

ilayaperumalg:
LGTM, rebased, modified the commit description to read `DataFlowServerProperties` and merged as fb36d9c08cdd8cca5a77370df397ee4607c3e8ce

PR #419, opened 2016-02-25
Add app launcher/configuration metadata

ilayaperumalg:


ilayaperumalg:
There are few changes that need to be done either iteratively or in the same PR:

1) Change the references of `module` to `app` in all the places including the class names. Currently, only the project and doc are updated with this change. The class name changes still require co-ordination from across the `deployer` implementations
2) The docker image name is changed to `springcloud/dataflow-app-launcher` which requires addition of CI docker push on `dataflow` repo and update of docker image name across all the `deployer` implementations.

markfisher:
I don't think we should invest more time in the items you list in point 1 above since the `app-launcher` (f.k.a. `module-launcher`) itself is going to be removed completely once we make the switch to `spring-cloud-deployer`. Likewise, I suppose we will no longer have a Docker image (your point 2) for this in the first place (although we may have other Docker images as a means to deploy stream apps in certain deployment environments).

mbogoevici:
One concern with this is that the package renaming needs coordinated with other repositories. YARN I believe.

ilayaperumalg:
Yeah, the co-ordination needs to happen across all the deployer implementations (either the package change or dependency name update)

spring-cloud-dataflow-app-launcher/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+From the `spring-cloud-dataflow/spring-cloud-dataflow-app-launcher` directory:
+
+````
+java -Dmodules=org.springframework.cloud.stream.module:time-source:jar:exec:1.0.0.BUILD-SNAPSHOT -Dspring.cloud.stream.bindings.output.destination=ticktock -jar target/spring-cloud-dataflow-app-launcher-1.0.0.BUILD-SNAPSHOT.jar
+java -Dmodules=org.springframework.cloud.stream.module:log-sink:jar:exec:1.0.0.BUILD-SNAPSHOT -Dargs.0.server.port=8081 -Dspring.cloud.stream.bindings.input.destination=ticktock -jar target/spring-cloud-dataflow-app-launcher-1.0.0.BUILD-SNAPSHOT.jar
+````
+
+Note that `server.port` needs to be specified explicitly for the log sink app as the time source app already uses the default port `8080`.
+The aoo launcher is able to launch several apps, hence the `args.0.` prefix.
```
markfisher: "aoo" should be "app"

spring-cloud-dataflow-app-launcher/README.md
```diff
@@ -0,0 +1,88 @@
...
+docker run -p 8080:8080 -e MODULES=org.springframework.cloud.stream.module:time-source:jar:exec:1.0.0.BUILD-SNAPSHOT \
+ -e spring.cloud.stream.bindings.output.destination=ticktock -e SPRING_REDIS_HOST=<Redis-Host-IP> springcloud/dataflow-app-launcher
+
+docker run -p 8081:8080 -e MODULES=org.springframework.cloud.stream.module:log-sink:jar:exec:1.0.0.BUILD-SNAPSHOT \
+  -e spring.cloud.stream.bindings.input.destination=ticktock -e SPRING_REDIS_HOST=<Redis-Host-IP> springcloud/dataflow-app-launcher
+````
+Note the binding name `ticktock` is specified for the source's output and sink's input.
+The port mapping is done so that individual apps' http endpoints can be accessed via the docker VM port.
+
+To run pub/sub apps individually on docker, the binding name has to start with `topic:`.
```
markfisher: This (`topic:`) is no longer valid. I think we should just delete from this line to the end of the file.

spring-cloud-dataflow-app-launcher/pom.xml
```diff
@@ -0,0 +1,155 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+
+	<artifactId>spring-cloud-dataflow-app-launcher</artifactId>
+	<packaging>jar</packaging>
+	<name>spring-cloud-dataflow-app-launcher</name>
+	<description>Application for launching spring-cloud-stream modules</description>
```
markfisher: should replace "modules" here, but might want to change "Application" as well (maybe "Bootstrap")

spring-cloud-dataflow-app-launcher/src/main/docker/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM anapsix/alpine-java:8
```
markfisher: is this the best java8 image to use?

mbogoevici: It's the most lightweight and stable. @markpollack has spend some time researching it.

spring-cloud-dataflow-completion/pom.xml
```diff
@@ -20,7 +20,8 @@
 		</dependency>
 		<dependency>
 			<groupId>org.springframework.cloud</groupId>
-			<artifactId>spring-cloud-stream-configuration-metadata</artifactId>
+			<artifactId>spring-cloud-dataflow-configuration-metadata</artifactId>
+			<version>1.0.0.BUILD-SNAPSHOT</version>
```
markfisher: we shouldn't need this explicit version - need to add to dependencyManagement?

spring-cloud-dataflow-deployer-local/pom.xml
```diff
@@ -17,7 +17,8 @@
 		</dependency>
 		<dependency>
 			<groupId>org.springframework.cloud</groupId>
-			<artifactId>spring-cloud-stream-module-launcher</artifactId>
+			<artifactId>spring-cloud-dataflow-app-launcher</artifactId>
+			<version>1.0.0.BUILD-SNAPSHOT</version>
```
markfisher: shouldn't need this explicit version

spring-cloud-dataflow-app-launcher/src/main/docker/Dockerfile
markfisher: ok, just making sure ;)

mbogoevici:
Merged w/minor changes https://github.com/spring-cloud/spring-cloud-dataflow/commit/151d54078b355f28bd8968c7813ca0c742fb2c8d

PR #422, opened 2016-02-26
Solve problem proxy without authentication

Mwea:
If a proxy doesn't need authentication but the user has no choice to use it, he can set a proxy with an authentication set as null


ilayaperumalg:
@Mwea thanks for moving the PR here and the formatting fixes. 

We are still waiting on couple of things:

1) Confirmation on the contributor agreement
2) Changes to this PR to include what Marius mentioned in the previous PR:

- we'd need to check if a proxy is configured at all, otherwise this will fail when no proxy is configured.

Mwea:
@ilayaperumalg I just signed to the contributor agreement but didn't know the name of the project lead so I wrote Alain Sahli which is the first name I got when I searched for spring cloud dataflow projet
I also added a test to see if the proxy is configured or not :+1:

ilayaperumalg:
@Mwea Thanks for adding the test. I believe you need to choose the correct name so that it will send the confirmation to the project lead. Hence, please choose `Mark Fisher` or `Mark Pollack` from the list as they are the project leads for Data Flow. Once you have done it, please post the confirmation number here.

Mwea:
Done 
> Your confirmation number is: 164620160226082353

ilayaperumalg:
LGTM, squashed, rebased and merged as b3e4d2f.

@Mwea thank you for the contribution!

Issue #273, opened 2015-12-22
REST-API returns 100s of k/v pairs as module properties


sabbyanandan:
As a user, I'm trying to use the REST-API to retrieve module specific properties; however, the returned document includes 100s of options that are not directly associated with the module, but inherited from Boot itself. 

This is related to spring-cloud/spring-cloud-dataflow#187. Though we partially fixed it with incremental tab notion in `shell`, this is still a problem for anyone who's directly using the REST-API. 

The immediate candidate impacted by this is `Flo`. The tool-tip for the `modules` listed in the palette loads very slow due to this behavior. It does eventually load though. Tracker story: [#110520316](https://www.pivotaltracker.com/n/projects/1266326/stories/110520316).

cc:/ @BoykoAlex

sabbyanandan:
@dsyer: We'd love to hear if you've other ideas how to filter application specific property overrides. There are three areas in SCDF with UX issues.

1. Shell: when you issue `module info source:time`, you'll see 100s of module options. We'd rather want to see only `time` specific metadata options
2. Flo: palette is loaded with all OOTB applications and the associated metadata options; it takes 10-20 seconds to load the page
3. Job/Task deployment page in `admin-ui` is dynamically created based on the job/task properties as form-fields with respective defaults pre-populated. Given all these options, >95% of the form-field elements in the UI are irrelevant for the job/task application itself

cc:/ @ghillert @ericbottard

dsyer:
The "time" module has a `TimeProperties`. I would think that might be a good source of minimal properties that are applicable. Some other modules piggy back on existing Spring Boot conventions (which is a good thing), but you could still support that by just having the module author list the "properties beans" in a config file (e.g. /META-INF/module.properties or something). If there is no explicit declaration it should be easy to scan for `@ConfigurationProperties` in the module, either at build tiume or run time.

dsyer:
Scanning the available modules I'd say the biggest problem is still the flat namespace - how are we going to ensure that there is no clash in an environment that wants to run 2 file sources, for instance? I think it's really quite important that we wean the modules off that stage of their development somehow. Probably there's another issue open somewhere for that?

ericbottard:
There is no clash. Modules get their own Environment

dsyer:
Maybe in Data Flow? Not sure what you mean otherwise because (for example) there is a plain vanilla `@EnableConfigurationProperties` in all modules.

ericbottard:
[...]

PR #432, opened 2016-03-02
Only show whitelisted configuration properties

ericbottard:
Fixes spring-cloud/spring-cloud-dataflow#273


ericbottard:
Once reviewed, I'll work on a PR for https://github.com/spring-cloud/spring-cloud-stream-modules that adds the `dataflow.properties` files

spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/stream/configuration/metadata/ModuleConfigurationMetadataResolver.java
```diff
@@ -154,4 +176,9 @@ protected ClassLoader createClassLoader() throws Exception {
 			return createClassLoader(classPathArchives);
 		}
 	}
+
+	public static void main(String[] args) {
+		ModuleConfigurationMetadataResolver resolver = new ModuleConfigurationMetadataResolver();
+		resolver.listPropertyGroups(new FileSystemResource("/Users/ericbottard/Documents/projects/spring-cloud-data-family/spring-cloud-stream-modules/trigger-source/target/trigger-source-1.0.0.BUILD-SNAPSHOT-exec.jar"));
```
markfisher: `"/Users/ericbottard"`

BoykoAlex:
I have tried this PR in my local branch and I don't get any options for any module now. Did the endpoint change?
GET http://0.0.0.0:8000/modules/source/file
Returns:
{"name":"file","type":"source","coordinates":"org.springframework.cloud.stream.module:file-source:jar:exec:1.0.0.BUILD-SNAPSHOT","shortDescription":null,"options":[]}

I started Redis and local SCDF server. Pretty sure i have applied changes in the PR correctly, since they seem to be present.

spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/stream/configuration/metadata/ModuleConfigurationMetadataResolver.java
ericbottard: Correct, this was a debug run. Will remove

ericbottard:
@BoykoAlex yes, you are correct. With this PR applied, modules need a `META-INF/dataflow.properties` file for any property to show up. I'll issue a PR for modules, but I'd like to have consensus on this before doing all the work. Let me push something for just one module, so you can have a feel of it.

ericbottard:
@BoykoAlex see https://github.com/spring-cloud/spring-cloud-stream-modules/pull/208

ghillert:
@ericbottard For the properties for modules on the UI side - I had been using as a work-around https://github.com/spring-cloud/spring-cloud-dataflow/issues/436 - To limit hundreds of properties to show up. Do you think that will be totally invalid with your PR?

ericbottard:
The PR is far from being merged, but if it is, then yes. Filtering should be controlled on the server side I would argue.

ghillert:
@ericbottard See comment by @markfisher  at https://github.com/spring-cloud/spring-cloud-stream-modules/pull/208

ghillert:
@ericbottard I seem to have problem getting this to work. I tried it for the `time` task at https://github.com/spring-cloud/spring-cloud-task-modules/tree/master/timestamp. Added `dataflow.properties` with `configuration.classes=org.springframework.cloud.task.timestamp.TimestampTaskProperties` but now no property is showing up. Looks like I am missing something trivial (?)

ericbottard:
@ghillert Are you sure you added the file in the correct folder?
If you grab spring-cloud/spring-cloud-stream-modules#208, do you see props for the cassandra sink?

ghillert:
@ericbottard I am making the following observations:

- without your PR I am getting hundreds of options
- with your PR (without `dataflow.properties`) I get zero options
- with your PR (**with** `dataflow.properties`), I get zero options.

The properties file is loaded. I see that in `ModuleConfigurationMetadataResolver#listPropertyGroups`:

```java
			if (propertiesFile.isReadable()) {
				properties.load(propertiesFile.getInputStream());
			}
```

property value is:

```
{configuration.classes=org.springframework.cloud.task.timestamp.TimestampTaskProperties}
```

But fails at:

```java
				if (!filter || (!sourceTypes.isEmpty() && classes.containsAll(sourceTypes))) {
					result.add(group);
				}
```

markpollack:
Instead of building this metadata by hand (and to echo mark's comment, the metafile name should not be dataflow related) can be piggy back on http://docs.spring.io/spring-boot/docs/current-SNAPSHOT/reference/htmlsingle/#configuration-metadata or make some customization to this current feature?  It seems that boot 2.0 might address our needs (if we can identify gaps) and this looks to be pretty much on target as a general purpose feature for all apps, not just those used in dataflow.

ghillert:
@ericbottard Also a question: Why do we need `dataflow.properties` - couldn't we derive this from the corresponding `spring-configuration-metadata.json`?

aclement:
@markpollack I had the same thought yesterday Mark, to build on that configuration metadata feature.

ericbottard:
@markpollack @ghillert no we can't, really, because there are several `spring-configuration-metadata.json` files visible at runtime, including ones coming from boot (hence the 100s of options visible). 

As I've been saying for quite some time now, we need a way to distinguish between options. Having a handcrafted metadata file makes me sick, but that is what seemed to be [the accepted answer](https://github.com/spring-cloud/spring-cloud-dataflow/issues/273#issuecomment-184600601). Are we back to square one again?

markpollack:
One other thought, assume the applications that are to be deployed by dataflow have a common prefix, e.g. `spring.cloud.stream`, having this prefix as a basic filter mechanism would take care of a great majority of the cases, leaving only some common boot properties such as `server.port` to be somehow idenetified (perhaps those are manually entered?)

markpollack:
What I am suggesting is not the exact reuse of `spring-configuration-metadata.json`, but to build something we need that uses that same approach and generates the metadata we would like to have.  I know I'm late to comment on this - so apologies.  The accepted answer :) also mentions `If there is no explicit declaration it should be easy to scan for @ConfigurationProperties in the module, either at build time or run time.` - I think if we start down a build time solution it will pay off in what can be incorporated into boot 2.0.  Similar thing is happening now with initalizr and our OOTB apps so it can generate some boilerplate pom/main.java code.

markpollack:
So basically modifying http://docs.spring.io/spring-boot/docs/current-SNAPSHOT/reference/htmlsingle/#configuration-metadata-annotation-processor to have either slightly different output/filename and to restrict to an specific `@ConfigurationProperites` annotation that has the majority of the properties.  Additional by-hand properties could be stubbed out in some way manually and merged into the final result.  Only reading up on this now....

ericbottard:
> One other thought, assume the applications that are to be deployed by dataflow have a common prefix, 

Sooo, this is the solution I advocated last _summer_ (although at the time I kept using `module.` in my proposals as an example, instead of `spring.cloud.stream`). 

This solution still has a problem though, as you point out. The problem can be seen from two angles really. I call it the "philosophical problem", as one could argue that _any_ property that may make sense to a boot app (including properties the app author doesn't think about at the time of writing, but which may get used later, _e.g._ because of composition) may be deemed valuable.
Or you can see the problem through a purely technical lens, and indeed, there are _some_ properties that boot provides (`server.port`, or `spring.datasource.jdbc.*`) that we may want to advertise.

ericbottard:
> it should be easy to scan for @ConfigurationProperties in the module

Having wandered down that road, I can tell you it is *not* easy. There is no easy way to know which classes will really be used at runtime. And accepting all that are in the überjar amounts to reading the whole assembled `spring-configuration-metadata.json`, which is what we do today

markpollack:
I discussed the composition aspect a bit with Marius, it maybe a red herring in that at least the current way modules are composed, each 'app' gets it's own application context.  Need to think it through.  

I think the main argument to have a prefix is that we encountered collisions with other infrastructure, e.g. `localRepository` and maven.  Otherwise, I think when we use config server, the config information is already scoped to the application name.  Perhaps having @ConfigurationProperties("${app.name}) would give us the flexibility we want, since we would like to reuse a class that has that annotation across multiple applications to provide consistency to application options.   Seems this isn't liked so much when discussed with the boot team last by my understanding.  @philwebb what do you think? 

I get the 5 steps forward, 8 steps back, 4 steps forward issues....

markpollack:
I don't see how reading `@ConfigurationProperties` in an annotation post processor is not easy....we are doing it now it seems.  I don't understand the comment `There is no easy way to know which classes will really be used at runtime`.  The base use case, take [TimeProperties](https://github.com/sobychacko/spring-cloud-stream-app-starters/blob/gh-202/time/spring-cloud-starter-stream-source-time/src/main/java/org/springframework/cloud/stream/app/time/TimeSourceProperties.java).  At build time, we know this is the `@ConfigurationProperties` that provides the main properties to expose.  Instead of creating a metadata file manually, we automate it to look for this specific annotation.  What am I missing here...?

markpollack:
The exposure of `server.port`, or `spring.datasource.jdbc.*` is also so common that much like there are already annotations to modify the generation of the metadatafile, we can have some files that have these properties ready to go and then `To support such cases and allow you to provide custom "hints", the annotation processor will automatically merge items from META-INF/additional-spring-configuration-metadata.json into the main meta-data file`

ericbottard:
I was not referring to the process of reading stuff using an annotation processor, but reading the *result* of something produced by an annotation processor (what we do today, basically).

So, seems like what you're proposing would be duplicating what boot already offers (at least in some interim scenario). Also, it seems to break as soon as we try to re-use an already-produced artifact (*i.e.* the app currently being built does not have the interesting `@ConfigurationProperties` annotated classes in source form)

markpollack:
Yes, it would be duplicating or adding functionality to what boot has such that it can eventually become part of boot since our use case can't be the only thing driving boot in this direction to have more customization of the metadata.  I'm sure some drop downs in the IDE now are overwhelming...  The breakage is interesting, I'm not familiar with annotation processors (only docklets ;)).  Does that mean that @ConfigurationProperties such as [FileConsumerProperties](https://github.com/sobychacko/spring-cloud-stream-app-starters/blob/gh-202/app-starters-common/src/main/java/org/springframework/cloud/stream/app/file/FileConsumerProperties.java) which live in a different .jar are not accessible?  I don't see us reusing the [common properties](https://github.com/sobychacko/spring-cloud-stream-app-starters/tree/gh-202/app-starters-common) yet in the new repo...

ericbottard:
Updated the PR to implement what is documented in the google doc.

In a nutshell,
* properties that use the `spring.cloud.stream.app` prefix are visible by default and can be passed in w/o a prefix
* other properties can be whitelisted (either by class or single property) to be visible
* some properties can be globally whitelisted by same means (currently only `server.port`)

A couple of rewarks:

* This is a dataflow-only feature
* The whitelist file is named `dataflow.properties`, which can be changed after discussion. But see bullet above
* The "short form" detection criteria is the absence of dot. This means that we can't use nested class `@ConfigurationProperties` approach.
* Still needs to add support for `exhaustive=true` at the REST/shell/ui level once this PR is agreed upon
* This uses a "constant" `spring.cloud.stream.app` prefix. Before we start complicating things, I'd really like people to assess this PR in its current state and pinpoint the tooling issues. The boot javadocs explicitly state that several classes can contribute to the same prefix, so...

ericbottard:
This is what we get once modules have been made "compatible" with this new scheme (basically just changing the prefix in their `@ConfigurationProperties`:

```
dataflow:>module info --name sink:log
Information about sink module 'log':
Resource URI: maven://org.springframework.cloud.stream.module:log-sink:jar:exec:1.0.0.BUILD-SNAPSHOT
╔══════════════════════════════╤══════════════════════════════╤══════════════════════════════╤══════════════════════════════╗
║         Option Name          │         Description          │           Default            │             Type             ║
╠══════════════════════════════╪══════════════════════════════╪══════════════════════════════╪══════════════════════════════╣
║level                         │The level at which to log     │<none>                        │org.springframework.integratio║
║                              │messages.                     │                              │n.handler.LoggingHandler$Level║
║name                          │The name of the logger to use.│<none>                        │java.lang.String              ║
║expression                    │A SpEL expression (against the│payload                       │java.lang.String              ║
║                              │incoming message) to evaluate │                              │                              ║
║                              │as the logged message.        │                              │                              ║
║server.port                   │Server HTTP port.             │<none>                        │java.lang.Integer             ║
╚══════════════════════════════╧══════════════════════════════╧══════════════════════════════╧══════════════════════════════╝

dataflow:>module info --name source:http
Information about source module 'http':
Resource URI: maven://org.springframework.cloud.stream.module:http-source:jar:exec:1.0.0.BUILD-SNAPSHOT
╔══════════════════════════════╤══════════════════════════════╤══════════════════════════════╤══════════════════════════════╗
║         Option Name          │         Description          │           Default            │             Type             ║
╠══════════════════════════════╪══════════════════════════════╪══════════════════════════════╪══════════════════════════════╣
║pathPattern                   │An Ant-Style pattern to       │/                             │java.lang.String              ║
║                              │determine which http requests │                              │                              ║
║                              │will be captured              │                              │                              ║
║server.port                   │Server HTTP port.             │<none>                        │java.lang.Integer             ║
╚══════════════════════════════╧══════════════════════════════╧══════════════════════════════╧══════════════════════════════╝
```

mbogoevici:
One possible suggestion here, before reading the code: instead of using the `spring.cloud.stream.app` prefix convention here, could we just assume that in the absence of a whitelist (whatever configuration triggers it), everything is whitelisted?

The benefit of doing that would be that new (e.g. custom) modules that do not opt into the mechanism would still get options advertised (point at which they may decide that it's better to restrict what they advertise), whereas with the current approach, nothing will be shown (which could be eventually more confusing). Also, this would allow to gradually transition the existing applications, while still having something workable. 

Also, I think it would be useful to whitelist patterns - e.g. `cassadra.sink.*` (unless this is being done already, but doesn't seem to). 

The drawback would be that, by default, you would still get a ton of properties.

markpollack:
The whitelist is only for 'existing boot' properties, all the properties under the prefix would automatically be available.  (as i understand it)

mbogoevici:
@markpollack 
>The whitelist is only for 'existing boot' properties, all the properties under the prefix would automatically be available.

Correct, I was suggesting a strawman alternative.

ericbottard:
Rebased on top of latest master

markpollack:
Let's catch up on this - not going to go with a fixed prefix, also need to expose all properties by default is no whitelist specified.  Property nesting seems to be working well. I have a wip on a local branch that I plan to push out later today.

ericbottard:
Closing in favor of https://github.com/spring-cloud/spring-cloud-dataflow/pull/672

PR #439, opened 2016-03-08
Support proxy config via DataFlowServer properties

### LINKED ISSUE #420: Spring failing to connect behind corporate proxy for the out-of-process deployer
Hi,

We are failing to deploy the basic ticktock stream on Spring cloud dataflow.

We are working behind a corporate proxy  and it seems that the repo system doesn't take any informations from our `dataflow-server.yml ` (it also fails by lauching it with -dHttp parameters)

`server:
  port: 9393
management:
  contextPath: /management
  security:
    enabled: false
info:
  app:
    name: "@project.artifactId@"
    description: "@project.description@"
    version: "@project.version@"
security:
  basic:
    enabled: false

spring:
  application:
    name: spring-cloud-dataflow-server
  cloud:
    config:
      uri: http://localhost:8888
  datasource:
    url: jdbc:h2:tcp://localhost:19092/mem:dataflow
    username: sa
    password:
    driverClassName: org.h2.Driver
  task:
    repo:
      initialize: true
aether:
  proxy:
    host: proxy.corp.fr
    port: XXXX`

Our ticktock logs : 

[stderr_0.txt](https://github.com/spring-cloud/spring-cloud-dataflow/files/146432/stderr_0.txt)
[stdout_0.txt](https://github.com/spring-cloud/spring-cloud-dataflow/files/146431/stdout_0.txt)

Has  someone  already experienced this problem or has any  suggestions on this issue ? 

Many thanks

ilayaperumalg:
- The module resolver and aether config properties are obtained from the deployer environment and updated for the out-of-process deployer's process builder environment
  - This make sure to pass any module resolver/aether properties such as proxy configs are passed along
  - Given the module launcher properties are set via dataflow already, it is sufficient to check for the module resolver and aether config properties here

This resolves #420


ngngsoftorg:
Ok, we will try downloading and redeploying the latest snapshot.  fingers crossed.

mbogoevici:
Wouldn't it make sense to support this via https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowServerProperties.java and related, since all the remote connectivity configuration (except proxies) is controlled by that. That would also mean that all the servers would support proxy settings automatically?

sabbyanandan:
@ngngsoftorg: As an FYI, Ilaya's fix is for the out-of-process `Local` deployer and it doesn't address the `CF` deployer yet. Similar but a more general solution will be implemented, so all the deployers inherit the functionality consistently.

ilayaperumalg:
Updated the PR that tries to use the `DataFlowServerProperties` for passing the aether proxy settings to the app launcher environment. This enables other deployer SPI implementations that use `DataFlowServerProperties` work without any changes to support proxy settings.

ilayaperumalg:
For instance, on the CF app deployer with the following properties

```
-jar spring-cloud-dataflow-server-cloudfoundry/target/spring-cloud-dataflow-server-cloudfoundry-1.0.0.BUILD-SNAPSHOT.jar --deployer.localRepository=test --deployer.aether.proxy.host=host1 --deployer.aether.proxy.auth.username=user1
```

I see the environment is set to use:

```
JBP_CONFIG_JAVA_MAIN
{arguments: '''--aether.proxy.host\=host1 --aether.proxy.auth.username\=user1 --args.0.spring.cloud.stream.bindings.input.destination\=a1.time
    --aether.proxy.protocol\=http --args.0.spring.cloud.stream.bindings.input.group\=default
    --args.0.dataflow.group-deployment-id\=a1-1457519014766 --args.0.spring.jmx.default-domain\=a1.log
    --modules\=org.springframework.cloud.stream.module:log-sink:jar:exec:1.0.0.BUILD-SNAPSHOT
    --args.0.spring.cloud.stream.bindings.input.durableSubscription\=true --localRepository\=test
    '''}
```

Once the `app-launcher` for the CF deployer is updated with this PR changes, then the app launcher will use the above proxy properties.

ilayaperumalg:
Verified the CF deployer (that uses the app-launcher with this PR changes) and the proxy configuration properties are applied appropriately when set via server startup command line args.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowServerProperties.java
```diff
@@ -32,6 +33,10 @@
 
 	static final String PREFIX = "deployer";
 
+	static final String AETHER_PROXY_PREFIX = "aether.proxy.";
```
ericbottard: as discussed in a call, maybe use `maven.proxy` (as aether can be seen as an impl detail).

Or maybe going even further, have those be very generic `http.proxy` settings (if e.g. targeting s3 in the future)

ngngsoftorg:
Understood.  We await the scdf fix.

PR #446, opened 2016-03-10
Move dataflow-server.yml to server-local

### LINKED ISSUE #445: DataFlow Server config updates
The `dataflow-server.yml ` as its name suggests the yml file should belong to `server-local` instead of  `deployer-local`. Also there is no need to enable `MavenProperties` in server core configuration `@EnableConfigurationProperties(MavenProperties.class)` in DataFlowServerConfiguration as this properties should belong to deployer configuration which already exists in respective deployer configuration.

ilayaperumalg:
- Move from `deployer-local`
 - as its name suggests the yml file should belong to server-local
 - Also remove `@EnableConfigurationProperties(MavenProperties.class)` in DataFlowServerConfiguration as this properties should belong to deployer configuration

This resolves #445

### DIFF
--- a/spring-cloud-dataflow-server-local/src/main/resources/dataflow-server.yml
+++ b/spring-cloud-dataflow-server-local/src/main/resources/dataflow-server.yml
@@ -15,7 +15,7 @@ security:
 
 spring:
   application:
-    name: spring-cloud-dataflow-server
+    name: spring-cloud-dataflow-server-local
   cloud:
     config:
       uri: http://localhost:8888
--- a/spring-cloud-dataflow-deployer-local/pom.xml
+++ b/spring-cloud-dataflow-deployer-local/pom.xml
@@ -35,7 +35,6 @@
 				<directory>src/main/resources</directory>
 				<filtering>true</filtering>
 				<includes>
-					<include>dataflow-server.yml</include>
 					<include>META-INF/spring.factories</include>
 				</includes>
 			</resource>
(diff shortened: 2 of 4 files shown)


spring-cloud-dataflow-server-local/src/main/resources/dataflow-server.yml
```diff
@@ -15,7 +15,7 @@ security:
 
 spring:
   application:
-    name: spring-cloud-dataflow-server
+    name: spring-cloud-dataflow-server-local
```
markfisher: I think this can remain as `spring-cloud-dataflow-server`... if we want it to be consistent regardless of the implementation, for things like config server. WDYT?

ilayaperumalg: Since this yml file is now in the implementation specific I thought having a specific name which is consistent with other SPI implementations. For instance CF server has [this] ( https://github.com/spring-cloud/spring-cloud-dataflow-admin-cloudfoundry/blob/master/spring-cloud-dataflow-server-cloudfoundry/src/main/resources/dataflow-server.yml#L28)

By config server consistency, do you mean the same config properties can be shared across the SPI implementations? If so, I think we can make that change (changing the spring.application.name to `spring-cloud-dataflow-server` across all the SPI implementations)

markfisher: That was my line of thinking (consistency meaning they all use the same name "spring-cloud-dataflow-server" which does describe what this is, in terms of its role). But perhaps the tradeoff favors indicating the implementation within the name. I'd like to hear @ericbottard's thoughts.

markfisher:
merged - we can revisit the `spring.application.name` later

PR #447, opened 2016-03-10
Add doc for maven/proxy configuration

### LINKED ISSUE #442: Document how to override remote maven repository
As a user, I'd like to override the default remote maven repository, so I can have the `dataflow-server` point to my custom maven repo. 

**Acceptance:**
* instructions how and where to override is included in the ref. guide
* all available options including `-DremoteRepositories` is listed

ilayaperumalg:
- Also update flag for in-process deployer

This resolves #442

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/getting-started.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/getting-started.adoc
@@ -29,14 +29,46 @@ You also need to have link:http://redis.io/[Redis] installed and running if you 
 wget http://repo.spring.io/milestone/org/springframework/cloud/spring-cloud-dataflow-server-local/1.0.0.BUILD-SNAPSHOT/spring-cloud-dataflow-server-local-1.0.0.BUILD-SNAPSHOT.jar
 wget http://repo.spring.io/milestone/org/springframework/cloud/spring-cloud-dataflow-shell/1.0.0.BUILD-SNAPSHOT/spring-cloud-dataflow-shell-1.0.0.BUILD-SNAPSHOT.jar
 ```
+
 [start=2]
+1. Running behind proxy/custom maven settings (remote repositories etc.,)
+
+If you run the Data Flow Server behind the proxy and want to override specific maven configuration properties, you need to specify those properties as command line arguments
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-docs/src/main/asciidoc/getting-started.adoc
```diff
@@ -29,14 +29,46 @@ You also need to have link:http://redis.io/[Redis] installed and running if you
...
+export MAVEN_PROXY_AUTH_PASSWORD=passwd
+```
+
+[start=3]
 1. launch the Data Flow Server:
 
 ```
 $ java -jar spring-cloud-dataflow-server-local-1.0.0.BUILD-SNAPSHOT.jar
 ```
+By default, this will start the local deployer that is `out-of-process`. If you want to use `in-process` deployer then
```
markfisher: I think we can leave out the "in-process" discussion here. As soon as we switch the local server to use the new `spring-cloud-deployer` SPI (hopefully today or tomorrow), it will only be out-of-process.

mbogoevici: We should probably remove the in-process thing altogether ...

ilayaperumalg: Any specific reason we want to remove `in-process` deployer altogether? Personally I like the `in-process` deployer (especially to get quick developer experience in terms of deploying/undeploying apps within the same JVM - like the single-node in XD). Other important benefit that helps very much is that the `debugging` things within the deployer is a lot easier.

markfisher:
we don't intend to support `in-process` going forward since the complexity (e.g. classpath issues) quickly outweighs the convenience... but we should consider options for debugging the out-of-process deployment

markfisher:
rebased and merged as https://github.com/spring-cloud/spring-cloud-dataflow/commit/df049ff425f31be7a12e61243b04bd6152b1b26c

along with a few changes here: https://github.com/spring-cloud/spring-cloud-dataflow/commit/3d6d4e18e71aaa95e525ce44df53e83c90f90e79

PR #454, opened 2016-03-14
Support aether connect/request timeout

ilayaperumalg:
- Add connect/request timeout as maven properties

The environment variables now need to be set as:

```
 export MAVEN_CONNECT_TIMEOUT=100000
 export MAVEN_REQUEST_TIMEOUT=150000
```

### DIFF
--- a/spring-cloud-dataflow-app-launcher/src/main/java/org/springframework/cloud/dataflow/app/resolver/AetherModuleResolver.java
+++ b/spring-cloud-dataflow-app-launcher/src/main/java/org/springframework/cloud/dataflow/app/resolver/AetherModuleResolver.java
@@ -84,9 +84,21 @@ public class AetherModuleResolver implements ModuleResolver {
 
 	private volatile boolean offline = false;
 
+	private MavenProperties mavenProperties;
+
 	private MavenProperties.Proxy proxyProperties;
 
 	private Authentication authentication;
+
+	/**
+	 * @see {@link org.eclipse.aether.ConfigurationProperties#CONNECT_TIMEOUT}
+	 */
+	private int connectTimeout;
+
+	/**
+	 * @see {@link org.eclipse.aether.ConfigurationProperties#REQUEST_TIMEOUT}
+	 */
+	private int requestTimeout;
 
 	/**
 	 * Create an instance specifying the locations of the local and remote repositories.
@@ -105,6 +117,7 @@ public class AetherModuleResolver implements ModuleResolver {
 				log.debug("Remote repositories: " + StringUtils.collectionToCommaDelimitedString(remoteRepositories.values()));
 			}
 		}
+		this.mavenProperties = mavenProperties;
 		if (mavenProperties != null) {
 			this.proxyProperties = mavenProperties.getProxy();
 		}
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-app-launcher/src/main/java/org/springframework/cloud/dataflow/app/resolver/AetherModuleResolver.java
```diff
@@ -193,6 +206,12 @@ private DefaultRepositorySystemSession newRepositorySystemSession(RepositorySyst
 		LocalRepository localRepo = new LocalRepository(localRepoPath);
 		session.setLocalRepositoryManager(system.newLocalRepositoryManager(session, localRepo));
 		session.setOffline(this.offline);
+		if (this.mavenProperties != null && this.mavenProperties.getConnectTimeout() > 0) {
+			session.setConfigProperty("aether.connector.connectTimeout", this.mavenProperties.getConnectTimeout());
```
markfisher: should use the constants in Aether's `ConfigurationProperties` instead of the String keys

spring-cloud-dataflow-app-launcher/src/main/java/org/springframework/cloud/dataflow/app/resolver/MavenProperties.java
```diff
@@ -198,6 +218,12 @@ public void setPassword(String password) {
 					this.getRemoteRepositories()));
 		}
 		properties.put(MAVEN_PREFIX + "offline", String.valueOf(this.isOffline()));
+		if (this.getConnectTimeout() > 0) {
```
markfisher: this means we would not pass thru a value of 0, which is valid... but maybe that's ok since that means NO timeout (although a user could pass any negative value to accomplish the same).... another option is to use `Integer` instead of `int` and check for `null` as the indicator to rely on the system's own default

markfisher:
I've created a corresponding PR in `spring-cloud-deployer`: https://github.com/spring-cloud/spring-cloud-deployer/pull/21

...but I would like to resolve the open question about `0` values here to ensure we do the same there

spring-cloud-dataflow-app-launcher/src/main/java/org/springframework/cloud/dataflow/app/resolver/MavenProperties.java
ilayaperumalg: yeah, I agree and will update the change.

ilayaperumalg:
Updated the PR.

spring-cloud-dataflow-app-launcher/src/main/java/org/springframework/cloud/dataflow/app/resolver/AetherModuleResolver.java
```diff
@@ -84,11 +85,23 @@
...
+	private MavenProperties mavenProperties;
+
 	private MavenProperties.Proxy proxyProperties;
 
 	private Authentication authentication;
 
 	/**
+	 * @see {@link org.eclipse.aether.ConfigurationProperties#CONNECT_TIMEOUT}
+	 */
+	private int connectTimeout;
```
markfisher: this is not used

spring-cloud-dataflow-app-launcher/src/main/java/org/springframework/cloud/dataflow/app/resolver/AetherModuleResolver.java
```diff
@@ -84,11 +85,23 @@
...
+
+	/**
+	 * @see {@link org.eclipse.aether.ConfigurationProperties#REQUEST_TIMEOUT}
+	 */
+	private int requestTimeout;
```
markfisher: this is not used

markfisher:
will remove unused properties while merging

markfisher:
rebased, polished, squashed and merged as https://github.com/spring-cloud/spring-cloud-dataflow/commit/c31a532f93deb1505455fcc872dfd695ecde7500

PR #455, opened 2016-03-15
Remove durability configuration

### LINKED ISSUE #411: Deployment of app using kafka binder fails
When deploying the sink app that uses `Kafka` binder, the following error occurs:

```
MessageChannelBinder does not support consumer property: durableSubscription
```

Since the `durableSubscription` is always set as the consumer property for the sink app and `Kafka` binder doesn't support this property, this error occurs.

mbogoevici:
The default set by Spring Cloud Stream is durable. 

Fixes #411


markfisher:
these same changes have been addressed by https://github.com/spring-cloud/spring-cloud-dataflow/pull/463

Issue #449, opened 2016-03-11
Implement spring-cloud-deployer for Local SPI


sabbyanandan:
As a developer, I'd like to refactor the current `OutOfProcessModuleDeployer` deployer implementation, so I can implement the deployer with `spring-cloud-deployer` contracts.

**Acceptance:**
- Existing `stream` and `task` unit tests run successfully
- The project successfully builds 
- Docs updated

PR #459, opened 2016-03-17
dataflow-server-local uses the new deployer SPI

markfisher:
- deleted the spring-cloud-dataflow-deployer-local project
- spring-cloud-starter-dataflow-server-local now depends on spring-cloud-deployer-local
- other dataflow-server implementations can now switch SPIs at their convenience
- all Controllers that relied on ModuleDeployer are now prefixed with "Deprecated"
- Controller beans are explicitly created (no component-scanning) in auto-configuration
- the above seems necessary due to the lifecycle relationship between scanning, auto-config, conditional checks, and the HATEOAS link registry
- even if there is a way to control the ordering, explicit bean configuration seems better for the Controllers
- 2 auto-config classes exist: one for the new Controllers (using the new SPI), another for the deprecated Controllers
- the auto-config classes use ConditionalOnBean, for AppDeployer/TaskLauncher (new) or ModuleDeployer (deprecated)
- once we have migrated all of the dataflow-server implementations, we can delete the deprecated controllers (and remove the ConditionalOnBean annotation from the remaining auto-config class)

Resolves spring-cloud/spring-cloud-dataflow#449


spring-cloud-dataflow-server-core/pom.xml
```diff
@@ -34,6 +34,16 @@
 			<artifactId>spring-cloud-dataflow-deployer-spi</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-deployer-spi</artifactId>
+			<version>1.0.0.BUILD-SNAPSHOT</version>
```
ericbottard: I think we'll want to externalize those versions somehow

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/CounterController.java
```diff
@@ -53,7 +52,6 @@
 
 	public static final String COUNTER_PREFIX = "counter.";
 
-	@Autowired
```
ericbottard: How is the repository injected now?

markfisher: all controllers are configured explicitly in the auto-configuration classes

spring-cloud-dataflow-server-core/pom.xml
markfisher: it should probably be added to the `spring-cloud-dataflow-dependencies` POM

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/CounterController.java
ericbottard: See `DataFlowControllerAutoConfiguration` line 109

ericbottard: Same for the deprecated one, unless I'm missing something obvious

markfisher: Thanks! - I see now... missed this one as a constructor arg. This is why field injection is such a bad idea, because otherwise this variable should be marked `final`. Will update.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowControllerAutoConfiguration.java
```diff
@@ -0,0 +1,156 @@
...
+ * Configuration for the Data Flow Server Controllers.
+ *
+ * @author Mark Fisher
+ */
+@Configuration
+@ConditionalOnBean({ AppDeployer.class, TaskLauncher.class })
+public class DataFlowControllerAutoConfiguration {
+
+	@Bean
+	public RootController rootController(EntityLinks entityLinks) {
```
jvalkeal: On yarn I'm building additional boot apps manually to get access to fresh beans with different configs. I got into trouble with `EntityLinks` because there we none to autowire. So it might be a good idea to add more conditions so that this autoconfig would not kick in if `EntityLinks` are not available. I got away from this problem by excluding this class. Trouble wasn't with a _main_ yarn dataflow server but with those extra `SpringApplication` instances I'm using which does same autoconfig magic behind a scenes.

It probably failed because I already excluded some other features like not enabling web meaning `hateoas` autoconfig probably didn't create anything.

jvalkeal: Might be simply by adding `EntityLinks` to above `@ConditionalOnBean`.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/CounterController.java
markfisher: I just pushed a change to this PR to address the missing `MetricRepository`. This reveals that we have no tests for this. I've asked Sabby to create a story.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowControllerAutoConfiguration.java
markfisher: `DataFlowServerConfiguration` itself has `@EnableHypermediaSupport(type = HAL)`, but are you bypassing that main configuration class as well?

It would seem appropriate to enable HATEOAS in the controller config, but I believe there are lifecycle issues if that is not available until the auto-config stage.

jvalkeal: For now this is not a big issue for me, just wanted to write it down here if it comes back to haunt us in a future.

markfisher: this is a future, no callbacks... we will need to remember ;)

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
```diff
@@ -191,13 +200,17 @@ private void deployStream(StreamDefinition stream, Map<String, String> streamDep
 
 			currentModule = postProcessLibraryProperties(currentModule);
 
-			this.deployer.deploy(new ModuleDeploymentRequest(currentModule, coordinates, moduleDeploymentProperties));
+			AppDefinition definition = new AppDefinition(currentModule.getLabel(), currentModule.getParameters());
+			MavenResource resource = MavenResource.parse(coordinates.toString(), mavenProperties);
+			AppDeploymentRequest request = new AppDeploymentRequest(definition, resource, moduleDeploymentProperties);
+			this.deployer.deploy(request);
```
jvalkeal: I know this is still broken and can be done later but just a note, id from deploy needs to be stored somewhere so that it can be use with undeploy()/status().

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowControllerAutoConfiguration.java
jvalkeal: I'll play with this issue when we get this PR merged.

jvalkeal:
Other than that, I played with(using Mark's branch) this locally yesterday and it was ok with yarn.

jvalkeal:
squashed/merged https://github.com/spring-cloud/spring-cloud-dataflow/commit/e0195b4ffbf8bb2fca2281a46c2e85bc57f2c952

PR #463, opened 2016-03-21
Aligns Spring Cloud Data Flow with Spring Cloud Stream 1.0.0.RC1

mbogoevici:


markfisher:
merged along with this fix (and test) for wildcard producer/consumer properties: https://github.com/spring-cloud/spring-cloud-dataflow/commit/55a0c484cad33efcadcd658e0ac955ada4f385e3

the above was actually addressing a pre-existing bug which was revealed during this review

PR #477, opened 2016-03-30
use UriRegistry in deployment controllers

markfisher:
also populating the registry on startup by default
(with "locations" and "enabled" as boot properties of the populator)

The DelegatingResourceLoader includes:
- MavenResourceLoader for 'maven:'
- DefaultResourceLoader for 'file:' and 'http:'

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/registry/DataFlowUriRegistryPopulator.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/registry/DataFlowUriRegistryPopulator.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.dataflow.server.registry;
+
+import org.springframework.beans.factory.InitializingBean;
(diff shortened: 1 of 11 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/registry/DataFlowUriRegistryPopulator.java
```diff
@@ -0,0 +1,56 @@
...
+import org.springframework.cloud.deployer.resource.registry.UriRegistry;
+import org.springframework.cloud.deployer.resource.registry.UriRegistryPopulator;
+import org.springframework.util.Assert;
+
+/**
+ * A {@link UriRegistryPopulator} implementation that provides stream and task app URIs for a Data Flow Server.
+ *
+ * @author Mark Fisher
+ */
+public class DataFlowUriRegistryPopulator extends UriRegistryPopulator implements InitializingBean {
```
pperalta: Instead of extending `UriRegistryPopulator` how about passing an instance in as a bean in the constructor?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/registry/DataFlowUriRegistryPopulatorProperties.java
```diff
@@ -0,0 +1,56 @@
...
+
+	/**
+	 * Indicates whether to populate the registry on startup. Default is {@literal true}.
+	 */
+	private boolean enabled = true;
+
+	/**
+	 * Resource locations for one or more properties files where the keys are stream or task app
+	 * names (e.g. "source.foo" or "task.bar") and the values are Resource URIs. Default location
+	 * is {@literal "classpath:META-INF/applications.properties"}.
```
pperalta: If this is meant to be fixed width, use `{@code ...}` instead of `{@literal }`.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/registry/DataFlowUriRegistryPopulator.java
markfisher: Will do, and will rename this `DataFlowUriRegistryInitializer`.

pperalta:
Merged as b323cefad8732d57dda8fa7f94fbb37f282a6716

PR #481, opened 2016-03-31
Add @ConditionalOnMissingBean for DelegatingResourceLoader

trisberg:


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowControllerAutoConfiguration.java
```diff
@@ -124,6 +125,7 @@ public MavenResourceLoader MavenResourceLoader(MavenProperties properties) {
 	}
 
 	@Bean
+	@ConditionalOnMissingBean
```
markfisher: I believe this should be: `@ConditionalOnMissingBean(DelegatingResourceLoader.class)`

markfisher: ...although strictly speaking, the deployment controller should not care if it's a `DelegatingResourceLoader`, and in a case like k8s, it may only need `DockerResourceLoader` anyways. But the only other way to have conditionality (since it won't be the only `ResourceLoader

trisberg: Yes, it should be: `@ConditionalOnMissingBean(DelegatingResourceLoader.class)`, I must have used the same name since it worked when I tested the k8s deployer

markfisher:
LGTM; merging

markfisher:
rebased, squashed, and merged as https://github.com/spring-cloud/spring-cloud-dataflow/commit/4729d534eaee5a197764c35b6d52549b1d8fc0f3

PR #487, opened 2016-04-04
remove ArtifactRegistry in favor of AppRegistry

markfisher:
- AppRegistry wraps UriRegistry and returns AppRegistration
- AppRegistration replaces ArtifactRegistration
- AppRegistration defines a getResource() method
- ModuleResolver is therefore no longer needed for resolving Resources
- ModuleResolver can be removed with ModuleLauncher when it's no longer used

(after this PR is merged, the "spring-cloud-dataflow-artifact-registry" sub-project will need to be renamed or else, the `AppRegistry` and `AppRegistration` classes could move to a lower-level, such as "spring-cloud-dataflow-core"... at that point the package name would also change, but for now it is still "artifact")


jvalkeal:
Looks good, did various things like zapping redis, local maven files, registered custom modules, played with streams, etc. I'll go and merge this.

Issue #489, opened 2016-04-04
Use persisted IDs to undeploy() and query for status()


sabbyanandan:
As a developer, I'd like to use the persisted IDs to undeploy() and query for status(), so I can pin and operate on the pipeline identified by its ID.

This depends on spring-cloud/spring-cloud-dataflow#446

**Acceptance:**
- `StreamDeploymentController` and `TaskDeploymentController` relies on the persisted ID for individual pipeline level interaction
- Unit tests included

Issue #466, opened 2016-03-22
Add deployment mapping repository


jvalkeal:
As a developer, I'd like to add support to persist stream and task deployment IDs at the time of stream and task deployment, so I can query the status and as well as undeploy or destroy pipelines using the unique IDs.

**Acceptance:**
- deploy() and launch() operations in `AppDeployer` and `TaskLauncher` APIs persists the deployment IDs
- construct to persist the IDs live in `spring-cloud-dataflow-server-core` project as this is specific to SCDF 
- Unit tests included

**Original Notes:**
AppDeployer SPI contract, use String returned from deploy() for undeploy() and status():
```
String deploy(AppDeploymentRequest request);
void undeploy(String id);
AppStatus status(String id);
```

In `StreamDeploymentController.deployStream(StreamDefinition)` we should store returned id's so that we can later map `StreamDefinition` back to those id's as `StreamDefinition` is only info available for `StreamDeploymentController.undeployStream(StreamDefinition)`. With streams we possibly get away from a complex mapping due to fact that we can have only one stream at any given time.

TaskLauncher SPI contract, use String returned from launch() for cancel() and status():
```
String launch(AppDeploymentRequest request);
void cancel(String id);
TaskStatus status(String id);
```

We're in trouble with cancelling tasks or getting status of a running task app because we can run multiple task instances at any given time. This means we have no way to map a launched task to an actual running task app.

PR #492, opened 2016-04-05
Add AppDeploymentRepository for deploy id tracking

jvalkeal:
NOTE: thought this works(on going work for tasks/redis), maybe for just a review for now.
- AppDeploymentRepository may be a bad name thinking that this is meant for streams and we'd need to add similar features for tasks too.
- Should do similar work for tasks.
- Should do Redis impl.
- connect #466 
- connect #489 

From commit log:
- New AppDeploymentRepository which is doing reverse mapping
  from AppDeploymentKey into a identifier returned from a deployer.
- Only InMemoryAppDeploymentRepository for now.
- AppDeploymentKey is a composition of StreamDefinition and
  ModuleDefinition which are used/needed when status() and
  undeploy() are used.
- Implemented hash/equal for StreamDefinition and ModuleDefinition
  for AppDeploymentKey to work as a key in maps.
- StreamDefinitionController/StreamDeploymentController now uses
  this repository to backmap id from current in-flight
  StreamDefinition and ModuleDefinition in use.


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/AppDeploymentKey.java
```diff
@@ -0,0 +1,94 @@
...
+import org.springframework.cloud.dataflow.core.ModuleDefinition;
+import org.springframework.cloud.dataflow.core.StreamDefinition;
+import org.springframework.util.Assert;
+
+/**
+ * Key used in an {@link AppDeploymentRepository}.
+ *
+ * @author Janne Valkealahti
+ */
+public class AppDeploymentKey {
```
markfisher: It would seem the `<streamName>.<moduleLabel>` as a *String* would be sufficient for the "key".

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/KeyRepository.java
```diff
@@ -0,0 +1,53 @@
...
+ * Interface for generic {@link Repository} which does a backmapping from key into
+ * a identifier.
+ *
+ * @author Janne Valkealahti
+ *
+ * @param <T> the type of entity
+ * @param <ID> the type of id
+ */
+@NoRepositoryBean
+public interface KeyRepository<T, ID extends Serializable> extends Repository<T, ID> {
```
markfisher: this feels like premature generalization - do we have another use-case in mind already?

markfisher:
I did some minor polishing in this commit if you want to merge it into this PR:
https://github.com/markfisher/spring-cloud-dataflow/commit/485254195f1efded3b90f1486b4de13e754bfe2c

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/AppDeploymentKey.java
jvalkeal: I initially thought to just create `Repository<String, String>` but hated idea that user would need to compile `key` manually out from `StreamDefinition` and `ModuleDefinition`. But I had similar thought when I quickly looked what a possible redis impl(to store as little as needed) might look like. As we don't need to serialize `AppDeploymentKey` back from repository, we could possibly just use its hash mapped to identifier.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/KeyRepository.java
jvalkeal: Yeah I added this middle guy here as idea was to also add repo for storing composite of `TaskDefinition` and its `ModuleDefinition`. Related to my above comment about redis, if hash from an `AppDeploymentKey` would be enough, then potentially we might be able to go with `Repository<Object,String>` and we could remove this guy. But then we lose all options to use this repo interface for anything else than id backmapping. What do you think?

markfisher: I might interpret "lose all options to use this repo interface for anything else" as *well-defined single responsibility* ;)

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/AppDeploymentKey.java
markfisher: `ModuleDefinition` has the stream (group) name in it, so maybe a compromise would be something like the old `String id = ModuleDeploymentId.fromModuleDefinition(module)` (possibly just a private method of the controller itself) and something similar for `TaskDefinition` instances.

jvalkeal:
@markfisher Right, think this PR is now ready.

markfisher:
I still think a simple String as the key in the repo would be sufficient, and unique keys can be generated directly from the `ModuleDefinition`.

@jvalkeal: Have a look at this commit and let me know what you think: https://github.com/markfisher/spring-cloud-dataflow/commit/bf4e94cf45e1dd69c927d0923e57ec3bc29e6d1e

jvalkeal:
Right I somehow thought StreamDefinition name is not necessarily guaranteed to be same as ModuleDefinition group. It's definitely simple this way and worked. I'll merge this.

markfisher:
rebased, squashed, and merged as https://github.com/spring-cloud/spring-cloud-dataflow/commit/d2bf897f6fda21e7c562dbe9a0f41d117aed8076

PR #495, opened 2016-04-06
renamed artifact-registry subproject

markfisher:
- now simply spring-cloud-dataflow-registry
- moved the URI registry related classes there as well


ghillert:
LTGM - Merged.

PR #498, opened 2016-04-07
add docs for module register command

### LINKED ISSUE #485: Document custom app registration
As a user, I'd like to have docs to guide with custom app contribution, so I can follow the right syntaxes and the expected format for registering new applications.

**Acceptance:**
- DSL syntax and string literal format is available
- Includes steps to verify (_in redis_) whether the contribution is successfully listed in the registry
- Example included

markfisher:
resolves #485

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
@@ -137,6 +137,24 @@ dataflow:> stream create --definition "http --server.port=8000 | log" --name myh
 ```
 The shell provides tab completion for module parameters and also the shell command `module info` provides some additional documentation.  For more comprehensive documentation on module parameters, please see the xref:modules#modules[Modules] chapter.
 
+=== Register a Stream App
+
+Register a Stream App with the App Registry using the Spring Cloud Data Flow Shell
+`module register` command. You must provide a unique name and a URI that can be
+resolved to the app artifact. For the type, specify "source", "processor", or "sink".
+Here are a few examples:
+
+```
+dataflow:>module register --name mysource --type source --uri maven://com.example:mysource:0.0.1-SNAPSHOT
+
+dataflow:>module register --name myprocessor --type processor --uri file:///Users/example/myprocessor-1.2.3.jar
+
+dataflow:>module register --name mysink --type sink --uri http://example.com/mysink-2.0.1.jar
+```
+
+If a stream app is already registered with the provided name and type, it will not be overridden by default.
+If you would like to override the pre-existing stream app, then include the `--force` option.
+
 == Advanced Features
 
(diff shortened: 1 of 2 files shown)


ericbottard:
Shall we explain that actual download/resolution may happen on the machine running the app (at least for now), and so that some schemes may not work as expected?

Also, what about the `exec` qualifier which we often use? Shall we include an explanation of the parts of the maven uri?

markfisher:
+1 on the suggestions, but what do you mean by "some schemes may not work as expected?"... if you mean that Docker apps won't work with auto-completion, I think that's a bigger (separate) issue.

ericbottard:
I meant fille:// is resolved on the 'module' machine filesystem, not the
dataflow server for example.

markfisher:
@ericbottard you said:

> fille:// is resolved on the 'module' machine filesystem, not the dataflow server for example.

What implementation of the dataflow-server are you referring to in that statement?

ericbottard:
I think this does not apply anymore actually, as the CF deployer resolves on the dataflow server side, and others are being rewritten to use specific docker images, AFAIU.

Not sure if we should add a mention still though, as there is nothing in the contract of `AppDeployer` that states how (and where) it should handle the resource

markfisher:
I think the details of each server will need to be described in their docs. For the common doc here, probably a vague description will suffice. Something like:

> In some cases the `Resource` is resolved on the server side, whereas in others the URI will be passed to a runtime container instance where it is resolved. Consult the specific documentation of each Data Flow Server for more detail.

ericbottard:
Seems fine to me. I can incorporate on merge if you wish, as it seems this is the only outstanding issue with this PR.

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -137,6 +137,30 @@ dataflow:> stream create --definition "http --server.port=8000 | log" --name myh
...
+
+dataflow:>module register --name myprocessor --type processor --uri file:///Users/example/myprocessor-1.2.3.jar
+
+dataflow:>module register --name mysink --type sink --uri http://example.com/mysink-2.0.1.jar
+```
+
+When providing a URI with the `maven` scheme, the format should conform to the following:
+
+```
+maven://<groupId>:<artifactId>[:<extension>[:<classifier>]]:>version>
```
ericbottard: typo on opening bracket for version, will fix

markfisher: good catch - thanks!

markfisher:
Sure if you can add that blurb to the doc for both `streams` and `tasks` in the corresponding `Register...` sections, that would be great. As you noticed, I did already push the commit with the Maven URI description.

Thanks!

ericbottard:
Merged with small addition as dc1701f

Issue #369, opened 2016-02-05
Add debug capabilities to LocalAppDeployer


ericbottard:

aclement:
A pair of PRs now enable debugging with local deployer. Works for me but there are open questions. I raised the PRs to drive the discussion:
PR on deployer: https://github.com/spring-cloud/spring-cloud-deployer/pull/35
PR on dataflow: https://github.com/spring-cloud/spring-cloud-dataflow/pull/502

Info I included in those PRs:

Created two pull requests (one on dataflow, the other on local deployer), these are for github issue #369 "Add debug capabilities to LocalAppDeployer".

With these two in place you can turn on debug by specifying "module.foo.debug-port=8800" as a deployment property.

The first pull request (on the deployer) recognizes a debug-port property and uses it to add a debug agent to the configuration for the Process it launches.

The second pull request (on dataflow) recognizes the deployment pattern module..debug-port=XXX and maps that to the debug-port property in the deployer.

This works but are the properties in the right place? Is it ok to have this as a deployment config property? (It is quite nice that turning debug on/off does not require meddling with the stream definition). What do we do for other deployers? Do we need a simpler 'debug=true/false' option alongside this for any that don't need a port, or do we need a deployer specific namespace subsection for this?

jvmlet:
Would be great to  be able to debug the **task** launched by SCDF as well, see `listen` debugger mode  approach described in [SO](https://stackoverflow.com/questions/47102545/how-to-remote-debug-a-spring-cloud-data-flow-task/47637252#47637252)

sabbyanandan:
This was addressed in spring-cloud/spring-cloud-deployer-local#85; docs [here](https://docs.spring.io/spring-cloud-dataflow/docs/1.3.0.RC1/reference/htmlsingle/#_remote_debugging).

PR #502, opened 2016-04-07
[For discussion] basic debug support for deployers

aclement:
Created two pull requests (one here on dataflow, the other on local deployer), these are for github issue https://github.com/spring-cloud/spring-cloud-dataflow/issues/369 "Add debug capabilities to LocalAppDeployer".

With these two in place you can turn on debug by specifying "module.foo.debug-port=8800" as a deployment property.

The first pull request (on the deployer) recognizes a debug-port property and uses it to add a debug agent to the configuration for the Process it launches. 

The second pull request (on dataflow) recognizes the deployment pattern module.<xxx>.debug-port=XXX and maps that to the debug-port property in the deployer.

This works but are the properties in the right place? Is it ok to have this as a deployment config property? (It is quite nice that turning debug on/off does not require meddling with the stream definition). What do we do for other deployers? Do we need a simpler 'debug=true/false' option alongside this for any that don't need a port, or do we need a deployer specific namespace subsection for this?

This resolves spring-cloud/spring-cloud-dataflow#369


LyashenkoGS:
Any updates on this PR status?

olegz:
@LyashenkoGS I am actually working on it and have a local prototype with possibly a simpler solution by adding fist class support for debugging to Local Deployer. Give me till next week as I am tied up now

olegz:
@LyashenkoGS One thing I forgot to mention which is already pushed. With local deployer you can now redirect logs to local server console which may not satisfy you fully with regard to debugging, but it will simplify the process quite a bit since you no longer have to go and tail multiple logs.
Basically during the deploy all you need to do is specify _inheritLogging=true_. For example
```stream deploy --name myStream --properties "deployer.*.local.inheritLogging=true"```
The above will redirect logs from all apps in the stream to local server console.
You can also replace '*' pattern with the name of the specific app you  want to track

pjannapu:
Hi, Just wondering if anyone's planning to get this rolled out soon. Currently, I have been facing an issue on how to (remote) debug my apps deployed on SCDF and came across this PR. Appreciate someone taking care of this soon. Thanks!

PR #513, opened 2016-04-12
Use persisted id for task status

### LINKED ISSUE #512: Use persisted ID to query task status
For some reason `TaskDefinitionController` was not modified to pass id to `TaskLauncher.status()` query. Original work done in #466 and #489.

jvalkeal:
- Add missing use of DeploymentIdRepository in a
  TaskDefinitionController when querying task
  status from a TaskLauncher.
- Fixes #512


markfisher:
LGTM; merging

PR #515, opened 2016-04-12
Update README.adoc

ractive:
Point "Local" deployer link to correct repo.


markfisher:
The existing link is intentionally pointing to the local *Data Flow Server* which itself has a dependency on the `spring-cloud-deployer-local` lib.

ractive:
ok

PR #518, opened 2016-04-14
Fix NPE for undeploy

### LINKED ISSUE #511: Destroying a partially deployed stream fails with NPE
As a user, I deployed a stream and the deployment failed, so when I tried to destroy the stream, I get a NPE. 

The stream deployment failed because of: spring-cloud/spring-cloud-deployer#38.

**list:**
```
dataflow:>stream list
╔═══════════╤═════════════════╤══════════╗
║Stream Name│Stream Definition│  Status  ║
╠═══════════╪═════════════════╪══════════╣
║test      │time | test-log │undeployed║
╚═══════════╧═════════════════╧══════════╝
```

**error:**
```
dataflow:>stream all destroy --force
Command failed org.springframework.cloud.dataflow.rest.client.DataFlowClientException: NullPointerException

NullPointerException

org.springframework.cloud.dataflow.rest.client.DataFlowClientException: NullPointerException

    at org.springframework.cloud.dataflow.rest.client.VndErrorResponseErrorHandler.handleError(VndErrorResponseErrorHandler.java:52)
    at org.springframework.web.client.RestTemplate.handleResponse(RestTemplate.java:641)
    at org.springframework.web.client.RestTemplate.doExecute(RestTemplate.java:597)
    at org.springframework.web.client.RestTemplate.execute(RestTemplate.java:557)
    at org.springframework.web.client.RestTemplate.delete(RestTemplate.java:429)
    at org.springframework.cloud.dataflow.rest.client.StreamTemplate.destroyAll(StreamTemplate.java:112)
    at org.springframework.cloud.dataflow.shell.command.StreamCommands.destroyAllStreams(StreamCommands.java:165)
    at java.lang.reflect.Method.invoke(Method.java:483)
    at org.springframework.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:216)
    at org.springframework.shell.core.SimpleExecutionStrategy.invoke(SimpleExecutionStrategy.java:64)
    at org.springframework.shell.core.SimpleExecutionStrategy.execute(SimpleExecutionStrategy.java:57)
    at org.springframework.shell.core.AbstractShell.executeCommand(AbstractShell.java:130)
    at org.springframework.shell.core.JLineShell.promptLoop(JLineShell.java:533)
    at org.springframework.shell.core.JLineShell.run(JLineShell.java:179)
    at java.lang.Thread.run(Thread.java:745)
```

jvalkeal:
- Now checking `ID` from `deploymentIdRepository` is null before
  trying to use it with deployer undeploy. Effectively
  assuming that module is not deployed.
- Added missing mock tests for trying to undeploy
  a stream which is not deployed.
- Fixes #511

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
@@ -368,10 +368,13 @@ public class StreamDeploymentController {
 		for (ModuleDefinition module : stream.getModuleDefinitions()) {
 			String key = DeploymentKey.forApp(module);
 			String id = this.deploymentIdRepository.findOne(key);
-			AppStatus status = this.deployer.status(id);
-			if (!EnumSet.of(DeploymentState.unknown, DeploymentState.undeployed)
-					.contains(status.getState())) {
-				this.deployer.undeploy(id);
+			// if id is null, assume nothing is deployed
+			if (id != null) {
+				AppStatus status = this.deployer.status(id);
+				if (!EnumSet.of(DeploymentState.unknown, DeploymentState.undeployed)
+						.contains(status.getState())) {
+					this.deployer.undeploy(id);
+				}
 			}
 		}
 	}
--- a/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/StreamControllerTests.java
+++ b/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/StreamControllerTests.java
@@ -421,6 +421,27 @@ public class StreamControllerTests {
 	}
 
 	@Test
+	public void testUndeployNonDeployedStream() throws Exception {
+		repository.save(new StreamDefinition("myStream", "time | log"));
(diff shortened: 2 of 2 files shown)


ericbottard:
I think this also calls for a review of what happens at `StreamDeploymentController#deployStream()` site when a concrete deployer throws an exception (well, should it, even?). Shouldn't we also continue deploying other modules? WDYT @markfisher ?

markfisher:
@ericbottard can you create a separate issue for reviewing the deploy side? (I'd like to go ahead and merge this PR without losing track of that). It's not necessarily consistent across implementations of the deployer SPI... For example,  an impl may be async, such that a deployment failure of a single app within a stream would never throw an `Exception`. Given that possibility, it doesn't seem right for a synchronous deployment to discontinue the deployment of other apps. Perhaps that means in `StreamDeploymentController` we should wrap the `deployer.deploy()` call in a `try/catch` and simply log at WARN level if a particular impl does perform the deploy synchronously and happens to throw an Exception.

ericbottard:
Done. Created #524. I agree the controller should continue deployment, as this seems to be the best way to somewhat unify the sync and async cases.

markfisher:
rebased and merged as https://github.com/markfisher/spring-cloud-dataflow/commit/3fef3fea82b76a5048448d1f7b41e65154052e8d

Issue #460, opened 2016-03-18
Add test coverage for CounterController


sabbyanandan:
As a developer, I'd like to add test coverage for `CounterController`, so I can incrementally evaluate the contracts responsible for `counter` sink interaction.

PR #527, opened 2016-04-19
Fix CounterController Metric value type & add tests

### LINKED ISSUE #526: CounterController should use Metric value type Long instead of Double
Currently, the CounterController uses `Metric` value as `Double` which needs to be `Long`.
and obviously, the `CounterService` persists metric value as `Long`.

ilayaperumalg:
- The metric value type used in the CounterController is set to use `Long` instead of `Double`
 - Add CounterControllerTests

Resolves spring-cloud/spring-cloud-dataflow#526 
Resolves spring-cloud/spring-cloud-dataflow#460

### DIFF
--- a/spring-cloud-dataflow-registry/pom.xml
+++ b/spring-cloud-dataflow-registry/pom.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
 	<modelVersion>4.0.0</modelVersion>
+	<name>spring-cloud-dataflow-registry</name>
 	<artifactId>spring-cloud-dataflow-registry</artifactId>
 	<packaging>jar</packaging>
 	<parent>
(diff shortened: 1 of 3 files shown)


markpollack:
The RedisMetricRepository in boot actually is returning a Metric<Double>, see https://github.com/spring-projects/spring-boot/blob/master/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/repository/redis/RedisMetricRepository.java#L182

ilayaperumalg:
It is weird that the `DefaultCounterService` uses `Double` and will have problem when using this as a `CounterService`. Since data flow explicitly uses `RedisCounterRepository`, I have updated this PR to use `Double` as the metric value type and updated the tests to use `RedisCounterRepository`.

markpollack:
Getting a test failure on testList

```
java.lang.AssertionError: JSON path "$.page.totalElements"
Expected: is <2>
     but: was <3>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.springframework.test.util.JsonPathExpectationsHelper.assertValue(JsonPathExpectationsHelper.java:103)
	at org.springframework.test.web.servlet.result.JsonPathResultMatchers$1.match(JsonPathResultMatchers.java:65)
	at org.springframework.test.web.servlet.MockMvc$1.andExpect(MockMvc.java:171)
	at org.springframework.cloud.dataflow.server.controller.CounterControllerTests.testList(CounterControllerTests.java:94)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
```

markpollack:
I cleaned redis and it worked, not sure the root cause as I didn't look at my keys/contents before flushing....

PR #530, opened 2016-04-19
Move application.properties out of the server core

### LINKED ISSUE #504: Move application.properties to spring-cloud-dataflow-server-local
As a developer, I'd like to move [`application.properties`](https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-server-core/src/main/resources/META-INF/applications.properties) to `spring-cloud-dataflow-server-local` project, so each deployer implementation can provide its own set of apps (_with URIs_) to bootstrap.

**Acceptance:**
Existing unit-tests run successfully

markfisher:
- The Local Data Flow Server provides its own application.properties file.
- Test classes that rely on the resolution of apps provide their own properties files also.

Fixes #504


markfisher:
fixing something - will resubmit

PR #533, opened 2016-04-20
Add Runtime command tests

### LINKED ISSUE #458: Add test coverage for RuntimeCommands
As a developer, I'd like to add test coverage for `RuntimeCommands`, so I can incrementally validate runtime state of the deployed applications through local builds and CI plans.

ilayaperumalg:
This resolves #458

### DIFF
--- a/spring-cloud-dataflow-shell/src/main/java/org/springframework/cloud/dataflow/shell/command/RuntimeCommands.java
+++ b/spring-cloud-dataflow-shell/src/main/java/org/springframework/cloud/dataflow/shell/command/RuntimeCommands.java
@@ -57,9 +57,13 @@ import org.springframework.stereotype.Component;
 public class RuntimeCommands implements CommandMarker {
 
 	private static final String LIST_MODULES = "runtime modules";
+	
+	private DataFlowShell dataFlowShell;
 
 	@Autowired
-	private DataFlowShell dataFlowShell;
+	public void setDataFlowShell(DataFlowShell dataFlowShell) {
+		this.dataFlowShell = dataFlowShell;
+	}
 
 	@CliAvailabilityIndicator({LIST_MODULES})
 	public boolean available() {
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-shell/src/main/java/org/springframework/cloud/dataflow/shell/command/RuntimeCommands.java
```diff
@@ -57,9 +57,13 @@
 public class RuntimeCommands implements CommandMarker {
 
 	private static final String LIST_MODULES = "runtime modules";
+	
+	private DataFlowShell dataFlowShell;
```
markfisher: other than the setter, this appears to be unused

spring-cloud-dataflow-shell/src/main/java/org/springframework/cloud/dataflow/shell/command/RuntimeCommands.java
```diff
@@ -57,9 +57,13 @@
...
 
 	@Autowired
-	private DataFlowShell dataFlowShell;
+	public void setDataFlowShell(DataFlowShell dataFlowShell) {
```
markfisher: a constructor might be better... and could we even avoid autowiring?

ilayaperumalg: I believe this is needed only for testing purpose where we use mocks and the `DataFlowShell` isn't available in the context already. I think we would need to `autowire` the `DataFlowShell` anyway as all the commands are scanned as components, is it not the case?

markfisher: ultimately I think we should avoid scanning, but that's part of a bigger effort

I'm just never a fan of seeing a change that's made purely for the tests

markfisher:
rebased and merged as https://github.com/spring-cloud/spring-cloud-dataflow/commit/b25146d288cbdc51b6ec2f72ebdd6a3158ce9a55

along with: https://github.com/spring-cloud/spring-cloud-dataflow/commit/982ca69930fe6e52eeb9dd54b3375c83ea7a69ef

PR #534, opened 2016-04-20
StreamDeploymentController to process all the app deployment requests

### LINKED ISSUE #524: StreamDeploymentController#deploy should deploy other apps on failure
See discussion at https://github.com/spring-cloud/spring-cloud-dataflow/pull/518#issuecomment-210475534

Currently, when deploying a stream, if an app fails to deploy and a concrete deployer throws an exception, the rest of the stream is not deployed. The for loop should wrap each call in a try/catch and act accordingly.

ilayaperumalg:
- Irrespective of the exception thrown during the app deployments when using synchronous deployer implementation, the stream deployment controller proceeds to deploy all the apps in request and log the warning message in case of exception thrown on a deployment
 - Add test

This resolves #524

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
@@ -22,6 +22,9 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
 import org.springframework.cloud.dataflow.core.ArtifactType;
 import org.springframework.cloud.dataflow.core.BindingPropertyKeys;
 import org.springframework.cloud.dataflow.core.ModuleDefinition;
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
```diff
@@ -197,8 +202,15 @@ private void deployStream(StreamDefinition stream, Map<String, String> streamDep
...
-			String id = this.deployer.deploy(request);
-			this.deploymentIdRepository.save(DeploymentKey.forApp(currentModule), id);
+			try {
+				String id = this.deployer.deploy(request);
+				this.deploymentIdRepository.save(DeploymentKey.forApp(currentModule), id);
+			}
+			// If the deployer implementation handles the deployment request synchronously, log warning message if
+			// any exception is thrown out of the deployment and proceed to the next deployment.
+			catch (Exception e) {
+				loggger.warn(String.format("Exception when deploying the app %s:%s", currentModule, e.getMessage()));
```
markfisher: should add a space after the colon

ilayaperumalg: Fixed.

markfisher:
rebased, squashed, and merged as https://github.com/spring-cloud/spring-cloud-dataflow/commit/e34669fa0b9ed4d8b7372ab1cad7c4515c2616c2

PR #545, opened 2016-04-22
RuntimeModulesController using deployment id

### LINKED ISSUE #544: RuntimeModulesController not using deployment id
`RuntimeModulesController` is still using old way to query status and should use new SPI style via `DeploymentIdRepository`.

jvalkeal:
- Fix RuntimeModulesController to use deployment id
  so that `runtime modules` command wont fail with
  deployers expecting correct format when status
  is queried.
- Generic polish, javadocs, sts warnings.
- Fixes #544


ilayaperumalg:
LGTM, merging.

PR #546, opened 2016-04-22
Fix display of name, version and description

### LINKED ISSUE #536: Server name and version not showing
Looking at the UI under About section the server name ad version aren't showing:

Admin Server Implementation
Name	@project.artifactId@
Version	@project.version@ (e34669f)
Description	@project.description@

Also, shouldn't we drop "Admin" here?

trisberg:
- added application.yml for spring-cloud-dataflow-server-local

- added description to pom for spring-cloud-starter-dataflow-server-local just in case an app using the starter dosen't averride "@project.description@"

Fixes #536

### DIFF
--- a/spring-cloud-starter-dataflow-server-local/pom.xml
+++ b/spring-cloud-starter-dataflow-server-local/pom.xml
@@ -38,9 +38,9 @@
 		</resources>
 		<plugins>
 			<plugin>
-	            <groupId>pl.project13.maven</groupId>
-	            <artifactId>git-commit-id-plugin</artifactId>
-	        </plugin>
-        </plugins>
+				<groupId>pl.project13.maven</groupId>
+				<artifactId>git-commit-id-plugin</artifactId>
+			</plugin>
+		</plugins>
 	</build>
 </project>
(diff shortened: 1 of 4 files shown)


spring-cloud-starter-dataflow-server-local/pom.xml
```diff
@@ -25,4 +26,21 @@
...
+			<resource>
+				<directory>src/main/resources</directory>
+				<filtering>true</filtering>
+				<includes>
+					<include>dataflow-server.yml</include>
+				</includes>
+			</resource>
+		</resources>
+		<plugins>
+			<plugin>
```
markfisher: nitpick: indent weirdness

spring-cloud-starter-dataflow-server-local/src/main/resources/dataflow-server.yml
```diff
@@ -37,6 +33,8 @@ spring:
 # that client apps all need a bootstrap.yml (or an environment variable) with the Eureka
 # server address, e.g. in eureka.client.serviceUrl.defaultZone
```
markfisher: do we need to keep all this commented out stuff here? (since it's moved to `application.yml` also?)

trisberg: wasn't sure, could probably be removed from the starter, although this might be where anyone crating their own server might look

spring-cloud-starter-dataflow-server-local/pom.xml
trisberg: spaces, fixed

markfisher:
rebased, squashed, and merged as https://github.com/spring-cloud/spring-cloud-dataflow/commit/4c5bb4d5ec499cfd3ef4e939ff59b41188b480eb

Issue #491, opened 2016-04-05
Add validation for app registration command


sabbyanandan:
As a user, I'm trying to upload a new application to dataflow-server. Though the feedback states `Successfully registered module 'app-name'`, I'm unable to use the newly registered application in stream definition. 

**Acceptance:**
- I'd like to see dataflow-server validate my registration request; if the `--uri` format is incorrect, I'd like to get alerted about the same
- I'm able to see the error message with hints to what went wrong (_eg: show an example registration command and `--uri` format_)
- I'm able to see different errors depending on `maven`, `docker`, `file`, or other URI variants

styjdt:
I don't know if it's intended, but imho during validation also resolution of the `--uri` target should be done so that not only the format is checked, but also if the app is indeed available and valid for the intended use.

sabbyanandan:
Addressed via spring-cloud/spring-cloud-dataflow#600

PR #549, opened 2016-04-22
support app import from uri and bulk register

### LINKED ISSUE #540: Add "apps import" command
As a developer, I'd like to add `apps import` command, so I can register Apps from a file. 

**Acceptance:**
- `apps import` (`--uri`) command allows me to point to a file with App URIs
- Invalid app format is rejected by the import command
- REST-API is published and available for use outside of shell 
- Unit tests included

### LINKED ISSUE #541: Add "apps register" command
As a developer, I'd like to add `apps register` command, so I can register Apps as a set of k/v pairs. 

**Acceptance:**
- `apps register` (`--map`) command allows me to pass App URIs as k/v pairs 
- Invalid app format is rejected by the register command
- REST-API is published and available for use outside of shell 
- Unit tests included

markfisher:
fixes #540
fixes #541 

requires https://github.com/spring-cloud/spring-cloud-deployer/pull/51

we might want to revisit error handling, but that could possibly be done as a follow-up post M3 instead of within the scope of this PR (cc @aclement )


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/ModuleController.java
```diff
@@ -157,6 +160,42 @@ public void unregister(@PathVariable("type") ArtifactType type, @PathVariable("n
...
 	}
 
+	/**
+	 * Register all apps listed in a properties file or provided as key/value pairs.
+	 * @param uri         URI for the properties file
+	 * @param apps        key/value pairs representing apps, separated by newlines
+	 * @param force       if {@code true}, overwrites any pre-existing registrations
+	 */
+	@RequestMapping(method = RequestMethod.POST)
+	@ResponseStatus(HttpStatus.OK)
```
ericbottard: May want to return CREATED or something similar instead

spring-cloud-dataflow-shell/src/main/java/org/springframework/cloud/dataflow/shell/command/ModuleCommands.java
```diff
@@ -61,14 +62,18 @@
 
 	private static final String REGISTER_MODULE = "module register";
 
+	private static final String IMPORT_APPS = "apps import";
```
ericbottard: When we rename `module` to `app`, we'll have commands that start with `app` and others that start with `apps`, which is not ideal for completion...

spring-cloud-dataflow-shell/src/main/java/org/springframework/cloud/dataflow/shell/command/ModuleCommands.java
```diff
@@ -191,14 +195,45 @@ public Object getValue(int row, int column) {
...
+					key = {"", "uri"},
+					help = "URI for the properties file")
+			String uri,
+			@CliOption(key = "force",
+				help = "force update if any app already exists (only if not in use)",
+				specifiedDefaultValue = "true",
+				unspecifiedDefaultValue = "false")
+			boolean force) {
+		moduleOperations().importFromResource(uri, force);
+		return String.format("Successfully registered apps from '%s'", uri);
```
ericbottard: May be useful to print the `size()` of the returned object

spring-cloud-dataflow-shell/src/main/java/org/springframework/cloud/dataflow/shell/command/ModuleCommands.java
```diff
@@ -191,14 +195,45 @@ public Object getValue(int row, int column) {
...
+	}
+
+	@CliCommand(value = REGISTER_APPS, help = "Register all apps provided in a map")
+	public String registerAll(
+			@CliOption(mandatory = true,
+					key = {"", "map"},
+					help = "apps as key/value pairs with key as 'type.name' and value as a URI")
+			Properties apps,
```
ericbottard: I did not expect support for inline props (which is likely unwieldy), but from a reference to a local file (may be `--propertiesFile` then). WDYT?

markfisher: I thought about this a bit too... using `\n` is not ideal in the shell (fine for the REST API though).

Actually makes me wonder if we need this command in the shell at all, since the other one (taking `--uri`) is already essentially the same thing as `--propertiesFile` (we could consider renaming that option in fact, since the latter is more explicit and matches what we do in other places, such as `stream deploy`)

Another option would be to allow a configurable `--delimiter`, with something more convenient as the default (maybe semicolon). I think it would need to be configurable in case the URIs being provided contain that character.

spring-cloud-dataflow-shell/src/main/java/org/springframework/cloud/dataflow/shell/command/ModuleCommands.java
markfisher: Right. I considered `app all` which would be similar to `stream all` but that didn't seem very intuitive. Maybe we should instead consider the other word in the command... I guess `app import` is okay though (could imply 1 *or more*), and maybe we just need to make the existing `app register` (once `module register` is renamed) accept 1 *or more* itself.

spring-cloud-dataflow-shell/src/main/java/org/springframework/cloud/dataflow/shell/command/ModuleCommands.java
ericbottard: It's not __really__ the same thing, as `--uri`is resolved from the dataflow server, while `--propertiesFile` would be resolved from the running shell filesystem, which is likely not accessible remotely from the dataflow server.

markfisher: Indeed. Good point. I've been running both locally too much ;)

sabbyanandan:
There's spring-cloud/spring-cloud-dataflow#491 to account for validation; yes, we could address it post-M3.

spring-cloud-dataflow-shell/src/main/java/org/springframework/cloud/dataflow/shell/command/ModuleCommands.java
markfisher: @ericbottard: what do you think of this plan (at least for M3)... we could support both `--properties` and `--propertiesFile`, but we could use comma as a delimiter in the case of inline `--properties` since that is at least consistent with deployment properties. My only concern is whether "properties" is intuitive at all when someone is providing apps via key/value pairs.

ericbottard: The thing is, I don't really see the value of inline properties. It takes 2 registrations to make sense (otherwise it's good ol' `module register`) and to me it gets very very hard to read. So people would most certainly have those props in a file somewhere (not to mention the need to do/redo registrations), so `--propertiesFile` is the thing we should promote IMHO.

That being said, I agree that if we're gonna use the `properties/propertiesFile` terminology, then we should somehow be consistent with stream deployment, and use a comma as separator.

markfisher: maybe we just need to be introducing 1 new command: `module import`

that can take `--uri` as it does now, but could also take a flag to indicate that it's local (resolved at the shell and passed to the server as opposed to resolved on the server). From a functionality perspective that seems to cover what you are suggesting as the only options that make sense

ericbottard: hmm, could do, although it could be confusing as well. What would `--local=true --uri=http://foo.bar` mean?

markfisher: that would mean `http://foo.bar` is resolved from the shell process... seems fine actually, since the server might have restrictions (I think `local` should be TRUE by default since it's more intuitive that you are executing the command IN the shell)

spring-cloud-dataflow-shell/src/main/java/org/springframework/cloud/dataflow/shell/command/ModuleCommands.java
markfisher: I'll keep just "module" for now (e.g. "module import")... these can all move to "app" at the same time

markfisher:
@ericbottard I just pushed a commit to address the review comments as well as the approach of always using a URI for the new bulk registering command.... where the "local" flag determines whether resolved in the shell process or deferred to the Data Flow server

spring-cloud-dataflow-shell/src/main/java/org/springframework/cloud/dataflow/shell/command/ModuleCommands.java
```diff
@@ -191,14 +208,48 @@ public Object getValue(int row, int column) {
...
+	@CliCommand(value = IMPORT_MODULES, help = "Register all modules listed in a properties file")
+	public String importFromResource(
+			@CliOption(mandatory = true,
+					key = {"", "uri"},
+					help = "URI for the properties file")
+			String uri,
+			@CliOption(key = "local",
+				help = "whether to resolve the URI locally (as opposed to on the server)",
+				specifiedDefaultValue = "true",
+				unspecifiedDefaultValue = "true")
```
ericbottard: seems like a typo here, should read false, shouldn't it?

spring-cloud-dataflow-shell/src/main/java/org/springframework/cloud/dataflow/shell/command/ModuleCommands.java
```diff
@@ -191,14 +208,48 @@ public Object getValue(int row, int column) {
...
+				help = "force update if any module already exists (only if not in use)",
+				specifiedDefaultValue = "true",
+				unspecifiedDefaultValue = "false")
+			boolean force) {
+		if (local) {
+			try {
+				Resource resource = this.resourceLoader.getResource(uri);
+				Properties modules = PropertiesLoaderUtils.loadProperties(resource);
+				moduleOperations().registerAll(modules, force);
+				return String.format("Successfully registered modules: " + modules.keySet());
```
ericbottard: This "result" is different from the pure uri variant.
I think this is ok for now, later we might want to revisit and display exactly what was registered as a nice Table for example.

ericbottard:
Didn't actually try it yet, but the overall approach seems ok to me after all

spring-cloud-dataflow-shell/src/main/java/org/springframework/cloud/dataflow/shell/command/ModuleCommands.java
markfisher: I agree. It's not trivial to display everything (since it's a paged resources return value). At least, in the case where the URI is resolved within the shell process, the app keys are known, hence I decided to show that information (whereas the server-side resolution of the URI now at least shows the number of apps registered per your suggestion). I didn't want to go much further for now since the follow-up work on error handling might affect what we display.

spring-cloud-dataflow-shell/src/main/java/org/springframework/cloud/dataflow/shell/command/ModuleCommands.java
markfisher: I think it should be TRUE by default (whether specified or not). That just seems more intuitive when pointing to a location *in a shell command*... that you'd expect that location to be resolved by the shell process itself.

markfisher:
something's not working properly for the non-local resolution... investigating

markfisher:
had to incorporate @ilayaperumalg's fix on spring-cloud/spring-cloud-deployer#51

also need to make one small change on the response

markfisher:
addressed review comments, squashed and merged as https://github.com/spring-cloud/spring-cloud-dataflow/commit/953bb6d997a45f69c59f660d4e39dbedeef33efb

ericbottard:
What bothered me is having both `specifiedDefaultValue = "true", 	unspecifiedDefaultValue = "true"`.

They usually have opposite values. Maybe what we need is to invert the option, like so
```
@CliOption(key = "remote",
				help = "whether to resolve the URI remotely (as opposed to on the shell)",
				specifiedDefaultValue = "true",
				unspecifiedDefaultValue = "false")
```

markfisher:
@ericbottard I still think it's correct as-is. My understanding (could be wrong) of those attributes is as follows:

- `specifiedDefaultValue`: the boolean value to use if the option *key* is included but without an explicit value, i.e. `--local` (would have same effect as `--local true`)
- `unspecifiedDefaultValue`: the boolean value to use if the option key is not even present (this would also have the same effect as `--local true`)

In other words, the only way to request non-local resolution is to explicitly add `--local false`, and I think that's fine. My testing verified that behavior.

PR #559, opened 2016-04-26
use generated apps with kafka binder by default

### LINKED ISSUE #558: Populate default registry with app-starter URIs
As a developer, I'd like to populate the registry with app-starter URIs by default. 

**Acceptance:**
- stream-modules/task-modules are replaced with `kafka` app-starter URIs
- add a note about `kafka` requirement for local deployer
- update relevant docs

markfisher:
fixes #558


sobychacko:
merged.

PR #566, opened 2016-04-28
Override EvoInflectorRelProvider with DefaultRelProvider

ilayaperumalg:
- Since Dataflow UI expects the `DefaultRelProvider` to be used, always override the RelProvider with the DefaultRelProvider


trisberg:
LGTM, tested running server locally and on PWS, deploying stream on PWS

Issue #557, opened 2016-04-26
Polish Docs Generation (Asciidoctor/DocBook)


ghillert:
Our Docs generation needs some Polishing for GA … e.g. the PDF is missing the Spring Logo. There are also some image-related warnings.

ghillert:
As we adapted the docs generation from Spring Boot it is noteworthy that the Spring Boot Docs [1] don't use images. Also regarding scaling of image, be aware of the following issue:

* https://github.com/asciidoctor/asciidoctor/issues/763
* https://github.com/asciidoctor/asciidoctor/issues/1059

[1] https://github.com/spring-projects/spring-boot/tree/master/spring-boot-docs

PR #577, opened 2016-05-02
gh-557 Polish Docs Generation

ghillert:
* Add cover page for epub
* Add SCDF logo to cover page (SVG)
* In `index.adoc` authors need to be separated by semicolons
* Fix author and copyright notice rendering in `pdf.xsl`
* Fix image sizing: For Html rendering, make images more responsive by setting their style `max-width` to `100%`
* Ensure that `docbkx-maven-plugin` uses version `2.0.15`. The latest version `2.0.17` breaks the epub rendering

Resolves spring-cloud/spring-cloud-dataflow#557


spring-cloud-dataflow-docs/pom.xml
```diff
@@ -92,21 +101,11 @@
...
 										</copy>
-										<copy todir="${basedir}/target/contents/reference/htmlsingle">
-											<fileset dir="${basedir}/src/main/docbook">
-												<include name="**/*.css" />
-												<include name="**/*.png" />
-												<include name="**/*.gif" />
-												<include name="**/*.jpg" />
-											</fileset>
+										<copy todir="${basedir}/target/contents/reference/htmlsingle/images">
+											<fileset dir="${basedir}/target/generated-docs/images"/>
```
ericbottard: This seems to render this step dependent on the docbook one. Couldn't we have images originally in a non-flavored directory (ie *not* `src/main/docbook/images`) and use that everywhere?

markpollack: Yes, that would seem to be desirable since we may eventually drop docbook as it is just a gateway to creating pdf, which we never link directly to anymore.

markpollack: Ah, turns out "There are 2 sets of images - the ones that belong to Asciidoctor (SCDF) and then the ones that belong to the general docs generation … e.g. src/main/docbook/images/ contains the “Note” and “Tip” (Admotion) images, the EPub Cover, and the first page Spring Logo used for the PDF cover."  
```
HTML: “src/main/docbook/images/background.png”
PDF: “src/main/docbook/images/logo.svg” (switched to SVG for sharper rendering)
EPub: “src/main/docbook/images/cover.png”
```

markpollack:
Added newline at end of css and merged as c0372497ef803972bd942f2d2e1f98494e258f5e

PR #580, opened 2016-05-05
Add commandline params for tasks

### LINKED ISSUE #503: Add command line args support for tasks
When work for https://github.com/spring-cloud/spring-cloud-deployer/issues/20 is done, its wingman here needs to similar things in a shell and controllers.

### LINKED ISSUE #523: Spring Cloud Task launch properties are not persisted
When launching a task, launch properties are not persisted to the database. 

```
task launch scdf-batch-simple --properties sample.jobParams={"throwError":"true","-secondParam(long)":"99999999"}
```

jvalkeal:
- Modify shell launch command.
- Add raw list of params throughout a chain
  from shell into a deployer SPI call.
- Added one test verifying params.
- In DefaultTaskJobService.restartJobExecution() params are
  coming from TaskExecution, theory at least, but this is not
  testable right now as we don't have restart command.
- Fixes #503
- Fixes #523

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDeploymentController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDeploymentController.java
@@ -16,6 +16,8 @@
 
 package org.springframework.cloud.dataflow.server.controller;
 
+import java.util.List;
+
 import org.springframework.cloud.dataflow.rest.resource.TaskDeploymentResource;
 import org.springframework.cloud.dataflow.rest.util.DeploymentPropertiesUtils;
 import org.springframework.cloud.dataflow.server.service.TaskService;
@@ -63,11 +65,13 @@ public class TaskDeploymentController {
 	 * @param taskName the name of the existing task to be executed (required)
 	 * @param properties the runtime properties for the task, as a comma-delimited list of
 	 * 					 key=value pairs
+	 * @param params the runtime commandline arguments
 	 */
 	@RequestMapping(value = "/{name}", method = RequestMethod.POST)
 	@ResponseStatus(HttpStatus.CREATED)
-	public void deploy(@PathVariable("name") String taskName, @RequestParam(required = false) String properties) {
-		this.taskService.executeTask(taskName, DeploymentPropertiesUtils.parse(properties));
+	public void deploy(@PathVariable("name") String taskName, @RequestParam(required = false) String properties,
+			@RequestParam(required = false) List<String> params) {
+		this.taskService.executeTask(taskName, DeploymentPropertiesUtils.parse(properties), params);
 	}
 
 }
--- a/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/TaskControllerTests.java
+++ b/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/TaskControllerTests.java
@@ -17,9 +17,11 @@
 package org.springframework.cloud.dataflow.server.controller;
 
 import static org.hamcrest.Matchers.instanceOf;
+import static org.hamcrest.Matchers.is;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertThat;
 import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.atLeast;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.delete;
@@ -210,7 +212,7 @@ public class TaskControllerTests {
 				.andExpect(status().isCreated());
 
 		ArgumentCaptor<AppDeploymentRequest> argumentCaptor = ArgumentCaptor.forClass(AppDeploymentRequest.class);
-		verify(this.taskLauncher).launch(argumentCaptor.capture());
+		verify(this.taskLauncher, atLeast(1)).launch(argumentCaptor.capture());
 
 		AppDeploymentRequest request = argumentCaptor.getValue();
 		assertThat(request.getResource(), instanceOf(MavenResource.class));
@@ -221,5 +223,32 @@ public class TaskControllerTests {
 		assertEquals("jar", mavenResource.getExtension());
 		assertEquals("1", mavenResource.getVersion());
 		assertEquals("myTask", request.getDefinition().getProperties().get("spring.cloud.task.name"));
+	}
+
+	@Test
+	public void testLaunchWithParams() throws Exception {
+		repository.save(new TaskDefinition("myTask2", "foo2"));
+		this.registry.register("task.foo2", new URI("maven://org.springframework.cloud:foo2:1"));
+
+		mockMvc.perform(
+				post("/tasks/deployments/{name}", "myTask2")
+				.param("params", "--foobar=jee")
+				.accept(MediaType.APPLICATION_JSON)).andDo(print())
+				.andExpect(status().isCreated());
+
+		ArgumentCaptor<AppDeploymentRequest> argumentCaptor = ArgumentCaptor.forClass(AppDeploymentRequest.class);
+		verify(this.taskLauncher, atLeast(1)).launch(argumentCaptor.capture());
+
+		AppDeploymentRequest request = argumentCaptor.getValue();
+		assertThat(request.getCommandlineArguments().size(), is(1));
+		assertThat(request.getCommandlineArguments().get(0), is("--foobar=jee"));
+		assertThat(request.getResource(), instanceOf(MavenResource.class));
+		MavenResource mavenResource = (MavenResource) request.getResource();
+		assertEquals("org.springframework.cloud", mavenResource.getGroupId());
+		assertEquals("foo2", mavenResource.getArtifactId());
+		assertEquals("", mavenResource.getClassifier());
+		assertEquals("jar", mavenResource.getExtension());
+		assertEquals("1", mavenResource.getVersion());
+		assertEquals("myTask2", request.getDefinition().getProperties().get("spring.cloud.task.name"));
 	}
 }
--- a/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/TaskOperations.java
+++ b/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/TaskOperations.java
@@ -16,6 +16,7 @@
 
 package org.springframework.cloud.dataflow.rest.client;
 
+import java.util.List;
 import java.util.Map;
 
 import org.springframework.cloud.dataflow.rest.resource.TaskDefinitionResource;
@@ -43,8 +44,8 @@ public interface TaskOperations {
 	/**
 	 * Launch an already created task.
 	 */
-	void launch(String name, Map<String, String> properties);
-	
+	void launch(String name, Map<String, String> properties, List<String> params);
+
 	/**
 	 * Destroy an existing task.
 	 */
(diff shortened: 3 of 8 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDeploymentController.java
```diff
@@ -63,11 +65,13 @@ public TaskDeploymentController(TaskService taskService) {
...
 	 * 					 key=value pairs
+	 * @param params the runtime commandline arguments
 	 */
 	@RequestMapping(value = "/{name}", method = RequestMethod.POST)
 	@ResponseStatus(HttpStatus.CREATED)
-	public void deploy(@PathVariable("name") String taskName, @RequestParam(required = false) String properties) {
-		this.taskService.executeTask(taskName, DeploymentPropertiesUtils.parse(properties));
+	public void deploy(@PathVariable("name") String taskName, @RequestParam(required = false) String properties,
+			@RequestParam(required = false) List<String> params) {
+		this.taskService.executeTask(taskName, DeploymentPropertiesUtils.parse(properties), params);
```
cppwfs: probably need to parse each of the parameters (delimited by space) out into a list.   Else it will send the all params as a single param instead of collection parameters.  
```
		List<String> paramsToUse = new ArrayList<>();
		if(params != null) {
			for (String paramsItem : params) {
				for (String param : StringUtils.delimitedListToStringArray(paramsItem, " ")) {
					paramsToUse.add(param);
				}
			}
		}
		this.taskService.executeTask(taskName, DeploymentPropertiesUtils.parse(properties), paramsToUse);
```

mminella: Why is this needed?

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/TaskControllerTests.java
```diff
@@ -222,4 +224,31 @@ public void testLaunch() throws Exception {
...
 	}
+
+	@Test
+	public void testLaunchWithParams() throws Exception {
+		repository.save(new TaskDefinition("myTask2", "foo2"));
+		this.registry.register("task.foo2", new URI("maven://org.springframework.cloud:foo2:1"));
+
+		mockMvc.perform(
+				post("/tasks/deployments/{name}", "myTask2")
+				.param("params", "--foobar=jee")
```
mminella: Two questions here:

* How does this look with multiple params?
* why would this go over the wire with `--`?  Is that how other options are sent (pardon my ignorance on this)?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDeploymentController.java
cppwfs: Good question!

When executing from the command line the following arguments `—foo=bar —baz=boo`  would have 2 elements in the arg array.
`—foo=bar`
`—baz=boo`
while a task launched from SCDF would have only 1 element in the arg array.
`—foo=bar —baz=boo`

[

jvalkeal: I thought about trying to parse args but as we really don't know what's going to get passed in I felt safer not trying split it into a list. It really is a raw string(having arbitry params) what user can pass in via shell.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/TaskControllerTests.java
jvalkeal: As mentioned above, we don't care a format what user is passing in. With `--` that'd be a boot property but I don't know if batch itself supports any type of command line args.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDeploymentController.java
cppwfs: I think the primary issue is that without us parsing the args in SCDF, the behavior of how the Task will have to parse the args will be different based on how the task was launched.( i.e. from command line will be multiple elements and from SCDF only one element.)   The way a task parses the args should be consistent, regardless of how it was launched.

jvalkeal: Lets take a default `format` property and pass it in via params as `--format="yyyy-MM-dd HH:mm:ss.SSS"`.

It has a space in it between `dd` and `HH`. We can't just start to split these with spaces because then one parameter would be splitted into to two parameters.

I'm now also seeing some weird behaviour with local deployer where multiple params gets mixed up somehow. Something to do with escaping or ways how stuff is started.

jvalkeal:
So trouble with `ProcessBuilder` in local deployer.
```
task launch --name footask --params "--format='yyyy-MM-dd HH:mm:ss.SSS'"
```
is causing broken format in logs
```
2016-05-12 13:58:31.829  INFO 31548 --- [           main] o.s.c.t.a.t.TimestampTaskConfiguration   : yyyy-MM-dd HH:mm:ss.SSS
```
Process launched as seeing from ps is
```
java -jar /home/jvalkealahti/.m2/repository/org/springframework/cloud/task/app/timestamp-task/1.0.0.M1/timestamp-task-1.0.0.M1.jar --format='yyyy-MM-dd HH:mm:ss.SSS'
```
If I run that command manually I get correct format in logs
```
2016-05-12 14:00:54.460  INFO 31723 --- [           main] o.s.c.t.a.t.TimestampTaskConfiguration   : 2016-05-12 14:00:54.460
```
If I add second argument after it breaks with error:
```
task launch --name footask --params "--format='yyyy-MM-dd HH:mm:ss.SSS' --foo=bar"
```
```
Caused by: java.lang.IllegalArgumentException: Illegal pattern character 'f'
	at java.text.SimpleDateFormat.compile(SimpleDateFormat.java:826) ~[na:1.8.0_45]
```

This may be something mentioned in SO http://stackoverflow.com/questions/12124935/processbuilder-adds-extra-quotes-to-command-line

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDeploymentController.java
cppwfs: I know there is a better way to do this but:
```
		Pattern regex = Pattern.compile("(\\s(?=([^\\\"']*[\\\"'][^\\\"']*[\\\"'])*[^\\\"']*$))");
		for( String paramsItem : params) {
			Matcher regexMatcher = regex.matcher(paramsItem);
			int start = 0;
			while (regexMatcher.find()) {
				paramsToUse.add(paramsItem.substring(start, regexMatcher.start()));
				start = regexMatcher.start();
			}
			if(paramsItem != null && paramsItem.length() > 0){
				paramsToUse.add(paramsItem.substring(start,paramsItem.length()));
			}
		}```
Launching this task:   task launch foo --params "--format=\"yyyy-MM-dd HH:mm:ss.SSS\" bifflediffle=snizersnot wombats=\"eat veggies\" noway=really"

displayed this in my sample task: 
 you got ==>--format="yyyy-MM-dd HH:mm:ss.SSS"
 you got ==> bifflediffle=snizersnot
 you got ==> wombats="eat veggies"
 you got ==> noway=really

**Comment by jvalkeal** [comment]
ok, so we need to find a way to split that raw command-line arguments string. Just checked a process manually:
```
$ pid=$(ps -ef|grep timestamp|grep -v grep | awk '{print $2}'); sudo strings /proc/$pid/cmdline
java
-jar
/home/jvalkealahti/.m2/repository/org/springframework/cloud/task/app/timestamp-task/1.0.0.M1/timestamp-task-1.0.0.M1.jar
--format='yyyy-MM-dd HH:mm:ss.SSS' --foo=bar
```
In above `--format` would get value `'yyyy-MM-dd HH:mm:ss.SSS' --foo=bar` and things go south.
If I run it from command line its
```
$ pid=$(ps -ef|grep timestamp|grep -v grep | awk '{print $2}'); sudo strings /proc/$pid/cmdline
java
-jar
/home/jvalkealahti/.m2/repository/org/springframework/cloud/task/app/timestamp-task/1.0.0.M1/timestamp-task-1.0.0.M1.jar
--format=yyyy-MM-dd HH:mm:ss.SSS
--foo=bar
```

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDeploymentController.java
jvalkeal: Do we need to support arbitrary args which are not in `key=value` format?

cppwfs:
@jvalkeal I did a similar test where I created a timestamp task that printed the format.   And there are 2 problems.  

[Successful Run with multiple parameters from command line]
* Running from the command line with the following `java -jar timestamp-task/target/timestamp-task-1.0.0.BUILD-SNAPSHOT.jar --format="yyyy-MM-dd HH:mm:ss.SSS" --foo=bar`
* Running from the command line with the following `java -jar timestamp-task/target/timestamp-task-1.0.0.BUILD-SNAPSHOT.jar --format='yyyy-MM-dd HH:mm:ss.SSS' --foo=bar`
The task completes with the expected results.  i.e. 2 parameters work

[Running from SCDF]
Problem 1: Using existing code it is placing both --format &  --foo parameters into the format property.  Using the code sample I enumerated above fixes that issue.  
Problem 2: Like you said when executing from SCDF the " or ' are left in the format field, thus causing the logging problem.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDeploymentController.java
cppwfs: I'm going to say yes.  Because a `task` is kind of the wild west it can be any boot app.  But I defer to @mminella  on this one.

jvalkeal:
Ok, I'll tweak that stuff based on what you did. That kinda locks in a format we can use so need to document it. I'll do that tomorrow.

jvalkeal:
I think we need to take this back into a drawing board. From dataflow shell if we need to have a space, we either need to use `--format=\"yyyy-MM-dd HH:mm:ss.SSS\"`, `--format='yyyy-MM-dd HH:mm:ss.SSS', or `'--format=yyyy-MM-dd HH:mm:ss.SS'`.

With all cases we've left double/single quotes which are then passed to ProcessBuilder which then puts those into a real args and things fail. If those are used in a real OS shell, system takes those away.

This also makes it difficult to really know how underlying deployer(local, pcf, yarn) should really handle those command line args. With local those goes directly into ProcessBuilder, with yarn it's really a shell command to be executed.

This raw command-line args is not going to fly, we need to find something more generic and reliable.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDeploymentController.java
mminella: I'd say we need to support anything that boot supports.  I may be wrong here, but I don't think they support flag args (`--foo`), but only properties (`--foo=bar`).  This is something we'd need to confirm with @philwebb

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/TaskControllerTests.java
mminella: When running a batch app via boot, all the args are passed in as job parameters.

jvalkeal:
wohoo, closing this one and planning for 3rd PR.

Issue #311, opened 2016-01-18
Persist stream/task definitions in DB


sabbyanandan:
As a developer, I'd like to add an option to persist stream/task definitions in the existing `datasource`, so I can retrieve  (from the underlying DB) and redeploy the definitions under fault scenarios. This is dependent on spring-cloud/spring-cloud-dataflow#307.

**Acceptance:**
* Ability to persist the definitions can be turned ON; default is OFF
* `datasource` can be backed by variety of DB options 
* Documentation available for reference

markpollack:
Note, will need to test and also update documentation for various admin impls, e.g on CF having the admin bind to mysql and not to redis.

trisberg:
Would also be convenient to have additional JDBC drivers in the build, similar to - https://github.com/spring-cloud/spring-cloud-stream-modules/issues/167

markpollack:
I guess it doesn't hurt i guess - how big are they?  However there isn't a t ask schema that can be automatically created for most of those databases listed.  

Generally speaking, how would the admin get the additional .jar file for a database driver such as oracle?  There is the java -u0f approach or this would be a case for "Josh's" @EnableAdminServer annotation to create a new uberjar with the required dep.

trisberg:
Yeah, we only need the drivers for the supported databases - PostgreSQL (650KB), MySQL (Maria DB driver, 358K), H2 (already included as default) and HSQLDB (1.5MB). So adds ~2MB to the size of the jar. For Oracle, they would have to either inject the jdbc-jar file into the uber-jar or build a new app either putting the Oracle jdbc-jar in `src/main/resources/lib` or adding it to a local Maven repo and referencing that in the pom.

markpollack:
The current repo impls do a check on save to see if an entity already exists.  This is different than the spring data jpa save contract (persist if new, merge if not).  The logic for this check will need to move into the task/stream controllers or into a service layer.  controller is prob enough...  See this related PR for some background. https://github.com/spring-cloud/spring-cloud-dataflow/pull/588

sabbyanandan:
Duplicate of #606 and #589

PR #588, opened 2016-05-11
Support redis stream definition repository

### LINKED ISSUE #586: Use Redis StreamDefinition Repository instead of in-memory
As a user, I would like to persist stream definitions and make them available between the data flow server restarts.

### LINKED ISSUE #587: Validate stream definition's existence before saving into repository
As a user I would like to have some validation when saving stream definitions into repository. If the stream definition already exists, then it should throw appropriate exception.

ilayaperumalg:
- Also use redis deploymentId repository
 - Fix stream definition existence before saving
 - Fix NPE on pageable sort

Resolves #586 
Resolves #587

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
@@ -141,6 +141,9 @@ public class StreamDefinitionController {
 					@RequestParam(value = "deploy", defaultValue = "false")
 					boolean deploy) {
 		StreamDefinition stream = new StreamDefinition(name, dsl);
+		if (this.repository.findOne(stream.getName()) != null) {
+			throw new DuplicateStreamDefinitionException("Stream definition " + name + " already exists.");
+		}
 		this.repository.save(stream);
 		if (deploy) {
 			deploymentController.deploy(name, null);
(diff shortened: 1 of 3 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
```diff
@@ -141,6 +141,9 @@ public void save(@RequestParam("name") String name,
 					@RequestParam(value = "deploy", defaultValue = "false")
 					boolean deploy) {
 		StreamDefinition stream = new StreamDefinition(name, dsl);
+		if (this.repository.findOne(stream.getName()) != null) {
```
markpollack: I think this logic should move down into the repository implementation, as that is what is done for the current in-memory implementation.  Also, the save(Iterable) method in the repo should delegate to the save() method so that the check for existence can be done on each of the definitions.

markpollack: When we shift to Spring Data JPA, the save method is going to do an persist (if new) or merge if it exists already.  So if we want to unsure that you can't save a stream/task def more than once, the logic is going to need to shift back into the controllers for stream/task - or introduce a service layer.  I'll make a note of this on the issue for when we switch to using JPA. https://github.com/spring-cloud/spring-cloud-dataflow/issues/311.

markpollack:
I've updated the PR to sync w/ the current impls of task repos.   Also modified the 'update' test to check for an exception if save is called twice w/ the same stream name.

PR #597, opened 2016-05-16
Add commandline params for tasks

### LINKED ISSUE #503: Add command line args support for tasks

jvalkeal:
Take 3:

Can now use quoting shown below which are then removed during a parsing assuming that deployer is then responsible to handle proper quoting if needed(i.e. local deployer should not use any type of quoting whatsoever).
```
dataflow:>task launch --name footask --params "'--format=yyyy-MM-dd -- HH:mm:ss.SSS' --foo=bar --jee.juu='asdf'"
dataflow:>task launch --name footask --params "\"--format=yyyy-MM-dd -- HH:mm:ss.SSS\" --foo=bar --jee.juu='asdf'"
dataflow:>task launch --name footask --params "--format=\"yyyy-MM-dd -- HH:mm:ss.SSS\" --foo=bar --jee.juu='asdf'"
```

- Modify shell launch command.
- Add raw list of params throughout a chain
  from shell into a deployer SPI call.
- Added one test verifying params.
- In DefaultTaskJobService.restartJobExecution() params are
  coming from TaskExecution, theory at least, but this is not
  testable right now as we don't have restart command.
- Add parsing logic to DeploymentPropertiesUtils together
  with tests.
- Fixes #503

### DIFF
--- a/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/TaskControllerTests.java
+++ b/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/TaskControllerTests.java
@@ -17,9 +17,11 @@
 package org.springframework.cloud.dataflow.server.controller;
 
 import static org.hamcrest.Matchers.instanceOf;
+import static org.hamcrest.Matchers.is;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertThat;
 import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.atLeast;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.delete;
@@ -210,7 +212,7 @@ public class TaskControllerTests {
 				.andExpect(status().isCreated());
 
 		ArgumentCaptor<AppDeploymentRequest> argumentCaptor = ArgumentCaptor.forClass(AppDeploymentRequest.class);
-		verify(this.taskLauncher).launch(argumentCaptor.capture());
+		verify(this.taskLauncher, atLeast(1)).launch(argumentCaptor.capture());
 
 		AppDeploymentRequest request = argumentCaptor.getValue();
 		assertThat(request.getResource(), instanceOf(MavenResource.class));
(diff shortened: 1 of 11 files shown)


cppwfs:
LGTM!

mminella:
I'm a bit confused on the syntax.  There are examples here that allow the quoting of just the value (what I'd expect) and quoting both the key and value together (something I wouldn't expect).  Am I missing something?

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/TaskControllerTests.java
```diff
@@ -222,4 +224,59 @@ public void testLaunch() throws Exception {
...
+		assertEquals("org.springframework.cloud", mavenResource.getGroupId());
+		assertEquals("foo2", mavenResource.getArtifactId());
+		assertEquals("", mavenResource.getClassifier());
+		assertEquals("jar", mavenResource.getExtension());
+		assertEquals("1", mavenResource.getVersion());
+		assertEquals("myTask2", request.getDefinition().getProperties().get("spring.cloud.task.name"));
+	}
+
+	@Test
+	public void testLaunchWithParams2() throws Exception {
```
mminella: Can we add one test that quotes the params as well (since we have all that logic to handle the quotes, it would be good to be sure that nothing in this chain of processing screws them up.

jvalkeal: Yeah I just added different ways user might use quoting and those what I added seem to work. I can add quoted params also in those controller tests.

jvalkeal: and btw, you can do both from a command-line so it's kinda expected behaviour/format
```
java -jar timestamp-task-1.0.0.M1.jar --format='yyyy-MM-dd -- HH:mm:ss.SSS'
java -jar timestamp-task-1.0.0.M1.jar '--format=yyyy-MM-dd -- HH:mm:ss.SSS'
```

mminella:
LGTM.  Merged.

PR #600, opened 2016-05-17
Add validation on app definitions

### LINKED ISSUE #574: Review stream/task parser validation
As a user, I'm able to create a stream without `sink` application in it. I should be prevented from creating this bogus stream definition. 

**example:**
```
dataflow:>stream create streama --definition "time | filter"
Created new stream 'streama'
```

ilayaperumalg:
- validate app names based on their type when the stream definition is created
  - this way we don't need to inject the app registry at the parser level but the validation is always performed when the defintion is created
  - use app registry to find if the names exist in the registry

 - Add test

This resolves #574

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
@@ -16,10 +16,19 @@
 
 package org.springframework.cloud.dataflow.server.controller;
 
+import java.util.ArrayList;
+import java.util.EnumSet;
+import java.util.List;
+import java.util.Set;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import org.springframework.cloud.dataflow.core.ArtifactType;
+import org.springframework.cloud.dataflow.core.BindingPropertyKeys;
 import org.springframework.cloud.dataflow.core.ModuleDefinition;
 import org.springframework.cloud.dataflow.core.StreamDefinition;
+import org.springframework.cloud.dataflow.registry.AppRegistry;
 import org.springframework.cloud.dataflow.rest.resource.StreamDefinitionResource;
 import org.springframework.cloud.dataflow.server.repository.DeploymentIdRepository;
 import org.springframework.cloud.dataflow.server.repository.DeploymentKey;
@@ -36,15 +45,13 @@ import org.springframework.hateoas.PagedResources;
 import org.springframework.hateoas.mvc.ResourceAssemblerSupport;
 import org.springframework.http.HttpStatus;
 import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.ResponseStatus;
 import org.springframework.web.bind.annotation.RestController;
-
-import java.util.EnumSet;
-import java.util.Set;
 
 /**
  * Controller for operations on {@link StreamDefinition}. This
@@ -78,6 +85,11 @@ public class StreamDefinitionController {
 	private final AppDeployer deployer;
 
 	/**
+	 * The app registry this controller will use to lookup apps.
+	 */
+	private final AppRegistry appRegistry;
+
+	/**
 	 * Assembler for {@link StreamDefinitionResource} objects.
 	 */
 	private final Assembler streamDefinitionAssembler = new Assembler();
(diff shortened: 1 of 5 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
```diff
@@ -16,10 +16,19 @@
 
 package org.springframework.cloud.dataflow.server.controller;
 
+import java.util.ArrayList;
```
markfisher: please adjust your import settings to avoid these changes

ilayaperumalg: Sorry. I will fix this. The `spring-cloud-stream` project now uses [this import settings] (https://github.com/spring-cloud/spring-cloud-stream/blob/master/src/checkstyle/checkstyle.xml#L85) and I got it continued here through my IDE settings. I think we need to decide on the common settings across the projects.

ilayaperumalg: Pushed the change to fix the import order.

markfisher: Actually - I'm sorry, and nevermind my comment. What you have in your PR is correct (*java.* should precede *org.*)... so apparently this is a case of someone else having different settings previously ;)

markfisher: ...or now I should say what you *had* in the PR originally was correct. Sorry about the confusion.

ilayaperumalg: oh ok :-)  no problem. I will revert the recent push.

artembilan: Yeah... We really should decide something common for all our projects.
Because we use in SI, SA, SK etc. these import options:
````xml
<module name="ImportOrder">
	<property name="groups" value="java,/^javax?\./,org,org.springframework,*" />
	<property name="ordered" value="true" />
	<property name="separated" value="true" />
	<property name="option" value="top" />
	<property name="sortStaticImportsAlphabetically" value="true" />
</module>
````
:cry:

markfisher: I think we should go with what @artembilan posted above. CC @mbogoevici

mbogoevici: Why? this is an old standard and not fitting neither the Boot nor Spring Framework standards. Other projects have adopted SF code conventions as well. How do we plan to provide a reference to contributors?

artembilan: Not really: https://github.com/spring-projects/spring-security/blob/master/config/src/main/java/org/springframework/security/config/websocket/WebSocketMessageBrokerSecurityBeanDefinitionParser.java. 
But we can reconsider.

Although we have reflected that in our [Code Style](https://github.com/spring-projects/spring-integration/wiki/Spring-Integration-Framework-Code-Style) rules already :smile: : 

CC @garyrussell

garyrussell: The most annoying change (to me) is boot moved the statics to the bottom - AFAIK, SF (and the rest of us) always have had them at the top.

artembilan: Not really. SF code is more ugly than we expect: https://github.com/spring-projects/spring-framework/blob/master/spring-core/src/test/java/org/springframework/core/ConventionsTests.java.

They allow to use *asterisk* !!! :open_mouth:

mbogoevici: Well, we do advertise that we follow the Spring Cloud code conventions on both https://github.com/spring-cloud/spring-cloud-stream and https://github.com/spring-cloud/spring-cloud-dataflow#contributing, which link back to https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-docs/src/main/asciidoc/appendix-contributing.adoc, which link back to the Spring Framework code conventions. It doesn't help that the eclipse settings in Spring Cloud Commons are even less standard than anything else: https://github.com/spring-cloud/spring-cloud-commons/blob/master/spring-cloud-commons/eclipse/org.eclipse.jdt.ui.prefs#L67. I guess we do need to be consistent within the Spring Cloud group of project. So, under the circumstances, I think that adopting the Spring Framework rules would give us a reference point. I am willing to adjust, but let's be consistent. Also, I have noticed that Spring Cloud Task has adopted  the Spring Framework rules as well (including imports). cc @mminella @dsyer

cppwfs: +1 to @mbogoevici 's comment

mminella: +1 for using Spring Framework's code conventions.  I'd be open to hearing a reason why we should deviate but in general, the order of the imports isn't something I'd expect to have to mess with as I move from Spring project to Spring project...

markfisher: We should continue the formatting discussion, but I'm going to merge this PR now ;)

markfisher:
rebased and merged as https://github.com/spring-cloud/spring-cloud-dataflow/commit/3ec4bb95c3ddb1e5063e434bb5ba8cda9d0794eb

PR #602, opened 2016-05-17
Use random port in datasource URL for tests

### LINKED ISSUE #601: Dataflow server tests should use random port for datasource URL
Some of the tests `LocalConfigurationTests` and all the `shell` tests use datasource URL which is fixed from `dataflow-server.yml`. We need to use random port for the datasource URL so that the tests can be run in parallel (as well as while the dataflow server is running)

ilayaperumalg:
This resolves #601

### DIFF
--- a/spring-cloud-dataflow-server-local/src/test/java/org/springframework/cloud/dataflow/server/local/LocalConfigurationTests.java
+++ b/spring-cloud-dataflow-server-local/src/test/java/org/springframework/cloud/dataflow/server/local/LocalConfigurationTests.java
@@ -16,16 +16,17 @@
 
 package org.springframework.cloud.dataflow.server.local;
 
-import static org.hamcrest.Matchers.instanceOf;
-import static org.hamcrest.Matchers.is;
-import static org.junit.Assert.assertThat;
-
 import org.junit.Test;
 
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-server-local/src/test/java/org/springframework/cloud/dataflow/server/local/LocalConfigurationTests.java
```diff
@@ -16,16 +16,17 @@
...
-
 import org.junit.Test;
 
 import org.springframework.boot.SpringApplication;
 import org.springframework.cloud.deployer.spi.local.LocalAppDeployer;
 import org.springframework.cloud.deployer.spi.local.LocalTaskLauncher;
 import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.util.SocketUtils;
+
+import static org.hamcrest.Matchers.instanceOf;
```
markfisher: static imports should remain at the top

markfisher:
adjusted static import order, rebased and merged as https://github.com/spring-cloud/spring-cloud-dataflow/commit/742546d40ee05b551250d4b6c811024fccfe9dd7

Issue #589, opened 2016-05-11
Persist Task definitions in TaskRespository


sabbyanandan:
As a developer, I'd like to persist task definitions using [`TaskDefinitionRepository`](https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/TaskDefinitionRepository.java), so we could list the tasks and as well query to merge the current status for each task running in various platforms. 

**Acceptance:**
* task definitions are persisted in `TaskRepository`, backed by an RDBMS
* `task list` from the shell retrieves all tasks from the `TaskRepository`
* i can query task by task definition name

PR #604, opened 2016-05-19
Persist Task definitions in RDBMSTaskRespository

cppwfs:
resolves spring-cloud/spring-cloud-dataflow#589


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/RdbmsTaskDefinitionRepository.java
```diff
@@ -0,0 +1,284 @@
...
+	private static final String DELETE_FROM_TASK_DEFINITION = "DELETE FROM " + FROM_CLAUSE;
+
+	private static final String DELETE_FROM_TASK_DEFINITION_BY_NAME = DELETE_FROM_TASK_DEFINITION
+			+ TASK_DEFINITION_NAME_WHERE_CLAUSE;
+
+	private String tablePrefix = DEFAULT_TABLE_PREFIX;
+
+	private JdbcOperations jdbcTemplate;
+
+	private NamedParameterJdbcTemplate namedParameterJdbcTemplate;
```
mminella: Why are we using both a `JdbcTemplate` and a `NamedParameterJdbcTemplate`?  Seems like just picking one would make sense...

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/RdbmsTaskDefinitionRepository.java
```diff
@@ -0,0 +1,284 @@
...
+	private String prefixSql(String base) {
+		return StringUtils.replace(base, "%PREFIX%", tablePrefix);
+	}
+
+	private Page<TaskDefinition> queryForPageableResults(Pageable pageable,
+														 String selectClause,
+														 String fromClause,
+														 String whereClause,
+														 Object[] queryParam,
+														 long totalCount) {
```
mminella: Formatting

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/RdbmsTaskDefinitionRepository.java
```diff
@@ -0,0 +1,284 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.dataflow.server.repository;
+
+import javax.sql.DataSource;
```
mminella: I believe java should go above javax...

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/RdbmsTaskDefinitionRepository.java
```diff
@@ -0,0 +1,284 @@
...
+	private static final String DELETE_FROM_TASK_DEFINITION_BY_NAME = DELETE_FROM_TASK_DEFINITION
+			+ TASK_DEFINITION_NAME_WHERE_CLAUSE;
+
+	private String tablePrefix = DEFAULT_TABLE_PREFIX;
+
+	private JdbcOperations jdbcTemplate;
+
+	private NamedParameterJdbcTemplate namedParameterJdbcTemplate;
+
+	private DataSource dataSource;
```
mminella: Why save off the datasource?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/RdbmsTaskDefinitionRepository.java
```diff
@@ -0,0 +1,284 @@
...
+		this.jdbcTemplate = new JdbcTemplate(dataSource);
+		this.namedParameterJdbcTemplate = new NamedParameterJdbcTemplate(dataSource);
+		this.dataSource = dataSource;
+		orderMap = new TreeMap<>();
+		orderMap.put("DEFINITION_NAME", Order.ASCENDING);
+	}
+
+	@Override
+	public Iterable<TaskDefinition> findAll(Sort sort) {
+		throw new UnsupportedOperationException();
```
mminella: Why is this unsupported?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/RdbmsTaskDefinitionRepository.java
```diff
@@ -0,0 +1,284 @@
...
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Types;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+
+import org.springframework.batch.item.database.Order;
```
mminella: This shouldn't have a batch dependency...

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/RdbmsTaskDefinitionRepository.java
cppwfs: This is not supported in the in-InMemoryTaskDefinitionRepository.  Do we wish to add this?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/RdbmsTaskDefinitionRepository.java
cppwfs: good catch

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/RdbmsTaskDefinitionRepository.java
mminella: If it's in the SPI...I'd think so...

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/RdbmsTaskDefinitionRepository.java
cppwfs: In this case we used the traditional '?' to denote traditional place holders for most of the queries thus the JDBCTemplate.  The only exception is when we needed to have a findAll(taskNames) where a "IN" clause and in this case we used the NamedParameterJdbcTemplate as it seemed a more elegant way to handle this scenario.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/RdbmsTaskDefinitionRepository.java
cppwfs: I'll be submitting a new commit that will remove this dependency from a couple of other files along with one other batch dependency.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/RdbmsTaskDefinitionRepository.java
cppwfs: On it!

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/RdbmsTaskDefinitionRepository.java
mminella: Why not just use the named everywhere then?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/RdbmsTaskDefinitionRepository.java
```diff
@@ -0,0 +1,297 @@
...
+
+	private Map<String, Order> orderMap;
+
+	public RdbmsTaskDefinitionRepository(DataSource dataSource) {
+		Assert.notNull(dataSource);
+		this.jdbcTemplate = new JdbcTemplate(dataSource);
+		this.namedParameterJdbcTemplate = new NamedParameterJdbcTemplate(dataSource);
+		this.dataSource = dataSource;
+		orderMap = new TreeMap<>();
+		orderMap.put("DEFINITION_NAME", Order.ASCENDING);
```
mminella: `this.`?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/RdbmsTaskDefinitionRepository.java
```diff
@@ -0,0 +1,297 @@
...
+
+		while (iter.hasNext()) {
+			Sort.Order order = iter.next();
+			query = query + order.getProperty() + " " + order.getDirection();
+			if (iter.hasNext()) {
+				query = query + ", ";
+			}
+		}
+		return Collections.unmodifiableCollection(
+				jdbcTemplate.query(prefixSql(query), new TaskDefinitionRowMapper()));
```
mminella: Why the unmodifiable collection here?  We need it in the in memory option because we wanted to control who updated the underlying datastore.  Here that isn't an issue.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/RdbmsTaskDefinitionRepository.java
```diff
@@ -0,0 +1,297 @@
...
+		delete(definition.getName());
+	}
+
+	@Override
+	public void delete(Iterable<? extends TaskDefinition> definitions) {
+		Assert.notNull(definitions, "definitions must not null");
+		for (TaskDefinition definition : definitions) {
+			delete(definition);
+		}
+	}
```
mminella: In most of these methods, can we do the call to `prefixSql` once and cache it so that we don't need to do that string manipulation for every call?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/RdbmsTaskDefinitionRepository.java
```diff
@@ -0,0 +1,297 @@
...
+		catch (EmptyResultDataAccessException e) {
+			result = false;
+		}
+		return result;
+	}
+
+	@Override
+	public Iterable<TaskDefinition> findAll() {
+		return Collections.unmodifiableCollection(
+				jdbcTemplate.query(prefixSql(FIND_ALL_DEFINITIONS), new TaskDefinitionRowMapper()));
```
mminella: Same unmodifiable collection question.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/RdbmsTaskDefinitionRepository.java
```diff
@@ -0,0 +1,297 @@
...
+		for (String name : taskNames) {
+			listOfNames.add(name);
+		}
+
+		MapSqlParameterSource namedParameters = new MapSqlParameterSource();
+		namedParameters.addValue(LIST_OF_NAMES, listOfNames);
+
+		return Collections.unmodifiableCollection(
+				namedParameterJdbcTemplate.query(prefixSql(FIND_DEFINITIONS_WHERE_IN_CLAUSE),
+				namedParameters, new TaskDefinitionRowMapper()));
```
mminella: Same unmodifiable collection question.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/RdbmsTaskDefinitionRepository.java
```diff
@@ -0,0 +1,297 @@
...
+		return StringUtils.replace(base, "%PREFIX%", tablePrefix);
+	}
+
+	private Page<TaskDefinition> queryForPageableResults(Pageable pageable,
+			String selectClause,
+			String fromClause,
+			String whereClause,
+			Object[] queryParam,
+			long totalCount) {
+		SqlPagingQueryProviderFactoryBean factoryBean = new SqlPagingQueryProviderFactoryBean();
```
mminella: Creating one of these for every call may be a bit heavy since it needs to do the work of determining the type of database it's talking to each time.  I'm not saying we need to do something now, but a note here saying that we may want to improve the performance of this would be a good idea.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/RdbmsTaskDefinitionRepository.java
```diff
@@ -0,0 +1,297 @@
...
+	 */
+	private final class TaskDefinitionRowMapper implements RowMapper<TaskDefinition> {
+
+		public TaskDefinitionRowMapper() {
+		}
+
+		@Override
+		public TaskDefinition mapRow(ResultSet rs, int rowNum) throws SQLException {
+			return new TaskDefinition(
+					rs.getString("DEFINITION_NAME"), rs.getString("DEFINITION"));
```
mminella: Constants?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/support/DefinitionRepositoryInitializer.java
```diff
@@ -0,0 +1,105 @@
...
+	/**
+	 * Path to the SQL file to use to initialize the database schema.
+	 */
+	private static String schema = DEFAULT_SCHEMA_LOCATION;
+
+	private DataSource dataSource;
+
+	private ResourceLoader resourceLoader;
+
+	@Value("${spring.cloud.dataflow.server.initialize.enable:true}")
```
mminella: is this property specific enough for what it does?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/support/DefinitionRepositoryInitializer.java
```diff
@@ -0,0 +1,105 @@
...
+ * {@link DataSource} is available in the current context, and functionality is enabled
+ * (as it is by default), this will initialize the database.
+ *
+ * By default, initialization of the database can be disabled by configuring the property
+ * <code>spring.cloud.dataflow.server.initialize.enable</code> to false.
+ *
+ * @author Glenn Renfro
+ */
+
+public final class DefinitionRepositoryInitializer implements InitializingBean {
```
mminella: Is there something within Spring that we can use instead of bringing this file around from project to project (or contribute it to core)?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/support/JdbcParameterUtils.java
```diff
@@ -0,0 +1,113 @@
...
+import java.util.List;
+
+/**
+ * Helper methods for SQL statement parameter parsing.
+ *
+ * Only intended for internal use.
+ *
+ * @author Thomas Risberg
+ * @author Juergen Hoeller
+ * @since 2.0
```
mminella: Where did this come from and why is it copied in (I'm assuming Juergen isn't working on DataFlow and DataFlow isn't 2.0...)?

mminella:
This PR has a lot of code copied over from SCT which was copied over from batch to support the pagination functionality.  We need to look at Spring Data to see how they do this because coping all this code around doesn't make sense.

cppwfs:
closed by mistake

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/RdbmsTaskDefinitionRepository.java
cppwfs: Good question.  In short the pagination code expects the '?' as a placeholder.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/support/JdbcParameterUtils.java
cppwfs: Makes sense.  But in this case, this code was pulled from Batch, to handle the gathering of placeholders for pagination.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfiguration.java
```diff
@@ -307,6 +302,18 @@ public TaskRepositoryInitializer taskRepositoryInitializerForDefaultDB(DataSourc
 			return taskRepositoryInitializer;
 		}
 
+		@Bean
+		public DefinitionRepositoryInitializer definitionRepositoryInitializerForDefaultDB(DataSource dataSource, Server server) {
```
ilayaperumalg: What is the reasoning behind having two separate `definitionRepositoryInitializerForDefaultDB` beans? I guess we don't use the H2 `Server` here?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfiguration.java
```diff
@@ -307,6 +302,18 @@ public TaskRepositoryInitializer taskRepositoryInitializerForDefaultDB(DataSourc
...
+			DefinitionRepositoryInitializer definitionRepositoryInitializer = new DefinitionRepositoryInitializer();
+			definitionRepositoryInitializer.setDataSource(dataSource);
+			return definitionRepositoryInitializer;
+		}
+
+		@Bean
+		@ConditionalOnMissingBean
+		public TaskDefinitionRepository taskDefinitionRepository(DataSource dataSource, Server server) throws Exception{
```
ilayaperumalg: Same is the case for the `taskDefinitionRepository` bean.

ilayaperumalg:
I believe this PR sets the base work for rest of the RDBMS based repository implementations (with some possible refactoring around reusing the jdbc interactions).

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfiguration.java
cppwfs: Good question.  In short when using the embedded h2 server (default) we have to wait for the H2 server to be available prior to trying to connect to it.  When not using the embedded server we use the other definitionRepositoryInitializerForDefaultDB (No server param) to connect to an external datasource.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/RdbmsTaskDefinitionRepository.java
cppwfs: updated with latest commit

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfiguration.java
ilayaperumalg: I wonder if this can be achieved via `@Autowired(required=false)` on `Server` ?

cppwfs: Not sure if this would help.  The issue is that the server and the datasource both must be up before  this bean can be created (and the database initialized). By stating that the sever autowiring is not required, the following scenario could occur when connecting to the H2:
* Datasource is available.  Server is not available.   The definitionRepositoryInitializerForDefaultDB gets called and the initialization fails because datasource failed to connect to the h2 server.

mminella:
LGTM.  Merged as ec70c92a2196c925ee96dc4e9b93f3e65b6a2d7b.

PR #609, opened 2016-05-20
Cleanup deploymentId repo after undeployment

### LINKED ISSUE #608: Cleanup deploymentId of the app after undeployment
When an app is deployed as a stream, each app is assigned with the deployment identifier and this identifier gets persisted on `DeploymentIdRepository`. This id needs to be deleted when the app is un-deployed.

ilayaperumalg:
- When the app is undeployed, the corresponding deployment id gets deleted from the deploymentIdRepository

This resolves #608

### DIFF
--- a/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/repository/RedisDeploymentIdRepositoryTests.java
+++ b/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/repository/RedisDeploymentIdRepositoryTests.java
@@ -16,10 +16,15 @@
 
 package org.springframework.cloud.dataflow.server.repository;
 
+import java.util.ArrayList;
+import java.util.List;
+
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.notNullValue;
+import static org.hamcrest.Matchers.nullValue;
 import static org.junit.Assert.assertThat;
 
+import org.junit.After;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
@@ -27,12 +32,15 @@ import org.springframework.cloud.dataflow.core.ModuleDefinition;
 import org.springframework.cloud.dataflow.core.StreamDefinition;
 import org.springframework.cloud.dataflow.core.TaskDefinition;
 import org.springframework.cloud.stream.test.junit.redis.RedisTestSupport;
+import org.springframework.data.redis.core.RedisTemplate;
+import org.springframework.data.redis.core.StringRedisTemplate;
 
 /**
  * Tests for {@link RedisDeploymentIdRepository}.
  *
  * @author Janne Valkealahti
  * @author Mark Fisher
+ * @author Ilayaperumal Gopinathan
  */
 public class RedisDeploymentIdRepositoryTests {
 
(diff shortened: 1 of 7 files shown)


cppwfs:
On potential scenario... And argue with me ;-)
1) User creates a stream `stream create woot --definition "time | log" --deploy`
2) User shuts down local server
3) User brings local server back up
4) User destroys the stream `stream destroy woot`
In this case the keys are still in the deployment-ids hash (i.e. orphaned).  So do we want to move the  line at https://github.com/ilayaperumalg/spring-cloud-dataflow/blob/b9431226d553c8f10eb3e8b6920e73db8c61d472/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java#L367 to outside of the `if (!EnumSet.of(DeploymentState.unknown, DeploymentState.undeployed).contains(status.getState()))`  to force a delete of the key.    

Thoughts?

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/repository/RedisDeploymentIdRepositoryTests.java
```diff
@@ -61,6 +77,7 @@ public void testSimpleOperations() {
 		String appDeploymentKey5 = DeploymentKey.forApp(taskDefinition1.getModuleDefinition());
 		String appDeploymentKey6 = DeploymentKey.forApp(taskDefinition2.getModuleDefinition());
 
+
```
cppwfs: nitpicky I can get rid of this blank line on merge

ilayaperumalg:
@cppwfs Even in the above case you mentioned, the deletion of `deployment-ids` would only happen when the `undeployStream` gets invoked. The shutdown of the server from step (2) will not invoke this method and hence the `deployment-ids` will remain intact.

Also, this specific issue is only with the server that uses `local` deployer as the other deployers have their ways to maintain the deployment status of the app. Let's assume we add a check on the `calculateStreamState` and cleanup deployment ids when the status is `unknown`, this might impact other deployers where the `unknown` might come back as `deployed` but the deployment-ids cleaned up. I guess we don't want to get into that state.

cppwfs:
@ilayaperumalg Just a quick clarification to my question.   The shutting down of the server was meant to simulate the abnormal termination of a server  and not meant to delete the deployment-ids.  The goal was to verify that I could remove the ids via an destroy/undeploy (thus cleaning up the ids) following the this scenario: create the stream, kill the server and then bring it back up.   

This test was brought about because following the run of our unit tests there was some orphaned data in the deployment-ids from other tests in our suite.  

To your second point, I guess the question is this:   If we allow for the deletion of a stream from the repository if the stream is in an unknown state, then why do we keep the deployment-ids?

ilayaperumalg:
@cppwfs 
> The goal was to verify that I could remove the ids via an destroy/undeploy (thus cleaning up the ids) following the this scenario

I believe the `deployment-id` should get cleaned up only during `undeploy` as the `destroy` could just be deleting the definition.

> If we allow for the deletion of a stream from the repository if the stream is in an unknown state, then why do we keep the deployment-ids?

right, in this case we could just check if the `deployment-ids` exist when the stream definition is deleted as we can't handle cleanup of `deployment-id` in `unknown` state. But this happens only in `local` server case.

What do you think?

cppwfs:
Agreed.  I think we are ready for a merge.

--Glenn

cppwfs:
LGTM Rebase Merged.

PR #611, opened 2016-05-20
Fix tab completion for taps DSL

### LINKED ISSUE #610: Shell tab completion strategy for tap isn't working
As a user, I would like to get tab completion for the stream definition for the tap. The tap stream currently uses `:` prefix and the tab completion is expected to provide list of stream names.

ilayaperumalg:
- Fix the DSL parser to throw appropriate error message based on the shell provided DSL text
 - If the DSL starts with tap prefix ":", then provide auto completion proposals
   - with the stream names from stream definition repositories
   - the app definition label names (except for the sink apps)
   - Move the app definition `determineModuleType` static method into util so that all the classes under server core could use it
 - Fix the DestinationNameYieldsModulesRecoveryStrategy with the appropriate samples
 - Fix tests

This resolves #610

### DIFF
--- a/spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/DestinationNameYieldsAppsRecoveryStrategy.java
+++ b/spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/DestinationNameYieldsAppsRecoveryStrategy.java
@@ -37,7 +37,7 @@ class DestinationNameYieldsAppsRecoveryStrategy extends
 	private final AppRegistry appRegistry;
 
 	public DestinationNameYieldsAppsRecoveryStrategy(AppRegistry appRegistry) {
-		super(CheckPointedParseException.class, "queue:foo >", "queue:foo > ");
+		super(CheckPointedParseException.class, ":foo >", ":foo > ");
 		this.appRegistry = appRegistry;
 	}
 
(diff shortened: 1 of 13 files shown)


markfisher:
I have an alternative commit that was rebased on top of PR #614:
https://github.com/markfisher/spring-cloud-dataflow/commit/36975f6e7b8b48f96c61fb79abc4a66a6e099d3b

markfisher:
@ilayaperumalg I've rebased/resolved-conflicts and added one commit, and I pushed those 2 commits to this branch: https://github.com/markfisher/spring-cloud-dataflow/commits/fix-tab-completion-for-taps

I tried it in the shell as follows:

```
> stream create ticktock --definition "time | log"

> stream create tap --definition "ticktock<TAB>...
```

That worked... it completed as "ticktock.time".

However, this did not offer any completions (TAB after a DOT):

```
> stream create tap --definition "ticktock.<TAB>...
```

Shouldn't that work as well?

ilayaperumalg:
@markfisher yeah, I think that should work as well. Since I didn't add a test for that case I might have missed it from my PR before rebase as well. I will check it out and update.

ilayaperumalg:
After rebasing, pushed the fix for completion after `streamName + DOT`. Also, realised there weren't specific tests for tab completions for taps. Added tests for the cases.

markfisher:
squashed and merged as https://github.com/spring-cloud/spring-cloud-dataflow/commit/0aad264ea15ddae812b9f796ed7f55c7230e7528

PR #622, opened 2016-05-24
Add TaskAppStarters to LocalServer app properties.

### LINKED ISSUE #581: Add OOTB task apps to registry
As a developer, I'd like to add OOTB `task` applications to the registry, so it is readily available for use from the shell.

**Acceptance:**
- OOTB apps from [task-app-starters](https://github.com/spring-cloud/spring-cloud-task-app-starters) repo are listed in `kafka-apps.properties` and `rabbit-apps.properties`
- `module list` command includes OOTB `task` applications

cppwfs:
This resolves #581


spring-cloud-dataflow-server-local/src/main/resources/META-INF/kafka-apps.properties
```diff
@@ -37,3 +37,8 @@ sink.tcp=maven://org.springframework.cloud.stream.app:tcp-sink-kafka:@registry.a
 sink.throughput=maven://org.springframework.cloud.stream.app:throughput-sink-kafka:@registry.app.defaultVersion@
 sink.websocket=maven://org.springframework.cloud.stream.app:websocket-sink-kafka:@registry.app.defaultVersion@
 task.timestamp=maven://org.springframework.cloud.task.app:timestamp-task:@registry.app.defaultVersion@
```
ilayaperumalg: Given these task app starters don't have any `binder` dependency, can we move them to a common `properties` file named `task-apps.properties` or something?

sabbyanandan: Not that they use binder the same way as stream applications, but the Task applications do publish events to brokers as well. While I like the idea of keeping `task` applications in a separate file, at the same time, this is going to add one extra step for users to register them. Also, this experience is only for the Local server, so I'd vote for keeping them in the same file.

cppwfs: I updated the PR with what Ilaya had requested.  Keep in mind that if a user wants to treat a task as a source then they will need to add it to {binder}-apps.properties.  But with the starter task apps here they will not be dispatching messages to downstream apps.  If we do change this then we can move them back to the {binder}-apps.properties.

ilayaperumalg: @cppwfs I am not sure I understand the above. I think it is just a convenient we just register them via explicit location (task-apps.properties in this case). Irrespective of the location, these apps are always registered as `task` apps.

@sabbyanandan I thought the reverse. By not keeping `task` applications in a separate file, we add extra step to add the tasks in all the binder properties files (and duplicating them).

cppwfs:
@ilayaperumalg I have already updated this PR based on your previous request.   Please verify the update is correct.

spring-cloud-dataflow-server-local/src/main/resources/META-INF/kafka-apps.properties
ilayaperumalg: @cppwfs I saw your changes and they look good. The one I referred above which is unclear to me is this:

> Keep in mind that if a user wants to treat a task as a source then they will need to add it to {binder}-apps.properties. But with the starter task apps here they will not be dispatching messages to downstream apps.

Could you clarify this.

cppwfs: I think we can disregard that statement.  After some additional thought the user will only use a task as exactly that and not a source.

ilayaperumalg:
LGTM, squashed, rebased and merged as ec49eb7cc07ce5251a13fcbdb0f91c09a8e8e898

Issue #618, opened 2016-05-23
Persist app-deployment-id in RDBMS


sabbyanandan:
As a user, I can persist application deployment-id in MySQL, so I don't have a dependency on `redis`. 

**Acceptance:**
- Application deployment-id's are now persisted in MySQL
- Unit tests included
- Documentation updated

Issue #606, opened 2016-05-19
Persist stream definitions in RDBMS


sabbyanandan:
As a user, I can persist stream definitions in MySQL, so I don't have a dependency on `redis`. 

**Acceptance:**
- [StreamDefinitionController](https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java#L153) doesn't use [RedisStreamDefinitionRepository](https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/RedisStreamDefinitionRepository.java) anymore
- Similar to `RdbmsTaskDefinitionRepository`, we have an implemementation for `RdbmsStreamDefinitionRepository`
- Unit tests included
- Documentation updated

Issue #548, opened 2016-04-22
Restart Task from Shell


ghillert:
For Job Execution Restarts we basically already support task restarts.

sabbyanandan:
This depends on spring-cloud/spring-cloud-dataflow#523

sabbyanandan:
We have the option to restart the batch-job from the Dashboard, but not from the Shell yet.

PR #634, opened 2016-06-01
Add RDBMS based repositories

### LINKED ISSUE #617: Generalize RDBMS repository infrastructure to support both streams and tasks
As a developer, I'd like to refactor Task repository infrastructure to support both streams and tasks.

ilayaperumalg:
- Refactor the current code to re-use the RDBMS operations
 - Add RDBMS StreamDefinition repository
 - Add RDBMS DeploymentId repository
 - Add tests

This resolves #617, #606, #618


spring-cloud-dataflow-server-local/src/main/resources/application.yml
```diff
@@ -10,7 +10,7 @@ spring:
       registry:
         populator:
           locations: classpath:META-INF/${binder}-apps.properties
-
+serverStore: redis
```
cppwfs: Just curious why are we defaulting to redis as a serverStore if we have an embedded h2 that we can use.   Granted its still in-memory, but that means that a user could technically start a DF Server without a Redis instance running (unless they wanted to do analytics.)  Thoughts?

markpollack: I'd actually be in favor of removing redis completely for this purpose.

ilayaperumalg: Yeah, I agree. I thought of having this as temporary until we move away from `redis`. But, that is not necessary as we can rely on H2 and use the default. I will update the PR to move the `serverStore` property completely. Also, we need to decide whether we really need the in-memory implementation of these repositories as they are never used.

markpollack: Maybe we move redis or memory into the testing directory so we can at least
test but not ship multiple impls.
M

On Jun 1, 2016, at 8:59 AM, Ilayaperumal Gopinathan <
notifications@github.com> wrote:

In spring-cloud-dataflow-server-local/src/main/resources/application.yml
<https://github.com/spring-cloud/spring-cloud-dataflow/pull/634#discussion_r65354656>
:

> @@ -10,7 +10,7 @@ spring:
>        registry:
>          populator:
>            locations: classpath:META-INF/${binder}-apps.properties
> -
> +serverStore: redis

Yeah, I agree. I thought of having this as temporary until we move away
from redis. But, that is not necessary as we can rely on H2 and use the
default. I will update the PR to move the serverStore property completely.
Also, we need to decide whether we really need the in-memory implementation
of these repositories as they are never used.

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/spring-cloud/spring-cloud-dataflow/pull/634/files/2e189a019d922508760c6ac242e322271b013636#r65354656>,
or mute the thread
<https://github.com/notifications/unsubscribe/AAPGqmzlo8bG8_Y9DGUfR-Aez_50s_DMks5qHYIYgaJpZM4IrffL>
.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfiguration.java
```diff
@@ -162,12 +164,28 @@ public Object postProcessAfterInitialization(Object bean, String s) throws Beans
...
+
+	@Bean
+	@ConditionalOnExpression("#{'${serverStore:redis}'.equals('db')}")
+	public DeploymentIdRepository rdbmsDeploymentIdRepository(DataSource dataSource) {
+		return new RdbmsDeploymentIdRepository(dataSource);
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	@ConditionalOnExpression("#{'${serverStore:redis}'.equals('redis')}")
```
cppwfs: if we are keeping redis as a repository the method name should be redisStreamDefinitionRepository.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfiguration.java
```diff
@@ -162,12 +164,28 @@ public Object postProcessAfterInitialization(Object bean, String s) throws Beans
 
 	@Bean
 	@ConditionalOnMissingBean
+	@ConditionalOnExpression("#{'${serverStore:redis}'.equals('db')}")
```
cppwfs: Copy paste error?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfiguration.java
```diff
@@ -162,12 +164,28 @@ public Object postProcessAfterInitialization(Object bean, String s) throws Beans
...
+	public StreamDefinitionRepository rdbmsStreamDefinitionRepository(DataSource dataSource) {
+		return new RdbmsStreamDefinitionRepository(dataSource);
+	}
+
+	@Bean
+	@ConditionalOnExpression("#{'${serverStore:redis}'.equals('db')}")
```
cppwfs: Copy paste error?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfiguration.java
ilayaperumalg: No, this wasn't a copy-paste error ;) I was just providing the default value for `serverStore`. Eventually, this property itself will be removed.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/RdbmsStreamDefinitionRepository.java
```diff
@@ -0,0 +1,65 @@
...
+import org.springframework.util.Assert;
+
+/**
+ * RDBMS implementation of {@link StreamDefinitionRepository}.
+ *
+ * @author Ilayaperumal Gopinathan
+ */
+public class RdbmsStreamDefinitionRepository extends AbstractRdbmsKeyValueRepository<StreamDefinition> implements StreamDefinitionRepository {
+
+	public RdbmsStreamDefinitionRepository(DataSource dataSource) {
```
cppwfs: probably want to have the user to pass in the prefix and suffix.  As they may already have tables with these names.  And if they don't use the defaults.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/RdbmsTaskDefinitionRepository.java
```diff
@@ -19,313 +19,48 @@
...
-		Assert.notNull(sort, "sort must not be null");
-		Iterator<Sort.Order> iter = sort.iterator();
-		String query = findAllDefinitionsQuery  + "ORDER BY " ;
-
-		while (iter.hasNext()) {
-			Sort.Order order = iter.next();
-			query = query + order.getProperty() + " " + order.getDirection();
-			if (iter.hasNext()) {
-				query = query + ", ";
+		super(dataSource, "TASK_", "DEFINITIONS", new RowMapper<TaskDefinition>() {
```
cppwfs: probably want to have the user to pass in the prefix and suffix.  As they may already have tables with these names.  And if they don't use the defaults.

spring-cloud-dataflow-server-core/src/main/resources/schema-hsqldb.sql
```diff
@@ -1,6 +1,15 @@
...
 
+CREATE TABLE STREAM_DEFINITIONS  (
+	DEFINITION_NAME VARCHAR(255) NOT NULL PRIMARY KEY,
+	DEFINITION CLOB DEFAULT NULL
+);
+
+CREATE TABLE DEPLOYMENT_IDS  (
+	DEPLOYMENT_KEY VARCHAR(255) NOT NULL PRIMARY KEY,
+	DEPLOYMENT_ID VARCHAR(255) NOT NULL
+);
```
cppwfs: Extra line.  (hate these things ;-) )

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/repository/RdbmsStreamDefinitionRepositoryTests.java
```diff
@@ -0,0 +1,320 @@
...
+	}
+	
+	@Test
+	public void testFindOne() {
+		StreamDefinition definition = new StreamDefinition("stream1", "time | log");
+		repository.save(definition);
+		repository.save(new StreamDefinition("stream2", "http | jdbc"));
+		repository.save(new StreamDefinition("stream3", "twitterstream | hdfs"));
+
+		assertEquals(definition, repository.findOne("stream1"));
```
cppwfs: probably need to check for all  stream definitions

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfiguration.java
cppwfs: I was thinking that we'd want db as default.   I'm such a trouble maker ;-)

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/RdbmsTaskDefinitionRepository.java
ilayaperumalg: I was thinking the reverse as we can specify the `prefix` and `suffix` for the table here in the repository class directly rather than specifying where they are instantiated. But, I am ok with both the ways.

cppwfs: sweeeet ;-)

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfiguration.java
trisberg: should it be 'rdbms' instead of 'db'? To me 'db' could mean MongoDB, Cassandra or some other database. Also, shouldn't this be the default?

ilayaperumalg: +1. Eventually we will remove this property itself :-)

spring-cloud-dataflow-server-core/src/main/resources/schema-hsqldb.sql
trisberg: How can we re-launch a task? I get a duplicate key error launching the task a second time:

```
Command failed org.springframework.cloud.dataflow.rest.client.DataFlowClientException: PreparedStatementCallback; SQL [INSERT into DEPLOYMENT_IDS (DEPLOYMENT_KEY, DEPLOYMENT_ID)values (?, ?)]; Duplicate entry 'timestamp.test3' for key 'PRIMARY'; nested exception is java.sql.SQLIntegrityConstraintViolationException: Duplicate entry 'timestamp.test3' for key 'PRIMARY'
```

sabbyanandan: I don't think we have relaunching supported yet. See here: https://github.com/spring-cloud/spring-cloud-dataflow/issues/548

ilayaperumalg:
Addressed review comments.

spring-cloud-dataflow-server-local/src/test/java/org/springframework/cloud/dataflow/server/local/LocalConfigurationTests.java
```diff
@@ -20,6 +20,7 @@
 import static org.hamcrest.Matchers.is;
 import static org.junit.Assert.assertThat;
 
+import org.junit.Ignore;
```
cppwfs: Will remove junit.Ignore during merge.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/repository/RdbmsDeploymentIdRepositoryTests.java
```diff
@@ -28,6 +28,7 @@
 
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.PropertyPlaceholderAutoConfiguration;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
```
cppwfs: Will remove unused imports on merge.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfiguration.java
```diff
@@ -82,6 +82,7 @@
 import org.springframework.context.annotation.ComponentScan;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
+import org.springframework.context.annotation.Lazy;
```
cppwfs: Will cleanup imports on merge

spring-cloud-dataflow-server-core/src/main/resources/schema-hsqldb.sql
cppwfs: Well... I think there is a problem.  Before we could do the following:
task create foo --definition "timestamp" 
task launch foo
task launch foo
Now we get an exception on the 2nd launch.  
We can put a try catch here: https://github.com/spring-cloud/spring-cloud-dataflow/blob/e03a2db23d67190ff95d2702e1dc911e18761c4e/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java#L166   
to ignore that exception.

ilayaperumalg: Yeah, I see. I think we can check if the deploymentId exists before saving. That way we can be sure of it.

ilayaperumalg:
Pushed the changes that fixes repeated task launch as well as organising package imports.

cppwfs:
LGTM

cppwfs:
Squashed, Merged & rebased

Issue #641, opened 2016-06-01
Support features endpoint to disable shell commands


markpollack:
See https://github.com/spring-cloud/spring-cloud-dataflow/issues/640

If the task support is disabled, the shell should not show task/job related commands.

PR #663, opened 2016-06-09
Feature toggle for stream, task and metric endpoints

### LINKED ISSUE #640: Create feature toggle for analytics, stream, and task endpoints
As a developer, I would like to individually enable/disable the analytics, stream, and job controller features.  This will enable users to 

* Avoid dependency on Redis if there is no need to use analytics functionality (common for batch centric use-case deployments)
* Not expose additional endpoints if they are not to be used, e.g. current spring-batch-admin users that would like to use SCDF's UI/endpoints and not be exposed to any stream API/concepts.  Similar case for stream only users.

The features would be defined by the config
```
spring:
  cloud:
    dataflow:
      features:
        analyticsEnabled: true
        streamsEnabled:  true
        tasksEnabled: true
```

by default.  Tasks support includes job support.

A `features/` endpoint will need to be created to will indicate which features are enabled/disabled so the UI/Shell can take the appropriate actions.  There is similar information for security already exposed under `/security/info` - we should consolidate the security information to be serves up under the `features` endpoint as well even though security is not configured under the features yaml section.

There will be related issues to handle this on a per-server basis and for the UI/Shell and also for PCF and Mesos servers so that task support is overridden to be false.

Only when enabling tasks or streams would corresponding schemas be created in the database, otherwise no schemas would be created.

ilayaperumalg:
- Refactor DataflowServerConfiguration and split specific configuration classes that can be conditionally loaded based on the features
 - Add FeaturesProperties that provides flags for streams, tasks and analytics
 - When disabled, the appropriate repositories, controllers and database schema are disabled as well
 - Add separate rdbms schema which get loaded conditionally based on what is enabled
 - Rename dataflow rdbms initializer property from `spring.cloud.dataflow.server.initialize.enable` to `spring.cloud.dataflow.rdbms.initialize.enable`
 - Add `/features/info` endpoint to display features properties
 - Add tests
 - Intiailize the REST endpoints conditionally based on the features properties
 - Fix Shell to conditionally initialize the required dataflow shell operations

This resolves #640


spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/DataFlowTemplate.java
```diff
@@ -51,62 +51,67 @@
 	/**
 	 * REST client for stream operations.
 	 */
-	private final StreamOperations streamOperations;
```
ericbottard: Why remove `final`? What happens if none of the feature flags are enabled? Seems those are left `null`, so why not make that an explicit `else` branch (if that's really what expected downstream)?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowControllerAutoConfiguration.java
```diff
@@ -113,17 +119,12 @@ public RootController rootController(EntityLinks entityLinks) {
...
-		return new RuntimeAppsController(repository, deploymentIdRepository, appDeployer);
-	}
-
-	@Bean
 	public AppInstanceController appInstanceController(AppDeployer appDeployer) {
 		return new AppInstanceController(appDeployer);
 	}
 
 	@Bean
+	@ConditionalOnBean(StreamDefinitionRepository.class)
```
ericbottard: If those can only work with some collaborators (which are driven by feature flags) then, in some cases at least, doesn't it make sense to declare in that same configuration class (that is driven by the flag)?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/FeaturesConfiguration.java
```diff
@@ -0,0 +1,47 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.cloud.dataflow.server.repository.DeploymentIdRepository;
+import org.springframework.cloud.dataflow.server.repository.RdbmsDeploymentIdRepository;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Import;
+
+/**
+ * Configuration class that conditionally imports stream, task and anaytics configuration classes based on the
```
ericbottard: This doesn't seem to do what the javadoc claims.
Seems like it only conditionally creates a single bean (DeploymentIdRepository), for which we don't really understand why it's treated as a first class citizen in a class named FeaturesConfiguration

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/FeaturesProperties.java
```diff
@@ -0,0 +1,57 @@
...
+package org.springframework.cloud.dataflow.server.config.features;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * Configuration properties for all the features that need to be enabled/disabled at the dataflow server.
+ *
+ * @author Ilayaperumal Gopinathan
+ */
+@ConfigurationProperties(prefix = "spring.cloud.dataflow.features")
```
ericbottard: This is really a boot problem in a way, but given that that prefix is used both a `@ConfigurationProperties` prefix and directly in some SpEL expressions, it should be made a constant somehow

spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/DataFlowTemplate.java
ilayaperumalg: I am fine with both ways. In fact, there is some more work to be done with #641 along with these things.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/RootController.java
```diff
@@ -74,58 +81,48 @@ public RootController(EntityLinks entityLinks) {
...
-		resourceSupport.add(entityLinks.linkToCollectionResource(AppStatusResource.class).withRel("runtime/apps"));
-		resourceSupport.add(unescapeTemplateVariables(entityLinks.linkForSingleResource(AppStatusResource.class, "{appId}").withRel("runtime/apps/app")));
-
-		resourceSupport.add(unescapeTemplateVariables(entityLinks.linkFor(AppInstanceStatusResource.class, UriComponents.UriTemplateVariables.SKIP_VALUE).withRel("runtime/apps/instances")));
-
-		templated = entityLinks.linkFor(CompletionProposalsResource.class).withSelfRel().getHref() + ("/stream{?start,detailLevel}");
-		resourceSupport.add(new Link(templated).withRel("completions/stream"));
-
+		String completionStreamTemplated = entityLinks.linkFor(CompletionProposalsResource.class).withSelfRel().getHref() + ("/stream{?start,detailLevel}");
+		resourceSupport.add(new Link(completionStreamTemplated).withRel("completions/stream"));
```
ericbottard: Shouldn't this be conditional on stream support?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowControllerAutoConfiguration.java
ilayaperumalg: Currently, this configuration class acts as `AutoConfiguration` class to instantiate all the REST controllers in a single place. Moving these into separate configuration class might still work but it might be too confusing at this point I think.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/FeaturesConfiguration.java
ilayaperumalg: The `DeploymentIdRepository` bean gets created based on the stream/task enablement flags and require a common place. That's why this bean is in this `FeaturesConfiguration` class.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/RootController.java
ilayaperumalg: I thought about it. But I don't see any dependencies from Stream for this completion provider. Moreover, I think this is needed for `tasks` commands as well. Hence, I kept it as common.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/FeaturesConfiguration.java
ericbottard: Still, this is not what the javadoc says

ilayaperumalg:
Updated the PR after addressing the review comments.

ericbottard:
Rebased, squashed and merged as baa8f3e

PR #672, opened 2016-06-10
Add configuration props whitelisting

ericbottard:
Fixes spring-cloud/spring-cloud-dataflow#273


ilayaperumalg:
@ericbottard Could you please add some commit description so that it would help reviewing the code better?

ericbottard:
Please don't look at individual commits, this is really a PR that emerged from many many back and forths on previous code. I'll send you a link to an updated google doc

spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/ApplicationConfigurationMetadataResolver.java
```diff
@@ -32,50 +37,95 @@
...
 /**
  * Used to retrieve metadata about the configuration properties that can alter an application's behavior.
- *
  * @author Eric Bottard
  */
 public class ApplicationConfigurationMetadataResolver {
 
+	private static final String CONFIGURATION_METADATA_PATTERN = "classpath*:/META-INF/spring-configuration-metadata.json";
+
+	private static final String DATAFLOW_PROPERTIES = "classpath*:/META-INF/dataflow.properties";
```
markpollack: should be something like `configuration-whitelist.properties`

spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/ApplicationConfigurationMetadataResolver.java
```diff
@@ -32,50 +37,95 @@
...
+
+			Collection<String> whiteListedClasses = new HashSet<>(globalWhiteListedClasses);
+			Collection<String> whiteListedProperties = new HashSet<>(globalWhiteListedProperties);
+
+			loadWhiteLists(moduleResourceLoader.getResources(DATAFLOW_PROPERTIES), whiteListedClasses, whiteListedProperties);
+
 			ConfigurationMetadataRepositoryJsonBuilder builder = ConfigurationMetadataRepositoryJsonBuilder.create();
-			ResourcePatternResolver applicationResourceLoader = new PathMatchingResourcePatternResolver(classLoader);
-			for (Resource r : applicationResourceLoader.getResources("classpath*:/META-INF/*spring-configuration-metadata.json")) {
+			for (Resource r : moduleResourceLoader.getResources(CONFIGURATION_METADATA_PATTERN)) {
```
markpollack: I was trying to use `additional-spring-configuration-metadata.json` to override the advertised default value for maxMessages (it needs to be 1 or -1 depending on the source) so that change was not picked up of course.  Is there some boot class we can delegate to that understands how to do the merge of these two categories of file?

ericbottard: The merge is done at annotation processing time, but I'm not sure one can *override* a property (only add new "manual" ones)

spring-cloud-dataflow-completion/src/test/resources/org/springframework/cloud/dataflow/completion/apps/log-sink/META-INF/dataflow.properties
```diff
@@ -0,0 +1 @@
+primary.prefix=spring.cloud.stream.app
```
markpollack: Is this left over from something before?

spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/ApplicationConfigurationMetadataResolver.java
markpollack: going with 
```
	private static final String WHITELIST_PROPERTIES = "classpath*:/META-INF/spring-configuration-metadata-whitelist.properties";

	public static final String CONFIGURATION_PROPERTIES_CLASSES = "configuration-properties.classes";

	public static final String CONFIGURATION_PROPERTIES_NAMES = "configuration-properties.names";
```

spring-cloud-dataflow-configuration-metadata/src/main/resources/META-INF/dataflow.properties
```diff
@@ -0,0 +1 @@
+configuration.properties = server.port
```
markpollack: This ends up being a global default.... I don't think we want to have `server.port` listed for every app.

spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/ApplicationConfigurationMetadataResolver.java
```diff
@@ -32,50 +37,95 @@
...
+
+	private static final String DATAFLOW_PROPERTIES = "classpath*:/META-INF/dataflow.properties";
+
+	public static final String CONFIGURATION_CLASSES = "configuration.classes";
+
+	public static final String CONFIGURATION_PROPERTIES = "configuration.properties";
+
+	private final Set<String> globalWhiteListedProperties = new HashSet<>();
+
+	private final Set<String> globalWhiteListedClasses = new HashSet<>();
```
markpollack: I don't see a way to set the values of these global whitelist fields.

ericbottard: Hey, thanks for merging this!
The properties file is read from `classpath*` and each file property is merged with the previous. So, to add a global whitelist, we just put a copy of the properties file in dataflow itself (it is read in the resolver constructor using a classloader that "sees" the dataflow classes, as opposed to a classloader that sees an app later). Hope that makes sense

markpollack:
Squashed commits and changed name of property file and keys.  Merged as 985e038d3b8f58da2dfab4d569cdddf3ce9a9475

PR #674, opened 2016-06-10
Remove TaskDefinitionRepositoryFactoryBean

### LINKED ISSUE #673: Remove TaskDefinitionRepositoryFactoryBean
Given we only use `rdbms` based task definition repository, there is no need to have factory bean as we can instantiate the `RDBMSTaskDefinitionRepository` directly.

ilayaperumalg:
- Use `RdbmsTaskDefinitionFactory` directly

This resolves #673


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/TaskConfiguration.java
```diff
@@ -126,7 +125,7 @@ public TaskRepositoryInitializer taskRepositoryInitializerForDefaultDB(DataSourc
 		@Bean
 		@ConditionalOnMissingBean
 		public TaskDefinitionRepository taskDefinitionRepository(DataSource dataSource, Server server) throws Exception {
```
markfisher: was already the case, but noticing now: the `Server` parameter is unused here

markfisher: or if this is intentional, because this bean requires that the `Server` has been instantiated already, that intention would be clearer via `@DependsOn`... and that would make it less likely that someone deletes it at some point ;)

ilayaperumalg: yeah, sure. Created https://github.com/spring-cloud/spring-cloud-dataflow/issues/675 for this.

markfisher: I'll go ahead and merge this PR since #675 was created

markfisher:
rebased and merged as https://github.com/spring-cloud/spring-cloud-dataflow/commit/34eab1bcd5abc2bd6656c842847a486b8302f14d

PR #676, opened 2016-06-10
Use DependsOn to depend on H2 server bean

### LINKED ISSUE #675: Use DependsOn annotation for H2 server dependency
Currently, H2 server dependency on `DataflowServerConfiguration` uses unused method argument to denote as a dependency bean. 

@markfisher suggested to use `@DependsOn` instead: https://github.com/spring-cloud/spring-cloud-dataflow/pull/674#discussion_r66619359

ilayaperumalg:
- Instead of using unused method argument, we can use `DependsOn`

Resolves #675


markfisher:
rebased and merged as https://github.com/spring-cloud/spring-cloud-dataflow/commit/7188a69ce13186829afd2510dc52a5a9c31206ac

PR #677, opened 2016-06-10
Move Redis/in-memory repositories into test package

### LINKED ISSUE #647: Move redis/in-memory definition repositories as test resources
Once all definition/deployment-id repositories are used as RDBMS based ones, we can move redis, in-memory repositories as test resources only.

ilayaperumalg:
- Since Redis and in-memory repository implementations are not used by the core classes, we can move these classes into test package
 - Also deprecate RedisUriRegistry

Resolves #647


markfisher:
LGTM; merged via https://github.com/spring-cloud/spring-cloud-dataflow/commit/97e198cf32a9a88bb4bf04f07c0f19de24486fcf

PR #684, opened 2016-06-12
Inlcude section links in docs

sabbyanandan:
This is a prerequisite to explicitly include (section) docs in each SCDF's server reference guides. 

With this change, we can now include sections (_eg: in SCDF's CF Server ref. guide_) like the following:

> include::https://raw.githubusercontent.com/spring-cloud/spring-cloud-dataflow/master/spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc#spring-cloud-dataflow-stream-intro[]


ilayaperumalg:
LGTM, cherry-picked and merged as 854db43b8bd6dca8d13b1669ffe92355563bebc9

Issue #661, opened 2016-06-08
Provide server configuration to pass in env-vars/application properties to passed to all deployed application runtimes


markpollack:
Something like
```
spring:
  cloud:
    dataflow:
      environmentVariables:
        SPRING_CLOUD_CONFIG_URI: http://localhost:8888
```
Would enable all apps deployed by the Data Flow server to have the config server uri set as an environment variable.
 
See https://github.com/spring-cloud/spring-cloud-deployer/issues/70

markpollack:
Can alternatively pass in properties that get used as application props (works well for boot apps).

mbogoevici:
Yes, I think that initially we should just be able to pass normal application properties (i.e. as a global, server-level, `app.*`).

PR #688, opened 2016-06-13
GH-661 Add support for passing global properties to applications

mbogoevici:
Resolves spring-cloud/spring-cloud-dataflow#661

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
@@ -364,3 +364,23 @@ This can be done using the DSL syntax `http > :mydestination` or `:mydestination
 Second, you may need to determine the output channel of a stream based on some information that is only known at runtime.
 In that case, a router may be used in the sink position of a stream definition. For more information, refer to the Router Sink starter's
 link:https://github.com/spring-cloud/spring-cloud-stream-app-starters/tree/master/router/spring-cloud-starter-stream-sink-router[README].
+
+[[spring-cloud-dataflow-global-properties]]
+=== Common application properties
+
+In addition to configuration via DSL, Spring Cloud Data Flow provides a mechanism for setting common properties to all the streaming applications that are launched by it.
+This can be done by adding properties prefixed with `spring.cloud.dataflow.applicationProperties.stream` when starting the server.
+When doing so, the server will pass all the properties, without the prefix, to the instances it launches.
+
+For example, all the launched applications can be configured to use a specific Kafka broker by launching the configuration server with the following options:
+
+```
+--spring.cloud.dataflow.applicationProperties.stream.spring.cloud.stream.kafka.binder.brokers=192.168.1.100:9092
+--spring.cloud.dataflow.applicationProperties.stream.spring.cloud.stream.kafka.binder.zkNodes=192.168.1.100:2181
+```.
+
+This will cause the properties `stream.spring.cloud.stream.kafka.binder.brokers` and `spring.cloud.stream.kafka.binder.zkNodes` to be passed to all the launched applications.
+
+[NOTE]
+Properties configured using this mechanism have lower precedence than stream deployment properties.
+They will be overridden if a property with the same key is specified at stream deployment time (e.g. `app.http.spring.cloud.stream.kafka.binder.brokers` will override the common property).
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/apps/CommonApplicationProperties.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/apps/CommonApplicationProperties.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.dataflow.server.config.apps;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * Common properties for applications deployed via Spring Cloud Data Flow.
+ *
+ * @author Marius Bogoevici
+ */
+@ConfigurationProperties("spring.cloud.dataflow.applicationProperties")
+public class CommonApplicationProperties {
+
+	private Map<String,String> stream = new ConcurrentHashMap<>();
+
+	public Map<String, String> getStream() {
+		return this.stream;
+	}
+
+	public void setStream(Map<String, String> stream) {
+		this.stream = stream;
+	}
+}
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
@@ -30,6 +30,7 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
 import org.springframework.boot.configurationmetadata.ConfigurationMetadataProperty;
+import org.springframework.cloud.dataflow.configuration.metadata.ApplicationConfigurationMetadataResolver;
 import org.springframework.cloud.dataflow.core.ApplicationType;
 import org.springframework.cloud.dataflow.core.BindingPropertyKeys;
 import org.springframework.cloud.dataflow.core.StreamAppDefinition;
@@ -40,6 +41,7 @@ import org.springframework.cloud.dataflow.registry.AppRegistry;
 import org.springframework.cloud.dataflow.rest.resource.StreamDeploymentResource;
 import org.springframework.cloud.dataflow.rest.util.DeploymentPropertiesUtils;
 import org.springframework.cloud.dataflow.server.DataFlowServerUtil;
+import org.springframework.cloud.dataflow.server.config.apps.CommonApplicationProperties;
 import org.springframework.cloud.dataflow.server.repository.DeploymentIdRepository;
 import org.springframework.cloud.dataflow.server.repository.DeploymentKey;
 import org.springframework.cloud.dataflow.server.repository.NoSuchStreamDefinitionException;
@@ -48,7 +50,6 @@ import org.springframework.cloud.deployer.spi.app.AppDeployer;
 import org.springframework.cloud.deployer.spi.app.AppStatus;
 import org.springframework.cloud.deployer.spi.app.DeploymentState;
 import org.springframework.cloud.deployer.spi.core.AppDeploymentRequest;
-import org.springframework.cloud.dataflow.configuration.metadata.ApplicationConfigurationMetadataResolver;
 import org.springframework.core.io.Resource;
 import org.springframework.hateoas.ExposesResourceFor;
 import org.springframework.http.HttpStatus;
@@ -103,6 +104,11 @@ public class StreamDeploymentController {
 	private final ApplicationConfigurationMetadataResolver metadataResolver;
 
 	/**
+	 * General properties to be applied to applications on deployment.
+	 */
+	private final CommonApplicationProperties commonApplicationProperties;
+
+	/**
 	 * Create a {@code StreamDeploymentController} that delegates
 	 * <ul>
 	 * <li>CRUD operations to the provided {@link StreamDefinitionRepository}</li>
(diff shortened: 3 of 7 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
```diff
@@ -103,19 +104,27 @@
...
+	private final CommonApplicationProperties appsProperties;
+
+	/**
 	 * Create a {@code StreamDeploymentController} that delegates
 	 * <ul>
 	 * <li>CRUD operations to the provided {@link StreamDefinitionRepository}</li>
 	 * <li>app retrieval to the provided {@link AppRegistry}</li>
 	 * <li>deployment operations to the provided {@link AppDeployer}</li>
 	 * </ul>
+<<<<<<< HEAD
```
ilayaperumalg: This needs to go ;)

mbogoevici: Snap :(

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/configuration/TestDependencies.java
```diff
@@ -65,25 +68,30 @@
 @EnableHypermediaSupport(type = HAL)
 @Import(CompletionConfiguration.class)
 @EnableWebMvc
+@EnableConfigurationProperties(CommonApplicationProperties.class)
 public class TestDependencies extends WebMvcConfigurationSupport {
 
 	private final MavenProperties mavenProperties = new MavenProperties();
 
+	@Autowired
```
ilayaperumalg: Not sure what is the reasoning behind `autowiring` here. May be we just want to initialize the `CommonApplicationProperties` without having to do `@EnableConfigurationProperties`?

mbogoevici: Actually. It's the initialization that doesn't make sense. The test modifies the properties bean.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/apps/CommonApplicationProperties.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * Common properties for applications deployed via Spring Cloud Data Flow.
+ *
+ * @author Marius Bogoevici
+ */
+@ConfigurationProperties("spring.cloud.dataflow.applicationProperties")
```
ilayaperumalg: We need to document this and created one for this: https://github.com/spring-cloud/spring-cloud-dataflow/issues/690

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
ilayaperumalg: will fix while merging.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
```diff
@@ -300,24 +310,20 @@ private StreamAppDefinition qualifyParameters(StreamAppDefinition original, Reso
 		String wildCardConsumerPropertyPrefix = "app.*.consumer.";
 		String wildCardPrefix = "app.*.";
 		// first check for wild card prefix
```
ilayaperumalg: This comment needs to be placed next line.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/StreamControllerTests.java
```diff
@@ -572,6 +580,40 @@ public void testDeployWithWildcardProperties() throws Exception {
 	}
 
 	@Test
+	public void testDeployWithCommonApplicationProperties() throws Exception {
+		repository.save(new StreamDefinition("myStream", "time | log"));
+		assertThat(appsProperties.getStream().values(), empty());
+		appsProperties.getStream().put("spring.cloud.stream.fake.binder.host","fakeHost");
```
mbogoevici: On second thought, we shouldn't do this, as there may be multiple tests modifying the state. Ideally, we should create a new deployer with common properties initialized.

mbogoevici: On the other hand, the tests already change the state of the repository, registering different stream definitions, so they couldn't normally run in parallel anyway. So probably this could be left like this.

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -233,6 +233,21 @@ configuration.classes=org.springframework.cloud.stream.app.file.sink.FileSinkPro
 configuration-properties.names=server.port
 ```
 
+=== Global application properties (for streams)
```
mbogoevici: could ditch the paranthesis

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -233,6 +233,21 @@ configuration.classes=org.springframework.cloud.stream.app.file.sink.FileSinkPro
...
+
+In addition to configuration via DSL, Spring Cloud Data Flow provides a mechanism for setting specific properties to all the streaming applications that are launched by it.
+This can be done by adding properties prefixed with `spring.cloud.dataflow.applicationProperties.stream` when starting the server.
+When doing so, the server will pass all the properties, without the prefix, to the instances it launches.
+
+For example, all the launched applications can be reconfigured to use a different Kafka broker by the default by launching the configuration server with the following options:
```
ilayaperumalg: @mbogoevici Given these common application properties are for the dataflow server and launching the `config` server with the new properties would cause the refresh on the dataflow server configuration only. Is that right? In that case, how can we expect the deployed apps to be refreshed with these properties. what am I missing?

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/StreamControllerTests.java
markfisher: Maybe just reinitialize `appsProperties` in a `@Before` method?

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
mbogoevici: I don't think I mentioned refresh anywhere. We can simply replace `reconfigured` with `configured` and it'll make more sense, I guess?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/apps/CommonApplicationProperties.java
```diff
@@ -0,0 +1,41 @@
...
+ * Common properties for applications deployed via Spring Cloud Data Flow.
+ *
+ * @author Marius Bogoevici
+ */
+@ConfigurationProperties("")
+public class CommonApplicationProperties {
+
+	private Map<String,String> stream = new ConcurrentHashMap<>();
+
+	public Map<String, String> getStream() {
```
markfisher: I presume the plan is to add `task` as well?

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
mbogoevici: e.g. "For example, all the launched applications can be configured to use a specific Kafka broker by launching the configuration server with the following options:"

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -233,6 +233,21 @@ configuration.classes=org.springframework.cloud.stream.app.file.sink.FileSinkPro
...
+
+```
+--spring.cloud.dataflow.applicationProperties.stream.spring.cloud.stream.kafka.binder.brokers=192.168.1.100:9092
+--spring.cloud.dataflow.applicationProperties.stream.spring.cloud.stream.kafka.binder.zkNodes=192.168.1.100:2181
+```.
+
+This will cause the properties `stream.spring.cloud.stream.kafka.binder.brokers` and `spring.cloud.stream.kafka.binder.zkNodes` to be passed to all the launched applications.
```
markfisher: Should mention here that IF the same property is provided explicitly with the `app.<name>.` prefix, that value would take precedence. And, on that note, would we possibly want a boolean flag to control that precedence? (thinking of the case where config-server might be used, and overriding explicit property values might be desirable).

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/apps/CommonApplicationProperties.java
mbogoevici: Eventually, but I'm not overly familiar with how task properties are set. I suppose that there are properties that are common to both, too.

markfisher: we should make sure we're ok with the prefix namespaces that we'll support (especially if there is a group of properties that would apply to either stream or task apps) cc @mminella

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/StreamControllerTests.java
mbogoevici: Not a concern if we leave `@EnableConfigurationProperties` in, since the context is marked dirty after each method, I guess?

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
markfisher: this does raise an interesting point... the apps themselves could be reading properties from `config-server` (not just indirectly via dataflow deploying the apps), including support for refreshable properties if implemented as such by the stream app developer

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
mbogoevici: Config server would aways take precedence, IIRC.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/StreamControllerTests.java
markfisher: right... it would be autowired each time due to `@DirtiesContext` being present on this test class

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
markfisher: but here the "common" properties are being set first, and then the `app.<name>.` properties are set next... thus overriding, right?: https://github.com/spring-cloud/spring-cloud-dataflow/pull/688/files#diff-e6925c2ec0e82e9d426e01d83d3051b1R313

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
mbogoevici: Sure, but that's a different concern (and we'd need to add spring-cloud-bus to starter apps as well). In this case it was just infortunate that the initial phrasing could be somehow misconstrued.

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
mbogoevici: Yes, it helps to explain that the "common" properties are lowest in precedence among the ones set explicitly (which is to be expected) - I think that's intuitive enough, wdyt? - e.g. "common" -> "app.*" -> "app.name".

What I meant was that config server is orthogonal and typically takes precedence due to the priority of its property source.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/apps/CommonApplicationProperties.java
mminella: I'm not sure I see how this applies on the task side.  Since a task is a singular thing from the perspective of SCDF, global properties would be the same as regular properties.  Tasks do have the ability to launch other tasks, but that's a function of the task itself and not of SCDF.

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
markfisher: but do we foresee any case where it might be desirable for the common properties to take precedence even over those provided via `stream deploy` ?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/apps/CommonApplicationProperties.java
markfisher: @mminella I believe this could still be relevant for tasks just as a convenience... i.e. not having to provide common properties every time a task is launched

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
mbogoevici: Honestly, I don't believe so - and certainly can't see a case where that would be controlled by yet another global setting. I believe that the rule that "the more general the property, the lower precedence it has would work just fine" - it's a simple and quite intuitive rule.

markfisher: ok

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/apps/CommonApplicationProperties.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * Common properties for applications deployed via Spring Cloud Data Flow.
+ *
+ * @author Marius Bogoevici
+ */
+@ConfigurationProperties("")
```
mbogoevici: Ugh. This got somehow deleted.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/apps/CommonApplicationProperties.java
mbogoevici: Documentation pushed as part of the PR.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
```diff
@@ -126,6 +134,7 @@ public StreamDeploymentController(StreamDefinitionRepository repository, Deploym
 		this.registry = registry;
 		this.deployer = deployer;
 		this.metadataResolver = metadataResolver;
+		this.appsProperties = commonProperties;
```
markfisher: we should assert this is not null, since otherwise we could hit an NPE on line 309

markfisher:
@mbogoevici can you please rebase this PR? (and either you or I, while merging, can add that one `assertNotNull`)... there are conflicts due to recent doc updates

mbogoevici:
Squashed, rebased on top of master. Addressed comments.

markfisher:
LGTM; merged

Issue #637, opened 2016-06-01
Add hint to indicate the requirement to register applications


sabbyanandan:
As a developer, I'd like to add a hint/warning message in shell/ui, so when the `app list` is empty, users will have the indication to register apps.

**Acceptance:**
- empty `app list` results with the message: `Empty application registry detected, please register applications`
- similar experience in UI when the `Apps` tab is empty

ericbottard:
I think a similar issue needs to be opened against the UI repo, wdyt @ghillert ?

PR #722, opened 2016-06-21
Add hint to 'app list' when no apps registered

ericbottard:
Fixes spring-cloud/spring-cloud-dataflow#637

### DIFF
--- a/spring-cloud-dataflow-shell/src/main/java/org/springframework/cloud/dataflow/shell/command/AppRegistryCommands.java
+++ b/spring-cloud-dataflow-shell/src/main/java/org/springframework/cloud/dataflow/shell/command/AppRegistryCommands.java
@@ -27,8 +27,8 @@ import org.springframework.boot.configurationmetadata.ConfigurationMetadataPrope
 import org.springframework.cloud.dataflow.core.ApplicationType;
 import org.springframework.cloud.dataflow.rest.client.AppRegistryOperations;
 import org.springframework.cloud.dataflow.rest.client.DataFlowOperations;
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-shell/src/main/java/org/springframework/cloud/dataflow/shell/command/AppRegistryCommands.java
```diff
@@ -192,6 +191,12 @@ public Table list() {
 			column.add(appRegistration.getName());
 			max = Math.max(max, column.size());
 		}
+		if (max == 0) {
+			return String.format("There are no registered apps at the moment.%n" +
+					"You can register new apps with the '%s' command.", REGISTER_APPLICATION);
```
markfisher: should mention `app import` as well?

spring-cloud-dataflow-shell/src/main/java/org/springframework/cloud/dataflow/shell/command/AppRegistryCommands.java
```diff
@@ -192,6 +191,12 @@ public Table list() {
 			column.add(appRegistration.getName());
 			max = Math.max(max, column.size());
 		}
+		if (max == 0) {
+			return String.format("There are no registered apps at the moment.%n" +
```
markfisher: could be more concise, e.g. "No apps registered."

ericbottard:
Addressed review comments

markfisher:
squashed and merged as https://github.com/spring-cloud/spring-cloud-dataflow/commit/682586426c5b8d29d602af7af001933d4841b26b

PR #724, opened 2016-06-22
Allow skipping of tasks related docs

ericbottard:
Fixes spring-cloud/spring-cloud-dataflow-server-cloudfoundry#121


markpollack:
I merged it but then realized that we probably also need to exclude the intro section that talks about Spring Cloud Task.  there are a few sentences elsewhere which are not correct w/o having tasks. .e.g  `The Spring Cloud Data Flow architecture consists of a server that deploys Streams and Tasks.`...

markpollack:
Added a few other ifdefs.

PR #729, opened 2016-06-22
Added docs that discuss launching tasks from streams

### LINKED ISSUE #619: Document how tasks can be launched by a stream
_From @sabbyanandan on May 23, 2016 17:10_

As a user, I can use the reference guide to follow steps to launch tasks from a stream. 

**Acceptance:**
- docs include steps to launch tasks from a stream
- sample included

_Copied from original issue: spring-cloud/spring-cloud-task#145_

cppwfs:
resolves #619

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/tasks.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/tasks.adoc
@@ -239,3 +239,40 @@ The default Task/Batch event and destination names on the broker are enumerated 
 |Item Write events|`item-write-events`
 |Skip events|`skip-events`
 |===
+
+[[spring-cloud-dataflow-launch-tasks-from-stream]]
+== Launching Tasks from a Stream
+
+You can launch a task from a stream by using one of the available `task-launcher` sinks. Currently the only available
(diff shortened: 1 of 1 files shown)


spring-cloud-dataflow-docs/src/main/asciidoc/tasks.adoc
```diff
@@ -239,3 +239,40 @@ The default Task/Batch event and destination names on the broker are enumerated
...
+
+The `task-launcher-local` can be added to the available sinks by executing the app register command as follows:
+
+```
+app register --name task-launcher-local --type sink --uri maven://org.springframework.cloud.stream.app:task-launcher-local-sink-kafka:jar:1.0.0.BUILD-SNAPSHOT
+```
+
+=== TriggerTask
+
+The easiest way to launch a task using the `task-launcher` is to use the `triggertask` source. The `triggertask` source
```
mminella: I'm not sure this is actually "easier" than just doing `task launch foo` from a shell.

mminella:
Merged as 31a8ecc69bab8ceca15e3f79ad652b418429ea11

PR #738, opened 2016-06-27
Add license friendly JDBC drivers into server-core

### LINKED ISSUE #732: Bundle license-friendly JDBC drivers
As a developer, I'd like to add license-friendly JDBC drivers as part of the core dependencies, so it is easy for end users to connect to the desired RDBMS. For other database types, we would want to have it documented describing the procedure.

**Acceptance:**
- Following drivers are bundled: PostgreSQL (650KB), MySQL (Maria DB driver, 358K), H2 (already included as default) and HSQLDB (1.5MB)
- Docs describe the procedure how to add other driver types

ilayaperumalg:
- Add MariaDB java client (that supports MariaDB, MySQL), PostgreSQL and HSQLDB drivers
 - Add doc that explains how to configure RDBMS

This resolves #732

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/getting-started.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/getting-started.adoc
@@ -149,6 +149,36 @@ to `WARNING` in `logback.xml`:
 
 ====
 
+[[configuring-rdbms]]
+== RDBMS configuration
+
+The JDBC drivers for MySQL (via MariaDB driver), HSQLDB, PostgreSQL along with embedded H2 are available out of the box.
+If you are using any other RDBMS, then the corresponding JDBC driver jar needs to be on the classpath of the server.
+
+The RDBMS properties can be passed as command-line arguments to the dataflow server.
+
+For instance,
+If you are using MySQL,
+
+```
+java -jar spring-cloud-dataflow-server-local/target/spring-cloud-dataflow-server-local-1.0.0.BUILD-SNAPSHOT.jar \
+ --spring.datasource.url=jdbc:mysql:<db-info> --spring.datasource.username=<user> --spring.datasource.password=<password> --spring.datasource.driver-class-name=org.mariadb.jdbc.Driver &
+```
+For PostgreSQL,
+
+```
+java -jar spring-cloud-dataflow-server-local/target/spring-cloud-dataflow-server-local-1.0.0.BUILD-SNAPSHOT.jar \
+--spring.datasource.url=jdbc:postgresql:<db-info> --spring.datasource.username=<user> --spring.datasource.password=<password> --spring.datasource.driver-class-name=org.postgresql.Driver &
+```
+
+For HSQLDB,
+
+```
+java -jar spring-cloud-dataflow-server-local/target/spring-cloud-dataflow-server-local-1.0.0.BUILD-SNAPSHOT.jar \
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-docs/src/main/asciidoc/getting-started.adoc
```diff
@@ -149,6 +149,36 @@ to `WARNING` in `logback.xml`:
 
 ====
 
+[[configuring-rdbms]]
+== RDBMS configuration
```
ilayaperumalg: This needs to be part of other server implementations as well. may be we can split this part into a separate doc and have it added in each server implementations?

ilayaperumalg:
I have verified connecting to MySQL (using MariaDB driver), PostgreSQL, HSQLDB with these driver additions.

spring-cloud-dataflow-server-core/pom.xml
```diff
@@ -119,6 +119,21 @@
 			<artifactId>h2</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>org.hsqldb</groupId>
+			<artifactId>hsqldb</artifactId>
+			<version>${hsqldb.version}</version>
```
trisberg: do we need the <version> tag here?

cppwfs: They can be removed.

spring-cloud-dataflow-docs/src/main/asciidoc/getting-started.adoc
```diff
@@ -149,6 +149,36 @@ to `WARNING` in `logback.xml`:
...
+If you are using any other RDBMS, then the corresponding JDBC driver jar needs to be on the classpath of the server.
+
+The RDBMS properties can be passed as command-line arguments to the dataflow server.
+
+For instance,
+If you are using MySQL,
+
+```
+java -jar spring-cloud-dataflow-server-local/target/spring-cloud-dataflow-server-local-1.0.0.BUILD-SNAPSHOT.jar \
+ --spring.datasource.url=jdbc:mysql:<db-info> --spring.datasource.username=<user> --spring.datasource.password=<password> --spring.datasource.driver-class-name=org.mariadb.jdbc.Driver &
```
cppwfs: the datasource should be `jdbc:mariadb:`

spring-cloud-dataflow-docs/src/main/asciidoc/getting-started.adoc
trisberg: I think the configuration would be different for different server implementations, CF adds it's own configuration and so does k8s, so switching DB would not be the same instructions across the board

spring-cloud-dataflow-docs/src/main/asciidoc/getting-started.adoc
ilayaperumalg: In this case, we are using just the MariaDB JDBC driver to connect to the MySQL server. Hence, the `url` still points to `jdbc:mysql`.

spring-cloud-dataflow-docs/src/main/asciidoc/getting-started.adoc
ilayaperumalg: ok, I see. @trisberg thank you for clarifying this.

spring-cloud-dataflow-docs/src/main/asciidoc/getting-started.adoc
cppwfs: coolness

ghillert:
Merged with commit 66fc696.

Did some Asciidoc polishing and made sure that we use the naming `Data Flow Server`.

PR #739, opened 2016-06-27
Add how-to guide to override maven repository settings

sabbyanandan:
Once this is merged, we will add this as includes to all server implementation docs.

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/howto.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/howto.adoc
@@ -16,3 +16,74 @@ We're also more than happy to extend this section; If you want to add a '`how-to
 can send us a {github-code}[pull request].
 --
 
+== Configure Remote Maven Repository
+
+If you'd like to connect to your organization's internal maven repository, you can override the default 
+repository location by passing the credentials via the following environment properties:
+
+```
+export MAVEN_LOCAL_REPOSITORY=mylocalMavenRepo
+export MAVEN_REMOTE_REPOSITORIES_REPO1_URL=https://repo1
+export MAVEN_REMOTE_REPOSITORIES_REPO1_AUTH_USERNAME=user1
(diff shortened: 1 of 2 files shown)


ghillert:
Merged with polishing with commit  9086c97

spring-cloud-dataflow-docs/src/main/asciidoc/howto.adoc
```diff
@@ -16,3 +16,74 @@ We're also more than happy to extend this section; If you want to add a '`how-to
...
 --
 
+== Configure Remote Maven Repository
+
+If you'd like to connect to your organization's internal maven repository, you can override the default 
+repository location by passing the credentials via the following environment properties:
+
+```
+export MAVEN_LOCAL_REPOSITORY=mylocalMavenRepo
+export MAVEN_REMOTE_REPOSITORIES_REPO1_URL=https://repo1
```
ilayaperumalg: This will not work as mentioned in the previous doc. We should not be using this way of specifying the environment properties.

spring-cloud-dataflow-docs/src/main/asciidoc/howto.adoc
```diff
@@ -16,3 +16,74 @@ We're also more than happy to extend this section; If you want to add a '`how-to
...
+export MAVEN_REMOTE_REPOSITORIES_REPO1_AUTH_USERNAME=user1
+export MAVEN_REMOTE_REPOSITORIES_REPO1_AUTH_PASSWORD=pass1
+export MAVEN_REMOTE_REPOSITORIES_REPO2_URL=https://repo2
+```
+
+If you are running behind a proxy, then you need to specify the proxy settings via environment properties
```
ilayaperumalg: We deliberately removed the following from the previous doc and have only `SPRING_APPLICATION_JSON` as the way to set the environment properties.

sabbyanandan:
Reworked this via spring-cloud/spring-cloud-dataflow#740

PR #747, opened 2016-06-30
Spring boot maven plugin to use underlying spring boot version

ilayaperumalg:
Since the spring-boot-maven-plugin version 1.4.x breaks the loading of nested archive due to
https://github.com/spring-projects/spring-boot/commit/87fe0b2adeef85c842c009bfeebac1c84af8a5d7,
this fix tries use underlying spring-boot version which is 1.3.x based on what `spring-cloud-build` brings in.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -56,6 +56,17 @@
 			</dependency>
 		</dependencies>
 	</dependencyManagement>
+	<build>
+		<pluginManagement>
+			<plugins>
+				<plugin>
+					<groupId>org.springframework.boot</groupId>
+					<artifactId>spring-boot-maven-plugin</artifactId>
+					<version>${spring-boot.version}</version>
+				</plugin>
+			</plugins>
+		</pluginManagement>
+	</build>
 	<profiles>
 		<profile>
 			<id>spring</id>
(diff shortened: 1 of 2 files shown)


jvalkeal:
should we add this to `spring-cloud-dataflow-parent` as `pluginManagement`? Then in theory all nested server would get it automatically.

jvalkeal:
ah yeah, I wonder how maven works with nested `pluginManagement` sections. Can `pluginManagement` just be used to define version while other plugin settings are coming from parent for a particular plugin.

jvalkeal:
just tried this, added this to `spring-cloud-dataflow-parent` pom and yarn server started to work with `app info` as well.
```
<build>
  <pluginManagement>
    <plugins>
      <plugin>
        <version>${spring-boot.version}</version>
      </plugin>
    </plugins>
  </pluginManagement>
</build>
```

ilayaperumalg:
yeah, that's a good idea @jvalkeal. Moved the changes into `pluginManagement`.

jvalkeal:
squashed/merged per https://github.com/spring-cloud/spring-cloud-dataflow/commit/a5df5b3b0ff0733a60c74549f46136e6dccb2128

PR #755, opened 2016-07-01
properly allocate deployment and app properties

### LINKED ISSUE #754: Conditionally bucket deployment vs. application properties
As a developer, I'd like to conditionally bucket `spring.cloud.deployer.*` properties into `AppDeploymentRequest.deploymentProperties` and all other properties (_should be only application specific_) to `AppDefinition.properties`. 

**Acceptance:**
- Conditional bucketing is implemented in `StreamDeploymentController`
- Unit tests to simulate conditional bucketing through assertions

markfisher:
fixes #754

### DIFF
--- a/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/DataFlowAppDefinition.java
+++ b/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/DataFlowAppDefinition.java
@@ -16,6 +16,7 @@
 
 package org.springframework.cloud.dataflow.core;
 
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/DataFlowAppDefinition.java
```diff
@@ -101,22 +102,37 @@ protected void setRegisteredAppName(String registeredAppName) {
...
+		Map<String, String> appProperties = new HashMap<>(this.appDefinition.getProperties());
+		Map<String, String> deploymentProperties = new HashMap<>();
+		for (Map.Entry<String, String> entry : properties.entrySet()) {
+			String key = entry.getKey();
+			String value = entry.getValue();
+			if (key.startsWith("spring.cloud.deployer.")) {
+				deploymentProperties.put(key, value);
+			}
+			else {
+				appProperties.put(key, value);
```
ilayaperumalg: The only deployment property that is now included into `appProperties` is the property with the key `count` (which comes from app.*.count). This is the left over after extracting and updating the SCSt `instanceCount` property. It does no harm though. Will create a separate PR which can optionally merged as part of GA.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Also verified the deployment vs app properties using the deployment of partitioned stream.

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/DataFlowAppDefinition.java
ilayaperumalg: Also, the `count` as a property into app properties might conflict if the app itself has a property named `count`. But this is just an edge case in the custom built apps as most of the app properties would either have SCSt prefix or the app name prefix.

ilayaperumalg: Created PR that handles the removal of `count` property after being extracted: https://github.com/spring-cloud/spring-cloud-dataflow/pull/757

Issue #751, opened 2016-06-30
Document that only boot 1.3.x apps support property whitelisting.


markpollack:

ghillert:
See also: https://github.com/spring-cloud/spring-cloud-dataflow/issues/748

PR #756, opened 2016-07-01
gh-751 Document that only boot 1.3.x apps support property whitelisting

ghillert:
Resolves spring-cloud/spring-cloud-dataflow#751

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
@@ -134,8 +134,43 @@ configuration-properties.names=server.port
 [IMPORTANT]
 ====
 As of Spring Cloud Data Flow `1.0.0.RELEASE` the whitelisting of application properties
-is only supported for Spring Boot `1.3.x` based application. Milestone releases
-of the upcoming `1.4.0` release are not supported, yet.
+is only explicitly supported for Spring Boot `1.3.x` based application. Milestone releases
+of the upcoming `1.4.0` release are not explicitly supported, yet.
+
+The cause is in the used `spring-boot-maven-plugin`. As a result you will observe
+all available application properties being listed, not just the whitelisted ones. As a temporary
+workaround, you can override the managed version of your app's `spring-boot-maven-plugin`
+explicitly and revert to a version of the latest 1.3.x release:
+
+For example, if your app's `pom.xml` specifies to use Spring Boot `1.4.0.M3`:
+
+[source,xml]
+----
+…
+<parent>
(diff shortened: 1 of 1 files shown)


spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -131,6 +131,13 @@ configuration.classes=org.springframework.cloud.stream.app.file.sink.FileSinkPro
 configuration-properties.names=server.port
 ```
 
+[IMPORTANT]
+====
+As of Spring Cloud Data Flow `1.0.0.RELEASE` the whitelisting of application properties
+is only supported for Spring Boot `1.3.x` based application. Milestone releases
+of the upcoming `1.4.0` release are not supported, yet.
+====
```
ilayaperumalg: I think we could also add what happens when someone uses `@EnableDataflowServer` that uses boot `1.4.0` with the apps using `1.3.x` or `1.4.x` boot versions.

ghillert:
@ilayaperumalg  Rebased and Expanded Docs

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -131,6 +131,48 @@ configuration.classes=org.springframework.cloud.stream.app.file.sink.FileSinkPro
...
 ```
 
+[IMPORTANT]
+====
+As of Spring Cloud Data Flow `1.0.0.RELEASE` the whitelisting of application properties
+is only explicitly supported for Spring Boot `1.3.x` based application. Milestone releases
+of the upcoming `1.4.0` release are not explicitly supported, yet.
+
+The cause is in the used `spring-boot-maven-plugin`. As a result you will observe
+all available application properties being listed, not just the whitelisted ones. As a temporary
```
ilayaperumalg: The problem is we don't list any properties. Since the app's nested archives aren't loaded, there will be empty list of properties.

ilayaperumalg: I have double-checked this and can confirm that we wouldn't see any property listing at all when the `spring-boot-maven` plugin version `1.4.x` is used.

ilayaperumalg: We need documentation that talks about `dataflow` server (enabled using @EnableDataflowServer) that uses boot version `1.4.x` as well.

ilayaperumalg:
LGTM, merged as 3371e344091a0d10e5521b53cc5b6ee364d81994 along with the changes that address the above comments. f676ed1f414fc97aa118f1542526999e3be2a272

PR #757, opened 2016-07-01
Exclude `count` as a property in app properties

### LINKED ISSUE #758: Remove `count` as a property from app properties
Please see the discussion here: 

https://github.com/spring-cloud/spring-cloud-dataflow/pull/755#discussion_r69254944

ilayaperumalg:
When the deployment property `count` is extracted and updated as `spring.cloud.deployer.count` as deployment property, only the SCSt `instanceCount` app property should be set for app property.
The property `count` should not part of app properties.

This resolves #758

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
@@ -224,6 +224,8 @@ public class StreamDeploymentController {
 			if (appDeploymentProperties.containsKey(INSTANCE_COUNT_PROPERTY_KEY)) {
 				appDeploymentProperties.put(AppDeployer.COUNT_PROPERTY_KEY,
 						appDeploymentProperties.get(INSTANCE_COUNT_PROPERTY_KEY));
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
```diff
@@ -224,6 +224,8 @@ private void deployStream(StreamDefinition stream, Map<String, String> streamDep
 			if (appDeploymentProperties.containsKey(INSTANCE_COUNT_PROPERTY_KEY)) {
```
markfisher: can we instead change the part of the code that adds the property in the first place to use the right key?

markfisher: I will push a commit to my branch that makes the change I mentioned in the previous comment.

ilayaperumalg: sure, thank you!

ilayaperumalg:
Closing this in favour of #760

Issue #770, opened 2016-07-05
Docs: Provide OAuth Configuration Example for GitHub


ghillert:
- Provide OAuth Configuration Example for GitHub
- Provide link to https://github.com/ghillert/oauth-test-server in order to show a quick example of how to use Spring Security OAuth's authentication server

sabbyanandan:
Perhaps can we also include spring-cloud/spring-cloud-dataflow#767 as part of this PR?

PR #773, opened 2016-07-05
gh-770 Provide OAuth Configuration Example for GitHub

ghillert:
- Provide OAuth Configuration Example for GitHub
- Provide link to https://github.com/ghillert/oauth-test-server in order to show a quick example of how to use Spring Security OAuth's authentication server

Resolves spring-cloud/spring-cloud-dataflow#770


jvalkeal:
We should mention that  `Authorization callback URL` in github OAuth application  should be login url i.e. `http://localhost:9393/login`.

Other than that is there any way to limit login to a members of a github org? I assume no as we don't have no detailed authorization, just authentication?

jvalkeal:
And we should document whether github works with dataflow shell or not. I couldn't get it working.

ghillert:
@jvalkeal Thanks will get that addressed in a little bit.

ghillert:
Looks like GitHub does not support password grant type. They use basic authentication if you need a token without the web-oauth-dance. So much for standards ... 

https://developer.github.com/v3/oauth/#non-web-application-flow

ghillert:
@jvalkeal Addressed your comments.

jvalkeal:
@ghillert bummer, looks like there is a merge conflict as ilaya also made some changes 2 hours ago. Would you mind fixing it.

ghillert:
@jvalkeal Rebased. Yeah, OAuth is basically quite a narrow standard. Does not deal with what we commonly understand as `authorization`. That's covered by OpenID Connect. Feels like OAuth is commonly a bit misunderstood.

GitHub also illustrates that there are provider differences. Maybe less so then in the olden days but still. In that regard I find Spring Social the better approach as it provides provider specific implementations. I hope @rwinch can eventually merge Spring Security OAuth and Spring Social (or create the proper abstractions between lower-level apis and higher-level implementations and remove overlap). Maybe call it `Spring Security Connect` :-)

spring-cloud-dataflow-docs/src/main/asciidoc/getting-started.adoc
```diff
@@ -390,3 +390,53 @@ dataflow config info
...
+If you rather like to use an existing OAuth2 provider, here is an example for GitHub.
+First you need to **Register a new application** under your GitHub account at:
+
+https://github.com/settings/developers[https://github.com/settings/developers]
+
+When running a default version of Spring Cloud Data Flow locally, your GitHub configuration
+should look like the following:
+
+.Register an OAuth Application for GitHub
+image::dataflow-security-github.png[Register an OAuth Application for GitHub , scaledwidth="100%"]
```
jvalkeal: Just realised this image is missing. Other than that I believe we need to use embed images directly from github.

jvalkeal:
squashed and merged.

PR #784, opened 2016-07-07
Remove short-form app properties after qualifying

### LINKED ISSUE #783: Short form app properties should be removed after qualified
When the white listed properties are qualified with their corresponding long forms, the short form names should be removed from app properties map. This will make sure to avoid any collision with the underlying properties (possibly from boot)

Currently, the short-form properties are [updated] (https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java#L286) along with the app properties.

ilayaperumalg:
- When the white listed properties are qualified with their corresponding long forms, the short form names should be removed from app properties map
This will make sure to avoid any collision with the underlying properties (possibly from boot)

This resolves #783


ericbottard:
Merged as 8cd140f

PR #785, opened 2016-07-07
Add partitioning paragraph to architecture section

ericbottard:
Fixes spring-cloud/spring-cloud-dataflow#765

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/architecture.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/architecture.adoc
@@ -133,7 +133,7 @@ For an application that will consume events, Spring Cloud stream exposes a concu
 [[arch-streams-partitioning]]
 === Partitioning
 In some situations (for example to benefit from scaling out), applications can be deployed using multiple instances and
-benefit from data partitioning (whereby a some subset of data always goes to the same instance according to some
+benefit from data partitioning (whereby some subset of data always goes to the same instance according to some
 business rule). Please refer to <<passing_stream_partition_properties>> to learn more.
 
 [[arch-streams-delivery]]
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/architecture.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/architecture.adoc
@@ -130,6 +130,12 @@ Taps can be used to ‘listen in’ to the data that if flowing across any of th
 === Concurrency
 For an application that will consume events, Spring Cloud stream exposes a concurrency setting that controls the size of a thread pool used for dispatching incoming messages.  See the link:http://docs.spring.io/spring-cloud-stream/docs/1.0.2.RELEASE/reference/htmlsingle/index.html#_consumer_properties[Consumer properties] documentation for more information.
 
+[[arch-streams-partitioning]]
+=== Partitioning
+In some situations (for example to benefit from scaling out), applications can be deployed using multiple instances and
+benefit from data partitioning (whereby a some subset of data always goes to the same instance according to some
+business rule). Please refer to <<passing_stream_partition_properties>> to learn more.
+
 [[arch-streams-delivery]]
 === Message Delivery Guarantees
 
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-docs/src/main/asciidoc/architecture.adoc
```diff
@@ -130,6 +130,12 @@ Taps can be used to ‘listen in’ to the data that if flowing across any of th
 === Concurrency
 For an application that will consume events, Spring Cloud stream exposes a concurrency setting that controls the size of a thread pool used for dispatching incoming messages.  See the link:http://docs.spring.io/spring-cloud-stream/docs/1.0.2.RELEASE/reference/htmlsingle/index.html#_consumer_properties[Consumer properties] documentation for more information.
 
+[[arch-streams-partitioning]]
+=== Partitioning
+In some situations (for example to benefit from scaling out), applications can be deployed using multiple instances and
+benefit from data partitioning (whereby a some subset of data always goes to the same instance according to some
```
markfisher: keep either 'a' or 'some' before 'subset' but not both

ericbottard: Urgh, fixed

spring-cloud-dataflow-docs/src/main/asciidoc/architecture.adoc
```diff
@@ -130,6 +130,12 @@ Taps can be used to ‘listen in’ to the data that if flowing across any of th
 === Concurrency
 For an application that will consume events, Spring Cloud stream exposes a concurrency setting that controls the size of a thread pool used for dispatching incoming messages.  See the link:http://docs.spring.io/spring-cloud-stream/docs/1.0.2.RELEASE/reference/htmlsingle/index.html#_consumer_properties[Consumer properties] documentation for more information.
 
+[[arch-streams-partitioning]]
+=== Partitioning
+In some situations (for example to benefit from scaling out), applications can be deployed using multiple instances and
+benefit from data partitioning (whereby some subset of data always goes to the same instance according to some
+business rule). Please refer to <<passing_stream_partition_properties>> to learn more.
+
```
markpollack: I'll add that we are doing 'static' partitioning ATM w/ Kafka 0.8 on merge.

markfisher: the rabbit binder doesn't support dynamic/elastic partitions either (at this point)... so it would be better to either make a general comment about the fixed partition limitation, or to wait until we do have some cases that support dynamic/elastic partitions (might be kafka 0.9+ binder only at first, unless we refactor the rabbit binder's partition support as well)

mbogoevici: I believe we should simply clarify that dynamic scaling is not supported with partitioning. This goes beyond a simple 'binder limitation'. Scaling up with semantic partitioning needs to be part of the more general stateful processing story. It's not only the binder that has to support it, but the state store as well (which needs to be part of the SCSt application model). When we scale up/down we effectively shuffle data across instances so we need to ensure that consumers are stopped, and the state is consistently flushed before all consumers begin consuming again. Also, in order for the process to be effective, the state store needs to be partitioned as well, and we need to ensure that the partitioning strategy of the state store after scaling matches the post-scaling data allocation scheme of the consumers. At this time, it is significantly simpler to rely on stopping and restarting the streams to ensure consistency in what concerns application lifecycle and state flushing. If we are to dynamically scaling with partitioning, we need to do this in conjunction with first-class stateful processing support.

markpollack:
I will add a link to http://docs.spring.io/spring-cloud-stream/docs/1.0.2.RELEASE/reference/htmlsingle/index.html#_configuring_output_bindings_for_partitioning for info on how partitioning is configured ATM.  Once we have new binder impl, we can update the SCDF docs as needed.

markpollack:
Updated and merged as 6cce6096b3548aa9ceed3414abc20624ff5965c3 and 7cbcc8bb8d770dd8f758326bee03028f5c2113eb

PR #789, opened 2016-07-07
Qualify app properties specified during stream deployment

### LINKED ISSUE #787: Qualify application properties when specified during stream deployment

ilayaperumalg:
- When application properties are specified during stream deployment, these properties need to be qualified.
This will make the `white-listed` properties specified in `short-form` resolving to fully qualified property names.
 - Add tests with overriding app properties during deployment

This resolves #787

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
@@ -237,8 +237,8 @@ public class StreamDeploymentController {
 					currentApp.getName(), type));
 			Resource resource = registration.getResource();
 			currentApp = qualifyParameters(currentApp, resource);
-
-			AppDeploymentRequest request = currentApp.createDeploymentRequest(resource, appDeploymentProperties);
+			AppDeploymentRequest request = currentApp.createDeploymentRequest(resource,
+					qualifyParameters(appDeploymentProperties, resource));
 			try {
 				String id = this.deployer.deploy(request);
 				this.deploymentIdRepository.save(DeploymentKey.forStreamAppDefinition(currentApp), id);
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
```diff
@@ -251,11 +251,7 @@ private void deployStream(StreamDefinition stream, Map<String, String> streamDep
 		}
 	}
 
-	/**
-	 * Return a copy of a given app definition where short form parameters have been expanded to their long form
-	 * (amongst the whitelisted supported properties of the app) if applicable.
-	 */
-	/*default*/ StreamAppDefinition qualifyParameters(StreamAppDefinition original, Resource resource) {
+	private Map<String, String> qualifyParameters(Map<String, String> appDeploymentProperties, Resource resource) {
```
ericbottard: Not too happy with the refactor generated parameter name `appDeploymentProperties`, will change myself

markfisher: That is the name of the variable in the code that calls out to this... it's a bit confusing but it's intention is to capture the fact that it is the groups of properties *from* `stream deploy` that apply to a particular *app*. What do you suggest as an alternative?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
```diff
@@ -290,7 +285,16 @@ private void deployStream(StreamDefinition stream, Map<String, String> streamDep
...
 		}
-		return builder.setProperties(mutatedProps).build(original.getStreamName());
+		return mutatedProps;
+	}
+
+	/**
+	 * Return a copy of a given app definition where short form parameters have been expanded to their long form
+	 * (amongst the whitelisted supported properties of the app) if applicable.
+	 */
+	/*default*/ StreamAppDefinition qualifyParameters(StreamAppDefinition original, Resource resource) {
```
ericbottard: Also changing this method name (and the other) to `qualifyProperties` for consistency

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
ilayaperumalg: Since this method is used in two different invocations with `app properties defined at creation time` and `app properties defined at deployment time along with other deployment properties`, may be we can just call this as `properties` ?

ericbottard:
Merged as https://github.com/spring-cloud/spring-cloud-dataflow/commit/31a68465a7ccb0be6d6fbe72753a008df5accfb4 which includes https://github.com/ericbottard/spring-cloud-dataflow/commit/da7afa7c6260e1df9bba6de442a25c505a8ca866 as a polishing commit

ilayaperumalg:
thank you @ericbottard

PR #791, opened 2016-07-08
Handle instanceCount and indexed properties

ilayaperumalg:
- Always set instanceCount when the app count is specified. By default it will use `1` specified in SCSt
 - Set `spring.cloud.deployer.indexed` property to `true` for the consumer applications (processor/sink/task)
 - Update tests

### DIFF
--- a/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/StreamControllerTests.java
+++ b/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/StreamControllerTests.java
@@ -442,7 +442,7 @@ public class StreamControllerTests {
 		assertEquals(2, requests.size());
 		AppDeploymentRequest logRequest = requests.get(0);
 		assertThat(logRequest.getDefinition().getName(), is("log"));
-		assertEquals(logRequest.getDeploymentProperties().get("spring.cloud.deployer.indexed"), "true");
+		assertEquals(logRequest.getDeploymentProperties().get(AppDeployer.INDEXED_PROPERTY_KEY), "true");
 		AppDeploymentRequest timeRequest = requests.get(1);
 		assertThat(timeRequest.getDefinition().getName(), is("time"));
 	}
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/StreamControllerTests.java
```diff
@@ -441,6 +442,7 @@ public void testDeploy() throws Exception {
 		assertEquals(2, requests.size());
 		AppDeploymentRequest logRequest = requests.get(0);
 		assertThat(logRequest.getDefinition().getName(), is("log"));
+		assertEquals(logRequest.getDeploymentProperties().get("spring.cloud.deployer.indexed"), "true");
```
markfisher: this should use the `AppDeployer` constant instead of a String

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/StreamControllerTests.java
```diff
@@ -613,6 +615,7 @@ public void testDeployWithProperties() throws Exception {
 		assertEquals("2", logAppProps.get("spring.cloud.stream.instanceCount"));
 		assertEquals("true", logAppProps.get("spring.cloud.stream.bindings.input.consumer.partitioned"));
 		assertEquals("3", logAppProps.get("spring.cloud.stream.bindings.input.consumer.concurrency"));
+		assertEquals("2", logAppProps.get(StreamPropertyKeys.INSTANCE_COUNT));
```
markfisher: maybe we should add this same assertion in other relevant places in this test class

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/StreamControllerTests.java
```diff
@@ -441,6 +442,7 @@ public void testDeploy() throws Exception {
 		assertEquals(2, requests.size());
 		AppDeploymentRequest logRequest = requests.get(0);
 		assertThat(logRequest.getDefinition().getName(), is("log"));
+		assertEquals(logRequest.getDeploymentProperties().get(AppDeployer.INDEXED_PROPERTY_KEY), "true");
```
markfisher: nitpick: all of these `assertEquals` should have the expected value ("true") on the left hand side

ilayaperumalg: I agree.

markfisher:
rebased, squashed and merged as https://github.com/spring-cloud/spring-cloud-dataflow/commit/1963124b6e0f0b9c5c9c2a31021fc046b862184e

PR #801, opened 2016-07-20
Remove EnableConfig prop at SecurityController

### LINKED ISSUE #366: Make SecurityProperties safe in SecurityController
As mentioned in #364 we may get into trouble:
```
expected single matching bean but found 2: securityProperties,security.CONFIGURATION_PROPERTIES
```

Effectively same issue per https://github.com/spring-projects/spring-boot/issues/3476 where spring-session autoconfig used `@EnableConfigurationProperties` in a same way we do causing bean to created twice.

We can either remove `@EnableConfigurationProperties(SecurityProperties.class)` altogether assuming it's gonna get created by boot, or to play safe do same trick boot do in https://github.com/spring-projects/spring-boot/blob/master/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/SessionAutoConfiguration.java. Commit which fixed it is https://github.com/spring-projects/spring-boot/commit/28edc95f18a864285cc8ec4ef46c01347a190650.

This time bomb was planted a while back by me and @ilayaperumalg

ilayaperumalg:
- Since the `SecurityProperties` is set at the constructor level via controller auto configuration, we don't need `EnableConfigurationProperties` for `SecurityProperties` at
SecurityController

This fixes #366

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/security/SecurityController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/security/SecurityController.java
@@ -46,7 +46,6 @@ import org.springframework.web.bind.annotation.RestController;
 @RestController
 @RequestMapping("/security/info")
 @ExposesResourceFor(SecurityInfoResource.class)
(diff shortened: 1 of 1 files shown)


ghillert:
Merged.

PR #839, opened 2016-08-26
Add REST endpoint for finding related streams

### LINKED ISSUE #822: Add the ability to query for related streams
As a user, I'd like to see the graph of all the related streams in Flo. This should include all the TAP'd streams as one cohesive view. 

For instance, the Flo graph at the create time looks like:
image

I'd like to see similar rendering on click of a button from the the list page, so I can review the entire topology. Currently, all the TAP'd streams are listed as individual rows in the list page and the Flo graph for the individual entries is not very useful.

**Acceptance:**
- a new REST-API that accepts stream-id/stream-name to query for related streams from the stored definitions
- Flo would use this API to build the graph with related nodes to show it in a dialog window - related story in [flo tracker](https://www.pivotaltracker.com/story/show/106697770)

ilayaperumalg:
The REST endpoing `/streams/definitions/{name}/related` will get all the stream definitions that are related (main/tap streams) of the given stream named by `name`

- This resolves #822

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
@@ -28,6 +28,9 @@ import org.slf4j.LoggerFactory;
 import org.springframework.cloud.dataflow.core.ApplicationType;
 import org.springframework.cloud.dataflow.core.StreamAppDefinition;
 import org.springframework.cloud.dataflow.core.StreamDefinition;
+import org.springframework.cloud.dataflow.core.dsl.StreamNode;
+import org.springframework.cloud.dataflow.core.dsl.StreamParser;
+import org.springframework.cloud.dataflow.core.dsl.TokenKind;
 import org.springframework.cloud.dataflow.registry.AppRegistry;
 import org.springframework.cloud.dataflow.rest.resource.StreamDefinitionResource;
 import org.springframework.cloud.dataflow.server.DataFlowServerUtil;
@@ -198,7 +201,7 @@ public class StreamDefinitionController {
 	@RequestMapping(value = "/{name}/related", method = RequestMethod.GET)
 	@ResponseStatus(HttpStatus.OK)
 	public PagedResources<StreamDefinitionResource> listRelated(@PathVariable("name") String name,
-																@RequestParam("nested") boolean nested,
+																@RequestParam(value="nested", required = false, defaultValue = "false") boolean nested,
 																PagedResourcesAssembler<StreamDefinition> assembler) {
 		Set<StreamDefinition> relatedDefinitions = new HashSet<>();
 		StreamDefinition currentStreamDefinition = repository.findOne(name);
@@ -214,18 +217,13 @@ public class StreamDefinitionController {
 	private Set<StreamDefinition> findRelatedDefinitions(StreamDefinition currentStreamDefinition, Iterable<StreamDefinition> definitions,
 														  Set<StreamDefinition> relatedDefinitions, boolean nested) {
 		relatedDefinitions.add(currentStreamDefinition);
+		String currentStreamName = currentStreamDefinition.getName();
+		String indexedStreamName = currentStreamName + ".";
 		for (StreamDefinition definition: definitions) {
-			String tapNamePrefix = ":" + currentStreamDefinition.getName();
-			String dslText = definition.getDslText();
-			if (dslText.startsWith(tapNamePrefix)) {
-				String[] dslSplits = dslText.split(">");
-				if (dslText.startsWith(tapNamePrefix + ".")) {
-					relatedDefinitions.add(definition);
-					if (nested) {
-						findRelatedDefinitions(definition, definitions, relatedDefinitions, true);
-					}
-				}
-				else if (dslSplits[0].trim().equalsIgnoreCase(tapNamePrefix)) {
+			StreamNode sn = new StreamParser(definition.getName(), definition.getDslText()).parse();
+			if (sn.getSourceDestinationNode() != null) {
+				String nameComponent = sn.getSourceDestinationNode().getDestinationName();
+				if (nameComponent.equals(currentStreamName) || nameComponent.startsWith(indexedStreamName)) {
 					relatedDefinitions.add(definition);
 					if (nested) {
 						findRelatedDefinitions(definition, definitions, relatedDefinitions, true);
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
@@ -25,7 +25,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import org.springframework.cloud.dataflow.core.ApplicationType;
-import org.springframework.cloud.dataflow.core.BindingPropertyKeys;
 import org.springframework.cloud.dataflow.core.StreamAppDefinition;
 import org.springframework.cloud.dataflow.core.StreamDefinition;
 import org.springframework.cloud.dataflow.registry.AppRegistry;
@@ -39,7 +38,7 @@ import org.springframework.cloud.dataflow.server.repository.StreamDefinitionRepo
 import org.springframework.cloud.deployer.spi.app.AppDeployer;
 import org.springframework.cloud.deployer.spi.app.AppStatus;
 import org.springframework.cloud.deployer.spi.app.DeploymentState;
-import org.springframework.cloud.deployer.spi.core.AppDefinition;
+import org.springframework.data.domain.PageImpl;
 import org.springframework.data.domain.Pageable;
 import org.springframework.data.web.PagedResourcesAssembler;
 import org.springframework.hateoas.ExposesResourceFor;
@@ -187,6 +186,40 @@ public class StreamDefinitionController {
 		}
 		deploymentController.undeploy(name);
 		this.repository.delete(name);
+	}
+
+	/**
+	 * Return a list of related stream definition resources based on the given stream name.
+	 * Related streams include the main stream and the tap stream(s) on the main stream.
+	 *
+	 * @param name the name of an existing stream definition (required)
+	 */
+	@RequestMapping(value = "/{name}/related", method = RequestMethod.GET)
+	@ResponseStatus(HttpStatus.OK)
+	public PagedResources<StreamDefinitionResource> listRelated(@PathVariable("name") String name,
+																PagedResourcesAssembler<StreamDefinition> assembler) {
+		List<StreamDefinition> relatedDefinitions = new ArrayList<>();
+		StreamDefinition currentStreamDefinition = repository.findOne(name);
+		if (currentStreamDefinition == null) {
+			throw new NoSuchStreamDefinitionException(name);
+		}
+		relatedDefinitions.add(currentStreamDefinition);
+		String tapNamePrefix = ":" + name;
+		Iterable<StreamDefinition> definitions = repository.findAll();
+
+		for (StreamDefinition definition: definitions) {
+			String dslText = definition.getDslText();
+			if (dslText.startsWith(tapNamePrefix)) {
+				String[] dslSplits = dslText.split(">");
+				if (dslText.startsWith(tapNamePrefix + ".")) {
+					relatedDefinitions.add(definition);
+				}
+				else if (dslSplits[0].trim().equalsIgnoreCase(tapNamePrefix)) {
+					relatedDefinitions.add(definition);
+				}
+			}
+		}
+		return assembler.toResource(new PageImpl<>(relatedDefinitions), streamDefinitionAssembler);
 	}
 
 	/**
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
@@ -18,6 +18,7 @@ package org.springframework.cloud.dataflow.server.controller;
 
 import java.util.ArrayList;
 import java.util.EnumSet;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
```diff
@@ -190,6 +189,40 @@ public void delete(@PathVariable("name") String name) {
 	}
 
 	/**
+	 * Return a list of related stream definition resources based on the given stream name.
+	 * Related streams include the main stream and the tap stream(s) on the main stream.
+	 *
+	 * @param name the name of an existing stream definition (required)
+	 */
+	@RequestMapping(value = "/{name}/related", method = RequestMethod.GET)
```
ilayaperumalg: @aclement @BoykoAlex It might be helpful to have the same REST endpoint `/streams/definitions/{name}` to serve the related streams (with some request attribute set) response as well. But that will make the response resulting in single/plural response based on the attribute. Hence, having this explicit REST endpoint might be better. What do you think?

BoykoAlex:
I like the REST endpoint `/streams/definitions/{name}/related`. I noticed, however, that it only returns level 1 tap streams but no nested tap streams.
For example here are my stream definitions:
```
main-stream=time | filter | log
tap-level-1=:main-stream.filter > transform | aggregate-counter
tap-level-2=:tap-level-1.transform > splitter | counter
```
After these are created the rest call:
`http://localhost:9393/streams/definitions/main-stream/related`
Would only list `main-stream` and `tap-level-1`
```
{"_embedded":{"streamDefinitionResourceList":[{"name":"main-stream","dslText":"time | filter | log","status":"undeployed","_links":{"self":{"href":"http://localhost:9393/streams/definitions/main-stream"}}},{"name":"tap-level-1","dslText":":main-stream.filter > transform | aggregate-counter","status":"undeployed","_links":{"self":{"href":"http://localhost:9393/streams/definitions/tap-level-1"}}}]},"_links":{"self":{"href":"http://localhost:9393/streams/definitions/main-stream/related"}},"page":{"size":0,"totalElements":2,"totalPages":1,"number":0}}
```

I think we may need all nested related streams as well for the graph. Am I right @aclement @sabbyanandan ?

ilayaperumalg:
We could possibly add an attribute to fetch `nested` streams using the flag `nested=true`.

ilayaperumalg:
@BoykoAlex @aclement @sabbyanandan what do you think about my suggestion above?

BoykoAlex:
`nested` attribute sounds good to me!-)

ilayaperumalg:
Pushed changes to include `nested` stream definitions as part of finding related streams.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
```diff
@@ -190,6 +190,53 @@ public void delete(@PathVariable("name") String name) {
...
+		List<StreamDefinition> result = new ArrayList<>(findRelatedDefinitions(currentStreamDefinition, definitions,
+				relatedDefinitions, nested));
+		return assembler.toResource(new PageImpl<>(result), streamDefinitionAssembler);
+	}
+
+	private Set<StreamDefinition> findRelatedDefinitions(StreamDefinition currentStreamDefinition, Iterable<StreamDefinition> definitions,
+														  Set<StreamDefinition> relatedDefinitions, boolean nested) {
+		relatedDefinitions.add(currentStreamDefinition);
+		for (StreamDefinition definition: definitions) {
+			String tapNamePrefix = ":" + currentStreamDefinition.getName();
```
ericbottard: This is very String oriented. Can we work with the object model instead (parse result or something)?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
```diff
@@ -190,6 +190,53 @@ public void delete(@PathVariable("name") String name) {
...
+			String dslText = definition.getDslText();
+			if (dslText.startsWith(tapNamePrefix)) {
+				String[] dslSplits = dslText.split(">");
+				if (dslText.startsWith(tapNamePrefix + ".")) {
+					relatedDefinitions.add(definition);
+					if (nested) {
+						findRelatedDefinitions(definition, definitions, relatedDefinitions, true);
+					}
+				}
+				else if (dslSplits[0].trim().equalsIgnoreCase(tapNamePrefix)) {
```
ericbottard: Why ignore case?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
```diff
@@ -190,6 +190,53 @@ public void delete(@PathVariable("name") String name) {
...
+														  Set<StreamDefinition> relatedDefinitions, boolean nested) {
+		relatedDefinitions.add(currentStreamDefinition);
+		for (StreamDefinition definition: definitions) {
+			String tapNamePrefix = ":" + currentStreamDefinition.getName();
+			String dslText = definition.getDslText();
+			if (dslText.startsWith(tapNamePrefix)) {
+				String[] dslSplits = dslText.split(">");
+				if (dslText.startsWith(tapNamePrefix + ".")) {
+					relatedDefinitions.add(definition);
+					if (nested) {
```
ericbottard: Why test for nested here? Aren't we always nested if we call `findRelatedDefinitions ` in the first place?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
ericbottard: Also, this concatenation can be taken out of the for loop

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/StreamControllerTests.java
```diff
@@ -178,6 +178,96 @@ public void testSave() throws Exception {
 	}
 
 	@Test
+	public void testFindRelatedStreams() throws Exception {
```
ericbottard: It would be good to have a non-web unit test for `findRelatedDefinitions()` IMO

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
ilayaperumalg: No, the `nested` is checked only during the `recursive` calls. We aren't always `nested` (at least the first time the `findReleatedDefinitions` is called).

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
ilayaperumalg: The stream definition needs to be case `insensitive` when compared to the related stream definitions.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
ilayaperumalg: > This is very String oriented. Can we work with the object model instead (parse result or something)?

May be. But, it turned out that using the simple `String` based check against the DSL text is easier for the tap prefix check(Given we have more deterministic way of specifying this) than the StreamParser approach. 
Do you see any issue?

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/StreamControllerTests.java
ilayaperumalg: Since the `actual` changes are in the `controller` side I would like to know what advantage we get by testing this without web based unit tests (where we verify the usecases).

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
ericbottard: Why? Where and when did we say that stream names were case insensitive? If they are, then a whole lot of places in the code need to be fixed.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
ericbottard: It's brittle and violates single responsibility principle. If the syntax for taps ever changes, both the parser AND this place would need to change

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
ilayaperumalg: I think it doesn't make sense to have case *sensitive* DSL where the tap syntax ":test1 > log" and ":TEST1 > log" aren't same. do you think otherwise?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
ericbottard: That was my point indeed. Although I see and understand how this method is crafted, it seems a bit weird to call a method named `findReleatedDefinitions()` even once, when you're not interested in related definitions.

By the way, that `nested` flag would be better renamed to `related` (ie why have two names)

ilayaperumalg: Both `nested` and `related` serve different purposes. Please see here for more detail: https://github.com/spring-cloud/spring-cloud-dataflow/pull/839#issuecomment-243482028

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/StreamControllerTests.java
ericbottard: mvc tests tend to be less readable IMO. Also, we should test the actual logic. The fact that business logic is constantly added to the *controller* is wrong in the first place if you ask me..

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/StreamControllerTests.java
```diff
@@ -178,6 +178,96 @@ public void testSave() throws Exception {
...
+		assertEquals(0, repository.count());
+		mockMvc.perform(
+				post("/streams/definitions/").param("name", "myStream1").param("definition", "time | log")
+						.accept(MediaType.APPLICATION_JSON)).andDo(print())
```
ericbottard: Please remove the `andDo(print())`s, we have enough garbage in the console while building already.

Also, I think there is the way to have `accept(MediaType.APPLICATION_JSON)` as a default for the whole test

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
ericbottard: I personally prefer case sensitiveness by default, but that is not the question. What I'm saying is that the rest of the system is not case-insensitive (eg definition repositories are not). So we should be consistent.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
ericbottard: I'm not sure I get it. Is related the same as "nested at level 1"? If we do indeed need that, why not have a nest*ing* depth parameter?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
ilayaperumalg: I will leave this to @aclement @BoykoAlex as I am not sure if **Flo** supports case insensitiveness.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/StreamControllerTests.java
ilayaperumalg: This specific method doesn't look like a business logic IMO as it more looks like search/filter changes at the `controller` side.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
ericbottard: It's not only Flo, it's the whole SCDF ecosystem. It has to be consistent throughout

ilayaperumalg: I think the reasoning here is different in the context of **Flo** (and in this use case) based on my comment [here] (https://github.com/spring-cloud/spring-cloud-dataflow/pull/839/files/547f1477f20639d42b2765f977862fe3cf2c5ec2#r77795124). Also, Flo supports case insensitiveness from the editor.

ericbottard: I don't understand what you mean.
If i'm not mistaken (having had a quick look at the rdbms implementation), I can do
```
stream create foo --definition "time | log"
stream create FOO --definition "some | other | apps"
stream created bar --definition "foo > log"
```

It doesn't make sense then that `FOO` comes up when looking up streams related to `bar`

ilayaperumalg: yeah, that is right. I get it now. Currently, we don't check for case insensitiveness explicitly and so let's not do it here as well :-)

ericbottard:
LGTM

ericbottard:
Rebased, Squashed and Merged as 1cf74b4

PR #850, opened 2016-09-06
Remove dependency on scs-test-support-internal

ericbottard:
Fixes spring-cloud/spring-cloud-dataflow#849

This will ease migrating SCDF to boot 1.4, as boot upgrades its spring-data version


ilayaperumalg:
LGTM, merging.

Issue #855, opened 2016-09-06
Introduce strategy for retrieving metadata


sabbyanandan:
As a developer, I'd like to introduce a strategy for retrieving application metadata. 

**Acceptance:**
- strategy takes both maven and docker artifacts into account

ilayaperumalg:
@sabbyanandan I missed the context for this story. Could you give an example of what kind of metadata we try to retrieve from the application? Is it different from what we currently do via `app info`?

ericbottard:
Nope, this is what we have for `app info`, but introducing a dynamic Strategy for boot apps for which we have access to the file vs Docker (impl TBD). Also, this could have been a way to support boot 1.3 vs 1.4, although the direction I'm heading towards uses dynamic selection a level below (once we know it's boot, selecting the layout as a second level strategy)

Issue #851, opened 2016-09-06
Create “property whitelisting” extension to support 1.3.x apps


sabbyanandan:
Continuing the research efforts from #748, as a developer, I'd like to add support for "property whitelisting" for 1.3.x based Boot apps. 

**Acceptance:**
- Spring Boot's [change](https://github.com/spring-projects/spring-boot/commit/87fe0b2adeef85c842c009bfeebac1c84af8a5d7) from 1.3.x to 1.4.x will be treated as an exception and the extension will be available in SCDF core 
- Any custom specifications to differentiate between the 1.3.x and 1.4.x apps that are end-user responsibility will be documented

Issue #852, opened 2016-09-06
Add test support to validate Boot 1.4.x and 1.3.x compatibility


sabbyanandan:
Continuing the research efforts from #748, as a developer, I'd like to add test coverage to validate 1.3.x and 1.4.x based Boot apps, so we could incrementally test backward compatibility via CI pipelines. 

**Acceptance:**
- Test coverage validates "property whitelisting" feature for both 1.3.x and 1.4.x apps
- Pointers to test infrastructure is included in the docs

PR #863, opened 2016-09-09
Refactor ApplicationMetadataResolver

ericbottard:
* introduce a dedicated subclass for boot 
* introduce a Delegating implementation that relies on supports(Resource app) to dynamically select impl
* refine boot support to cope with 1.3 and 1.4 layouts 
* refactor tests 
* remove ClassLoader related code that is no longer needed (dated back from the ModuleLauncher days)

Fixes spring-cloud/spring-cloud-dataflow#855
Fixes spring-cloud/spring-cloud-dataflow#851


spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/ConfigurationPropertyValueHintRecoveryStrategy.java
```diff
@@ -85,7 +86,7 @@ public void addProposals(String dsl, CheckPointedParseException exception, int d
 
 					File appFile = appResource.getFile();
 					Archive jarFileArchive = appFile.isDirectory() ? new ExplodedArchive(appFile) : new JarFileArchive(appFile);
-					classLoader = new ClassLoaderExposingJarLauncher(jarFileArchive).createClassLoader();
+					classLoader = new BootClassLoaderCreation(jarFileArchive, Thread.currentThread().getContextClassLoader()).createClassLoader();
```
ilayaperumalg: Do we really need to set the entire current context loader as `parent` here? I thought the URLs derived from the `jarFileArchive` are sufficient to create the class-loader and to load the metadata files.

spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/BootApplicationConfigurationMetadataResolver.java
```diff
@@ -0,0 +1,211 @@
...
+	}
+
+	public List<ConfigurationMetadataProperty> listProperties(Archive archive, boolean exhaustive) {
+		ClassLoader moduleClassLoader = null;
+		try {
+
+			List<ConfigurationMetadataProperty> result = new ArrayList<>();
+			moduleClassLoader = new BootClassLoaderCreation(archive, parent).createClassLoader();
+			ResourcePatternResolver moduleResourceLoader = new PathMatchingResourcePatternResolver(moduleClassLoader);
+
```
ilayaperumalg: Are these extra lines for readability? I think they can be avoided and probably move into some private methods if needed.

spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/ConfigurationPropertyValueHintRecoveryStrategy.java
ericbottard: As I'm working on adding tests for this, I'll come up for a better strategy here. But long story short: for the current state of this PR, this is the only way I had the tests pass

spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/BootApplicationConfigurationMetadataResolver.java
ericbottard: I don't think they're new (code copy/pasted). They can be removed

ilayaperumalg:
LGTM, merging while the fix for https://github.com/spring-cloud/spring-cloud-dataflow/pull/863#discussion_r78370305 is taken care at https://github.com/spring-cloud/spring-cloud-dataflow/issues/852

ilayaperumalg:
Squashed the commits, rebased and merged as https://github.com/spring-cloud/spring-cloud-dataflow/commit/543fdfe6dcf5e883e50e8ed992e88f13512221f3 along with fixing the `1.1.0` version 3fae7e9

spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/ConfigurationPropertyValueHintExpansionStrategy.java
```diff
@@ -104,7 +105,7 @@ public boolean addProposals(String text, StreamDefinition parseResult,
 				try {
 					File file = appResource.getFile();
 					Archive jarFileArchive = file.isDirectory() ? new ExplodedArchive(file) : new JarFileArchive(file);
-					classLoader = new ClassLoaderExposingJarLauncher(jarFileArchive).createClassLoader();
+					classLoader = new BootClassLoaderCreation(jarFileArchive, Thread.currentThread().getContextClassLoader()).createClassLoader();
```
markpollack: Should you use Spring's `ClassUtils.getDefaultClassLoader()` here instead?

ilayaperumalg: This will be removed eventually once the fix for #852 is in.

spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/BootClassLoaderCreation.java
```diff
@@ -0,0 +1,97 @@
...
+import org.springframework.boot.loader.archive.Archive;
+import org.springframework.boot.loader.util.AsciiBytes;
+
+/**
+ * Strategy interface for creating a ClassLoader that mimics the one used when
+ * a boot uber-jar runs.
+ *
+ * @author Eric Bottard
+ */
+public class BootClassLoaderCreation {
```
markpollack: Maybe call it BootClassLoaderFactory?

spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/BootClassLoaderCreation.java
```diff
@@ -0,0 +1,97 @@
...
+
+	private final Archive archive;
+
+	private final ClassLoader parent;
+
+	private static final AsciiBytes BOOT_13_LIBS_LOCATION = new AsciiBytes("lib/");
+
+	private static final AsciiBytes BOOT_14_LIBS_LOCATION = new AsciiBytes("BOOT-INF/lib/");
+
+	public BootClassLoaderCreation(Archive archive, ClassLoader parent) {
```
markpollack: ctor javadoc?

spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/DelegatingApplicationConfigurationMetadataResolver.java
```diff
@@ -0,0 +1,56 @@
...
+import org.springframework.core.io.Resource;
+
+/**
+ * An {@link ApplicationConfigurationMetadataResolver} that tries several delegates in turn,
+ * returning the results of the first one that {@link ApplicationConfigurationMetadataResolver#supports(Resource)}
+ * an app. Returns an empty list of properties as a last resort.
+ *
+ * @author Eric Bottard
+ */
+public class DelegatingApplicationConfigurationMetadataResolver extends ApplicationConfigurationMetadataResolver {
```
markpollack: It isn't clear why we need this class as the only place it is used has code with just one delegate, in `ApplicationConfigurationMetadataResolverAutoConfiguration`

```
	@Bean
	@ConditionalOnMissingBean(ApplicationConfigurationMetadataResolver.class)
	public ApplicationConfigurationMetadataResolver metadataResolver() {
		return new DelegatingApplicationConfigurationMetadataResolver(new BootApplicationConfigurationMetadataResolver());
	}
```

ilayaperumalg: I think the idea is to have this in place now and add the upcoming metadata resolvers based on other `resources` that use `docker` etc.,

ericbottard: Correct

PR #872, opened 2016-09-13
Refactor CL creation in metadata resolver

ericbottard:
Allows some corner cases to work in commpletion (mainly, being able
to "see" enum classes that are part of any boot 1.3/1.4 uberjar.)

Fixes spring-cloud/spring-cloud-dataflow#852

### DIFF
--- a/spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/BootApplicationConfigurationMetadataResolver.java
+++ b/spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/BootApplicationConfigurationMetadataResolver.java
@@ -16,13 +16,12 @@
 
 package org.springframework.cloud.dataflow.configuration.metadata;
 
-import java.io.Closeable;
 import java.io.File;
 import java.io.IOException;
+import java.net.URLClassLoader;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Properties;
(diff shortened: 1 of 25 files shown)


spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/BootApplicationConfigurationMetadataResolver.java
```diff
@@ -106,29 +97,18 @@ public boolean supports(Resource app) {
...
-			throw new RuntimeException("");
+			throw new RuntimeException("Failed to list properties for " + app, e);
 		}
 	}
 
 	public List<ConfigurationMetadataProperty> listProperties(Archive archive, boolean exhaustive) {
-		ClassLoader moduleClassLoader = null;
-		try {
-
+		try (URLClassLoader moduleClassLoader = new BootClassLoaderFactory(archive, parent).createClassLoader()) {
```
ilayaperumalg: Since the parent class loader is always 'null' and no longer needed to use any of parent classloaders to load the metadata resources, can we simplify the `BootClassLoaderFactory` as well as `BootApplicationConfigurationMetadataResolver` not to use the `parent` class loader at all? and just use the 'null' option when creating `LaunchedURLClassLoader`?

ericbottard: In some of the tests, it's not `null`

ilayaperumalg: which tests are those that use non-null parent class loaders? I don't seem to find them via usages of the class references.

ericbottard: https://github.com/spring-cloud/spring-cloud-dataflow/pull/872/files/349131983ca59b01e4474481fb2c56ba3140dcfb#diff-0f08760dc293294b1be2538fae1a613fR151 for example. Needed to see the enum that is in `src/test/java` there

ilayaperumalg:
LGTM, merging. Also, verified stream completion and metadata retrieval of spring boot `1.3.x` and `1.4.0` based apps.

ilayaperumalg:
Squashed, rebased and merged as db67b8e

PR #875, opened 2016-09-14
gh-874 Fix OAuth Integration Setup

ghillert:
Resolves https://github.com/spring-cloud/spring-cloud-dataflow/issues/874


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/security/support/OnOAuth2Disabled.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * {@link Condition} that is only valid if {@code security.basic.enabled} is
+ * {@code true} and the property {@code security.oauth2} exists.
+ *
+ * @author Gunnar Hillert
+ * @since 1.1.0
+ *
+ */
+public class OnOAuth2Disabled extends NoneNestedConditions {
```
ericbottard: I don't really understand the relation between the name of this class ("disabled") and the actual condition. Is there a copy/paste error?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/security/support/OnSecurityEnabledAndOAuth2Disabled.java
```diff
@@ -0,0 +1,43 @@
...
+public class OnSecurityEnabledAndOAuth2Disabled extends AllNestedConditions {
+
+	public OnSecurityEnabledAndOAuth2Disabled() {
+		super(ConfigurationPhase.PARSE_CONFIGURATION);
+	}
+
+	@ConditionalOnProperty("security.basic.enabled")
+	static class BasicSecurityEnabledEnabled { }
+
+	@Conditional(OnOAuth2Disabled.class)
```
ericbottard: Likewise here, this is very hard to follow if it is legit: Condition says disabled, but class name says the contrary

jvalkeal:
looks like this restores oauth back to where it was, merging.

markpollack:
Janne, did you fix the naming as pointed out by Eric on merge?

Issue #870, opened 2016-09-12
Provide DataflowTemplate Constructor without the RestTemplate


ghillert:
In order to improve the UX when using `DataFlowTemplate`, we should provide a constructor that does not require a `RestTemplate` to be passed in. Currently, not only must a `RestTemplate` be presently provided, but it also must fulfill the specific requirements of the `DataFlowTemplate`:

```java
	@Bean
	public static RestTemplate restTemplate() {
		RestTemplate restTemplate = new RestTemplate();
		restTemplate.setErrorHandler(new VndErrorResponseErrorHandler(restTemplate.getMessageConverters()));
		for(HttpMessageConverter<?> converter : restTemplate.getMessageConverters()) {
			if (converter instanceof MappingJackson2HttpMessageConverter) {
				final MappingJackson2HttpMessageConverter jacksonConverter = (MappingJackson2HttpMessageConverter) converter;
				jacksonConverter.getObjectMapper().registerModule(new Jackson2HalModule());
				jacksonConverter.getObjectMapper().addMixIn(JobExecution.class, JobExecutionJacksonMixIn.class);
				jacksonConverter.getObjectMapper().addMixIn(JobParameters.class, JobParametersJacksonMixIn.class);
				jacksonConverter.getObjectMapper().addMixIn(JobParameter.class, JobParameterJacksonMixIn.class);
				jacksonConverter.getObjectMapper().addMixIn(JobInstance.class, JobInstanceJacksonMixIn.class);
				jacksonConverter.getObjectMapper().addMixIn(ExitStatus.class, ExitStatusJacksonMixIn.class);
				jacksonConverter.getObjectMapper().addMixIn(StepExecution.class, StepExecutionJacksonMixIn.class);
				jacksonConverter.getObjectMapper().addMixIn(ExecutionContext.class, ExecutionContextJacksonMixIn.class);
				jacksonConverter.getObjectMapper().addMixIn(StepExecutionHistory.class, StepExecutionHistoryJacksonMixIn.class);
			}
		}
		return restTemplate;
	}
```

This configuration is taken from the Spring Cloud Dataflow Shell module.

ghillert:
See also: pring-cloud/spring-cloud-dataflow/issues/871

Issue #871, opened 2016-09-12
DataflowTemplate - Assert that passed-in RestTemplate fulfills requirements


ghillert:
As described in https://github.com/spring-cloud/spring-cloud-dataflow/issues/870, a `RestTemplate` with specific requirement is needed when creating a new `DataflowTemplate`. 

When a `RestTemplate` is passed into a `DataflowTemplate`, we should make assertion whether the passed in `RestTemplate` fulfills the requirements. Currently, when someone passes naïvely a `new RestTemplate()`, the user will end up with `NullPointer` exceptions.

ericbottard:
This is no longer an issue given the way we took care of #870

PR #876, opened 2016-09-14
gh-870 Provide DataflowTemplate Constructor without the RestTemplate

ghillert:
Resolves https://github.com/spring-cloud/spring-cloud-dataflow/issues/870


ericbottard:
I think it would be better to accept any RestTemplate, and apply the mixins that we want.
This would take care of https://github.com/spring-cloud/spring-cloud-dataflow/issues/871 as well

ghillert:
@ericbottard Made some changes - If a RestTemplate is provided, the MixIns will be applied now. Also provide substantially more JavaDoc documentation and also provide JUnit Tests for the DataFlowTemplate. I hope this satisfies most use-cases.

ericbottard:
Getting test errors on `testDataFlowTemplateContructorWithNonExistingUri` (different error message than expected, unknown host vs timeout). I will remove that part of the test and merge.

ericbottard:
Merged with a couple of small changes as 762878e

PR #881, opened 2016-09-15
Fix REST Docs, broken because of disabled analytics

ericbottard:


ericbottard:
@ilayaperumalg You break it, you merge it ;)

ilayaperumalg:
@ericbottard sure ;). Before that help me understand the problem. What is actually broken and how can I see it is broken?

ericbottard:
The REST docs (that only get generated with `-Pfull`, hence you not realizing) need analytics enabled.

So, to reproduce: checkout master, `./mvnw clean install -Pfull`

PR #891, opened 2016-09-19
Maintain order of related streams

BoykoAlex:
Flo requires stream definitions to come in specific order to be displayed correctly graphically. Server side already assembles stream definitions in the acceptable order, but returns a `HashSet` rather than a `LinkedHashSet`.
The fix is switch to `LinkedHashSet`


trisberg:
Merged as commit 15d07016cc958d4c5101bde352aeefeddd21fc55

PR #897, opened 2016-09-26
ui-gh-118 Add form-based non-oauth login page

ghillert:
* Provide server-side functionality, migrated from Spring XD
* See also respective UI PR

Resolves https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/118


jvalkeal:
Logging out when oauth is enabled goes to page `http://localhost:9393/dashboard/logout-success-oauth.html` which doesn't exists.

```
Whitelabel Error Page

This application has no explicit mapping for /error, so you are seeing this as a fallback.

Wed Sep 28 12:48:38 BST 2016
There was an unexpected error (type=Not Found, status=404).
No message available
```

Other than that LGTM and I can even logout and login with a diffrent user. Even file/ldap work if both are enabled. Nice!

jvalkeal:
Squashed and merged per https://github.com/spring-cloud/spring-cloud-dataflow/commit/dc604f9da42bb3ba7b40240928ee40d6271972a0

PR #901, opened 2016-09-28
DefaultTaskService to use DataSourceProperties

### LINKED ISSUE #879: Tasks don't always get the correct driver-class-name passed in
Seems like the default driver `org.h2.Driver` gets passed in even when the SCDF server is started using `--spring.datasource.driver-class-name=org.mariadb.jdbc.Driver`.  This doesn't happen when the SCDF server is started with `--spring.datasource.driverClassName=org.mariadb.jdbc.Driver`. 

Something is off the way these properties are handled.

ilayaperumalg:
- Instead of using the property string, autowire the already configured DataSourceProperties

This resolves #879


cppwfs:
LGTM tested using spring.datasource.driver-class-name successfully

cppwfs:
Rebased, added changes merged.

PR #929, opened 2016-10-06
Add REST-API to fetch an existing task definition

### LINKED ISSUE #886: Add REST-API to fetch an existing task definition
As a user, I'd like to have an API to fetch an existing task definition. Currently, when requesting `http://localhost:9393/tasks/definitions/foo` for a task definition with the name `foo`, I get `Request method 'GET' not supported` error message. 

**Acceptance:**
- I should be able to resolve and fetch the definition of the task by hitting `http://localhost:9393/tasks/definitions/foo`
- The new API is documented
- Tests included

cppwfs:
resolves #886


ericbottard:
Rebased and merged as c4fd047

Issue #438, opened 2016-03-07
Shell: UX issue - standardize app type/name order


ghillert:
Currently the following commands exist:

`app unregister timestamp:task`
`app info task:timestamp`

We should standardize the order of the module type/name combo.

ericbottard:
Agreed. Note that for completion purposes, it may be better to use `<type>:<name>`

ghillert:
+1 should be `<type>:<name>`

PR #931, opened 2016-10-07
Harmonize shell commands to use id=type:name notation

ericbottard:
Fixes https://github.com/spring-cloud/spring-cloud-dataflow/issues/438

### DIFF
--- a/spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/AppRegistryCommands.java
+++ b/spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/AppRegistryCommands.java
@@ -166,16 +166,15 @@ public class AppRegistryCommands implements CommandMarker, ResourceLoaderAware {
 			@CliOption(mandatory = true,
 					key = {"", "name"},
 					help = "name of the application to unregister")
-			String name,
-			@CliOption(mandatory = false,
+					String name,
+			@CliOption(mandatory = true,
 					key = {"type"},
 					help = "type of the application to unregister")
-			ApplicationType type) {
+					ApplicationType type) {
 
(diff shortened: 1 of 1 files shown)


jvalkeal:
Shouldn't we change `register` as well as we now use:
```
dataflow:>app register --name foo --type sink --uri http://foo.com/foo
Successfully registered application 'sink:foo'

dataflow:>app unregister --name foo --type sink
Successfully unregistered application 'foo' with type sink
```

I find in confusing if we change only `unregister` command.

ericbottard:
Not sure. I think it could be ok to add a `--id <type>:<name>` as an _alternative_. To me, the first driver for id in the unregister case, is so that we can easily autocomplete an existing id (because Spring Shell has no way of completing `--name` on its own, even if `--type` would have been set earlier). This does not apply to the register case (we could autocomplete the first part of `--id` though).

markpollack:
Well despite the fact that it might help with autocomplete to use the id format for unregister since we can provide that guidance, it makes for a real asymmetry between register and unregister, which should be the most similar vs. the info command.  I'd actually suggest just leaving it as it is.  I thought the original motivation for this issue was that in some places we had used `type:name` and others `name:type`.

ericbottard:
> I thought the original motivation for this issue was that in some places we had used type:name and others name:type.

We kind of have, as the current implementation uses `--name` in a lax way and accepts `<name>:<type>` in it. So at the minimum we should get rid of that behavior IMHO

markpollack:
that sound good to me.   I don't think there are any folks clamoring for an overload `id` option for reg/unreg that takes type:name - right?

ericbottard:
@markpollack Changed the PR to reflect latest discussion. Now `--name` is just a (mandatory) name

markpollack:
merged as 3fba890fc8405d9fb0dfecec6b0b885ae3935d46

PR #945, opened 2016-10-14
Add dashboard link to RootController

### LINKED ISSUE #807: Dashboard link is not present in root HATEOAS link page.
Just like there is a link for stream/definitions

```
_links: {
streams/definitions: {
href: "http://mlp-dataflow-server.cfapps.io/streams/definitions"
},
```
and so on, there should be a link to the dashboard.

ilayaperumalg:
- Resolves #807

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/security/OAuthSecurityConfiguration.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/security/OAuthSecurityConfiguration.java
@@ -17,6 +17,7 @@ package org.springframework.cloud.dataflow.server.config.security;
 
 import org.springframework.boot.autoconfigure.security.oauth2.client.EnableOAuth2Sso;
 import org.springframework.cloud.dataflow.server.config.security.support.OnSecurityEnabledAndOAuth2Enabled;
+import org.springframework.cloud.dataflow.server.controller.UiController;
 import org.springframework.cloud.dataflow.server.service.impl.ManualOAuthAuthenticationProvider;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Conditional;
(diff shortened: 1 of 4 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/RootController.java
```diff
@@ -81,6 +81,7 @@ public RootController(EntityLinks entityLinks) {
 	@RequestMapping("/")
 	public ResourceSupport info() {
 		ResourceSupport resourceSupport = new ResourceSupport();
+		resourceSupport.add(new Link("/dashboard", "dashboard"));
```
ericbottard: I know this is not strictly part of this PR, but we should try to limit the occurence of the magic string "/dashboard".
So maybe introduce a CONSTANT for it in `UiController` and reference that here, and also in `OAuthSecurityConfiguration` and `BasicAuthSecurityConfiguration`

ilayaperumalg: Good idea! will fix that here.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/UiController.java
```diff
@@ -28,9 +28,11 @@
 @Controller
 public class UiController {
 
-	@RequestMapping("/dashboard")
+	public static final String WEB_UI_INDEX_PAGE_NAME = "dashboard";
```
ericbottard: I would make the starting slash part of the constant, that 1) makes more sense (it's a part, not just a "word") and 2) would simplify its usage quite a bit

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/security/OAuthSecurityConfiguration.java
```diff
@@ -42,12 +43,12 @@ protected void configure(HttpSecurity http) throws Exception {
 		http.antMatcher("/**")
 		.authorizeRequests()
 			.antMatchers(
-				"/security/info**", "/login**", "/dashboard/logout-success-oauth.html",
-				"/dashboard/styles/**", "/dashboard/images/**", "/dashboard/fonts/**",
-				"/dashboard/lib/**").permitAll()
+					"/security/info**", "/login**", "/"+ UiController.WEB_UI_INDEX_PAGE_NAME +"/logout-success-oauth.html",
```
ericbottard: creating a method that does the concatenantion (maybe even accepting varargs) would make this more readable IMO

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/security/OAuthSecurityConfiguration.java
```diff
@@ -43,15 +45,23 @@ protected void configure(HttpSecurity http) throws Exception {
...
+					"/security/info**", "/login**", StringUtils.arrayToCommaDelimitedString(appendIndexPageRoute("/logout-success-oauth.html",
+							"/styles/**", "/images/**", "/fonts/**", "/lib/**"))).permitAll()
 			.anyRequest().authenticated()
 		.and().httpBasic()
-		.and().logout().logoutSuccessUrl("/"+ UiController.WEB_UI_INDEX_PAGE_NAME +"/logout-success-oauth.html")
+		.and().logout().logoutSuccessUrl(UiController.WEB_UI_INDEX_PAGE_ROUTE + "/logout-success-oauth.html")
 		.and().csrf().disable();
 	}
 
+	private String[] appendIndexPageRoute(String... patterns) {
```
ericbottard: Will change as well, but this is not technically the "index" page route. It's the whole *path* to the UI app.

ericbottard:
@ilayaperumalg WDYT about https://github.com/ericbottard/spring-cloud-dataflow/commit/45458378e08867db709386e813787551fe76ea90 ?

ilayaperumalg:
@ericbottard looks much better.

ericbottard:
> @ericbottard looks much better.
Thank you, I just turned 38.

ericbottard:
All jokes aside, I will merge this then

ericbottard:
Merged as 1d1ce57

ericbottard:
Grrrr, I updated the actual link to dashboard just after I ran a full build, and actually introduced a test failure. Will fix in a new commit

PR #950, opened 2016-10-14
Support YAML based propertiesFile option

### LINKED ISSUE #288: Study the YAML file approach for deployment properties
As a developer, I'd like to study the YAML file approach to resolve deployment properties, so I don't have to send the properties as inline params, which is very verbose and error prone.

Refer to #224 and #271 for more details.

### LINKED ISSUE #889: Explore options to simplify stream binding deployment properties
As a developer, I'd like to study approaches to simplify the use of stream binding properties at the time of stream deployment.

ilayaperumalg:
- Stream deploy command now supports YAML file as `--propertiesFile` option

This resolves #889
This fixes #288

### DIFF
--- a/spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/StreamCommands.java
+++ b/spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/StreamCommands.java
@@ -24,12 +24,17 @@ import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Properties;
 
+import org.apache.commons.io.FilenameUtils;
+
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.config.YamlPropertiesFactoryBean;
+import org.springframework.boot.env.YamlPropertySourceLoader;
 import org.springframework.cloud.dataflow.rest.client.DataFlowOperations;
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/StreamCommands.java
```diff
@@ -116,9 +121,19 @@ public String deployStream(
 				propertiesToUse = DeploymentPropertiesUtils.parse(properties);
 				break;
 			case 1:
-				Properties props = new Properties();
-				try (FileInputStream fis = new FileInputStream(propertiesFile)) {
-					props.load(fis);
+				String extension = FilenameUtils.getExtension(propertiesFile.getName());
```
ericbottard: It may (or may not) be worth having a look at the `PropertySourceLoader` class and friends in boot

ilayaperumalg: The `PropertySourceLoader` is useful when we deal with the property sources directly. But in this case, we need to get the `Properties` type. Hence, the current approach is slightly better.

markpollack:
edited docs and removed unused import.  Merged as 81e0c2ad5befd27c8bfd29a4f58eb4e8e74190e3

Issue #681, opened 2016-06-10
Document how to incorporate the spring-boot-admin project


markpollack:
This will bring in some nice functionality.  See https://github.com/codecentric/spring-boot-admin

Issue #632, opened 2016-05-31
Create individual security section vs. including in getting started


markpollack:
Security is an important enough topic to have a separate section.

PR #965, opened 2016-10-24
Add documentation for Managment and Monitoring

markpollack:
* Describe how to use Spring Boot Admin
* Move security and database configuration to a new configuraiton section
* Remove executable file permission on some png images

Resolves spring-cloud/spring-cloud-dataflow#681
Resolves spring-cloud/spring-cloud-dataflow#632


spring-cloud-dataflow-docs/src/main/asciidoc/index.adoc
```diff
@@ -22,6 +22,7 @@ include::preface.adoc[]
 include::spring-cloud-dataflow-overview.adoc[]
 include::architecture.adoc[]
 include::getting-started.adoc[]
+include::configuration.adoc[]
```
ericbottard: I think this file will need to be included in other projects (eg CF, YARN, k8s, etc) as well

PR #967, opened 2016-10-25
Add property to disable embedded db

### LINKED ISSUE #926: DataSource and H2 server configuration looks wrong
There are 2 alternative configurations for the data source transaction manager and initialization configuration. One sets up an H2 embedded server as well, and the server has an extra (different) condiftion. Both have identical bean definitions (with different names) for the transaction manager and data source initializer, so the duplication is confusing. Also the two configurations and guarded by conditions that are not the inverse of each other.

All of this might be intentional, but it's hard to reason about because the conditions are long:

In `DataFlowServerConfiguration` in the H2 server bean:

```
@ConditionalOnExpression("#{'${spring.datasource.url:}'.startsWith('jdbc:h2:tcp://localhost:') && '${spring.datasource.url:}'.contains('/mem:')}")
```

On the config class surrounding it:

```
@ConditionalOnExpression("#{'${spring.datasource.url:}'.startsWith('jdbc:h2:tcp://localhost:') && '${spring.datasource.url:}'.contains('/mem:')}")
```

(these look identical, so one is either redundant or wrong).

And on the other configuration class, where there is a transaction manager, and an initializer:

```
@ConditionalOnExpression("#{!'${spring.datasource.url:}'.startsWith('jdbc:h2:tcp://localhost:') && !'${spring.datasource.url:}'.contains('/mem:')}")
```

So if there is an H2 database with TCP but not in memory, then I think this means the database is never initialized (it's tricky to know what was intended, but maybe it is better to always initialize and ignore errors in that case?).

There is also a `TaskConfiguration` with the same pair of conditions:

```
@ConditionalOnExpression("#{'${spring.datasource.url:}'.startsWith('jdbc:h2:tcp://localhost:') && '${spring.datasource.url:}'.contains('/mem:')}")
```

and

```
@ConditionalOnExpression("#{!'${spring.datasource.url:}'.startsWith('jdbc:h2:tcp://localhost:') && !'${spring.datasource.url:}'.contains('/mem:')}")
```

So if there is an H2 database with TCP but not in memory, then there is no `JobRepository`. This seems like a mistake (and it causes the app to barf on start up).

I think make some custom annotations for the conditions could be used to make their intention clearer. Also some javadocs might help. And there should always be a `JobRepository` right?

cppwfs:
resolves #926

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/configuration.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/configuration.adoc
@@ -74,6 +74,13 @@ upgrading from version `1.0.x` to `1.1.x`.  Migration scripts for specific
 database types can be found
 https://github.com/spring-cloud/spring-cloud-task/tree/master/spring-cloud-task-core/src/main/resources/org/springframework/cloud/task/migration[here].
 
+NOTE: If you wish to use an external H2 database instance instead of the one
+embedded with Spring Cloud Data Flow set the
+`spring.dataflow.embedded.database.enabled` property to false.  If
+`spring.dataflow.embedded.database.enabled` is set to false or a database
+other than h2 is specified as the datasource the embedded database will not
+start.
+
 [[configuration-security]]
 == Security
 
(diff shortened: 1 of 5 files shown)


spring-cloud-dataflow-docs/src/main/asciidoc/getting-started.adoc
```diff
@@ -206,6 +206,10 @@ upgrading from version `1.0.x` to `1.1.x`.  Migration scripts for specific
 database types can be found
 https://github.com/spring-cloud/spring-cloud-task/tree/master/spring-cloud-task-core/src/main/resources/org/springframework/cloud/task/migration[here].
 
+NOTE: If you wish to use an external H2 database instance instead of the one
```
mminella: Is setting this property to false required to use *any* external db or only an external H2?

cppwfs: Just the H2

cppwfs:
Added tests and updated docs based on code review.  Also moved the docs to the new location in configuration.adoc.

PR #973, opened 2016-10-27
Completion for task

aclement:
This is submitted for feedback, not necessarily immediate integration. Basically I've copied the pattern from stream content assist. Due to them not sharing a parser this meant a bunch of very similar classes getting created.  It may be possible to refactor what is there for stream to support both task and stream but they would get uglier and possibly when composed task DSL comes along the fact that task completion is separate _might_ be useful.  I've confirmed that this all works with client side changes too.

scdf_27oct

### DIFF
--- a/spring-cloud-dataflow-completion/src/test/java/org/springframework/cloud/dataflow/completion/TaskCompletionProviderTests.java
+++ b/spring-cloud-dataflow-completion/src/test/java/org/springframework/cloud/dataflow/completion/TaskCompletionProviderTests.java
@@ -1,0 +1,246 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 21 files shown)


ericbottard:
Fixes #921

spring-cloud-dataflow-completion/src/test/java/org/springframework/cloud/dataflow/completion/TaskCompletionProviderTests.java
```diff
@@ -0,0 +1,246 @@
...
+		assertThat(completionProvider.complete("foo --some-option=prefix", 1), empty());
+	}
+
+	/*
+	 * http --use-ssl=tr<TAB> => must be true or false, no need to present "...=tr --other.prop"
+	 */
+	@Test
+	public void testClosedSetValuesShouldBeExclusive() {
+		assertThat(completionProvider.complete("http --use-ssl=tr", 1), not(hasItems(
+				proposalThat(startsWith("http --port=12 --use-ssl=tr "))
```
ericbottard: This seems off, the `--port=12` part should come after. May be from stream completion. Will check and fix myself while merging

ericbottard:
Merged with a couple of fixes to javadoc and tests as 38f41ba

PR #981, opened 2016-11-02
Fix spring-cloud-dataflow-dependencies BOM

### LINKED ISSUE #976: `spring-cloud-dataflow-dependencies` is not a valid BOM
The bom for Spring Cloud Dataflow is invalid. It is pretty much the same thing as https://github.com/spring-cloud/spring-cloud-task/issues/177

You need a BOM where you provide dependency management for only spring-cloud-dataflow. Any opinion you have on Spring Cloud or other dependencies could prevent to use a different version as you're actually forcing that Spring Cloud version to be used.

ilayaperumalg:
- Move external dependencies out of `spring-cloud-dataflow-dependencies` BOM and to `spring-cloud-dataflow` parent
 - Fix BOM only to have dataflow project dependencies
 - Use `project.version` wherever appropriate

Resolves #976

### DIFF
--- a/spring-cloud-dataflow-dependencies/pom.xml
+++ b/spring-cloud-dataflow-dependencies/pom.xml
@@ -12,139 +12,62 @@
 	<packaging>pom</packaging>
 	<name>spring-cloud-dataflow-dependencies</name>
 	<description>Spring Cloud Data Flow Dependencies BOM designed to support consumption of Spring Cloud Data Flow from the Spring Initializr.</description>
-	<properties>
-		<spring-analytics.version>1.1.0.M1</spring-analytics.version>
-		<spring-cloud-task.version>1.1.0.BUILD-SNAPSHOT</spring-cloud-task.version>
-		<spring-cloud-commons.version>1.1.3.RELEASE</spring-cloud-commons.version>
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-dependencies/pom.xml
```diff
@@ -12,139 +12,62 @@
...
-			<dependency>
-				<groupId>org.springframework.analytics</groupId>
-				<artifactId>spring-analytics</artifactId>
-				<version>${spring-analytics.version}</version>
-			</dependency>
-			<dependency>
-				<groupId>org.springframework.cloud</groupId>
 				<artifactId>spring-cloud-dataflow-shell</artifactId>
-				<version>1.1.0.BUILD-SNAPSHOT</version>
+				<version>${project.version}</version>
```
snicoll: Any reason why you don't stick with the hard-coded value? This is replaced automatically on release and is clearer when reading the pom IMO

ilayaperumalg: @snicoll there is no specific reason for that. We could go by using explicit version name for better readability. thanks!

PR #986, opened 2016-11-03
Separate deployer build versions

### LINKED ISSUE #985: Different deployer build versions
`spring-cloud-starter-dataflow-server-local/pom.xml` define `spring-cloud-deployer-resource-maven` and `spring-cloud-deployer-local` and uses `spring-cloud-deployer.version`. After local deployer has been moved to its own repo, we should have own version for these.

jvalkeal:
- Use different build properties for deployer spi and
  local versions as those have a different lifecycle.
- Fixes #985

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -26,6 +26,7 @@
 		<java.version>1.7</java.version>
 		<spring-cloud-dataflow-ui.version>1.1.0.M2</spring-cloud-dataflow-ui.version>
 		<spring-cloud-deployer.version>1.1.0.BUILD-SNAPSHOT</spring-cloud-deployer.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -26,6 +26,7 @@
 		<java.version>1.7</java.version>
 		<spring-cloud-dataflow-ui.version>1.1.0.M2</spring-cloud-dataflow-ui.version>
 		<spring-cloud-deployer.version>1.1.0.BUILD-SNAPSHOT</spring-cloud-deployer.version>
```
trisberg: should this be `spring-cloud-deployer-spi.version` now?

jvalkeal: I think we loosely follow git repo name for these versions. `spring-cloud-deployer.version` is used elsewhere so we prob should not change it.

PR #994, opened 2016-11-07
Reference Spring Cloud Release train directly vs. individual jars

### LINKED ISSUE #988: Reference Spring Cloud Release train directly vs. individual jar versions
We should reference a spring cloud release train bom to pick up associate spring cloud libraries vs specific versions as listed in https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/pom.xml#L31

markpollack:
* Add additional plugin versions to avoid maven project warnings

Fixes #988


markpollack:
Fixes #988

PR #1003, opened 2016-11-09
Use explicit version in dataflow dependencies in BOM

ilayaperumalg:


spring-cloud-dataflow-dependencies/pom.xml
```diff
@@ -17,57 +17,57 @@
...
 				<groupId>org.springframework.cloud</groupId>
 				<artifactId>spring-cloud-dataflow-server-core</artifactId>
-				<version>${project.version}</version>
+				<version>1.1.0.BUILD-SNAPSHOT</version>
 			</dependency>
 			<dependency>
 				<groupId>org.springframework.cloud</groupId>
 				<artifactId>spring-cloud-dataflow-server-local</artifactId>
-				<version>${project.version}</version>
+				<version>1.1.0.BUILD-SNAPSHOT</version>
```
jvalkeal: I tried with `./mvnw versions:set -DnewVersion=1.2.3.RELEASE` and `spring-cloud-dataflow-server-local` was kept in `1.1.0.BUILD-SNAPSHOT`. and everything else got changed to `1.2.3.RELEASE`. This is probably because `spring-cloud-dataflow-server-local` has a different parent. I'm not 100% sure if maven's versions plugin does exactly same as that happens during a release.

jvalkeal:
ok, I'm gonna merge this. Lets hope those version issue above ^^^ is caused by maven versions plugin and bamboo release plugin works better.

ericbottard:
I don't see why the release mechanism would change those versions? How can it know the intent? Also, why are we changing this in the first place? Using `${project.version}` more clearly conveys the intent of "use the version of this very project" more clearly than the actual value IMHO

ilayaperumalg:
There is also a related discussion here: https://github.com/spring-cloud/spring-cloud-dataflow/pull/981#discussion_r86301755

Issue #958, opened 2016-10-18
Switch to asciidoctor's pdf generation


markpollack:
Getting a bunch of errors in the doc generation process, see https://build.spring.io/browse/SCD-BMASTER-JOB1-843/log

PR #1009, opened 2016-11-11
gh-958 Use Asciidoctor's PDF generation

ghillert:
* Remove docbook-based Maven tool-chain
* Ensure Spring Cloud Data Flow logo shows up on PDF cover-page
* Ensure Spring Cloud Data Flow version-number shows up on PDF cover-page
* Ensure that PDF and HTML docs are bundled up correctly

Resolves https://github.com/spring-cloud/spring-cloud-dataflow/issues/958


ghillert:
Just for reference - Looks like there are a ton of fixes in the underlying `asciidoctor-pdf` https://github.com/asciidoctor/asciidoctorj/issues/505 incl. better support for `SVG`.

ericbottard:
The `docbook` should be removed, shouldn't it? So should the `index-docinfo.xml` file

ericbottard:
We definitely need to figure out what's wrong with Part (title 1) titles. They're lacking a pagebreak before:
screen shot 2016-11-14 at 12 43 39

spring-cloud-dataflow-docs/pom.xml
```diff
@@ -361,13 +214,16 @@
 								</goals>
 								<configuration>
 									<target>
+										<copy file="${project.build.directory}/generated-docs/index.pdf" tofile="${project.build.directory}/contents/reference/pdf/spring-cloud-dataflow-reference.pdf" />
+										<copy todir="${project.build.directory}/contents/reference/html">
```
ericbottard: Could even argue that the assembly plugin is a more natural fit...

ericbottard:
See https://github.com/asciidoctor/asciidoctor-pdf/issues/74

ericbottard:
Could force a page break with `<<<` for now: http://asciidoctor.org/docs/user-manual/#page-break

ghillert:
Looks like there is a way to configure custom dependency versions:

https://github.com/asciidoctor/asciidoctor-maven-examples/blob/master/asciidoctor-pdf-with-theme-example/pom.xml

ghillert:
Made some changes:

* Remove docbook contents
* Add basic theme-template
* Polish theme

Not sure yet on the section titles.

Also, not sure about using the `generated-docs` directly - Asciidoctor is placing some not-needed files there.

ericbottard:
> Also, not sure about using the generated-docs directly - Asciidoctor is placing some not-needed files there.

Still, I think we should follow the principle of least astonishment. We can cherry pick the files we want in the zip easily.

Will have a look at this new commit and maybe incorporate the final changes myself (looks like conditional pagebreaks is the best workaround for now)

spring-cloud-dataflow-docs/pom.xml
```diff
@@ -168,6 +168,8 @@
 									<backend>pdf</backend>
 									<sourceHighlighter>rouge</sourceHighlighter>
 									<attributes>
+										<pdf-stylesdir>${project.basedir}/src/main/asciidoc/pdf-theme</pdf-stylesdir>
```
ericbottard: We could move the theme out of the `asciidoc` directory (maybe as a sibling `asciidoc-pdf-theme`) so that it does not end up in the `generated-docs` dir

ericbottard:
Merged as f707c9b, which includes https://github.com/ericbottard/spring-cloud-dataflow/commit/2e1b093a9765261bf8fd41ff578f106219a1bf4b

Seems like asciidoctor-pdf-alpha13 includes a fix for pagebreaks. Sadly, its -J version is not published on maven yet.

PR #1025, opened 2016-11-18
Add documentation for custom stream applications

### LINKED ISSUE #815: Add docs section for creating custom source, processor, sink and task applications

ilayaperumalg:
- Doc with examples and steps

Resolves #815 (for streams)

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
@@ -158,6 +158,428 @@ Make sure to add 'spring-boot-configuration-processor' as an optional dependency
 ----
 ====
 
+[[custom-applications]]
+== Creating custom applications
+
+While there are out of the box source, processor, sink applications available, one can extend these applications or write a custom link:https://github.com/spring-cloud/spring-cloud-stream[Spring Cloud Stream] application.
+
+To create any Spring Cloud Stream application, based on the requirement one or more Spring Cloud Stream binder dependencies need to be added as dependencies for the project.
+
+If there are multiple binders used by the application, refer the instructions <<passing_producer_consumer_properties>> on how to configure them.
+
+In the following section, step by step instructions are provided with an example on how to create custom source, processor and sink applications.
+
+[[custom-source-application]]
+=== Creating custom source
+
+Create a Spring Cloud Stream "Source" application that is responsible for sending the source events into the stream.
+
+To get started, let's assume this custom source application emits text messages at a configurable time interval and uses Spring Cloud Stream Kafka binder to send the messages to Kafka.
+
+The source application maven POM may look like this:
+
+[source,xml]
+----
+     <groupId>source</groupId>
+        <artifactId>custom-source</artifactId>
+        <version>1.0-SNAPSHOT</version>
+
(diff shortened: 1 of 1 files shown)


spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -158,6 +158,428 @@ Make sure to add 'spring-boot-configuration-processor' as an optional dependency
...
+To create any Spring Cloud Stream application, based on the requirement one or more Spring Cloud Stream binder dependencies need to be added as dependencies for the project.
+
+If there are multiple binders used by the application, refer the instructions <<passing_producer_consumer_properties>> on how to configure them.
+
+In the following section, step by step instructions are provided with an example on how to create custom source, processor and sink applications.
+
+[[custom-source-application]]
+=== Creating custom source
+
+Create a Spring Cloud Stream "Source" application that is responsible for sending the source events into the stream.
```
mbogoevici: Shouldn't we point them to Initializr ? What's the preferred way of creating apps now ?

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -158,6 +158,428 @@ Make sure to add 'spring-boot-configuration-processor' as an optional dependency
...
+----
+
+Now, we can register this custom source application with Spring Cloud Data Flow server as follows:
+
+```
+dataflow:>app register --name text-emitter --type source --uri maven://source:custom-source:jar:exec:1.0-SNAPSHOT
+Successfully registered application 'source:text-emitter'
+```
+For the custom applications, the maven 'artifacts' need to be available in the 'Remote' maven repository configured.
+If the application isn't yet uploaded into a remote repository and to proceed with locally installed maven artifact one can
```
mbogoevici: We should explain that this only works for the 'local' deployer.

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -158,6 +158,428 @@ Make sure to add 'spring-boot-configuration-processor' as an optional dependency
 ----
 ====
 
+[[custom-applications]]
```
mbogoevici: This can be simplified by pointing to the stream documentation for details on how to actually create Spring Cloud Stream applications. Also, registration of custom apps and creating streams is already covered by other sections in the documentation, so not sure why do we need this content again ? A simple paragraph pointing to the SCSt app tutorial should suffice ?

mbogoevici:
Added a simplified version here: https://github.com/mbogoevici/spring-cloud-dataflow/tree/custom-apps-doc. I think we should make the step-by-step guide part of the samples or move it into an annex.

markpollack:
I think the simplied doc is good enough for now.  I'd suggest having the spring cloud stream docs include a custom processor (uppercase) and then clearly label each section in those docs for a source, processor, sink.  that can be done out of band with this change.

mbogoevici:
Closed as a shortened version in: https://github.com/spring-cloud/spring-cloud-dataflow/commit/dc5199ac88f5a3a152a1dbd0de2801a53e39f080

Issue #1027, opened 2016-11-18
Bring back docbook-based Maven tool-chain


sabbyanandan:
As a developer, I'd like to add the dockbook-based maven tool-chain, so we could support multi-page documentation format. We would have to undo what was addressed via: https://github.com/spring-cloud/spring-cloud-dataflow/pull/1009.

**Acceptance:**
- `/html` and `/htmlsingle` doc variants are available
- `/html` includes "multi-page" doc format
- `/htmlsingle` includes "single-page" doc format with TOC on the left pane

ghillert:
Hopefully we can revisit that once the following Asciidoctor issue has been resolved: https://github.com/asciidoctor/asciidoctor/issues/626

PR #1031, opened 2016-11-18
gh-1027 Bring back docbook-based Maven tool-chain

ghillert:
This reverts commit f707c9b9582874bacbe76ba456ac2fe8b9034c39 (gh-958 Use Asciidoctor's PDF generation)

Resolves https://github.com/spring-cloud/spring-cloud-dataflow/issues/1027


ericbottard:
Merged as a016e41 which also:
- removes asciidoctor-pdf generation
- removes docbook htmlsingle generation
- replaces that in the zip by the result of asciidoctor html5 backend (with left TOC)

PR #1033, opened 2016-11-21
Update banner.txt at server core

### LINKED ISSUE #1008: Modify the banner.txt in core to use  MANIFEST.MF values
We should change it to pull info from MANIFEST.MF using:

`${application.title} ${AnsiColor.GREEN}${application.formatted-version}${AnsiColor.DEFAULT}`

That way it is easier to manage this from the generated app

ilayaperumalg:
- corresponding fixes are at Data Flow Server implementations

Fixes #1008


ericbottard:
I really think we should either
- have only one banner in core and none in the server implementations (and it works as we want, which I'm not sure of)
or
- have banners ONLY in server impls (including local) but NOT in core

ericbottard:
And then I saw the other PRs.... 

Sorry :(

https://github.com/spring-cloud/spring-cloud-dataflow-server-kubernetes/pull/145/files

ilayaperumalg:
Yeah, I think with this PR we are going with option (1)

ericbottard:
Merged as fbbbf13

PR #1037, opened 2016-11-22
Update app.info.description to be maven project.name

markpollack:


trisberg:
We need to change the pom as well to have

```
	<name>Spring Cloud Data Flow Local Server</name>
```

ilayaperumalg:
LGTM, merged as aee2766589e79039aa1b00f7dbf7b39ed4b1d6b1 with the change to fix the `project.name`. Also, verified the updates in the UI -> About page.

Issue #974, opened 2016-10-28
Introduce dedicated prefix for props that are meant for the *deployer*


ericbottard:
Currently, any **app** property that starts with `spring.cloud.deployer.` is re-routed to the deployer used to deploy that app (see [here](https://github.com/spring-cloud/spring-cloud-dataflow/blob/50a8d6b57388c58baf5e310f9e348f02461bb6db/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/DataFlowAppDefinition.java#L120-L138)).

We propose to introduce an explicit **separate** bucket when issuing the `stream deploy` command (and possibly `task launch`) that would be dedicated for props that target the deployer:

```
deployer.<appname>.xxx
```

As a consequence, any property like `spring.cloud.deployer.foo` that is passed as an **app** property would then indeed reach the app (and not the deployer). This is a desired behavior for apps like the task-launcher sink.

PR #1047, opened 2016-11-30
[1.2.x] Introduce dedicated prefix for props that are meant for the *deployer*

ericbottard:
Changes from current behavior:

* deployment properties are now split between
  * `app.[appname].key = value` that go to the _app_
   * `deployer.[appname].key = value` that go to the _deployer_ that will deploy the app.
* the latter see their key automatically prefixed with `spring.cloud.deployer.`. For the former, the previous magic with `.producer. / .consumer.` still applies.

Fixes https://github.com/spring-cloud/spring-cloud-dataflow/issues/974

### DIFF
--- a/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
+++ b/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
@@ -111,7 +111,11 @@ public final class DeploymentPropertiesUtils {
 					appName, kv.getKey(), kv.getValue(),
 					appName, kv.getKey().substring("spring.cloud.deployer.".length()), kv.getValue());
 			});
-			return deprecated;
+			if (result.isEmpty()) {
+				return deprecated;
+			} else {
+				return result;
+			}
 		}
 
 	}
--- a/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
+++ b/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
@@ -41,7 +41,7 @@ public final class DeploymentPropertiesUtils {
 	/**
 	 * Pattern used for parsing a String of comma-delimited key=value pairs.
 	 */
-	private static final Pattern DEPLOYMENT_PROPERTIES_PATTERN = Pattern.compile(",\\s*app\\.[^\\.]+\\.[^=]+=");
+	private static final Pattern DEPLOYMENT_PROPERTIES_PATTERN = Pattern.compile(",\\s*(app|deployer)\\.[^\\.]+\\.[^=]+=");
 
 	/**
 	 * Pattern used for parsing a String of command-line arguments.
--- a/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
+++ b/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
@@ -105,7 +105,7 @@ public final class DeploymentPropertiesUtils {
 			return result;
 		} else {
 			deprecated.entrySet().forEach(kv -> {
-				logger.warn("Usage of application property prefix 'spring.cloud.deployer' to pass properties to the deployer has been deprecated and will be removed soon\n" +
+				logger.warn("Usage of application property prefix 'spring.cloud.deployer' to pass properties to the deployer has been deprecated and will be removed in a future release\n" +
 					"Instead of 'app.{}.{} = {}', please use\n" +
 					"           'deployer.{}.{} = {}'",
 					appName, kv.getKey(), kv.getValue(),
--- a/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
+++ b/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
@@ -26,6 +26,9 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import org.springframework.util.StringUtils;
 
 /**
@@ -35,6 +38,8 @@ import org.springframework.util.StringUtils;
  * @author Mark Fisher
  */
 public final class DeploymentPropertiesUtils {
+
+	private static final Logger logger = LoggerFactory.getLogger(DeploymentPropertiesUtils.class);
 
 	private DeploymentPropertiesUtils() {
 		// prevent instantiation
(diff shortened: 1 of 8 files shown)


trisberg:
Seems this change doesn't allow me to specify `--properties` when launching a task in the shell, they don't seem to reach the task app. (This feature actually seems to be a bit broken in master right now (see #1076) but I can at least pass in one property).

spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
```diff
@@ -94,6 +99,39 @@ private DeploymentPropertiesUtils() {
 				(fromWildcard, fromApp) -> fromApp
 				)
 			);
+
+		Map<String, String> deprecated = extractDeprecatedDeployerProperties(input, appName);
+		if (deprecated.isEmpty()) {
+			return result;
+		} else {
+			deprecated.entrySet().forEach(kv -> {
+				logger.warn("Usage of application property prefix 'spring.cloud.deployer' to pass properties to the deployer has been deprecated and will be removed soon\n" +
```
trisberg: Should we say "in a future release" instead of "soon"?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
```diff
@@ -139,15 +142,14 @@ public void executeTask(String taskName, Map<String, String> runtimeProperties,
...
-				taskDefinition.getRegisteredAppName()));
+		URI uri = this.registry.find(String.format("task.%s", taskDefinition.getRegisteredAppName()));
 		Resource resource = this.resourceLoader.getResource(uri.toString());
 
-		taskDefinition = qualifyProperties(taskDefinition, resource);
 
-		AppDeploymentRequest request = taskDefinition.createDeploymentRequest(resource,
-				deploymentProperties, runtimeParams);
+		Map<String, String> appDeploymentProperties = extractAppProperties(taskDefinition, taskDeploymentProperties);
+		Map<String, String> deployerDeploymentProperties = DeploymentPropertiesUtils.extractAndQualifyDeployerProperties(taskDeploymentProperties, taskDefinition.getName());
```
trisberg: We should use `getRegisteredAppName()` here since we want the name of the app and not the task definition

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
```diff
@@ -160,13 +162,25 @@ public void executeTask(String taskName, Map<String, String> runtimeProperties,
 		}
 	}
 
+	private Map<String, String> extractAppProperties(TaskDefinition taskDefinition, Map<String, String> taskDeploymentProperties) {
+		final String prefix = "app." + taskDefinition.getName() + ".";
```
trisberg: Same here, we should use `getRegisteredAppName()` since we want the name of the app and not the task definition

spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
```diff
@@ -34,14 +39,16 @@
...
+
 	private DeploymentPropertiesUtils() {
 		// prevent instantiation
 	}
 
 	/**
 	 * Pattern used for parsing a String of comma-delimited key=value pairs.
 	 */
-	private static final Pattern DEPLOYMENT_PROPERTIES_PATTERN = Pattern.compile(",\\s*app\\.[^\\.]+\\.[^=]+=");
+	private static final Pattern DEPLOYMENT_PROPERTIES_PATTERN = Pattern.compile(",\\s*(app|deployer)\\.[^\\.]+\\.[^=]+=");
```
trisberg: There is a general issue with the parsing that existed before this change. If I use the following properties: `deployer.*.memory=512m,server.port=8787` then I get a 500 error saying "java.lang.IllegalArgumentException: Could not parse '512m,server.port=8787' as a byte size. Expected a number with optional 'm' or 'g' suffix" -- looks like the last property gets appended to the value of the first one. It works as expected if I add the "app.*." prefix, but this is an easy mistake to make and we should be able to handle it.

ericbottard: I agree the parsing code has always been brittle. IMHO, we should first split on `,` (with support for some kind of escape mechanism, _e.g._ `\,`) and **then** look for `app.x` and `deployer.x`.

This can (and should) be the subject of another issue/PR though

spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
ericbottard: I'd go as far as saying "in release 1.3.x"...

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
ericbottard: To be even more correct, we want the label (even this does not commonly apply for tasks). But looking at the source, I think `TaskDefinition` should be fixed: one of the constructors does not call `super`, and I fear this can cause inconsistencies. See [here](https://github.com/spring-cloud/spring-cloud-dataflow/blob/28e62fd2b6c20202331c8448d75b3f4e50789345/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/TaskDefinition.java#L40-L58)

ericbottard:
> Also, some weirdness if I mix and match like this:
dataflow:>stream deploy ticktock --properties "app.log.spring.cloud.deployer.memory=1024m,deployer.log.count=2"; I only see the memory getting picked up

Fixed via ceb17d95 (I think), please advise.

spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
trisberg: Yes, agree that we can solve this in a separate PR since this PR didn't cause the existing problem.

spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
trisberg: wouldn't want to be that specific in a deprecation note, as long as we warn anyone using this we can remove it in a future release

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
trisberg: I see your point, TaskDefinition doesn't have a label but looks like the registeredAppName should be the label, see here: https://github.com/spring-cloud/spring-cloud-dataflow/blob/28e62fd2b6c20202331c8448d75b3f4e50789345/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/DataFlowAppDefinition.java#L62

spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
ericbottard: Best way to force us to remove it ;)
But I get your point. Will change to "in a future release"

spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
ericbottard: Created https://github.com/spring-cloud/spring-cloud-dataflow/issues/1097

trisberg:
Looks like most issues are resolved. Found one additional issue, using the old `app.log.count=2` style doesn't seem to get picked up while `app.log.spring.cloud.deployer.count=2` does.

ericbottard:
> using the old app.log.count=2 style doesn't seem to get picked up while app.log.spring.cloud.deployer.count=2 does.

Yes, thought about this. Do we want to handle this as a special case?

trisberg:
I would vote yes, we should handle this

trisberg:
LGTM now, will merge into master once we release 1.1.1 and create the 1.1.x maintenance branch

trisberg:
Squashed and rebased, merged as commit 1ea276c97a9cbdc8bfb642b0fb34d6d3b5bb64ac

PR #1068, opened 2016-12-19
Refactor for full auto-config

### LINKED ISSUE #1026: Dataflow doesn't work as @SpringBootApplication
Initializr will create app using `@SpringBootApplication` which breaks DB config as `@EnableDataFlowServer` imports `DataFlowServerConfiguration`.

```
@SpringBootApplication
public class LocalDataFlowServer {}
```

So you need to fix it like `LocalDataFlowServer` and other impls define a main class.

```
@EnableDataFlowServer
public class LocalDataFlowServer {}
```

I made an attempt to create `DataFlowServerAutoConfiguration` to automate this:

```
@Configuration
@EnableDataFlowServer
public class DataFlowServerAutoConfiguration {
}
```

This them blows up because `DataFlowServerConfiguration` has:

```
@EnableAutoConfiguration(exclude=SessionAutoConfiguration.class)
```

And we get error:
```
org.springframework.beans.factory.parsing.BeanDefinitionParsingException: Configuration problem: A circular @Import has been detected: Illegal attempt by @Configuration class 'DataFlowServerConfiguration' to import class 'DataFlowServerAutoConfiguration' as 'DataFlowServerAutoConfiguration' is already present in the current import stack [DataFlowServerConfiguration->DataFlowServerConfiguration->DataFlowServerAutoConfiguration->LocalDataFlowServer]
```

Also `SessionAutoConfiguration` is not compatible for something else we do in a dataflow(there's some weird MBean error). Culprint is that auto-config class should not have `@EnableAutoConfiguration` in its import stack. Without fixing issues with `SessionAutoConfiguration` we can't have a simple app with `@SpringBootApplication`

jvalkeal:
- This commit has one goal, Removal of @EnableAutoConfiguration
  from DataFlowServerConfiguration so that dataflow server can be
  build and run by just having needed deps and @SpringBootApplication
  in a main class.
- NOTE: This will be a breaking change with our use of @EnableDataFlowServer
  as it will no longer enable auto configuration thus it will now be
  required to use i.e. @SpringBootApplication.
- Add DataFlowServerAutoConfiguration which is a passthrough for
  @EnableDataFlowServer. For conveniance 'dataflow.server.enabled' is
  added for conditionally disable this autoconfig. JacksonAutoConfiguration
  needs to happen after as otherwise we hit issues with
  github.com/spring-projects/spring-boot/issues/5984. Previously
  DataFlowServerConfiguration enabled autoconfig and pretty much did
  same thing because of a use of @EnableDataFlowServer in a
  main class.
- Adding package for LocalDataFlowServerAutoConfiguration to get
  DelegatingResourceLoader(for docker and maven) created as previously
  this was imported manually from LocalDataFlowServer. Property
  'dataflow.server.local.enabled' is also added.
- Refactor some of a tests which previously relied auto-config to kick
  in automatically via DataFlowServerConfiguration.
- SessionAutoConfiguration used to be explicitely exluded via annotation
  as it will cause failure if redis is not available and we need to have
  a way to disable things around that. Boot guys recommended to set
  'spring.session.store-type=none' instead of disabling that autoconfig.
  This is now explicitely on DefaultEnvironmentPostProcessor which
  adds new yml file META-INF/dataflow-server-defaults.yml where we
  can place internal default props.
- Fixes #1026

Squashed, fixes for session autoconfig and use of @EnableDataFlowServer

- dataflow-server-defaults.yml now uses spring.autoconfigure.exclude which seems to
  work for excluding SessionAutoConfiguration instead of setting session store
  type to none.
- I still kept actual dataflow server configuration to happen via auto-config but it
  is now enabled conditionally if @EnableDataFlowServer is used. Without use of
  @EnableDataFlowServer auto-config doesn't happen. Logic for this is same as done in
  Spring Cloud ConfigServer
  github.com/spring-cloud/spring-cloud-config/blob/master/spring-cloud-config-server/src/main/java/org/springframework/cloud/config/server/EnableConfigServer.java
  which adds a marker bean and auto-config conditionally checks if marker bean exists.
  Benefit for this is that we get same ordering of bean creation in all cases. If @EnableDataFlowServer
  would simply cause raw @Import of configs conditionally bean ordering might be different because
  @EnableDataFlowServer would fire imports from app's main class.

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DefaultEnvironmentPostProcessor.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DefaultEnvironmentPostProcessor.java
@@ -21,8 +21,8 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
 
+import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
-
 import org.springframework.beans.factory.config.YamlPropertiesFactoryBean;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.env.EnvironmentPostProcessor;
@@ -35,39 +35,53 @@ import org.springframework.core.io.ClassPathResource;
 import org.springframework.core.io.Resource;
 
 /**
- * Contributes the values from {@code dataflow-server.yml} if it exists, before any of Spring Boot's normal
+ * Contributes the values from {@code META-INF/dataflow-server-defaults.yml} and
+ * {@code dataflow-server.yml} if it exists, before any of Spring Boot's normal
  * configuration contributions apply. This has the effect of supplying overridable defaults to the
  * various Spring Cloud Data Flow Deployer SPI implementations that in turn override the defaults
  * provided by Spring Boot.
  *
  * @author Josh Long
+ * @author Janne Valkealahti
  */
 public class DefaultEnvironmentPostProcessor implements EnvironmentPostProcessor, Ordered {
 
-	private Resource resource = new ClassPathResource("/dataflow-server.yml");
+	private static Log logger = LogFactory.getLog(DefaultEnvironmentPostProcessor.class);
+	private final Resource serverResource = new ClassPathResource("/dataflow-server.yml");
+	private final Resource serverDefaultsResource = new ClassPathResource("META-INF/dataflow-server-defaults.yml");
 
 	@Override
 	public void postProcessEnvironment(ConfigurableEnvironment environment, SpringApplication application) {
+		Map<String, Object> internalDefaults = new HashMap<>();
 		Map<String, Object> defaults = new HashMap<>();
 		MutablePropertySources existingPropertySources = environment.getPropertySources();
 
-		this.contributeDefaults(defaults);
+		contributeDefaults(internalDefaults, serverDefaultsResource);
+		contributeDefaults(defaults, serverResource);
 
 		String defaultPropertiesKey = "defaultProperties";
 
 		if (!existingPropertySources.contains(defaultPropertiesKey) ||
 				existingPropertySources.get(defaultPropertiesKey) == null) {
+			existingPropertySources.addLast(new MapPropertySource(defaultPropertiesKey, internalDefaults));
 			existingPropertySources.addLast(new MapPropertySource(defaultPropertiesKey, defaults));
 		}
 		else {
 			PropertySource<?> propertySource = existingPropertySources.get(defaultPropertiesKey);
 			@SuppressWarnings("unchecked")
 			Map<String, Object> mapOfProperties = Map.class.cast(propertySource.getSource());
+			for (String k : internalDefaults.keySet()) {
+				Set<String> setOfPropertyKeys = mapOfProperties.keySet();
+				if (!setOfPropertyKeys.contains(k)) {
+					mapOfProperties.put(k, internalDefaults.get(k));
+					logger.info(k + '=' + internalDefaults.get(k));
+				}
+			}
 			for (String k : defaults.keySet()) {
 				Set<String> setOfPropertyKeys = mapOfProperties.keySet();
 				if (!setOfPropertyKeys.contains(k)) {
 					mapOfProperties.put(k, defaults.get(k));
-					LogFactory.getLog(getClass()).info(k + '=' + defaults.get(k));
+					logger.info(k + '=' + defaults.get(k));
 				}
 			}
 		}
@@ -79,10 +93,10 @@ public class DefaultEnvironmentPostProcessor implements EnvironmentPostProcessor
 		return 0;
 	}
 
-	private void contributeDefaults(Map<String, Object> defaults) {
-		if (this.resource.exists()) {
+	private static void contributeDefaults(Map<String, Object> defaults, Resource resource) {
+		if (resource.exists()) {
 			YamlPropertiesFactoryBean yamlPropertiesFactoryBean = new YamlPropertiesFactoryBean();
-			yamlPropertiesFactoryBean.setResources(this.resource);
+			yamlPropertiesFactoryBean.setResources(resource);
 			yamlPropertiesFactoryBean.afterPropertiesSet();
 			Properties p = yamlPropertiesFactoryBean.getObject();
 			for (Object k : p.keySet()) {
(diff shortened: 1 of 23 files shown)


markpollack:
Hi.  I am not following the root issue in #1026.  Nevertheless, the end usage model is unlike that of all the other spring-cloud projects where after you create the project from Initializr you have to add an `@Enable` annotation, e.g. 

```
@SpringBootApplication
@EnableEurekaServer
public class Application {
...
}
```

I do not thing we should be changing that usage model but of course if we are creating a broken app from Initialzr, we should fix that.

jvalkeal:
`@EnableEurekaServer` just import `EurekaServerConfiguration` which doesn't do dummy things like `@EnableAutoConfiguration` as what we've being doing.

jvalkeal:
And as a side note, this never really worked as both ended up adding `@EnableAutoConfiguration`.
```
@SpringBootApplication
@EnableDataFlowServer
public class Application {
...
}
```

trisberg:
couldn't we just fix the double @EnableAutoConfiguration issue so that you could have booth @SpringBootApplication and @EnableDataFlowServer on your main class

joshlong:
yah +1 to @trisberg  and @markpollack  - it would be better to allow both `@EnableDataFlowServer` and `@SpringBotoApplication`. It's what developers expect when using config server, zipkin, eureka, the various dashboards out in the community, hystrix dashboard, etc. Hell, the config server is a very good example: the entire purpose of the `@EnableConfigServer` annotation is to register a dummy bean which other auto-configurations then use as a marker (with `@COnditionalOnBean`) to determine whether they get activated or not. The `@Enable*` annotation just serves to sort of 'opt-in' in this case because having that opt-in for full blown services is so important

jvalkeal:
I didn't say you could not use `@EnableDataFlowServer`. My point with this was to give a choice to user to either get fully working server with only `@SpringBootApplication` if auto-config handles things behind a scenes or if auto-config is not used then combination of `@EnableDataFlowServer` with `@SpringBootApplication` would do same thing.

spring-cloud-dataflow-server-local-autoconfig/src/main/java/org/springframework/cloud/dataflow/autoconfigure/local/LocalDataFlowServerAutoConfiguration.java
```diff
@@ -14,24 +14,32 @@
...
+ * Auto-configuration for local dataflow server.
+ *
+ * @author Janne Valkealahti
  *
- * @author Ilayaperumal Gopinathan
  */
-public class LocalDataFlowServerConfiguration {
+@Configuration
+@AutoConfigureBefore(DataFlowControllerAutoConfiguration.class)
+@ConditionalOnProperty(prefix = "dataflow.server.local", name = "enabled", havingValue = "true", matchIfMissing = true)
```
markpollack: Why do we need this property?  The only role this project has is to be a server.  What would be the meaning of setting this property to false?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DefaultEnvironmentPostProcessor.java
```diff
@@ -35,39 +35,53 @@
...
 		}
 		else {
 			PropertySource<?> propertySource = existingPropertySources.get(defaultPropertiesKey);
 			@SuppressWarnings("unchecked")
 			Map<String, Object> mapOfProperties = Map.class.cast(propertySource.getSource());
+			for (String k : internalDefaults.keySet()) {
+				Set<String> setOfPropertyKeys = mapOfProperties.keySet();
+				if (!setOfPropertyKeys.contains(k)) {
+					mapOfProperties.put(k, internalDefaults.get(k));
+					logger.info(k + '=' + internalDefaults.get(k));
```
markpollack: Should be debug.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DefaultEnvironmentPostProcessor.java
```diff
@@ -35,39 +35,53 @@
...
  *
  * @author Josh Long
+ * @author Janne Valkealahti
  */
 public class DefaultEnvironmentPostProcessor implements EnvironmentPostProcessor, Ordered {
 
-	private Resource resource = new ClassPathResource("/dataflow-server.yml");
+	private static Log logger = LogFactory.getLog(DefaultEnvironmentPostProcessor.class);
+	private final Resource serverResource = new ClassPathResource("/dataflow-server.yml");
+	private final Resource serverDefaultsResource = new ClassPathResource("META-INF/dataflow-server-defaults.yml");
```
markpollack: Something about using an extra config file for this feels a bit dirty.  Can we use `@ImportAutoConfiguration(exclude=SessionAutoConfiguration.class)` on `DataFlowServerAutoConfiguration` instead?

spring-cloud-dataflow-server-local-autoconfig/src/main/java/org/springframework/cloud/dataflow/autoconfigure/local/LocalDataFlowServerAutoConfiguration.java
jvalkeal: You're right it's not needed to be exact but I've just been accustomed to always add these flags. I suppose you can take it away during a merge if you like.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DefaultEnvironmentPostProcessor.java
jvalkeal: I didn't change logging levels here as those were on info already.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DefaultEnvironmentPostProcessor.java
jvalkeal: Whole point of this PR was to remove `@ImportAutoConfiguration` as if we put it back, main class will break things up again if `@SpringBootApplication` is in place.

markpollack: ah, I thought it was to remove `@EnableAutoConfiguration`   An  `@ImportAutoConfiguration` with just an exclude would maybe achieve the same thing as using the dataflow-server-defaults.yml file.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DefaultEnvironmentPostProcessor.java
markpollack: Yea, though we should be able to see the properties via the actuator endpoint if needed...so while not related to this PR, I think I'll change it to debug on merge.

jvalkeal:
Looks like this PR was merged.

Issue #1081, opened 2016-12-30
Add support in shell to skip SSL validation


sabbyanandan:
As a user, I'd like to disable SSL cert validation when connecting to the server running in a cloud environment such as Cloud Foundry, where it is possible to have the DEV PCF instance deployed with self-signed certificates. 

This option would be similar to what we have in CF CLI (_i.e: `cf login ..... --skip-ssl-validation`_), but for SCDF shell.

**Acceptance:**
- SSL validation is enabled by default
- a new flag (eg: `skip-ssl-validation` ?) can be used to disable the validation on demand
- docs updated

trisberg:
I think the additional concerns are handled in https://github.com/spring-cloud/spring-cloud-dataflow/issues/1114

PR #1082, opened 2017-01-03
Add support in shell to skip SSL validation

ericbottard:
Fixes https://github.com/spring-cloud/spring-cloud-dataflow/issues/1081


ghillert:
@ericbottard Needs documentation in the reference guide

ericbottard:
I updated the usage.txt file.

When it comes to the reference docs, where would you see that new documentation (we don't really have a section pertaining to the shell)?

ghillert:
We have a section in the reference guide under chapter `Using Self-Signed Certificates`. It already contains a few pointers in regards to the shell as well.

ghillert:
LGTM - Merging. Will add reference documentation during merge.

ghillert:
Merged with commits 032e453916c68a08de9821fef9d92b8f2cc8f982 and 59097274fc3500cdd9046e5a3c2b3ef4e17f417f

Issue #1072, opened 2016-12-21
REST: CompletionController - Missing "start" parameter should not trigger a 500 Status Code


ghillert:
While the `start` Request Parameter is mandatory it should not result in a `500` Internal Server Error Http status code. A `400` status would be better.

PR #1083, opened 2017-01-03
Return HTTP 400 error in case of missing request parameter

ericbottard:
Fixes https://github.com/spring-cloud/spring-cloud-dataflow/issues/1072


viniciusccarvalho:
Added tests, merged and squashed as 3e786f1291612c6cf37d22909550046a34c8ad57

Issue #1043, opened 2016-11-28
LoginController instantiated by default


ghillert:
Should only be instantiated when security is enabled

ghillert:
This could be an issue in Junit Tests that bootstrap a _Spring Cloud Data Flow_ application, e.g. in the case of a _Spring Initializr_ generated application. The work-around is to use **@SpringBootTest** instead of the deprecated **@SpringApplicationConfiguration** (Used by Spring Initializr):

```
@RunWith(SpringJUnit4ClassRunner.class)
@SpringBootTest(classes = Demo123Application.class)
public class DemApplicationTests {

	@Test
	public void contextLoads() {
	}

}
```

**Observation:**

It seems like there is some unintended Boot magic going on … despite security being off (From SCDF point of view), Spring Boot still generates an AuthenticationManager Bean - 

Therefore, add tests in order to test for the Boot Magic e.g. create test-cases to explicitly test for the non-existence of certain beans if security is disabled.

PR #1084, opened 2017-01-04
gh-1043 Ensure that the `LoginController` is only instantiated with enabled security

ghillert:
Resolves https://github.com/spring-cloud/spring-cloud-dataflow/issues/1043


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/security/LoginController.java
```diff
@@ -49,13 +63,4 @@ public String authorize(
 		return session.getId();
 	}
 
-	@RequestMapping("/admin-ui/")
```
markpollack: Why remove these two?

ghillert: Hangover from Spring XD days ... we now have `/dashboard` handled in class `UiController`

PR #1090, opened 2017-01-04
Composed Task Spike [Do Not Merge]

### LINKED ISSUE #1085: Spike for ComposedTasks
User wants to see how a base project for the Composed Tasks feature would work.
Includes :
* ComposedTaskRunner
* TaskLauncherTasklet
* Auto Register ComposedTaskRunner
* ComposedTaskStepExecutionListener
* Create child project under Dataflow project

cppwfs:
# Composed Task Spike

The purpose of this spike is to show how the Composed Tasks will be implemented in
the Data Flow project.

## Design Document

If you hadn't had a chance please review the design document.  It can be found 
[here](https://docs.google.com/document/d/10FiyNrKH1KmfDj7WRVmREc9VwzSZx3XnMBP8VU8y99U/edit#heading=h.4h9zsajkx156)

### What is included

Based on the design document this story created a spring-cloud-dataflow-composed-task sub-project
under the Spring Cloud Data Flow Project.  In this project you will see that the TaskLaunchTasklet, ComposedTaskRunner
along with the supporting ComposedTaskConfiguration and ComposedTaskStepExecutionListener are included.
Also you will note that the DataFlowServerConfiguration has been updated to include a EventListener to register the ComposedTaskRunner as a task in the repository (as specified in the design document) at Data Flow Server startup.  Thus you will see that a `composed-task-runner` task appears upon executing a `app list`.  This registration can be disabled by setting the spring.cloud.dataflow.composed.task.enabled property to false.

### What is not included

In this spike the following has not be included:

* Controllers for task compose
* Parser for the DSL within the ComposedTaskRunner (you can only include a single task definition for the DSL at the moment).
* Parent Task / Child Task relationship.

#### Kicking the tires

`task create mytask --definition "composed-task-runner --composedTaskDSL='AAA'"`  Where *AAA* is your task definition.
`task launch mytask`

resolves #1085


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfiguration.java
```diff
@@ -134,4 +141,32 @@ public DataflowRdbmsInitializer dataflowRdbmsInitializer(DataSource dataSource,
...
+		private DataSource dataSource;
+
+		@EventListener
+		public void handleContextRefresh(ContextRefreshedEvent event) {
+			RdbmsUriRegistry registry = new RdbmsUriRegistry(dataSource);
+			try {
+				registry.register(COMPOSED_TASK_RUNNER_NAME, new URI(
+						COMPOSED_TASK_RUNNER_URI_BASE
+								+ registry.getClass().getPackage()
+								.getImplementationVersion()));
```
ericbottard: This is wrong, version of the app has nothing to do with the version of the jar holding `RdbmsUriRegistry.class`

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfiguration.java
```diff
@@ -134,4 +141,32 @@ public DataflowRdbmsInitializer dataflowRdbmsInitializer(DataSource dataSource,
...
+				"task.composed-task-runner";
+
+		private static final String COMPOSED_TASK_RUNNER_URI_BASE =
+				"maven://org.springframework.cloud:spring-cloud-dataflow-composed-task:";
+
+		@Autowired
+		private DataSource dataSource;
+
+		@EventListener
+		public void handleContextRefresh(ContextRefreshedEvent event) {
```
ericbottard: Registration of this runner should not happen magically IMHO, for the same reason we don't register other apps by default.
In any case, should inject the registry and use it, not create a new one.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfiguration.java
jvalkeal: We should not create additional `RdbmsUriRegistry` as currently it's already created in `DataFlowControllerAutoConfiguration`.

spring-cloud-dataflow-composed-task/pom.xml
```diff
@@ -0,0 +1,104 @@
...
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>spring-cloud-dataflow-composed-task</artifactId>
```
jvalkeal: Is there a reason why this fatjar app would be kept in a dataflow core and not in where other task apps are kept? Am I correct to assume it was just done for this PR scope? Maybe it was mentioned in drive docs but I don't yet have access to it.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfiguration.java
jvalkeal: My above comment was for lines above for registry creation. I got back here from nitpicking about keeping composed tasks in a dataflow and I agree with Eric. Trying to register hard coded version here like this is definitely wrong and anyway we moved away for dataflow not registering any apps on default.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfiguration.java
cppwfs: While I agree with you this was one of the original tenants that was discussed during our design strategy.  Please comment on this in the design document so we can adjust it there.   And then once we all come to agreement, we can change this in the PR or submit a new PR.

spring-cloud-dataflow-composed-task/pom.xml
jvalkeal: ok, I understand a motivation but how would it work if user is using oracle DB? Generally speaking that is when appstarters comes into play if user needs to bundle jdbc drivers we don't/can't bundle OOB.

trisberg:
@cppwfs can we open up the permissions to the design doc so we all can read it?

cppwfs:
Based on the comments above we will be going back to the design doc to flush out the details above.  In short we will not include ComposedTaskRunner as a part of the SCDF project based on the comments made by Janne (i.e. user wants a different database that is not included OOTB).
And also not pre-registering the ComposedTaskRunner.

Issue #1073, opened 2016-12-21
Completion Controller - Able to provide negative detail level


ghillert:
Currently, users can provide a negative `detailLevel` request parameter. Should this be disallowed with the appropriate HTTP status code, e.g. `400`?

PR #1112, opened 2017-01-11
gh-1073 Completion Controller - Fix Able to provide negative detail level

ghillert:
* Use JSR303 bean validation to validate method parameters (`@RequestParam`)
* Enable bean validation by adding a `MethodValidationPostProcessor` to `WebConfiguration`
* Add tests

Resolves https://github.com/spring-cloud/spring-cloud-dataflow/issues/1073


jvalkeal:
looks good, merging.

PR #1113, opened 2017-01-11
Improving error handling for app registration

### LINKED ISSUE #1103: SCDF app import fails when using redirect
Using scdf shell and server local dashboard latest 1.1.1.BUILD-SNAPSHOT today, when importing streams using 
app import --uri https://bit.ly/Avogadro-GA-stream-applications-rabbit-maven
an errors happens and no apps are registered.
Same thing happens when trying to import via URL in dashboard.  

Example:
dataflow:>app import --uri https://bit.ly/Avogadro-GA-stream-applications-rabbit-maven
Command failed org.springframework.cloud.dataflow.rest.client.DataFlowClientException: 1

However, taking the location of the redirect in the bit.ly link, the following works:
dataflow:>app import --uri http://repo.spring.io/libs-release-local/org/springframework/cloud/stream/app/spring-cloud-stream-app-descriptor/Avogadro.RELEASE/spring-cloud-stream-app-descriptor-Avogadro.RELEASE.stream-apps-rabbit-maven
Successfully registered 0 applications from [sink.task-launcher-yarn, source.tcp, sink.jdbc, source.http, sink.rabbit, source.rabbit, source.ftp, sink.gpfdist, processor.transform, source.loggregator, source.sftp, processor.filter, source.file, sink.cassandra, processor.groovy-filter, sink.router, source.trigger, sink.hdfs-dataset, processor.splitter, source.load-generator, sink.sftp, sink.file, processor.tcp-client, sink.tasklaunchrequest-transform, source.time, source.gemfire, source.twitterstream, sink.tcp, source.jdbc, sink.field-value-counter, sink.redis-pubsub, sink.hdfs, sink.task-launcher-local, processor.bridge, processor.pmml, processor.httpclient, sink.ftp, source.s3, sink.log, sink.gemfire, sink.aggregate-counter, sink.throughput, source.triggertask, sink.s3, source.gemfire-cq, source.jms, source.tcp-client, processor.scriptable-transform, sink.counter, sink.websocket, source.mongodb, source.mail, processor.groovy-transform, source.syslog]

Also, the count for successfully registered is 0 and the app list shows all of them registered.

Attached is the stack trace from the exception on the scdf server local.  
[exception.txt](https://github.com/spring-cloud/spring-cloud-dataflow/files/691215/exception.txt)

Here are the first several lines:

2017-01-06 21:46:13.783 ERROR 58676 --- [nio-9393-exec-1] o.s.c.d.s.c.RestControllerAdvice         : Caught exception while handling a request

java.lang.ArrayIndexOutOfBoundsException: 1
	at org.springframework.cloud.dataflow.registry.AppRegistry.createAppRegistration(AppRegistry.java:108) ~[spring-cloud-dataflow-registry-1.1.1.BUILD-SNAPSHOT.jar!/:1.1.1.BUILD-SNAPSHOT]
	at org.springframework.cloud.dataflow.registry.AppRegistry.importAll(AppRegistry.java:81) ~[spring-cloud-dataflow-registry-1.1.1.BUILD-SNAPSHOT.jar!/:1.1.1.BUILD-SNAPSHOT]
	at org.springframework.cloud.dataflow.server.controller.AppRegistryController.registerAll(AppRegistryController.java:177) ~[spring-cloud-dataflow-server-core-1.1.1.BUILD-SNAPSHOT.jar!/:1.1.1.BUILD-SNAPSHOT]

In the future (new card/issue?) consider adding the git commit id in the MANIFEST Implementation Version to get some additional information as to what BUILD-SNAPSHOT source generated the jar.  This jar was created around "Fri Jan 06 22:25:31 UTC 2017" based on the pom.properties file in the META-INF directory.

trisberg:
- if provided properties file entries don't resolve to a valid format then throw the best available error message

- require that app registrations have a URI

Resolves #1103

### DIFF
--- a/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/AppRegistry.java
+++ b/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/AppRegistry.java
@@ -114,7 +114,8 @@ public class AppRegistry {
 	private AppRegistration createAppRegistration(String key, URI uri) {
 		String[] tokens = key.split("\\.", 2);
 		if (tokens.length != 2) {
-			throw new IllegalArgumentException("Invalid application key: " + key);
+			throw new IllegalArgumentException("Invalid application key: " + key +
+					"; the expected format is <name>.<type>");
 		}
 		return new AppRegistration(tokens[1], ApplicationType.valueOf(tokens[0]), uri, this.resourceLoader);
 	}
--- a/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/AppRegistry.java
+++ b/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/AppRegistry.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 the original author or authors.
+ * Copyright 2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -34,6 +34,7 @@ import org.springframework.core.io.ResourceLoader;
  *
  * @author Mark Fisher
  * @author Gunnar Hillert
+ * @author Thomas Risberg
  */
 public class AppRegistry {
 
@@ -75,10 +76,17 @@ public class AppRegistry {
 
 	public List<AppRegistration> importAll(boolean overwrite, String... resourceUris) {
 		List<AppRegistration> apps = new ArrayList<>();
-		Map<String, URI> registered = this.uriRegistryPopulator.populateRegistry(
-				overwrite, this.uriRegistry, resourceUris);
-		for (Map.Entry<String, URI> entry : registered.entrySet()) {
-			apps.add(createAppRegistration(entry.getKey(), entry.getValue()));
+		for (String uri : resourceUris) {
+			try {
+				Map<String, URI> registered = this.uriRegistryPopulator.populateRegistry(
+						overwrite, this.uriRegistry, uri);
+				for (Map.Entry<String, URI> entry : registered.entrySet()) {
+					apps.add(createAppRegistration(entry.getKey(), entry.getValue()));
+				}
+			}
+			catch (Exception e) {
+				throw new IllegalStateException("Error when registering applications from " + uri + ": " + e.getMessage(), e);
+			}
 		}
 		return apps;
 	}
@@ -105,6 +113,9 @@ public class AppRegistry {
 
 	private AppRegistration createAppRegistration(String key, URI uri) {
 		String[] tokens = key.split("\\.", 2);
+		if (tokens.length != 2) {
+			throw new IllegalArgumentException("Invalid application key: " + key);
+		}
 		return new AppRegistration(tokens[1], ApplicationType.valueOf(tokens[0]), uri, this.resourceLoader);
 	}
 }
--- a/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/RdbmsUriRegistry.java
+++ b/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/RdbmsUriRegistry.java
@@ -82,8 +82,11 @@ public class RdbmsUriRegistry implements UriRegistry {
 
 	@Override
 	public void register(String name, URI uri) {
-		Assert.notNull(uri, "Error when registering " + name + ": uri is required");
-		Assert.hasText(uri.toString(), "Error when registering " + name + ": uri must not be blank");
+		Assert.notNull(uri, "Error when registering " + name + ": URI is required");
+		Assert.hasText(uri.getScheme(), "Error when registering " + name + " with URI " + uri +
+				": URI scheme must be specified");
+		Assert.hasText(uri.getSchemeSpecificPart(), "Error when registering " + name + " with URI " + uri +
+				": URI scheme-specific part must be specified");
 		String uriString = uri.toString();
 		if (find(name) != null) {
 			jdbcTemplate.update(UPDATE_SQL, new Object[] { uriString, name },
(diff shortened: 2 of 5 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppRegistryController.java
```diff
@@ -179,6 +180,9 @@ public void unregister(@PathVariable("type") ApplicationType type, @PathVariable
 		else if (!CollectionUtils.isEmpty(apps)) {
 			for (String key : apps.stringPropertyNames()) {
 				String[] tokens = key.split("\\.", 2);
+				if (tokens.length != 2) {
+					throw new IllegalArgumentException("Invalid application key: " + key);
```
ericbottard: Could add a message saying that the expected format is `<name>.<type>`

spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/AppRegistry.java
```diff
@@ -105,6 +113,9 @@ private String key(String name, ApplicationType type) {
 
 	private AppRegistration createAppRegistration(String key, URI uri) {
 		String[] tokens = key.split("\\.", 2);
+		if (tokens.length != 2) {
+			throw new IllegalArgumentException("Invalid application key: " + key);
```
ericbottard: Could add a message saying that the expected format is `<name>.<type>`

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppRegistryController.java
trisberg: Yes, makes sense

spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/RdbmsUriRegistry.java
```diff
@@ -81,6 +82,8 @@ public URI find(String name) {
 
 	@Override
 	public void register(String name, URI uri) {
+		Assert.notNull(uri, "Error when registering " + name + ": uri is required");
+		Assert.hasText(uri.toString(), "Error when registering " + name + ": uri must not be blank");
```
ilayaperumalg: Given the usage of `ResourceLoader` when loading from the URI, can we just validate if the URI string has a valid URI Scheme (at least starting with a string with colon)?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppRegistryController.java
```diff
@@ -179,6 +180,10 @@ public void unregister(@PathVariable("type") ApplicationType type, @PathVariable
 		else if (!CollectionUtils.isEmpty(apps)) {
 			for (String key : apps.stringPropertyNames()) {
 				String[] tokens = key.split("\\.", 2);
+				if (tokens.length != 2) {
```
ilayaperumalg: Can we have some tests covering these exception cases?

spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/AppRegistryCommands.java
```diff
@@ -248,7 +249,13 @@ public String importFromResource(
 			try {
 				Resource resource = this.resourceLoader.getResource(uri);
 				Properties applications = PropertiesLoaderUtils.loadProperties(resource);
-				PagedResources<AppRegistrationResource> registered = appRegistryOperations().registerAll(applications, force);
+				PagedResources<AppRegistrationResource> registered = null;
+				try {
+					registered = appRegistryOperations().registerAll(applications, force);
+				}
+				catch (Exception e) {
+					return "Error when registering applications from " + uri + ": " + e.getMessage();
```
ilayaperumalg: It might be worth to consider some SCDF specific exception (something like AppRegistrationException) and have it defined in `RestControllerAdvice`?

ilayaperumalg: Might as well include `AppRegistration` exception in other cases where the runtime exceptions are thrown during the app registration.

spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/RdbmsUriRegistry.java
trisberg: Adding additional asserts

spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/AppRegistryCommands.java
trisberg: If we want to add a specific `AppRegistrationException` then I think that should live in the spring-cloud-deployer project where `UriRegistry` and `UriRegistryPopulator`are defined. I really don't see a strong need for that though.

markpollack:
Added tests that read from good/bad property files.
Squashed and merged as 6d58b157190f7750eeaf104ed872aa9bf1adde07

spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/AppRegistryCommands.java
ilayaperumalg: I think the REST controller advice for the exceptions that are caught at the REST controller side would still be better in this case. In any case, we can add it later if we find it necessary.

PR #1116, opened 2017-01-12
Fix task launching properties handling

### LINKED ISSUE #1076: Unable to specify multiple properties for task launch
Haven't been able to find any docs for how the `--properties` option for launching a task via the shell is supposed to work.  Tried separating multiple entries with both space and comma, seems neither one works as expected and I can't provide more than one property. Also tried this via the UI and get the second property expression concatenated with the value of the first property. This is the same I get if I separate the properties with a comma using the shell.

jvalkeal:
- Fixed DefaultTaskService to strip i.e. "app.timestamp" out from
  a passed properties string so that this prefix is not passed
  into an task app itself. This is how stream apps are handled.
- Added tests to verify properties handling.
- Added docs how properties can be used during a task launch.
- Effectively i.e with local server launch
  task launch --name footask1 --properties "app.timestamp.foo1=bar1,app.timestamp.foo2=bar2,app.timestamp.spring.cloud.deployer.foo3=bar3"
  will add 'foo1=bar1' and 'foo2=bar2' as task app environment variables
  and 'foo3=bar3' as a deployment property.
- Fixes #1076

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/tasks.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/tasks.adoc
@@ -160,6 +160,22 @@ to the task application can be set when launching the task as follows:
 dataflow:>task launch mytask --arguments "--server.port=8080,--foo=bar"
 ```
 
+Additional properties meant for a `TaskLauncher` itself can be passed
+in using a `--properties` option. Format of this option is a comma
+delimited string of propertys prefixed with `app.<task definition
+name>.<property>`.
+
+If actual property is prefixed with `spring.cloud.deployer` it is
+passed to `TaskLauncher` as a deployment property and its meaning may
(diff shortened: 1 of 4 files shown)


spring-cloud-dataflow-docs/src/main/asciidoc/tasks.adoc
```diff
@@ -160,6 +160,22 @@ to the task application can be set when launching the task as follows:
 dataflow:>task launch mytask --arguments "--server.port=8080,--foo=bar"
 ```
 
+Additional properties meant for a `TaskLauncher` itself can be passed
+in using a `--properties` option. Format of this option is a comma
+delimited string of propertys prefixed with `app.<task definition
```
cppwfs: properties instead of propertys

trisberg:
Fixed typos and merged as commit aad1616b21d844437500b9b967147308d9fa8bbf

Issue #1114, opened 2017-01-12
Shell - Use HttpClient for all operations


ghillert:
Currently, we are in a situation where we use different Http clients for the SCDF Shell's Rest calls.

When we use authentication (Basic Auth), we use  Apache Http Client. However, for the default (non-security), the `RestTemplate` will use the JDKs Http client.

This causes issues when dealing with certificates (Https) as the use-case of skipping certificate validation only works with the JDK's Http client.

We should standardize on Apache Http Client.

PR #1117, opened 2017-01-13
gh-1114 Shell: Use HttpClient for all operations

ghillert:
* Ensure that the Shell uses Apache HttpClient for all "target" opererations
* Make `--skip-ssl-validation` (command-line) part of the Target command
* Add a `skip-ssl-validation` to the Target command (Within Shell)

Resolves https://github.com/spring-cloud/spring-cloud-dataflow/issues/1114

### DIFF
--- a/spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/ConfigCommands.java
+++ b/spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/ConfigCommands.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2015 the original author or authors.
+ * Copyright 2014-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,20 +16,25 @@
 
 package org.springframework.cloud.dataflow.shell.command;
 
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.security.cert.CertificateException;
+import java.security.cert.X509Certificate;
+import java.util.Map;
+import java.util.TreeMap;
+
+import javax.net.ssl.SSLContext;
+
 import org.apache.http.auth.AuthScope;
 import org.apache.http.auth.UsernamePasswordCredentials;
+import org.apache.http.conn.ssl.NoopHostnameVerifier;
 import org.apache.http.impl.client.BasicCredentialsProvider;
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.HttpClientBuilder;
+import org.apache.http.ssl.SSLContexts;
+import org.apache.http.ssl.TrustStrategy;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.springframework.batch.core.ExitStatus;
-import org.springframework.batch.core.JobExecution;
-import org.springframework.batch.core.JobInstance;
-import org.springframework.batch.core.JobParameter;
-import org.springframework.batch.core.JobParameters;
-import org.springframework.batch.core.StepExecution;
-import org.springframework.batch.item.ExecutionContext;
 import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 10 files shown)


spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/ConfigCommands.java
```diff
@@ -270,17 +296,27 @@ public void afterPropertiesSet() throws Exception {
...
 			initialized = true;
-			target(this.serverUri, this.userName, this.password);
+			target(this.serverUri, this.userName, this.password, isSkipSslValidation(this.environment));
 		}
 	}
 
-	@Bean
-	public static RestTemplate restTemplate() {
-		return DataFlowTemplate.getDefaultDataflowRestTemplate();
+	private boolean isSkipSslValidation(Environment env) {
```
ericbottard: Doesn't need to pass env as a parameter, since it is a member field

spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/ConfigCommands.java
```diff
@@ -169,35 +169,55 @@ public String target(
 			@CliOption(mandatory = false, key = {"password"},
 					help = "the password for authenticated access to the Admin REST endpoint (valid only with a username)",
 					specifiedDefaultValue = Target.DEFAULT_SPECIFIED_PASSWORD,
-					unspecifiedDefaultValue = Target.DEFAULT_UNSPECIFIED_PASSWORD) String targetPassword) {
+					unspecifiedDefaultValue = Target.DEFAULT_UNSPECIFIED_PASSWORD) String targetPassword,
+			@CliOption(mandatory = false, key = {"skip-ssl-validation"},
+					help = "accept any SSL certificate (even self-signed)",
+					specifiedDefaultValue = "false",
```
ericbottard: should be `true`

spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/ConfigCommands.java
```diff
@@ -169,35 +169,55 @@ public String target(
...
+					unspecifiedDefaultValue = Target.DEFAULT_UNSPECIFIED_PASSWORD) boolean skipSslValidation){
```
ericbottard: should be `false` (currently copy/paste error)

spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/ConfigCommands.java
```diff
@@ -256,12 +283,11 @@ public String info() {
 		return sb.toString();
 	}
 
-
 	@Override
 	public void onApplicationEvent(ApplicationReadyEvent event) {
 		//Only invoke if the shell is executing in the same application context as the data flow server.
 		if (!initialized) {
-			target(this.serverUri, this.userName, this.password);
+			target(this.serverUri, this.userName, this.password, isSkipSslValidation(this.environment));
```
ericbottard: Why aren't we using the same pattern as for uri, username and password here (and line 299)?

spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/ConfigCommands.java
ericbottard: Depending on answer to my comment for line 290, this could be removed altogether

spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/ConfigCommands.java
markpollack: What pattern do you mean?  Line 299 looks to be the same at 290

spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/ConfigCommands.java
markpollack: Yes, no need to pass the env.

spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/ConfigCommands.java
ericbottard: I meant the same pattern as what was there before for uri, username and password (a member field)

ghillert:
@markpollack @ericbottard Updated the PR.

- Make `skip-ssl-validation` part of the `target` command
- Change command-line argument `skip-sslvalidation` to `dataflow.skip-sslvalidation`
- Make command-line argument `dataflow.skip-sslvalidation` more explicit by requiring a value, such as:
  - `dataflow.skip-sslvalidation=true` (after discussion with @markpollack)
- Implement `username`, `password` and `skip-sslvalidation` parameters for the `http` shell command
- The `http` shell command uses Apache Http Client

spring-cloud-dataflow-shell-core/src/main/resources/usage.txt
```diff
@@ -2,7 +2,7 @@ Data Flow Options:
   --dataflow.uri=<uri>                  Address of the Data Flow Server [default: http://localhost:9393].
   --dataflow.username=<USER>            Username of the Data Flow Server [no default].
   --dataflow.password=<PASSWORD>        Password of the Data Flow Server [no default].
-  --spring.shell.historySize=<SIZE>     Default size of the shell log file [default: 3000]
+  --dataflow.skip-ssl-validation=true   Accept any SSL certificate (even self-signed) [default: no].
```
ericbottard: defaults are specified in square brackets, not inline. So should read `=true|false` or something similar

ericbottard:
Can fix my only last comment on merge

ericbottard:
Merged as 12abd159

PR #1118, opened 2017-01-13
Error handling on non existent app

### LINKED ISSUE #1070: REST: Retrieving non-existing app details should return a 404
The issue is in class `AppRegistryController#DetailedAppRegistrationResource`

When executing a **GET** `/apps/{type}/{name}` a `200` HTTP Status is returned.

But it should produce a `404`.

We need to ensure that when making this minor **breaking change** that the shell and web UI are working.

ilayaperumalg:
- When the application info is queried, non existent app should throw appropriate exception i.e, NoSuchAppRegistryException
 - Update AppRegistry command to handle the exception appropriately
 - Add test

Resolves #1070


jvalkeal:
Tried it and looks good, merging.

PR #1119, opened 2017-01-13
Exception handling at JobInstance list endpoint

### LINKED ISSUE #1074: REST JobInstanceController vs JobExecutionController 404 inconsistency
Some minor inconsistency - 

* GET `/jobs/instances?name=job-does-not-exist` returns an empty collection with HTTP status code `200`
* GET `/jobs/executions?name=job-does-not-exist` returns with HTTP status `404`

Name is a **mandatory** parameter. 

**What behavior should we standardize on in case the respective job does not exist?**

ilayaperumalg:
- Throw NoSuchJobException when there is no job found of the given name
 - This will make it consistent with other endpoints
 - update test

Resolves #1074

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/JobInstanceController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/JobInstanceController.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 the original author or authors.
+ * Copyright 2016-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


jvalkeal:
ok so couldn't do real testing other than using curl
```
{"_links":{"self":{"href":"http://localhost:9393/jobs/instances"}},"page":{"size":0,"totalElements":0,"totalPages":1,"number":0}}
```
vs.
```
[{"logref":"NoSuchJobException","message":"No Job with that name either current or historic: [BAZ]","links":[]}]
```
response looks consistent compared to what was done for streams, so I think this PR looks good, merging.

jvalkeal:
Cherry-picked per https://github.com/spring-cloud/spring-cloud-dataflow/commit/b0eaeec38b18ecb0093aa174616233b7f150c6ed.

PR #1120, opened 2017-01-13
Exception handling at TaskExecution controller

### LINKED ISSUE #1089: REST: Retrieving task executions by non-existing task name should return a 404
For REST endpoint:

**GET** `/tasks/executions?name=my-task-name` (`TaskExecutionController#retrieveTasksByName`)

We currently return a **200** HTTP status code when providing a non-existing task-name (provided by `name` request parameter).

Instead we should return a **404**.

This would be consistent with the REST endpoint:

**GET** `/jobs/executions?name=my-job-name`
(`JobExecutionController#retrieveJobsByName`)

which throws a `NoSuchJobException`, resulting in a **404**.

ilayaperumalg:
- When the task executions are retrieved by the task `name`, check if the task exists
     - If the task doesn't exist, throw NoSuchTaskException

  - Use TaskDefinitionRepository at TaskExecutionController to find task definition by name
  - Update test

Resolves #1089

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowControllerAutoConfiguration.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowControllerAutoConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 the original author or authors.
+ * Copyright 2016-2017 the original author or authors.
  *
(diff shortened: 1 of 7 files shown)


jvalkeal:
Did same type of tests than with other similar PR's, looks good, merging.

jvalkeal:
Cherry-picked per https://github.com/spring-cloud/spring-cloud-dataflow/commit/7c82212d86e160d5128891e2705efb9a7dc7caf5.

PR #1127, opened 2017-01-17
Fix RDBMS registry find method

### LINKED ISSUE #1126: Fix RDBMSRegistry to handle IllegalArgumentException on find()
Based on this https://github.com/spring-cloud/spring-cloud-deployer/issues/176, we need to fix the RDBMS registry find(key) method to throw `IllegalArgumentException` when there is no app found by the key

ilayaperumalg:
- Throw IllegalArgumentException when the URI for the given app name is not found

Resolves #1126

### DIFF
--- a/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/RdbmsUriRegistry.java
+++ b/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/RdbmsUriRegistry.java
@@ -89,11 +89,13 @@ public class RdbmsUriRegistry implements UriRegistry {
 		Assert.hasText(uri.getSchemeSpecificPart(), "Error when registering " + name + " with URI " + uri +
 				": URI scheme-specific part must be specified");
 		String uriString = uri.toString();
-		if (find(name) != null) {
-			jdbcTemplate.update(UPDATE_SQL, new Object[] { uriString, name },
-					new int[] { Types.VARCHAR, Types.VARCHAR });
+		try {
+			if (find(name) != null) {
+				jdbcTemplate.update(UPDATE_SQL, new Object[]{uriString, name},
+						new int[]{Types.VARCHAR, Types.VARCHAR});
+			}
 		}
-		else {
+		catch (IllegalArgumentException e) {
 			jdbcTemplate.update(INSERT_SQL, new Object[] { name, uriString },
 					new int[] { Types.VARCHAR, Types.VARCHAR });
 		}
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/RdbmsUriRegistry.java
```diff
@@ -65,9 +66,9 @@ public URI find(String name) {
 			uriString = jdbcTemplate.queryForObject(SELECT_URI_SQL, String.class, name);
 		}
 		catch (EmptyResultDataAccessException e) {
-			return null;
+			throw new IllegalArgumentException("No URI is registered for ["+ name + "] "+ e);
 		}
-		return uriString != null ? toUri(uriString) : null;
```
ilayaperumalg: I believe `null` or `non-text` values need not be checked here as the `register` operation has validation for those before registering.

jvalkeal: Well yes, now it does what interface javadoc says.

jvalkeal:
I do get test failures with this
```
Tests in error: 
  testRegisterOverwrite(org.springframework.cloud.dataflow.server.registry.RdbmsUriRegistryTests): No URI is registered for [source.test1] org.springframework.dao.EmptyResultDataAccessException: Incorrect result size: expected 1, actual 0
  testFindAll(org.springframework.cloud.dataflow.server.registry.RdbmsUriRegistryTests): No URI is registered for [source.test1] org.springframework.dao.EmptyResultDataAccessException: Incorrect result size: expected 1, actual 0
  testFind(org.springframework.cloud.dataflow.server.registry.RdbmsUriRegistryTests): No URI is registered for [source.test1] org.springframework.dao.EmptyResultDataAccessException: Incorrect result size: expected 1, actual 0
  testUnregister(org.springframework.cloud.dataflow.server.registry.RdbmsUriRegistryTests): No URI is registered for [source.test1] org.springframework.dao.EmptyResultDataAccessException: Incorrect result size: expected 1, actual 0

Tests run: 229, Failures: 0, Errors: 4, Skipped: 0
```

ilayaperumalg:
@jvalkeal Sorry, I missed it with some changes just before push. will fix it.

ilayaperumalg:
Pushed the changes to handle `IllegalArgumentException` at the `register` method.

As mentioned [here] (https://github.com/spring-cloud/spring-cloud-dataflow/issues/1126#issuecomment-273196002) it might be better to handle all this at the deployer SPI level by returning null on the apps that are not found. But, is that ok to change this now at the deployer level?

jvalkeal:
I think this is ok as I would not just change SPI(maybe later but not in a point release).

jvalkeal:
Merged/squashed per https://github.com/spring-cloud/spring-cloud-dataflow/commit/3148fbf5ced0043e0f68fdd676c900a7b6556a3d

PR #1131, opened 2017-01-18
Fix docs, used to have continuation markers not needed

ericbottard:


ericbottard:
Merged as 9775275860331ea4c5601e5b4a9080aa7a234970

Issue #1055, opened 2016-12-12
add "task execution cleanup <id>" shell command


ericbottard:
+ REST server implementation

PR #1132, opened 2017-01-20
Implement task execution cleanup

ericbottard:
Fixes https://github.com/spring-cloud/spring-cloud-dataflow/issues/1055

### DIFF
--- a/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/TaskOperations.java
+++ b/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/TaskOperations.java
@@ -71,4 +71,10 @@ public interface TaskOperations {
 	 */
 	TaskExecutionResource taskExecutionStatus(long id);
 
+	/**
+	 * Cleanup any resources associated with the execution for the id specified.
+	 *
+	 * @param id identifier of the task execution
+	 */
+	void cleanup(long id);
 }
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
@@ -30,8 +30,6 @@ import org.springframework.cloud.dataflow.core.TaskDefinition;
 import org.springframework.cloud.dataflow.core.TaskDefinition.TaskDefinitionBuilder;
 import org.springframework.cloud.dataflow.rest.util.DeploymentPropertiesUtils;
 import org.springframework.cloud.dataflow.server.controller.WhitelistProperties;
-import org.springframework.cloud.dataflow.server.repository.DeploymentIdRepository;
-import org.springframework.cloud.dataflow.server.repository.DeploymentKey;
 import org.springframework.cloud.dataflow.server.repository.NoSuchTaskDefinitionException;
 import org.springframework.cloud.dataflow.server.repository.TaskDefinitionRepository;
 import org.springframework.cloud.dataflow.server.service.TaskService;
@@ -39,6 +37,9 @@ import org.springframework.cloud.deployer.resource.registry.UriRegistry;
 import org.springframework.cloud.deployer.spi.core.AppDefinition;
 import org.springframework.cloud.deployer.spi.core.AppDeploymentRequest;
 import org.springframework.cloud.deployer.spi.task.TaskLauncher;
+import org.springframework.cloud.task.repository.TaskExecution;
+import org.springframework.cloud.task.repository.TaskExplorer;
+import org.springframework.cloud.task.repository.TaskRepository;
 import org.springframework.core.io.Resource;
 import org.springframework.core.io.ResourceLoader;
 import org.springframework.util.Assert;
@@ -60,10 +61,18 @@ public class DefaultTaskService implements TaskService {
 	private final DataSourceProperties dataSourceProperties;
 
 	/**
-	 * The repository this service will use for deployment IDs.
+	 * Used to create TaskExecutions.
 	 */
-	private final DeploymentIdRepository deploymentIdRepository;
+	private final TaskRepository taskExecutionRepository;
 
+	/**
+	 * Used to read TaskExecutions.
+	 */
+	private final TaskExplorer taskExplorer;
+
+	/**
+	 * Used to launch apps as tasks.
+	 */
 	private final TaskLauncher taskLauncher;
 
 	/**
@@ -76,7 +85,7 @@ public class DefaultTaskService implements TaskService {
 	 */
 	private final ResourceLoader resourceLoader;
 
-	private final TaskDefinitionRepository repository;
+	private final TaskDefinitionRepository taskDefinitionRepository;
 
 	private final WhitelistProperties whitelistProperties;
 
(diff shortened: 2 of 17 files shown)


spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/TaskOperations.java
```diff
@@ -71,4 +71,10 @@
 	 */
 	TaskExecutionResource taskExecutionStatus(long id);
 
+	/**
```
ilayaperumalg: For all the java files changed, please update year 2017 in copyright header.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
```diff
@@ -76,90 +85,85 @@
...
+			TaskDefinitionRepository taskDefinitionRepository,
+			TaskExplorer taskExplorer,
+			TaskRepository taskExecutionRepository, UriRegistry registry,
 			ResourceLoader resourceLoader, TaskLauncher taskLauncher,
 			ApplicationConfigurationMetadataResolver metaDataResolver) {
 		Assert.notNull(dataSourceProperties, "DataSourceProperties must not be null");
-		Assert.notNull(repository, "TaskDefinitionRepository must not be null");
-		Assert.notNull(deploymentIdRepository, "DeploymentIdRepository must not be null");
+		Assert.notNull(taskDefinitionRepository, "TaskDefinitionRepository must not be null");
+		Assert.notNull(taskExecutionRepository, "DeploymentIdRepository must not be null");
```
ilayaperumalg: DeploymentIdRepository -> TaskRepository

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
```diff
@@ -183,4 +187,22 @@ private AppDefinition mergeAndExpandAppProperties(TaskDefinition original, Resou
...
+			dataSourceProperties.getUsername());
+		// password may be empty
+		if (StringUtils.hasText(dataSourceProperties.getPassword())) {
+			builder.setProperty("spring.datasource.password",
+				dataSourceProperties.getPassword());
+		}
+		builder.setProperty("spring.datasource.driverClassName",
+			dataSourceProperties.getDriverClassName());
+
+		builder.setProperty("spring.cloud.task.executionid", String.valueOf(taskExecution.getExecutionId()));
```
ilayaperumalg: This may have been discussed already. I'd like to know why this property name isn't `spring.cloud.task.executionId` but `...executionid`? It might be better with `executionId` (especially when using SPRING_CLOUD_TASK_EXECUTION_ID) \cc @cppwfs @mminella

ericbottard: Can't agree more, but obviously this reaches far beyond the scope of this PR

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
```diff
@@ -76,90 +85,85 @@
...
 		AppDefinition revisedDefinition = mergeAndExpandAppProperties(taskDefinition, resource, appDeploymentProperties);
+
 		AppDeploymentRequest request = new AppDeploymentRequest(revisedDefinition, resource, deployerDeploymentProperties, commandLineArgs);
 
 		String id = this.taskLauncher.launch(request);
 		if (!StringUtils.hasText(id)) {
 			throw new IllegalStateException("Deployment ID is null for the task:"
 					+ taskName);
 		}
-		String deploymentKey = DeploymentKey.forTaskDefinition(taskDefinition);
```
ilayaperumalg: Looks like this specific change requires more cleanup/discussion. The reasons are:

1) The DeploymentIdRepository is used by the `runtime` apps to show the running apps. This means, the task apps will no longer be displayed with this change
2) The Task destroy operation still removes the id from `DeploymentIdRepository`. If we don't want to go with the `DeploymentIdRepository` for `tasks` then I think all references of it need to be removed for task side of things.

ericbottard: Overlook from my side, especially 2).

The rationale, which we arrived at with @mminella is that for one given task, there may be several executions. The use of `DeploymentKey.forTaskDefinition()` as it were was hence completely flawed.

I suggest to do the minimum in this PR to avoid major breakage and open a second, separate PR, to get rid of `DeploymentKey.forTaskDefinition()`

ilayaperumalg: ok, sure.

spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/TaskCommands.java
```diff
@@ -198,6 +200,15 @@ public Table display(
 		return builder.build();
 	}
 
+	@CliCommand(value = TASK_EXECUTION_CLEANUP, help = "Clean up any platform specific resources linked to a task execution")
+	public String cleanup(
+		@CliOption(key = { "", "id" },
+			help = "the task execution id",
+			mandatory = true) long id) {
+		taskOperations().cleanup(id);
+		return String.format("Cleaned up resources for task execution %s", id);
```
ilayaperumalg: I guess the `cleanup` may be synchronous/asynchronous based on the implementation. Can we have the return message something like `Cleanup request for task execution <id> is submitted`

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
ilayaperumalg: Created https://github.com/spring-cloud/spring-cloud-dataflow/issues/1134

spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/TaskCommands.java
```diff
@@ -206,7 +206,7 @@ public String cleanup(
 			help = "the task execution id",
 			mandatory = true) long id) {
 		taskOperations().cleanup(id);
-		return String.format("Cleaned up resources for task execution %s", id);
+		return String.format("Request to clean up resources for task execution %s has been submitted", id);
```
ilayaperumalg: This requires the change in Test class as well :-) Can fix while merging.

ericbottard: Ahhh, sorry. Quick push ;)

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
cppwfs: One simple case is if the user wanted to use pass in the commandLine arg `executionId` for their own application. The prefix of spring.cloud.task marks the executionID specifically for the use by Spring Cloud Task.

ilayaperumalg: yeah, but my question was using `spring.cloud.task.executionId` vs `spring.cloud.task.executionid` not the explicit property `executionId` without prefix. Sorry if that wasn't clear.

ilayaperumalg:
LGTM, merging.

Issue #1133, opened 2017-01-23
Provide `/app-starters` rest endpoint


ghillert:
Provide a REST endpoint that returns a list of available App Starters:

GET '/app-starters'

This should be a collection of name/uri for the various app starters, e.g:

```
...
[name: 'Maven based Stream Applications with RabbitMQ Binder',
uri:  'http://bit.ly/1-0-2-GA/stream-applications-rabbit-maven'],
...
```
Is depended on by https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/146

PR #1140, opened 2017-01-24
gh-1133 Provide `/app-starters` rest endpoint

ghillert:
* Ensure that the rest endpoint is `features-aware` - E.g. if tasks are disabled, then don't return the task-specific app starters
* Add tests

resolves https://github.com/spring-cloud/spring-cloud-dataflow/issues/1133

Is depended on by https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/146

### DIFF
--- a/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/resource/AppStarter.java
+++ b/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/resource/AppStarter.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.cloud.dataflow.rest.resource;
+
+/**
+ * Provides meta information for an App Starter, such as name and a uri.
+ *
+ * @author Gunnar Hillert
+ */
+public class AppStarter {
+
+	private String name;
+	private String uri;
+
+	public AppStarter() {
+		super();
+	}
+
+	public String getName() {
+		return name;
+	}
+	public void setName(String name) {
+		this.name = name;
+	}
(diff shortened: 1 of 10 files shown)


spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/resource/AppStarter.java
```diff
@@ -0,0 +1,50 @@
...
+ * Provides meta information for an App Starter, such as name and a uri.
+ *
+ * @author Gunnar Hillert
+ */
+public class AppStarter {
+
+	private String name;
+	private String uri;
+
+	public AppStarter() {
```
ilayaperumalg: I think this can be removed.

ilayaperumalg: will fix when merging.

ilayaperumalg:
LGTM, merging.

ericbottard:
Call me stupid, and sorry for being late to the party, but doesn't this re-introduce coupling between the server and the ootb apps?

ilayaperumalg:
No problem as I haven't pushed it yet ;)

To your question, I guess it is useful to have this kind of endpoint but yes, it does re-introduce the term 'app-starters' into SCDF

ericbottard:
Not only that, but it couples the version of the server with the version of the starters (or apps, to be more precise).

As a matter of fact, I think that if we push through with this, the endpoint should definitely not be `/app-starters`. What we reference are not *starters*, they're full fledged apps. I'd be curious to hear your thoughts on all this @markfisher @markpollack

ilayaperumalg:
I think it doesn't have to tie to a specific version of server vs apps. We can list all supported app starter URLs as part of this REST endpoint as this endpoint could `only` act as the info for all the supported URLs.

markfisher:
I agree with Eric. This reintroduces the exact coupling we moved away from earlier. An endpoint for the starters should *not* be included within the server itself IMO.

ghillert:
@markfisher I only store meta-information in the yaml file. The issue is that until now I hard-coded the versions in the UI code (bad). Having it on the server, also allows users to provided their own "links" for streams and tasks, so it is not only be our starters being stored here. @sabbyanandan

markfisher:
It sounds to me like this should be taking the shape of the "catalog" we've discussed in the past, but the metadata itself should be hosted elsewhere and imported by the UI (in the same way that a user would import any set of available apps, even their own). Also, as Eric mentioned, it's definitely not right to refer to these as "starters".

sabbyanandan:
Given the core premise of SCDF and the idea of promoting microservice patterns, this feature specifically deviates from this thinking. The tight coupling of app-starter releases (_with the server_) and the bulk import workflow creates confusion, which we would like to avoid; hence, this UI feature will be removed.

PR #1144, opened 2017-01-30
TaskService returns task execution id upon task launch

### LINKED ISSUE #1143: TaskLaunch should return taskExecutionID
User needs access to the TaskExecutionId that was generated by the executeTask method.  After changes made by https://github.com/spring-cloud/spring-cloud-dataflow/pull/1132/files#diff-f321e2778c26b9cb6d278c722f6f9409R143 SCDF took ownership of task execution generation and needs to return it via the REST API and TaskOperations.executeTask.

cppwfs:
resolves #1143

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
@@ -128,7 +128,7 @@ public class DefaultTaskService implements TaskService {
 	}
 
 	@Override
-	public void executeTask(String taskName, Map<String, String> taskDeploymentProperties,
+	public long executeTask(String taskName, Map<String, String> taskDeploymentProperties,
 			List<String> commandLineArgs) {
 		Assert.hasText(taskName, "The provided taskName must not be null or empty.");
 		Assert.notNull(taskDeploymentProperties,
@@ -155,6 +155,7 @@ public class DefaultTaskService implements TaskService {
 					+ taskName);
 		}
 		taskExecutionRepository.updateExternalExecutionId(taskExecution.getExecutionId(), id);
+		return taskExecution.getExecutionId();
 	}
 
 	@Override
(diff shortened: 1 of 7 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
```diff
@@ -155,6 +155,7 @@ public void executeTask(String taskName, Map<String, String> taskDeploymentPrope
 					+ taskName);
 		}
 		taskExecutionRepository.updateExternalExecutionId(taskExecution.getExecutionId(), id);
+		return taskExecution.getExecutionId();
```
ilayaperumalg: Given we already return the external executionId (Deployment ID) when launching and now returning the `executionId`, do these all mean the client has to wait for the launch request to complete at least in case of `synchronous` task launching?

Also, based on the `javadoc` for the TaskLauncher's `launch`:

> Launch a task for the provided AppDeploymentRequest. The returned id may later be used with cancel(String) or status(String) to cancel a task or get its status, respectively. Implementations may perform this operation asynchronously; therefore a successful launch may not be assumed upon return. To determine the status of a launch, invoke status(String).

I believe `launching` of the task always needs to be `asynchrounous` in order for the task to be **canceled**

cppwfs: This primary need for this executionId  is so that the ComposedTaskRunner knows what TaskExecution to monitor.  SCDF took ownership of creating the TaskExecutionId located [here](https://github.com/cppwfs/spring-cloud-dataflow/blob/45c6cff964474dc77620c3140205450c31ce132b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java#L143).  Any client that wants to monitor that TaskExecution has to get that taskExecutionId from Data Flow after task launch.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Rebased and merged as 3f26bd9

Issue #1136, opened 2017-01-23
Change AppRegistration to include companion metadata artifact


ericbottard:
This extends to amending the REST API to allow population of the new field, as well as change to the shell command, etc

PR #1145, opened 2017-01-31
Add AppRegistration.metadataUri

ericbottard:
Also enhance REST API and Shell to support that new field.

Fixes https://github.com/spring-cloud/spring-cloud-dataflow/issues/1136

### DIFF
--- a/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/AppRegistry.java
+++ b/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/AppRegistry.java
@@ -16,18 +16,31 @@
 
 package org.springframework.cloud.dataflow.registry;
 
+import java.io.IOException;
+import java.io.InputStream;
 import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.AbstractMap;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Properties;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import org.springframework.cloud.dataflow.core.ApplicationType;
 import org.springframework.cloud.dataflow.registry.support.NoSuchAppRegistrationException;
 import org.springframework.cloud.deployer.resource.registry.UriRegistry;
-import org.springframework.cloud.deployer.resource.registry.UriRegistryPopulator;
 import org.springframework.core.io.Resource;
 import org.springframework.core.io.ResourceLoader;
 import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
 
 /**
  * Convenience wrapper for the {@link UriRegistry} that operates on higher level
@@ -45,7 +58,9 @@ import org.springframework.util.Assert;
  */
 public class AppRegistry {
 
-	public static final String METADATA_KEY_SUFFIX = "metadata";
+	private static final Logger logger = LoggerFactory.getLogger(AppRegistry.class);
+
+	private static final String METADATA_KEY_SUFFIX = "metadata";
 
 	private final UriRegistry uriRegistry;
 
@@ -58,13 +73,9 @@ public class AppRegistry {
 
 	public AppRegistration find(String name, ApplicationType type) {
 		try {
-			URI uri = this.uriRegistry.find(key(name, type));
-			URI metadataUri = null;
-			try {
-				metadataUri = this.uriRegistry.find(metadataKey(name, type));
-			}
-			catch (IllegalArgumentException ignored) {
-			}
+			String key = key(name, type);
+			URI uri = this.uriRegistry.find(key);
+			URI metadataUri = metadataUriFromRegistry().apply(key);
 			return new AppRegistration(name, type, uri, metadataUri, this.resourceLoader);
 		}
 		catch (IllegalArgumentException e) {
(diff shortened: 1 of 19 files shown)


markfisher:
we need to decide if we're moving the populator logic completely out of deployer (as discussed in https://github.com/spring-cloud/spring-cloud-deployer/pull/182) before we merge this one

spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/AppRegistry.java
```diff
@@ -64,28 +75,56 @@ public AppRegistration find(String name, ApplicationType type) {
...
+	public List<AppRegistration> importAll(boolean overwrite, Resource... resources) {
 		List<AppRegistration> apps = new ArrayList<>();
-		for (String uri : resourceUris) {
+		for (Resource resource :resources) {
 			try {
-				Map<String, URI> registered = this.uriRegistryPopulator.populateRegistry(
-						overwrite, this.uriRegistry, uri);
+				Map<String, URI> registered = UriRegistryPopulator.populateRegistry(
+					overwrite, this.uriRegistry, resource);
 				for (Map.Entry<String, URI> entry : registered.entrySet()) {
```
ilayaperumalg: Looks like we perform validation of the app after they are registered? what happens if the assertion fails after populating the registry?

ericbottard: True. Not sure there is a way to do it earlier without duplicating logic of validation or shifting knowledge about key format etc. Ideas?

ilayaperumalg: May be we can `delete` the entries that don't pass the key validation based on the data flow app requirement?

ericbottard: Other methods validate after reading from the UriRegistry, so I think it's actually fine to leave bad entries in there, as we assume that this class may not be the sole way of writing to the registry anyway

ilayaperumalg: It may not be ok leaving the bad entries in the registry as I already see some tests failing after rebasing.

ericbottard: Then I think there really is a responsibility issue with the classes we currently have.

We should either:
* not fatally fail if keys don't match the expected pattern (just disregard them)
* or move key validation earlier, but then UriRegistryPopulator makes even less sense and definitely belongs to Data Flow, as it has knowledge of Data Flow specifics.

I'd vote for option 1

ericbottard: Moreover, UriRegistryPopulator does some validation in the case of bulk import that we don't do in the case of `save`.
I'd definitely vote for getting rid of UriRegistryPopulator altogether, and having that logic belong to AppRegistry

ilayaperumalg: Given the `Data Flow` is the only user of the `UriRegistryPopulator` currently we might as well move the class to Data Flow and deal with the data flow specific validation (validating key etc.,). Disregarding the erroneous data might have an impact from the end user (at least some confusion)

ericbottard: Working on a change that gets rid of UriRegistryPopulator altogether

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/registry/DataFlowUriRegistryPopulator.java
```diff
@@ -27,34 +30,39 @@
...
 
 	@Override
 	public void afterPropertiesSet() {
 		if (!ObjectUtils.isEmpty(this.locations)) {
-			this.populator.populateRegistry(true, this.registry, this.locations);
+			Resource[] resources = new Resource[locations.length];
+			for (int i = 0; i < resources.length; i++) {
+				resources[i] = resourceLoader.getResource(locations[i]);
+			}
+			UriRegistryPopulator.populateRegistry(true, this.registry, resources);
```
ilayaperumalg: I am confused why we still use `UriRegistryPopulator` here.

ericbottard: Good catch. Pushed a new commit on top

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/registry/DataFlowAppRegistryPopulator.java
```diff
@@ -27,34 +30,39 @@
  *
  * @author Mark Fisher
  */
-public class DataFlowUriRegistryPopulator implements InitializingBean {
+public class DataFlowAppRegistryPopulator implements InitializingBean, ResourceLoaderAware {
```
ilayaperumalg: I see that this class is not being used and there is no explicit way to specify the `import` of apps list during the SCDF server startup now. Am I missing something?

ilayaperumalg: After some discussion, we are going ahead merging this as is with the conclusion that `DataFlowAppRegistryPopulator` is only used by the tests.

ilayaperumalg:
LGTM, merging.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/registry/DataFlowAppRegistryPopulator.java
ericbottard: As discussed offline, no you're not. This class is only used in tests and has always been

ilayaperumalg:
Squashed, rebased and merged as c0d902a

Issue #1152, opened 2017-02-02
runtime apps and stream list should query AppStatus in parallel


ericbottard:
Currently, both those operations issue sequential queries to obtain individual AppStatus.
On platforms that include roundtrips and other delays, this could turn `runtime apps` and `stream list`/`task list` into very long running operations, while parallelism could be leveraged

PR #1160, opened 2017-02-03
Use parallel computing when doing runtime apps

ericbottard:
Fixes https://github.com/spring-cloud/spring-cloud-dataflow/issues/1152

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/RuntimeAppsController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/RuntimeAppsController.java
@@ -20,8 +20,11 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ForkJoinPool;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
-import org.springframework.cloud.dataflow.core.StreamAppDefinition;
 import org.springframework.cloud.dataflow.core.StreamDefinition;
 import org.springframework.cloud.dataflow.rest.resource.AppInstanceStatusResource;
 import org.springframework.cloud.dataflow.rest.resource.AppStatusResource;
@@ -81,6 +84,8 @@ public class RuntimeAppsController {
 
 	private final ResourceAssembler<AppStatus, AppStatusResource> statusAssembler = new Assembler();
 
+	private ForkJoinPool forkJoinPool = new ForkJoinPool(8);
+
 	/**
 	 * Instantiates a new runtime apps controller.
 	 *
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/RuntimeAppsController.java
```diff
@@ -99,24 +104,26 @@ public RuntimeAppsController(StreamDefinitionRepository streamDefinitionReposito
...
 		}
-		Collections.sort(values, new Comparator<AppStatus>() {
-			@Override
-			public int compare(AppStatus o1, AppStatus o2) {
-				return o1.getDeploymentId().compareTo(o2.getDeploymentId());
-			}
-		});
-		return assembler.toResource(new PageImpl<>(values), statusAssembler);
+
+		List<AppStatus> statuses = forkJoinPool.submit(() ->
```
ilayaperumalg: @ericbottard Is [this](https://blog.krecan.net/2014/03/18/how-to-specify-thread-pool-for-java-8-parallel-streams/) the reason you are using ForkJoinPool + Parallel Stream approach?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/RuntimeAppsController.java
```diff
@@ -99,24 +104,26 @@ public RuntimeAppsController(StreamDefinitionRepository streamDefinitionReposito
...
+			asList.stream()
+				.flatMap(sd -> sd.getAppDefinitions().stream())
+				.flatMap(sad -> {
```
ilayaperumalg: unfortunate choice of short hand name... sad ;)

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/RuntimeAppsController.java
ericbottard: Yes

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/RuntimeAppsController.java
```diff
@@ -81,6 +84,8 @@ public int compare(AppInstanceStatus i1, AppInstanceStatus i2) {
 
 	private final ResourceAssembler<AppStatus, AppStatusResource> statusAssembler = new Assembler();
 
+	private ForkJoinPool forkJoinPool = new ForkJoinPool(8);
```
jvalkeal: I'd recommend checking out `org.springframework.scheduling.concurrent.ForkJoinPoolFactoryBean`. As a library we should not just create threads which are beyond what user can control and that's why `ForkJoinPoolFactoryBean` allows user to set thread factory. Also we don't exactly handle any errors as that factory bean also have `uncaughtExceptionHandler`. Maybe this could be a shared component so that pool gets proper shutdown and possibly allows to complete running threads.

I can't say that I'm that familiar with `ForkJoinPool` but just throwing out default value `8` kinda smells a little.

ericbottard: Updated the PR.
For the record, 8 is the empirical value I came to when testing on a backend that has ~1s roundtrip time (CF) with a non trivial amount of apps (20). conclusion is that the FJP is very reluctant to steal work on low number of tasks (but which are IO blocked). Hardly saw 6 working at the same time. So 8 seemed like a good compromise.

Anyway, made that a bean in latest commit

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/RuntimeAppsController.java
ericbottard: Added comment in the code to clarify

jvalkeal:
ok, I think this look good. Merging.

jvalkeal:
Squashed/merged per https://github.com/spring-cloud/spring-cloud-dataflow/commit/3716c3cf515fb54ac2c60f6582bd821c73b61640

Issue #1158, opened 2017-02-03
Provide the User-associated roles via the `/security/info`


ghillert:
Provide the User-associated roles via `/security/info` (`SecurityController`)

alternatively we may also refactor the `LoginController` and instead of just returning the authentication-token, we may return dedicated user-details incl. the list of the users roles.

PR #1161, opened 2017-02-06
gh-1158 Provide the User-associated roles via the `/security/info`

ghillert:
Resolves https://github.com/spring-cloud/spring-cloud-dataflow/issues/1158


jvalkeal:
LGTM, merging

Issue #1153, opened 2017-02-02
Improve status messages from "stream list"


ericbottard:
We're currently displaying the raw AppDeploymentStatus enum in the shell when doing stream list / runtime apps.
This is worrisome now that we're returning `error` from some implementations. `error` means that we couldn't compute the status, **NOT** that the app is in error (that is `failed`).

We should map those codes to something more user friendly

trisberg:
Not able to cherry-pick this commit to master branch: https://github.com/spring-cloud/spring-cloud-dataflow/commit/97bfd2ed7c29d6840d734985b02865e0217014d8

Seems there are some other changes made in the RuntimeAppsController that conflict with these changes. The tests fail after I apply the cherry-picked commit.

We need to create a separate PR for these changes and apply them to the master branch.

trisberg:
Create new issue for master branch PR - 1174

PR #1169, opened 2017-02-13
[1.1.x+master] Clarify meaning of DeploymentStates when displayed

ericbottard:
Fixes https://github.com/spring-cloud/spring-cloud-dataflow/issues/1153

Please also merge on master

### DIFF
--- a/spring-cloud-dataflow-server-local/pom.xml
+++ b/spring-cloud-dataflow-server-local/pom.xml
@@ -80,6 +80,11 @@
 			<version>${spring-cloud-deployer.version}</version>
 		</dependency>
 		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-deployer-local</artifactId>
+			<version>${spring-cloud-deployer.version}</version>
+		</dependency>
+		<dependency>
 			<groupId>org.springframework.boot</groupId>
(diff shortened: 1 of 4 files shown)


spring-cloud-dataflow-server-local/pom.xml
```diff
@@ -80,6 +80,11 @@
 			<version>${spring-cloud-deployer.version}</version>
 		</dependency>
 		<dependency>
+			<groupId>org.springframework.cloud</groupId>
```
trisberg: Any dependencies we need to add in this pom to me indicates that we have an upstream issue, either in `spring-cloud-starter-dataflow-server-local` or in `spring-cloud-deployer-local`.

trisberg:
Removed the dependency added for `spring-cloud-deployer-local`, correct version should already be pulled in after this PR is merged: https://github.com/spring-cloud/spring-cloud-dataflow/pull/1172

Merged as commit 97bfd2ed7c29d6840d734985b02865e0217014d8

trisberg:
Not able to cherry-pick this commit, seems there are some other changes made in the `RuntimeAppsController` that conflict with these changes. The tests fail after I apply the cherry-picked commit.

trisberg:
Added comment on issue #1153

PR #1178, opened 2017-02-14
Handle RuntimeAppsController ResourceNotFoundException

### LINKED ISSUE #1077: RuntimeAppsController - AppStatus Never Null
When invoking:
- **GET** `/runtime/apps/123/instances`  (`RuntimeAppsController.AppInstanceController#list()`)
- **GET** `/runtime/apps/123/instances/123` (`RuntimeAppsController.AppInstanceController#display()`)
- **GET** `/runtime/apps/123` (`RuntimeAppsController#display()`)

the controller handles a _NULL_ `AppStatus` but the `AppStatus` will never be _NULL_. In case of a non-existing application, an `AppStatus` with `DeploymentState.unknown` is returned.  

Therefore, for a non-existing application, a HTTP status of `200` is returned. **Wouldn't a status of `404` the more appropriate status?** The code section:

```java
throw new ResourceNotFoundException();
```
seems to never be executed and would in fact return a `404` HTTP status.

For `/runtime/apps/123/instances/456` the `ResourceNotFoundException` is actually triggered but not handled correctly - resulting in a `500` Http status to be returned.

ilayaperumalg:
- Make sure to return 404 HttpStatus
 - Created NoSuchAppException and NoSuchAppInstanceException; added them into RestControllerAdvice for ResourceNotFound
 - Add tests

Resolves #1077

### DIFF
--- a/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/RuntimeAppsControllerTests.java
+++ b/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/RuntimeAppsControllerTests.java
@@ -1,0 +1,104 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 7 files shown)


spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/RuntimeAppsControllerTests.java
```diff
@@ -0,0 +1,104 @@
...
+import static org.springframework.test.web.servlet.result.MockMvcResultHandlers.print;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.jsonPath;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+/**
+ * @author Ilayaperumal Gopinathan
+ */
+@RunWith(SpringRunner.class)
+@SpringBootTest(classes = TestDependencies.class)
+@DirtiesContext(classMode = ClassMode.AFTER_EACH_TEST_METHOD)
```
ericbottard: Is this really necessary?

ericbottard:
Rebased and merged as b3a6e67b after removing unused imports (that referenced deleted classes)

Issue #1105, opened 2017-01-10
Make the Shell Role-aware


ghillert:
Currently, neither the **Shell** nor the **Dashboard** are aware of the user-associated roles. E.g. if the user executes operation that require non-associated roles, the user will get at best some server-generated error returned, but it may be cryptic at best.

We should make the Shell and Dashboard role-aware, meaning if a user has the View-role but lacks the Create-role, don't show the buttons that causes "create-role-requiring" actions to be executed.

Also, the shell and dashboard should return meaningful error messages pertaining to the lack of roles. For instance, the shell provides a useless message:

![image](https://cloud.githubusercontent.com/assets/685363/21796851/8253b420-d6af-11e6-9e82-dcda7695e028.png)

while the server-log provides at least:

```
org.springframework.security.access.AccessDeniedException: Access is denied
```

Also, if a user does not have the View-role (theoretically users should always have that) but nonetheless, if not associated, the users should be pretty much locked out of the application.

Issue depends on https://github.com/spring-cloud/spring-cloud-dataflow/issues/810
Issue depends on https://github.com/spring-cloud/spring-cloud-dataflow/issues/1158

ghillert:
Depends on https://github.com/spring-projects/spring-shell/issues/115

Issue #1175, opened 2017-02-14
REST - VndErrors and Error Responses - Standardize on single VndError or Array Notation


ghillert:
Currently, users of the REST API may encounter errors being reported as a single VndError:

```json
    {
        "timestamp": "2017-02-14T03:18:30.889Z",
        "status": 403,
        "error": "Forbidden",
        "message": "Access is denied",
        "path": "/streams/definitions"
    }
```

or being reported using the array notation:

```json
    [
        {
            "logref": "MissingServletRequestParameterException",
            "message": "Required String parameter 'name' is not present",
            "links":
            [
            ]
        }
    ]
```

Should we:

- support 1 or the other?
- support both?

Also, the message structure is a bit different. Should that be further standardized? This causes issues on the Shell client where we currently have only a `VndErrorResponseErrorHandler` that handles the array notation.

PR #1179, opened 2017-02-14
gh-1105 Make the Shell Role-aware

ghillert:
* Add annotation `@HasRole(RoleType.VIEW)` for `CommandMarker` classes
  - Ensure rudimentary role behavior and checking on the class-level of `CommandMarker` classes
  - Does not enforce authorization on the method-level
* Improve slightly handling of `403` (Forbidden) error responses in `VndErrorResponseErrorHandler`
  - See also https://github.com/spring-cloud/spring-cloud-dataflow/issues/1175

Resolves https://github.com/spring-cloud/spring-cloud-dataflow/issues/1105


spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/ConfigCommands.java
```diff
@@ -163,11 +174,37 @@ public String target(
...
+
+			if (securityInfoResource.isAuthenticationEnabled() && securityInfoResource.isAuthorizationEnabled()) {
+				List<CommandMarker> commandMarkersToRemove = new ArrayList<>(0);
+				for (CommandMarker commandMarker : jLineShellComponent.getSimpleParser().getCommandMarkers()) {
+					final HasRole hasRoleAnnotation = commandMarker.getClass().getAnnotation(HasRole.class);
+
+					if (hasRoleAnnotation != null && hasRoleAnnotation.value().length > 0) {
+						boolean found = false;
+						for (RoleType roleType : hasRoleAnnotation.value()) {
+							if (RoleType.VIEW.equals(roleType)) {
```
ericbottard: I don't understand this. Seems like ROLE_VIEW is the only thing supported?

ericbottard:
Couldn't this be implemented via `@CliAvailabilityIndicator` which, while not ideal, is better suited for this kind of thing?

The problem with the current approach (on top of its very intrusive behavior, which we both agree on I think) is that its granularity is __per class__ as opposed to __per command__.

Now, it would be even better if we could avoid logic duplication altogether. Back when I was a kid, there were methods in Spring Security that could tell if the current Principal had access. Could we for example imagine a way to enhance the list of returned `Links` with access (`yes|no`) info for the current principal? Only throwing ideas up in the air right now, but this is something worse considering I think.

spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/ConfigCommands.java
jvalkeal: you know, from a security perspective I don't really understand why client-side code would check any security parameters as it's fundamentally pointless. If we impose any security checks it should be done on a server.

ericbottard: I think the idea is not to __enforce__ security constraints (which __of course__ are enforced on the server), but rather to prevent the user from seeing things he doesn't have access to (remove clutter).

jvalkeal: ok, fair enough

ghillert: Correct - basically visual sugar. Same for the UI, you're downloading the whole code anyway.

ghillert:
@ericbottard This is the best I could come up with with the current REST API and current Spring Shell version. I agree it is intrusive and limited. For instance I did not see a way to limit the methods of the CommandMarker classes. Adding this support to Spring Shell proper would be ideal, e.g. as part of `@CliAvailabilityIndicator`.

ericbottard:
You can have several `@CliAvailabilityIndicator` annotated methods per command class. I'm not sure I understand what you mean by 
> For instance I did not see a way to limit the methods of the CommandMarker classes

ghillert:
@ericbottard In the context of authorization. In `jLineShellComponent` we can do `jLineShellComponent.getSimpleParser().remove(commandMarker);` and thus create filtered list of commandMarkers, I was not so sure how to accomplish that with the individual commands - It looks like the `SimpleParser#availabilityIndicators` might be the way to go. In any event that would be a more major change to Spring Shell.

ghillert:
@ericbottard I think I understand what you mean. Let me rework the PR accordingly.

PR #1183, opened 2017-02-14
Update existing docs using the new 'deployer' prefix

### LINKED ISSUE #1182: Update existing docs using the new 'deployer' prefix
This is just updating the docs based on work done in this commit: 1ea276c.

There is another #1163 issue to improve the docs for using deployer properties in general.

trisberg:
- Also some additional changes
1.  Removing getting started doc change for disabling security for http
2.  Removing duplicated content of using YAML for the deployment properties

Resolves #1182

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
@@ -309,11 +309,11 @@ For instance, one can specify how many instances need to be deployed for the spe
 ==== Passing instance count as deployment property
 
 If you would like to have multiple instances of an application in the stream, you
-can include a property with the deploy command:
+can include a deployer property with the deploy command:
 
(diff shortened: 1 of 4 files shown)


spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -356,12 +356,29 @@ stream deploy foo --propertiesFile myprops.properties
...
+[source,bash]
+----
+stream deploy foo --propertiesFile myprops.yaml
+----
+
+where `myprops.yaml` contains:
+
+```
+app:
+  transform:
```
sabbyanandan: Shouldn't the reference to `count` be tagged under `deployer` prefix?

trisberg: ahh, yes, that didn't show up in my search, will fix

Issue #224, opened 2015-12-10
Parsing deployment properties is unreliable


jvalkeal:
Creating a stream as:
```
dataflow:>stream create --name foo --definition "time|log"
dataflow:>stream deploy --name foo --properties "module.time.dataflow.yarn.app.streamcontainer.memory=1024m,module.log.dataflow.yarn.app.streamcontainer.memory=512m,foo=bar"
```
is sending `deploymentProperties` as shown from logs:
```
YarnStreamModuleDeployer   : deploymentProperties: {dataflow.yarn.app.streamcontainer.memory=512m,foo=bar}
YarnStreamModuleDeployer   : deploymentProperties: {dataflow.yarn.app.streamcontainer.memory=1024m}
```

for `log` module key `dataflow.yarn.app.streamcontainer.memory` is getting value `512m,foo=bar` and not `512m` as expected.

1. We should do a correct parsing(it's using regex pattern which assumes presence of `module` keywork). I think contract is a comma delimited list.
2. I think properties not prefixed with`module.<id>` should be passed to all modules or silently discarded.
3. Properties work ok if I change `foo=bar` to `module.*.foo=bar`.

ilayaperumalg:
I believe the `deployment` properties are expected to be specified with a `module.` prefix. If we want to add support for properties that don't use this prefix, we can look for a delimiter `,` and parse the individual properties. But this might get complex when the property value itself has a comma (say comma separated list as a value). I think it is better we add `module.*` prefix for the global properties like `foo=bar` above. what do you think? cc @markfisher @ericbottard

ericbottard:
Yes, I too believe the original intent for deployment properties was to always target some module. Which, in retrospect, makes the `module.` prefix redundant (if not lengthy, at least).

ilayaperumalg:
Though the prefix `module.` seems redundant, it makes it easier to parse through the list of all deployment properties without the need to parse based on a delimiter which makes it complex with the scenario I mentioned above.

markpollack:
[...]

PR #1184, opened 2017-02-15
Make deployment properties parsing more reliable

### LINKED ISSUE #1097: Revisit deployment properties parsing
Currently, if properties don't start with `app.` (and soon with `deployer.`), the ill formed property gets appended to the previous one, resulting in surprising behavior. This should be changed and fail earlier

See https://github.com/spring-cloud/spring-cloud-dataflow/pull/1047#discussion_r94823700

jvalkeal:
It's easier to play with used regex by using saved online editor at https://regex101.com/r/rPObYn/1

- Modify used regex to be more clever by using
  capturing groups.
- Fixes #1097

### DIFF
--- a/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
+++ b/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2016 the original author or authors.
+ * Copyright 2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -20,6 +20,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Properties;
 import java.util.TreeMap;
 import java.util.regex.Matcher;
@@ -48,7 +49,7 @@ public final class DeploymentPropertiesUtils {
 	/**
 	 * Pattern used for parsing a String of comma-delimited key=value pairs.
 	 */
-	private static final Pattern DEPLOYMENT_PROPERTIES_PATTERN = Pattern.compile(",\\s*(app|deployer)\\.[^\\.]+\\.[^=]+=");
+	private static final Pattern DEPLOYMENT_PROPERTIES_PATTERN = Pattern.compile("(?:,{0,1}\\s*)((\\.*(\\w|\\*|-)+)+)(\\s*=\\s*)");
 
 	/**
 	 * Pattern used for parsing a String of command-line arguments.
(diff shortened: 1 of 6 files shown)


ericbottard:
I think the aim was really to enforce either `app.` or `deployer.` so that we can fail early. I haven't tested this change, but it seems that it doesn't do that.
What about my proposal to split on comma (and support escaping for legitimate commas in values) in addition to making sure that each key starts with either `app.` or `deployer.`?

jvalkeal:
you can't split by comma as value may itself be comma delimited list. I think it should be responsibility of a caller of this method to bark if parsing returns something else. Only searching props having `app|deployer` was the reason why regex didn't really work.

jvalkeal:
and btw, this ticket is actually dup of #224 which never got fixed.

ericbottard:
> you can't split by comma as value may itself be comma delimited list.

This is why I wrote that comma should be escapable

jvalkeal:
but hold on with this PR, I need to make some other tweaks as I forget to run full test build. *sigh*

ericbottard:
This PR doesn't really answer the initial issue of not surprising the user in case of malformed properties.

IMHO, we should:
* support commas (or any character, for that matter) in values. This means that if the logic is to split on commas, then a mechanism to escape commas is needed
* optionally, fail early (ie in the parser class) if the keys don't start with either `app.` or `deployer.`. If we don't fail there (which is ok, but then the value add of this class is low) then we need to make sure that that responsibility is indeed handled by the caller (there are numerous call sites, sadly) and the to the user, this "fails early" with a clear message

jvalkeal:
Did you check tests I added? Splitting by commas already works and we don't need to escape anything.

```
props = DeploymentPropertiesUtils.parse("foo1=bar1,app.foo2=bar2,foo3=bar3,xxx3");
assertThat(props, hasEntry("foo1", "bar1"));
assertThat(props, hasEntry("app.foo2", "bar2"));
assertThat(props, hasEntry("foo3", "bar3,xxx3"));
```
you can even put those comma-delimited lists in a middle and it still works(looks crazy) thought I didn't add test for that case.
```
props = DeploymentPropertiesUtils.parse("foo1=bar1,foo3=bar3,xxx3,app.foo2=bar2");
assertThat(props, hasEntry("foo1", "bar1"));
assertThat(props, hasEntry("app.foo2", "bar2"));
assertThat(props, hasEntry("foo3", "bar3,xxx3"));
```

This `parse` function has always been to just parse properties out from longer string, not to impose restrictions of what is accepted. `StreamDeploymentController` will then take what it supports out from those properties passed in.

ericbottard:
How do I pass a value that contains a comma then?

jvalkeal:
Not sure I follow, do you mean something else than `foo3=bar3,xxx3` in tests?

jvalkeal:
I'd open the link in description to online regex editor which explains how pattern searches keys. As key always may start with comma, have multiple sequences separated with dot and ends with equal sign which may be pre/postfixed with space. Then we know position of keys in a string and we can try to calculate where matching keys/values are.

It's same as current one which assumes to find `app|deployer` in a key which then caused parser to blow up with any other normal key/value. This parser will blow up if value has something which looks like a valid key which imho will always be a limitation unless we move over to proper dsl parser.

ericbottard:
Testing this on a real case scenario (@trisberg's [comment](https://github.com/spring-cloud/spring-cloud-dataflow/pull/1047#discussion_r94823700)), seems like we end up in a worse position than we currently have, as we don't even get an error now).

jvalkeal:
If you refer `deployer.*.memory=512m,server.port=8787` what error would you expect. Didn't try this but I'd expect parser to return two pairs, `deployer.*.memory=512m` and `server.port=8787`. Did you got something else?

ericbottard:
I believe that's what we got, but given that we don't longer fail early in this class, these two pairs were passed as is (and `server.port` was not even considered, as it neither starts with `app.` or `deployer.`)

jvalkeal:
One problem in this is that we haven't really documented what is a format of this deployment properties string. I think it should be responsibility of a caller of that function to fail if parser returns something it doesn't like. I know that in case of tasks, using `deployer.*` or `deployer.<taskname>` is redundant as there is only one task but you are still forced to use same format as in streams.

jvalkeal:
@ericbottard ok, rebased, combined and force pushed this branch.

btw, I noticed for y'day discussion regarding that yml file mode should be used for more complex properties combinations vs. inlined is very much pointless commend as stuff from file is just converted to same comma-delimited format. This means that inline format needs to support exactly same stuff as file format.

ericbottard:
Yeah, well I'm not sure I understand the role of [this call](https://github.com/spring-cloud/spring-cloud-dataflow/blob/e99d360e3ccb4a2779a3d6e332f81fa49dd187cb/spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/StreamCommands.java#L145), as the parsing to key-value pairs is already handled by `.properties / .yml` syntax.

jvalkeal:
nope that's not what I meant. I didn't realise this earlier but yml syntax is simply converted into inline format and then sent of the wire meaning it doesn't help at all when we try to parse those properties. Properties in yml are still parsed as inlined format on a server side.

ericbottard:
I don't follow. In the case of `.yml`, they're converted to a `Properties` object (thus obeying the yml rules for escaping, etc) and sent as already separated key-value pairs. The yml case doesn't go through `DeploymentPropertiesUtils.parse(String)`

jvalkeal:
ah, you're right. Don't know why I thought that yml from shell would be handled by `DeploymentPropertiesUtils.format` but it uses `DeploymentPropertiesUtils.convert` indeed.

jvalkeal:
no, hold on it's `StreamTemplate` which still does a conversion to inline format from map`
```
	@Override
	public void deploy(String name, Map<String, String> properties) {
		MultiValueMap<String, Object> values = new LinkedMultiValueMap<>();
		values.add("properties", DeploymentPropertiesUtils.format(properties));
		restTemplate.postForObject(deploymentLink.expand(name).getHref(), values, Object.class);
	}
```

ericbottard:
Urgh, indeed. And the parsing happens on the other side, on the server.
We should change this so that that a `Properties` / JSON object goes on the wire.

Not as part of this PR, though

ericbottard:
Rebased and merged as 28957632.

I still found some cases that go through this though:
```
dataflow:>stream deploy foo --properties " app. http.bar=3,deployer.goo.koko=4"
Command failed java.lang.IllegalArgumentException: Only deployment property keys starting with 'app.' or 'deployer.' allowed, got 'http.bar'
dataflow:>stream deploy foo --properties " app. app.http.bar=3,deployer.goo.koko=4"
Deployment request has been sent for stream 'foo'
```

Nevertheless, created #1228. The sooner that issue is taken care of, the sooner we can change the behavior of `#parse()` to be dead trivial and warn the user to leverage `.properties/.yml` files for "complicated" cases. Especially now that we have a dedicated method for looking at keys that don't start with the correct prefix

PR #1190, opened 2017-02-17
gh-1105 Make the Shell Role-aware

ghillert:
* Add role handling to methods annotated with `@CliAvailabilityIndicator`
* Add access handling rules to `DataFlowShell#hasAccess`
* Add tests
* Improve slightly handling of `403` (Forbidden) error responses in `VndErrorResponseErrorHandler`
  - See also https://github.com/spring-cloud/spring-cloud-dataflow/issues/1175

Resolves https://github.com/spring-cloud/spring-cloud-dataflow/issues/1105
Depends on https://github.com/spring-projects/spring-shell/issues/115

### DIFF
--- a/spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/ConfigCommands.java
+++ b/spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/ConfigCommands.java
@@ -23,6 +23,8 @@ import java.util.TreeMap;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.beans.BeansException;
+import org.springframework.beans.factory.InitializingBean;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.context.event.ApplicationReadyEvent;
@@ -34,11 +36,11 @@ import org.springframework.cloud.dataflow.shell.TargetHolder;
 import org.springframework.cloud.dataflow.shell.command.support.HttpClientUtils;
 import org.springframework.cloud.dataflow.shell.command.support.RoleType;
 import org.springframework.cloud.dataflow.shell.config.DataFlowShell;
-import org.springframework.context.ApplicationEvent;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
 import org.springframework.context.ApplicationListener;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
-import org.springframework.context.event.ContextRefreshedEvent;
 import org.springframework.core.env.Environment;
 import org.springframework.hateoas.config.EnableHypermediaSupport;
 import org.springframework.hateoas.config.EnableHypermediaSupport.HypermediaType;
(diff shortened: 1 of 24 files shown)


spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/config/DataFlowShell.java
```diff
@@ -16,25 +16,125 @@
...
+	 * @param role Can be null.
+	 * @param opsType Must not be null.
+	 * @return If true the indicated operation is accessible and the user has all credentials to execute the operation.
+	 */
+	public boolean hasAccess(RoleType role, OpsType opsType) {
+		if (this.dataFlowOperations != null) {
+			boolean operationsAvailable = false;
+			switch(opsType) {
+				case AGGREGATE_COUNTER:
+					operationsAvailable = this.dataFlowOperations.aggregateCounterOperations() != null;
```
ericbottard: I feel like all these "available" checks belong to the commands, where they were initially. Why not `&&` them in the `@CliAvailabilityIndicator` methods?

spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/ConfigCommands.java
```diff
@@ -234,25 +245,15 @@ public String info() {
...
-		//Only invoke if the shell is executing in the same application context as the data flow server.
+	public void onApplicationEvent(ApplicationEvent event) {
 		if (!initialized) {
-			target(this.serverUri, this.userName, this.password, this.skipSslValidation);
-		}
-	}
-
-	@Override
-	public void afterPropertiesSet() throws Exception {
-		//Only invoke this lifecycle method if the shell is executing in stand-alone mode.
```
ericbottard: I can't say I understood the rationale for the previous logic and the comment that went with it, but haven't the semantics changed with the commit you're proposing?

ericbottard: My bad, I realize now that you even removed the `afterPropertiesSet ()` method. But then, isn't `initialized` always false?

spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/config/DataFlowShell.java
ghillert: I wanted to decouple the checking from the commands as much as possible for 2 reasons:

* Realizing that adding the access logic is quite repetitive, I wanted to make the code in the command-class look at simple as possible - 1 line:

```
return dataFlowShell.hasAccess(RoleType.VIEW, OpsType.AGGREGATE_COUNTER);
```

* ability to test the logic in insolation

spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/ConfigCommands.java
ghillert: Ooops, that is a hang-over from the previous PR. Wish that code would have been a bit cleaner. Will fix.

ghillert:
@ericbottard  Made the changes to Config Command. Let me know if any further changes are needed.

ericbottard:
Squashed and merged as c9f37779

Issue #1192, opened 2017-02-17
Add API version checking to shell


trisberg:
The shell should check the version of the server API via "api":"version" from `/management/info` endpoint ( see #1191 )

If the server major.minor version (1.2.x, 1.3.x etc.) is not the same as the shells and is 1.2 or greater then we should redirect to the about page for the server at `/dashboard/index.html#/about` for a download link. If the server API isn't available or is less than "1.2" then just show an error message to download a compatible shell based on the "app":"version".

Issue #1191, opened 2017-02-17
Add API version to '/management/info' endpoint


trisberg:
The `/management/info` endpoint currently has this:

```
{
  "app": {
    "name": "spring-cloud-dataflow-server-local",
    "description": "Spring Cloud Data Flow Local Server",
    "version": "1.2.0.M1"
  },
  "git": {
    "commit": {
      "time": "2017-02-15T00:06:45.000Z",
      "id": "06674a6"
    },
    "branch": "master"
  }
}
```

We could add an "api":"version" entry as well, this way the shell can query the compatible version and if it's not deemed compatible post a link to download the shell via the about page at `/dashboard/index.html#/about`

PR #1202, opened 2017-02-27
Implement REST API revision numbering and check by shell.

ericbottard:
Fixes https://github.com/spring-cloud/spring-cloud-dataflow/issues/1192
Fixes https://github.com/spring-cloud/spring-cloud-dataflow/issues/1191

### DIFF
--- a/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/DataFlowTemplate.java
+++ b/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/DataFlowTemplate.java
@@ -27,6 +27,7 @@ import org.springframework.batch.core.JobParameter;
 import org.springframework.batch.core.JobParameters;
 import org.springframework.batch.core.StepExecution;
 import org.springframework.batch.item.ExecutionContext;
+import org.springframework.cloud.dataflow.rest.Version;
 import org.springframework.cloud.dataflow.rest.client.support.ExecutionContextJacksonMixIn;
 import org.springframework.cloud.dataflow.rest.client.support.ExitStatusJacksonMixIn;
 import org.springframework.cloud.dataflow.rest.client.support.JobExecutionJacksonMixIn;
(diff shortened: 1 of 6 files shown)


jvalkeal:
Unfortunately api info cannot be behind management endpoints.

1. I think we don't require `management` to be active, thus starting server with `--management.port=-1` gives no error unless you manually do `dataflow config server` and even then all you get is `Unable to contact Data Flow Server at 'http://localhost:9393': 'org.springframework.cloud.dataflow.rest.client.DataFlowClientException: '.`
2. User can change management port and context path so we can't hard code those.
3. Management may have totally different security settings, etc, so shell should also have different settings.

ericbottard:
That's all valid points. Do you agree @trisberg ? I got carried away with your example that used boot, but i think it would be better indeed to rather use our "root" endpoint for this

jvalkeal:
I wonder if we could just expose api info via `SecurityController` or via something similar as i.e. `SecurityController` is not secured ever. I assume a shell should also be able to access api version if even if whole rest is secured.

Doing things other way around like validating client version on a server side with every request is one other option but that adds a lot of complexity and there is a tons of discussions in SO how right or wrong that is.

jvalkeal:
Maybe @ghillert can comment how UI JS code works with rest? In a sense that communication level should also be on par with rest api revision. What I'm saying that if we change api and try to make it easier for user to understand that client/server don't speak same level of language anymore, should this same rule also happen with UI code? I'd assume this will happen if we update rest api and forget to change UI.

trisberg:
I agree we should avoid the `/management` path if that's not always accessible, so maybe plain `/info` is best

trisberg:
What we are trying to guard against is a user mismatch of server/shell, we control the server/UI match-up so we don't need to worry about that here I think

ericbottard:
I'm adding it to the `RootController`. I agreed with your 3 points @jvalkeal, but when the server itself is secure, that piece of information can (and should IMO) be secured as well.

jvalkeal:
Do you think we could somehow expose a correct message to a user in a shell? I don't remember where actual exception message gets lost in a shell but with this you literally only get this if management is disabled or not accessible.
```
server-unknown:>dataflow config server --uri http://localhost:9393
Unable to contact Data Flow Server at 'http://localhost:9393': 'org.springframework.cloud.dataflow.rest.client.DataFlowClientException: '.
server-unknown:>
```

Just tried to change version and then used *older* shell version and I get proper message.
```
server-unknown:>dataflow config server --uri http://localhost:9393
Unable to contact Data Flow Server at 'http://localhost:9393': 'java.lang.IllegalStateException: Incompatible version of Data Flow server detected. Trying to use shell which supports revision 10, while server revision is 11. Both revisions should be aligned'.
```

I think what happens is that call
```
Map<String, Object> info = restTemplate.getForObject(baseURI + "/management/info", Map.class);
```
fails and shell only sees `DataFlowClientException`. Maybe we should catch that as well and at minimum set version to `[unknown]` what current code does if rest call succeed but api info is not there?

jvalkeal:
You can't add it to `RootController` as it only handles root and that just for hal links.

ericbottard:
^^ the whole handling of the exception is wrong at the moment IMO, as it is not thrown but converted to a String result. Also behavior when you start or do `dataflow config server` is different.

But this should be part of a different PR IMO.


I pushed a new commit with a different approach

ericbottard:
> You can't add it to RootController as it only handles root and that just for hal links.

I'm sorry, what?

`RootController` exposes a resource, which happened to be empty until now. There's nothing wrong with having info in there, quite the contrary actually

jvalkeal:
Sorry I thought you were planning to add new mapping as you mentioned `RootController`.

jvalkeal:
squashed and fixed conflict per https://github.com/spring-cloud/spring-cloud-dataflow/commit/e99d360e3ccb4a2779a3d6e332f81fa49dd187cb

PR #1204, opened 2017-02-27
gh-ui-158 Add property `formLogin` to `SecurityInfoResource`

ghillert:
Expand the `SecurityInfoResource` by a property `formLogin` that indicates that formLogin is being used. This is needed in cases where the application needs to be "reccreated" using the xauth-token and the `SecurityInfoResource` (aka the user presses refresh) only. If Outh2 authentication is being used, this property will be `false`.

Is depended on by https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/158


jvalkeal:
@ghillert would you mind force push changes to fix conflicts?

jvalkeal:
Merged/squashed/fix conflict per https://github.com/spring-cloud/spring-cloud-dataflow/commit/c015b39d0e0211718605fafe1b8da244ce15b01e

Issue #1146, opened 2017-01-31
StreamOperations create method should be void


viniciusccarvalho:
StreamOperations currently has a signature to return `StreamDefinitionResource` at https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/StreamOperations.java#L41

The issue is that the server returns void for the create operation, and thus leaving this as  null value that clients using it could end up in NPE

ericbottard:
Same applies to `TaskDefinitionController#save()`. Whatever the decision here, those should be aligned

PR #1207, opened 2017-03-01
Make stream.save and task.save return the created resource

ericbottard:
Fixes https://github.com/spring-cloud/spring-cloud-dataflow/issues/1146


spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/StreamControllerTests.java
```diff
@@ -458,6 +458,9 @@ public void testSinkDestinationWithTwoApps() throws Exception {
 	public void testDestinationsOnBothSides() throws Exception {
 		assertEquals(0, repository.count());
 		String definition = ":bar > filter > :foo";
+		AppStatus status = mock(AppStatus.class);
```
ilayaperumalg: Not sure how does this change helps testing the return of `stream` and `task`definition resources on `save` operations.

ericbottard: I doesn't, but it prevents a NPE in a previous test

cppwfs:
LGTM.

PR #1210, opened 2017-03-03
Fix the wrong link in Docs for Getting Started

linzhaoming:
Fix the wrong link for Getting Started


ghillert:
@linzhaoming Thanks for your PR. I have merged your other docs PR that also has those changes in this PR. 

**Can this PR be just closed?**

It also contains some other deletions that may not be needed. Please let me know. Thanks.

PR #1217, opened 2017-03-06
Document the usage of metadata artifact

ericbottard:
Fixes https://github.com/spring-cloud/spring-cloud-dataflow/issues/1180

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
@@ -89,12 +89,30 @@ List of available Stream Application Starters:
 [width="100%",frame="topbot",options="header"]
 |======================
 |Artifact Type |Stable Release |SNAPSHOT Release
-|RabbitMQ + Maven |link:http://bit.ly/Avogadro-SR1-stream-applications-rabbit-maven[http://bit.ly/Avogadro-SR1-stream-applications-rabbit-maven] |link:http://bit.ly/Bacon-BUILD-SNAPSHOT-stream-applications-rabbit-maven[http://bit.ly/Bacon-BUILD-SNAPSHOT-stream-applications-rabbit-maven]
-|RabbitMQ + Docker|link:http://bit.ly/Avogadro-SR1-stream-applications-rabbit-docker[http://bit.ly/Avogadro-SR1-stream-applications-rabbit-docker] | N/A ]
-|Kafka 0.9 + Maven 	  |link:http://bit.ly/Avogadro-SR1-stream-applications-kafka-09-maven[http://bit.ly/Avogadro-SR1-stream-applications-kafka-09-maven] |link:http://bit.ly/Bacon-BUILD-SNAPSHOT-stream-applications-kafka-09-maven[http://bit.ly/Bacon-BUILD-SNAPSHOT-stream-applications-kafka-09-maven]
-|Kafka 0.9 + Docker   |link:http://bit.ly/Avogadro-SR1-stream-applications-kafka-09-docker[http://bit.ly/Avogadro-SR1-stream-applications-kafka-09-docker] | N/A ]
-|Kafka 0.10 + Maven 	  |link:http://bit.ly/Avogadro-SR1-stream-applications-kafka-10-maven[http://bit.ly/Avogadro-SR1-stream-applications-kafka-10-maven] |link:http://bit.ly/Bacon-BUILD-SNAPSHOT-stream-applications-kafka-10-maven[http://bit.ly/Bacon-BUILD-SNAPSHOT-stream-applications-kafka-10-maven]
-|Kafka 0.10 + Docker   |http://bit.ly/Avogadro-SR1-stream-applications-kafka-10-docker[http://bit.ly/Avogadro-SR1-stream-applications-kafka-10-docker] | N/A ]
+
+|RabbitMQ + Maven
+|http://bit.ly/Avogadro-SR1-stream-applications-rabbit-maven
+|http://bit.ly/Bacon-BUILD-SNAPSHOT-stream-applications-rabbit-maven
+
+|RabbitMQ + Docker
+|http://bit.ly/Avogadro-SR1-stream-applications-rabbit-docker
+|N/A
+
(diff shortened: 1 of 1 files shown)


spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -89,24 +89,48 @@ List of available Stream Application Starters:
...
-|Kafka 0.10 + Maven 	  |link:http://bit.ly/Avogadro-SR1-stream-applications-kafka-10-maven[http://bit.ly/Avogadro-SR1-stream-applications-kafka-10-maven] |link:http://bit.ly/Bacon-BUILD-SNAPSHOT-stream-applications-kafka-10-maven[http://bit.ly/Bacon-BUILD-SNAPSHOT-stream-applications-kafka-10-maven]
-|Kafka 0.10 + Docker   |http://bit.ly/Avogadro-SR1-stream-applications-kafka-10-docker[http://bit.ly/Avogadro-SR1-stream-applications-kafka-10-docker] | N/A ]
+
+|RabbitMQ + Maven
+|http://bit.ly/Avogadro-SR1-stream-applications-rabbit-maven
+|http://bit.ly/Bacon-BUILD-SNAPSHOT-stream-applications-rabbit-maven
+
+|RabbitMQ + Docker
+|http://bit.ly/Avogadro-SR1-stream-applications-rabbit-docker
+|N/A
```
ericbottard: Why is this N/A btw @sabbyanandan ?

sabbyanandan: Ah, in Docker land it is either latest or versioned tags. There's no concept of SNAPSHOTs if I'm not wrong. cc:/ @trisberg

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -164,6 +188,86 @@ Make sure to add 'spring-boot-configuration-processor' as an optional dependency
...
+ 	</executions>
+ </plugin>
+----
+
+NOTE: This plugin comes in _addition_ to the `spring-boot-configuration-processor` that creates the individual JSON files.
+Be sure to configure the two!
+
+The benefits of a companion artifact are manifold:
+
+1. being way lighter (usually a few kilobytes, as opposed to megabytes for the actual app), they are quicker to download,
```
sabbyanandan: Perhaps it might be beneficial to also list the GUI experience either in the same or separate bullet. That's where we have had folks reporting UX issues.

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -164,6 +188,86 @@ Make sure to add 'spring-boot-configuration-processor' as an optional dependency
...
+allowing quicker feedback when using _e.g._ `app info`
+2. as a consequence of the above, they can be used in resource constrained environments (such as PaaS) when metadata is
+the only piece of information needed
+3. finally, for environments that don't deal with boot uberjars directly (for example for Docker based runtimes), this
```
sabbyanandan: Might be good to list few runtime examples in the parenthesis. 

> (for example, Docker-based runtimes such as Kubernetes or Mesos)

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -164,6 +188,86 @@ Make sure to add 'spring-boot-configuration-processor' as an optional dependency
...
+is the only way to provide metadata about the properties supported by the app.
+
+Remember though, that this is entirely optional when dealing with uberjars. The uberjar itself is a source of metadata.
```
sabbyanandan: This is unclear -> "The uberjar itself is a source of metadata". If the idea is to tell that the uberjar _also_ includes the metadata in it already, maybe we could say it so?

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -164,6 +188,86 @@ Make sure to add 'spring-boot-configuration-processor' as an optional dependency
...
+
+When registering several files using the `app import` command, the file should contain a `<type>.<name>.metadata` line
+in addition to each `<type>.<name>` line. This is optional (_i.e._ if some apps have it but some others don't, that's fine).
+Here is an example for a Dockerized app:
+
+[source, properties]
+----
+...
+source.http=docker:springcloudstream/http-source-rabbit:latest
+source.http.metadata=maven://org.springframework.cloud.stream.app:http-source-rabbit:jar:metadata:1.2.0.BUILD-SNAPSHOT
```
sabbyanandan: Since the companion artifact is a "maven artifact", could we consider elaborating that in the previous section? Someone who's all about Docker-based runtimes, they might see the "maven" reference ^^ and possibly get confused.

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
ericbottard: My point, so why isn't this `latest` (with no BACON, which is always good for cholesterol)?
Somehow I feel like I've already had this question, but fail to see why we wouldn't provide it

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
ericbottard: sure, will do

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
ericbottard: will rephrase

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
ericbottard: > Since the companion artifact is a "maven artifact"

It is in my example, but might as well be `http://` or `file://`. I'll clarify nonetheless

ericbottard:
Addressed review comments

ericbottard:
Squashed and merged as b1e794f3

PR #1218, opened 2017-03-06
Adds Task Composed Restful API

### LINKED ISSUE #1141: Add task compose  command to REST API
User wants the ability to compose tasks via the restful API
Once the DSL has been submitted it will be validated and then upon  a successful submission the following will be stored in the task definition: `composed_runner --graph='<your definition>'`
If validation fails an error will be returned to the user stating the reason for the failure.  
Please refer to the design document for further details: https://docs.google.com/document/d/10FiyNrKH1KmfDj7WRVmREc9VwzSZx3XnMBP8VU8y99U/edit#heading=h.34iilmvdzjp8

### LINKED ISSUE #1142: Need to support task compose command for SCDF Shell
User wants the ability to create composed tasks via the SCDF Shell.  
The command should look like :
`task compose baz --definition “your definition”`

cppwfs:
resolves #1141 
resolves #1142

Added support for task compose to the shell
This PR includes registering the task-composed-runner at SCDF start.

### DIFF
--- a/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/TaskDefinition.java
+++ b/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/TaskDefinition.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2016 the original author or authors.
+ * Copyright 2015-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -21,6 +21,8 @@ import java.util.Map;
 
 import org.springframework.cloud.dataflow.core.dsl.ArgumentNode;
 import org.springframework.cloud.dataflow.core.dsl.AppNode;
+import org.springframework.cloud.dataflow.core.dsl.ComposedTaskNode;
+import org.springframework.cloud.dataflow.core.dsl.ComposedTaskParser;
 import org.springframework.cloud.dataflow.core.dsl.TaskParser;
 import org.springframework.cloud.deployer.spi.core.AppDefinition;
 import org.springframework.core.style.ToStringCreator;
@@ -29,6 +31,7 @@ import org.springframework.util.Assert;
 /**
  * @author Michael Minella
  * @author Mark Fisher
+ * @author Glenn Renfro
  */
 public class TaskDefinition extends DataFlowAppDefinition {
 
@@ -38,17 +41,7 @@ public class TaskDefinition extends DataFlowAppDefinition {
 	private final String dslText;
 
 	public TaskDefinition(String name, String dsl) {
-		this.dslText = dsl;
-		AppNode taskNode = new TaskParser(name, dsl).parse();
-		setRegisteredAppName(taskNode.getName());
-		Map<String, String> properties = new HashMap<>();
-		if (taskNode.hasArguments()) {
-			for (ArgumentNode argumentNode : taskNode.getArguments()) {
-				properties.put(argumentNode.getName(), argumentNode.getValue());
-			}
-		}
-		properties.put("spring.cloud.task.name", name);
-		this.appDefinition = new AppDefinition(name, properties);
+		this(name, dsl, false);
 	}
 
 	TaskDefinition(String registeredAppName, String label, Map<String, String> properties) {
@@ -56,6 +49,27 @@ public class TaskDefinition extends DataFlowAppDefinition {
 		this.dslText = "";
 		properties.put("spring.cloud.task.name", registeredAppName);
 	}
+
+	public TaskDefinition(String name, String dsl, boolean isComposedTask) {
+		this.dslText = dsl;
+		Map<String, String> properties = new HashMap<>();
+		if(isComposedTask) {
+			new ComposedTaskParser().parse(name, dsl);//validate taskName & dsl
+			setRegisteredAppName(name);
+		}
+		else {
+			AppNode taskNode = new TaskParser(name, dsl).parse();
+			setRegisteredAppName(taskNode.getName());
+			if (taskNode.hasArguments()) {
+				for (ArgumentNode argumentNode : taskNode.getArguments()) {
+					properties.put(argumentNode.getName(), argumentNode.getValue());
+				}
+			}
+		}
+		properties.put("spring.cloud.task.name", name);
+		this.appDefinition = new AppDefinition(name, properties);
+	}
+
 
 	public String getDslText() {
 		return dslText;
--- a/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/TaskDefinition.java
+++ b/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/TaskDefinition.java
@@ -35,6 +35,7 @@ import org.springframework.util.Assert;
  */
 public class TaskDefinition extends DataFlowAppDefinition {
 
+	public static final String SPRING_CLOUD_TASK_NAME = "spring.cloud.task.name";
 	/**
 	 * DSL text for the module.
 	 */
@@ -47,7 +48,7 @@ public class TaskDefinition extends DataFlowAppDefinition {
 	TaskDefinition(String registeredAppName, String label, Map<String, String> properties) {
 		super(registeredAppName, label, properties);
 		this.dslText = "";
-		properties.put("spring.cloud.task.name", registeredAppName);
+		properties.put(SPRING_CLOUD_TASK_NAME, registeredAppName);
 	}
 
 	public TaskDefinition(String name, String dsl, boolean isComposedTask) {
@@ -66,7 +67,7 @@ public class TaskDefinition extends DataFlowAppDefinition {
 				}
 			}
 		}
-		properties.put("spring.cloud.task.name", name);
+		properties.put(SPRING_CLOUD_TASK_NAME, name);
 		this.appDefinition = new AppDefinition(name, properties);
 	}
 
--- a/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/ComposedTaskTokenizer.java
+++ b/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/ComposedTaskTokenizer.java
@@ -104,7 +104,7 @@ class ComposedTaskTokenizer extends AbstractTokenizer {
 					case '\\':
 						raiseException(DSLMessage.UNEXPECTED_ESCAPE_CHAR);
 					default:
-						raiseException(DSLMessage.UNEXPECTED_DATA,Character.valueOf(ch).toString());
+						raiseException(DSLMessage.COMPOSED_TASK_UNEXPECTED_DATA,Character.valueOf(ch).toString());
 				}
 			}
 		}
--- a/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/DSLMessage.java
+++ b/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/DSLMessage.java
@@ -104,7 +104,9 @@ public enum DSLMessage {
 	COMPOSED_TASK_ARROW_SHOULD_BE_PRECEDED_BY_CODE(ERROR, 157, "Transition arrow must be preceeded by the exit code that should drive the transition"), //
 	CT_VALIDATION_SECONDARY_SEQUENCES_MUST_BE_NAMED(ERROR, 158, "Secondary sequences must have labels or are unreachable"), //
 	CT_VALIDATION_DUPLICATE_LABEL(ERROR, 159, "This label has already been defined"), //
-	CT_VALIDATION_TRANSITION_TARGET_LABEL_UNDEFINED(ERROR, 160, "Transition specifies an undefined label")
+	CT_VALIDATION_TRANSITION_TARGET_LABEL_UNDEFINED(ERROR, 160, "Transition specifies an undefined label"), //
+	COMPOSED_TASK_DEFINITION_NOT_PRESENT(ERROR, 161, "Task in composed task definition does not exist"), //
+	COMPOSED_TASK_UNEXPECTED_DATA(ERROR, 162, "unexpected data in composed task definition ''{0}''") //
 	;
 
 	private Kind kind;
--- a/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/DSLMessage.java
+++ b/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/DSLMessage.java
@@ -105,8 +105,9 @@ public enum DSLMessage {
 	CT_VALIDATION_SECONDARY_SEQUENCES_MUST_BE_NAMED(ERROR, 158, "Secondary sequences must have labels or are unreachable"), //
 	CT_VALIDATION_DUPLICATE_LABEL(ERROR, 159, "This label has already been defined"), //
 	CT_VALIDATION_TRANSITION_TARGET_LABEL_UNDEFINED(ERROR, 160, "Transition specifies an undefined label"), //
-	COMPOSED_TASK_DEFINITION_NOT_PRESENT(ERROR, 161, "Task in composed task definition does not exist"), //
-	COMPOSED_TASK_UNEXPECTED_DATA(ERROR, 162, "unexpected data in composed task definition ''{0}''") //
+	CT_ELEMENT_IN_COMPOSED_DEFINITION_DOES_NOT_EXIST(ERROR, 161, "Task in composed task definition does not exist"), //
+	COMPOSED_TASK_UNEXPECTED_DATA(ERROR, 162, "Unexpected data in composed task definition ''{0}''"), //
+	COMPOSED_TASK_MORE_INPUT(ERROR, 163, "After parsing a valid composed task, there is still more data: ''{0}''") //
 	;
 
 	private Kind kind;
--- a/spring-cloud-dataflow-docs/src/test/java/org/springframework/cloud/dataflow/server/rest/documentation/ApiDocumentation.java
+++ b/spring-cloud-dataflow-docs/src/test/java/org/springframework/cloud/dataflow/server/rest/documentation/ApiDocumentation.java
@@ -93,7 +93,7 @@ public class ApiDocumentation extends BaseDocumentation {
 					linkWithRel("runtime/apps/instances").description("Provides the status for app instances"),
 					linkWithRel("tasks/definitions").description("Provides the task definition resource"),
 					linkWithRel("tasks/definitions/definition").description("Provides details for a specific task definition"),
-					linkWithRel("tasks/definitions/compose").description("Requests the creation of a composed task definition"),
+					linkWithRel("tasks/composed-definitions/compose").description("Requests the creation of a composed task definition"),
 					linkWithRel("tasks/executions").description("Returns Task executions and allows lanching of tasks"),
 					linkWithRel("tasks/executions/name").description("Returns all task executions for a given Task name"),
 					linkWithRel("tasks/executions/execution").description("Provides details for a specific task execution"),
--- a/spring-cloud-dataflow-docs/src/test/java/org/springframework/cloud/dataflow/server/rest/documentation/ApiDocumentation.java
+++ b/spring-cloud-dataflow-docs/src/test/java/org/springframework/cloud/dataflow/server/rest/documentation/ApiDocumentation.java
@@ -93,6 +93,7 @@ public class ApiDocumentation extends BaseDocumentation {
 					linkWithRel("runtime/apps/instances").description("Provides the status for app instances"),
 					linkWithRel("tasks/definitions").description("Provides the task definition resource"),
 					linkWithRel("tasks/definitions/definition").description("Provides details for a specific task definition"),
+					linkWithRel("tasks/definitions/compose").description("Requests the creation of a composed task definition"),
 					linkWithRel("tasks/executions").description("Returns Task executions and allows lanching of tasks"),
 					linkWithRel("tasks/executions/name").description("Returns all task executions for a given Task name"),
 					linkWithRel("tasks/executions/execution").description("Provides details for a specific task execution"),
--- a/spring-cloud-dataflow-docs/src/test/java/org/springframework/cloud/dataflow/server/rest/documentation/ApiDocumentation.java
+++ b/spring-cloud-dataflow-docs/src/test/java/org/springframework/cloud/dataflow/server/rest/documentation/ApiDocumentation.java
@@ -93,7 +93,7 @@ public class ApiDocumentation extends BaseDocumentation {
 					linkWithRel("runtime/apps/instances").description("Provides the status for app instances"),
 					linkWithRel("tasks/definitions").description("Provides the task definition resource"),
 					linkWithRel("tasks/definitions/definition").description("Provides details for a specific task definition"),
-					linkWithRel("tasks/composed-definitions").description("Requests the creation of a composed task definition"),
+					linkWithRel("tasks/composed-definitions/compose").description("Requests the creation of a composed task definition"),
 					linkWithRel("tasks/executions").description("Returns Task executions and allows lanching of tasks"),
 					linkWithRel("tasks/executions/name").description("Returns all task executions for a given Task name"),
 					linkWithRel("tasks/executions/execution").description("Provides details for a specific task execution"),
--- a/spring-cloud-dataflow-docs/src/test/java/org/springframework/cloud/dataflow/server/rest/documentation/ApiDocumentation.java
+++ b/spring-cloud-dataflow-docs/src/test/java/org/springframework/cloud/dataflow/server/rest/documentation/ApiDocumentation.java
@@ -93,7 +93,7 @@ public class ApiDocumentation extends BaseDocumentation {
 					linkWithRel("runtime/apps/instances").description("Provides the status for app instances"),
 					linkWithRel("tasks/definitions").description("Provides the task definition resource"),
 					linkWithRel("tasks/definitions/definition").description("Provides details for a specific task definition"),
-					linkWithRel("tasks/composed-definitions/compose").description("Requests the creation of a composed task definition"),
+					linkWithRel("tasks/composed-definitions").description("Requests the creation of a composed task definition"),
 					linkWithRel("tasks/executions").description("Returns Task executions and allows lanching of tasks"),
 					linkWithRel("tasks/executions/name").description("Returns all task executions for a given Task name"),
 					linkWithRel("tasks/executions/execution").description("Provides details for a specific task execution"),
--- a/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/TaskTemplate.java
+++ b/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/TaskTemplate.java
@@ -20,6 +20,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
+import org.springframework.cloud.dataflow.rest.resource.ComposedTaskResource;
 import org.springframework.cloud.dataflow.rest.resource.TaskDefinitionResource;
 import org.springframework.cloud.dataflow.rest.resource.TaskExecutionResource;
 import org.springframework.cloud.dataflow.rest.util.DeploymentPropertiesUtils;
@@ -41,7 +42,7 @@ public class TaskTemplate implements TaskOperations {
 
 	/*default*/ static final String DEFINITIONS_RELATION = "tasks/definitions";
 
-	private static final String COMPOSE_RELATION = "tasks/definitions/compose";
+	private static final String COMPOSE_RELATION = "tasks/composed-definitions/compose";
 
 	private static final String DEFINITION_RELATION = "tasks/definitions/definition";
 
@@ -105,12 +106,12 @@ public class TaskTemplate implements TaskOperations {
 	}
 
 	@Override
-	public TaskDefinitionResource compose(String name, String definition) {
+	public ComposedTaskResource compose(String name, String definition) {
 		MultiValueMap<String, Object> values = new LinkedMultiValueMap<String, Object>();
 		values.add("name", name);
 		values.add("definition", definition);
-		TaskDefinitionResource task = restTemplate.postForObject(
-				composeLink.expand().getHref(), values, TaskDefinitionResource.class);
+		ComposedTaskResource task = restTemplate.postForObject(
+				composeLink.expand().getHref(), values, ComposedTaskResource.class);
 		return task;
 	}
 
--- a/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/TaskTemplate.java
+++ b/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/TaskTemplate.java
@@ -42,7 +42,7 @@ public class TaskTemplate implements TaskOperations {
 
 	/*default*/ static final String DEFINITIONS_RELATION = "tasks/definitions";
 
-	private static final String COMPOSE_RELATION = "tasks/composed-definitions";
+	private static final String COMPOSE_RELATION = "tasks/composed-definitions/compose";
 
 	private static final String DEFINITION_RELATION = "tasks/definitions/definition";
 
--- a/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/TaskTemplate.java
+++ b/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/TaskTemplate.java
@@ -42,7 +42,7 @@ public class TaskTemplate implements TaskOperations {
 
 	/*default*/ static final String DEFINITIONS_RELATION = "tasks/definitions";
 
-	private static final String COMPOSE_RELATION = "tasks/composed-definitions/compose";
+	private static final String COMPOSE_RELATION = "tasks/composed-definitions";
 
 	private static final String DEFINITION_RELATION = "tasks/definitions/definition";
 
--- a/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/TaskTemplate.java
+++ b/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/TaskTemplate.java
@@ -41,6 +41,8 @@ public class TaskTemplate implements TaskOperations {
 
 	/*default*/ static final String DEFINITIONS_RELATION = "tasks/definitions";
 
+	private static final String COMPOSE_RELATION = "tasks/definitions/compose";
+
 	private static final String DEFINITION_RELATION = "tasks/definitions/definition";
 
 	private static final String EXECUTIONS_RELATION = "tasks/executions";
@@ -52,6 +54,8 @@ public class TaskTemplate implements TaskOperations {
 	private final RestTemplate restTemplate;
 
 	private final Link definitionsLink;
+
+	private final Link composeLink;
 
 	private final Link definitionLink;
 
@@ -66,6 +70,7 @@ public class TaskTemplate implements TaskOperations {
 		Assert.notNull(resources.getLink(EXECUTIONS_RELATION), "Executions relation is required");
 		Assert.notNull(resources.getLink(DEFINITIONS_RELATION), "Definitions relation is required");
 		Assert.notNull(resources.getLink(DEFINITION_RELATION), "Definition relation is required");
+		Assert.notNull(resources.getLink(COMPOSE_RELATION), "Compose relation is required");
 		Assert.notNull(restTemplate, "RestTemplate must not be null");
 		Assert.notNull(resources.getLink(EXECUTIONS_RELATION), "Executions relation is required");
 		Assert.notNull(resources.getLink(EXECUTION_RELATION), "Execution relation is required");
@@ -78,6 +83,7 @@ public class TaskTemplate implements TaskOperations {
 		this.executionsLink = resources.getLink(EXECUTIONS_RELATION);
 		this.executionLink = resources.getLink(EXECUTION_RELATION);
 		this.executionByNameLink = resources.getLink(EXECUTION_RELATION_BY_NAME);
+		this.composeLink = resources.getLink(COMPOSE_RELATION);
 
 	}
 
@@ -95,6 +101,16 @@ public class TaskTemplate implements TaskOperations {
 		values.add("definition", definition);
 		TaskDefinitionResource task = restTemplate.postForObject(
 				definitionsLink.expand().getHref(), values, TaskDefinitionResource.class);
+		return task;
+	}
+
+	@Override
+	public TaskDefinitionResource compose(String name, String definition) {
+		MultiValueMap<String, Object> values = new LinkedMultiValueMap<String, Object>();
+		values.add("name", name);
+		values.add("definition", definition);
+		TaskDefinitionResource task = restTemplate.postForObject(
+				composeLink.expand().getHref(), values, TaskDefinitionResource.class);
 		return task;
 	}
 
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/ComposedTaskProperties.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/ComposedTaskProperties.java
@@ -27,6 +27,11 @@ import org.springframework.boot.context.properties.ConfigurationProperties;
 @ConfigurationProperties(prefix = ComposedTaskProperties.COMPOSED_TASK_PREFIX)
 public class ComposedTaskProperties {
 
+	public static final String COMPOSED_TASK_PREFIX = "spring.cloud.dataflow.composed.task";
+
+	private URI composedTaskRunnerUri;
+
+	private String taskName = "composed-task-runner";
 
 	public ComposedTaskProperties() {
 		try {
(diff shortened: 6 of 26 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
```diff
@@ -113,9 +131,50 @@ public TaskDefinitionResource save(@RequestParam("name") String name,
...
+						new ComposedTaskValidationProblem(graph,
+						composedTaskValidator.getInvalidDefinitions().get(taskName),
+								DSLMessage.COMPOSED_TASK_DEFINITION_NOT_PRESENT));
+			}
+			throw new ComposedTaskValidationException(composedTaskNode,
+					composedTaskValidationProblems);
+		}
+	}
+
+	private String getComposedDSL(String graph) {
```
ericbottard: This method name is a bit misleading. I did not know what to expect until I saw the actual implementation. Something that mentions "arguments" would be better IMO

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
```diff
@@ -113,9 +131,50 @@ public TaskDefinitionResource save(@RequestParam("name") String name,
...
+		return String.format("%s --graph=\"%s\"", composedTaskProperties.getTaskName(), graph);
```
ericbottard: `graph` needs to have its double quotes escaped.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
```diff
@@ -113,9 +131,50 @@ public TaskDefinitionResource save(@RequestParam("name") String name,
...
+	}
+
+
+	/**
 	 * Delete the task from the repository so that it can no longer be executed.
 	 *
-	 * @param name name of the task to be deleted
+	 * @param name of the task to be deleted
```
ericbottard: `name` should not have been removed

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/support/ComposedTaskValidator.java
```diff
@@ -0,0 +1,56 @@
...
+	private Map<String, Integer> invalidDefinitions = new HashMap<>();
+
+	public ComposedTaskValidator(TaskDefinitionRepository repository) {
+		Assert.notNull(repository, "repository must not be null");
+		this.repository = repository;
+	}
+
+	@Override
+	public void visit(TaskAppNode taskApp) {
+		TaskDefinition taskDefinition = repository.findOne(taskApp.getName());
```
ericbottard: It would be better to make this stateful and grabbing all available names at once at start, otherwise we get N queries to the DB

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/TaskControllerTests.java
```diff
@@ -151,6 +157,59 @@ public void testSave() throws Exception {
...
 
 	@Test
+	public void testCompose() throws Exception {
+		assertEquals(0, repository.count());
+		TaskDefinition testTask = new TaskDefinition("task", "task");
+		repository.save(testTask);
+		appRegistry.save("task", ApplicationType.task, new URI("http://fake.example.com/"),null);
+		mockMvc.perform(
+				post("/tasks/definitions/compose").param("name", "myTask").param("definition", "task")
+						.accept(MediaType.APPLICATION_JSON)).andDo(print())
```
ericbottard: Please don't `print()`, we have enough of those already (which should be removed IMHO)

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/TaskControllerTests.java
```diff
@@ -151,6 +157,59 @@ public void testSave() throws Exception {
 	}
 
 	@Test
+	public void testCompose() throws Exception {
+		assertEquals(0, repository.count());
+		TaskDefinition testTask = new TaskDefinition("task", "task");
```
ericbottard: Could make the arguments a bit less ambiguous by not being `task` in both cases

ericbottard: Moreover, I understand this is not testing the parser etc, but a non-trivial composition would be welcome and would make it more clear what the test does

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/TaskControllerTests.java
```diff
@@ -151,6 +157,59 @@ public void testSave() throws Exception {
...
+				.andExpect(status().isOk());
+
+		assertEquals(2, repository.count());
+
+		TaskDefinition myTask = repository.findOne("myTask");
+
+		assertEquals(2, myTask.getProperties().size());
+		assertEquals("myTask", myTask.getProperties().get("spring.cloud.task.name"));
+		assertEquals("task", myTask.getProperties().get("graph"));
+		assertEquals("composed-task-runner --graph=\"task\"", myTask.getDslText());
```
ericbottard: This would be a good place to test proper escaping, hence calling for a non-trivial composition again

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/TaskControllerTests.java
```diff
@@ -151,6 +157,59 @@ public void testSave() throws Exception {
...
+	@Test
+	public void testComposedInvalidParse() throws Exception {
+		assertEquals(0, repository.count());
+		appRegistry.save("task", ApplicationType.task, new URI("http://fake.example.com/"),null);
+		mockMvc.perform(
+				post("/tasks/definitions/compose").param("name", "myTask").
+						param("definition", "task%%$$")
+						.accept(MediaType.APPLICATION_JSON)).andDo(print())
+				.andExpect(status().is5xxServerError())
+				.andExpect(content().json(
```
ericbottard: This is too specific of an assertion IMO. Better focus on interesting bits rather than the actual whole

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
ericbottard: This is according to DSL parser rules btw, not shell rules, obviously

spring-cloud-dataflow-shell-core/src/test/java/org/springframework/cloud/dataflow/shell/command/TaskCommandTests.java
```diff
@@ -105,6 +105,17 @@ public void testCreateTask() throws InterruptedException {
...
+
+	@Test
+	public void testComposeTask() throws InterruptedException {
+		logger.info("Compose Task Test");
+		String taskName = generateUniqueName();
+		task().create("AAA778877", "timestamp");
+		task().create("BBB889988", "timestamp");
+
+		task().compose(taskName, "AAA778877 && BBB889988");
+	}
```
ericbottard: Cleanup?

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/TaskDefinition.java
```diff
@@ -38,24 +41,35 @@
...
+
+	public TaskDefinition(String name, String dsl, boolean isComposedTask) {
 		this.dslText = dsl;
-		AppNode taskNode = new TaskParser(name, dsl).parse();
-		setRegisteredAppName(taskNode.getName());
 		Map<String, String> properties = new HashMap<>();
-		if (taskNode.hasArguments()) {
-			for (ArgumentNode argumentNode : taskNode.getArguments()) {
-				properties.put(argumentNode.getName(), argumentNode.getValue());
+		if(isComposedTask) {
```
ericbottard: There is still something that eludes me in this whole story: why isn't the "normal" task definition a special case of a composed task definition. Maybe @aclement can elaborate as well?

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/DSLMessage.java
```diff
@@ -104,7 +104,9 @@
 	COMPOSED_TASK_ARROW_SHOULD_BE_PRECEDED_BY_CODE(ERROR, 157, "Transition arrow must be preceeded by the exit code that should drive the transition"), //
 	CT_VALIDATION_SECONDARY_SEQUENCES_MUST_BE_NAMED(ERROR, 158, "Secondary sequences must have labels or are unreachable"), //
 	CT_VALIDATION_DUPLICATE_LABEL(ERROR, 159, "This label has already been defined"), //
-	CT_VALIDATION_TRANSITION_TARGET_LABEL_UNDEFINED(ERROR, 160, "Transition specifies an undefined label")
+	CT_VALIDATION_TRANSITION_TARGET_LABEL_UNDEFINED(ERROR, 160, "Transition specifies an undefined label"), //
+	COMPOSED_TASK_DEFINITION_NOT_PRESENT(ERROR, 161, "Task in composed task definition does not exist"), //
+	COMPOSED_TASK_UNEXPECTED_DATA(ERROR, 162, "unexpected data in composed task definition ''{0}''") //
```
cppwfs: fixed

spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/TaskTemplate.java
```diff
@@ -41,6 +41,8 @@
 
 	/*default*/ static final String DEFINITIONS_RELATION = "tasks/definitions";
 
+	private static final String COMPOSE_RELATION = "tasks/definitions/compose";
```
cppwfs: fixed

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/TaskConfiguration.java
```diff
@@ -63,11 +69,18 @@
...
 @Configuration
 @ConditionalOnProperty(prefix = FeaturesProperties.FEATURES_PREFIX, name = FeaturesProperties.TASKS_ENABLED, matchIfMissing = true)
+@EnableConfigurationProperties({ComposedTaskProperties.class})
 public class TaskConfiguration {
 
 	@Autowired
 	DataSourceProperties dataSourceProperties;
 
+	@Autowired
+	private UriRegistry uriRegistry;
```
cppwfs: fixed

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/TaskConfiguration.java
```diff
@@ -181,4 +194,17 @@ public TaskDefinitionRepository taskDefinitionRepository(DataSource dataSource)
 			return new RdbmsTaskDefinitionRepository(dataSource);
 		}
 	}
+
+	@EventListener
```
cppwfs: Datastore access is not available at PostConstruct time. And is needed for registration.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/ComposedTaskProperties.java
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * @author Glenn Renfro
+ */
+@ConfigurationProperties(prefix = ComposedTaskProperties.COMPOSED_TASK_PREFIX)
+public class ComposedTaskProperties {
+
+	public static final String COMPOSED_TASK_PREFIX = "spring.cloud.dataflow.composed.task";
+
+	private String composedTaskRunnerUri = "maven://org.springframework.cloud.task.app:composedtaskrunner-task:1.0.0.BUILD-SNAPSHOT";
```
cppwfs: Make composedTaskRunnerURI a URI.  Done.
Shoot to add metadata-uri for app in RC.

ericbottard: Please note that I'm not even sure this makes sense, just adding a thought. Seems the only property to the runner is `--graph`, correct?

cppwfs: good question  actually here are the other options:
maxWaitTime
intervalTimeBetweenChecks
dataFlowUri

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/RootController.java
```diff
@@ -101,6 +101,7 @@ public RootResource info() {
 		if (featuresProperties.isTasksEnabled()) {
 			root.add(entityLinks.linkToCollectionResource(TaskDefinitionResource.class).withRel("tasks/definitions"));
 			root.add(unescapeTemplateVariables(entityLinks.linkToSingleResource(TaskDefinitionResource.class, "{name}").withRel("tasks/definitions/definition")));
+			root.add(unescapeTemplateVariables(entityLinks.linkToSingleResource(TaskDefinitionResource.class, "/compose").withRel("tasks/definitions/compose")));
```
cppwfs: done

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
```diff
@@ -113,9 +131,50 @@ public TaskDefinitionResource save(@RequestParam("name") String name,
...
+	 * @param name of the composed task
+	 * @param graph DSL definition for the composed task
+	 */
+	@RequestMapping(value = "/compose", method = RequestMethod.POST)
+	public TaskDefinitionResource saveComposedTask(@RequestParam("name") String name,
+			@RequestParam("definition") String graph) {
+		TaskDefinition taskDefinition = new TaskDefinition(name, graph, true);
+		TaskDefinition composedTaskDefinition = new TaskDefinition(
+				taskDefinition.getName(),getComposedDSL(taskDefinition.getDslText()));
+		validateTaskDefinitionExist(name, graph);
```
cppwfs: done

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
```diff
@@ -113,9 +131,50 @@ public TaskDefinitionResource save(@RequestParam("name") String name,
...
+
+	private void validateTaskDefinitionExist(String name, String graph) {
+		ComposedTaskValidator composedTaskValidator =
+				new ComposedTaskValidator(repository);
+		ComposedTaskNode composedTaskNode = new ComposedTaskParser().parse(name, graph);
+		composedTaskNode.accept(composedTaskValidator);
+		if(composedTaskValidator.getInvalidDefinitions().size() > 0) {
+			List<ComposedTaskValidationProblem> composedTaskValidationProblems =
+					new ArrayList<>();
+			for(String taskName : composedTaskValidator.getInvalidDefinitions().keySet()) {
```
cppwfs: Your crystal ball is incredibly accurate.  Does it pick winning lottery numbers ;-)

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
```diff
@@ -113,9 +131,50 @@ public TaskDefinitionResource save(@RequestParam("name") String name,
...
+				composedTaskValidationProblems.add(
+						new ComposedTaskValidationProblem(graph,
+						composedTaskValidator.getInvalidDefinitions().get(taskName),
+								DSLMessage.COMPOSED_TASK_DEFINITION_NOT_PRESENT));
```
cppwfs: done

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
cppwfs: Noob at work here ;-)  Not sure I understand, the double quotes are already escaped.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
cppwfs: Done

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
cppwfs: done

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
ericbottard: No, they don't seem to be. What I'm saying is that if `graph` is a string containing `aaa "failed"->bbb`, then quotes need to be escaped and become `aaa ""failed""->bbb` IIRC

cppwfs:
I see.  Makes sense.

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/ComposedTaskTokenizer.java
```diff
@@ -104,7 +104,7 @@ else if (isTwoCharToken(TokenKind.ARROW)) {
 					case '\\':
 						raiseException(DSLMessage.UNEXPECTED_ESCAPE_CHAR);
 					default:
-						raiseException(DSLMessage.UNEXPECTED_DATA,Character.valueOf(ch).toString());
+						raiseException(DSLMessage.COMPOSED_TASK_UNEXPECTED_DATA,Character.valueOf(ch).toString());
```
ericbottard: Why do we need a special version of this?

cppwfs: The error refers to it as a "stream definition".

ericbottard: Ah, my bad. Thanks for the clarification

cppwfs: no worries.  thanks for the review!

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/support/ComposedTaskValidator.java
cppwfs: done

cppwfs: done

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/TaskControllerTests.java
cppwfs: done

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/TaskControllerTests.java
cppwfs: done

spring-cloud-dataflow-shell-core/src/test/java/org/springframework/cloud/dataflow/shell/command/TaskCommandTests.java
cppwfs: done

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
cppwfs: Not sure this a valid case in that the parser fails on it.  But asked the question anyway ;-)  Here: https://github.com/spring-cloud/spring-cloud-dataflow/issues/1221

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/TaskControllerTests.java
cppwfs: added 2 transition tests.  One for double quote and the other single quote.

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/TaskDefinition.java
aclement: This is never something we particularly talked about really but I can see your point of view. From my point of view, the DSL is complicated enough, if you supported/allowed arguments inside the DSL too everywhere a task could be specified, you are perhaps steering folks further down writing incomprehensible strings (especially as tasks seem to have far more options set). Spring XD didn't do this, but not sure how much that impacts choices made in SCDF.

cppwfs: I think the primary issue is that a task definition is comprised of an app with supporting properties.  While a composed task is comprised of task definitions.

ericbottard: That's all valid justification to me. This was really a naive question, the first gut feeling being that this looked like an extension of what we already had. But as Glenn pointed at, it's really not.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/ComposedTaskProperties.java
```diff
@@ -0,0 +1,71 @@
...
+	private URI composedTaskRunnerUri;
+
+
+	private String taskName = "composed-task-runner";
+
+	public URI getComposedTaskRunnerUri() {
+		return composedTaskRunnerUri;
+	}
+
+	public void setComposedTaskRunnerUri(String composedTaskRunnerUri) {
```
ericbottard: This setter should accept a URI, that was the whole point of my previous comment: let boot binding do the conversion. If this is not possible, then I'd argue this brings little value (at the very least, should throw IllegalArgumentException instead)

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/ComposedTaskProperties.java
```diff
@@ -0,0 +1,71 @@
...
+		try {
+			composedTaskRunnerUri = new URI("maven://org.springframework.cloud.task.app:composedtaskrunner-task:1.0.0.BUILD-SNAPSHOT");
+		}
+		catch (URISyntaxException e) {
+			throw new IllegalStateException(
+					"Invalid URI specified for composedTaskRunnerUri");
+		}
+	}
+
+	public static final String COMPOSED_TASK_PREFIX = "spring.cloud.dataflow.composed.task";
```
ericbottard: should move up

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
```diff
@@ -113,9 +131,50 @@ public TaskDefinitionResource save(@RequestParam("name") String name,
 	}
 
 	/**
+	 * Register a composed task definition for future execution.
+	 *
+	 * @param name of the composed task
+	 * @param graph DSL definition for the composed task
+	 */
+	@RequestMapping(value = "/compose", method = RequestMethod.POST)
```
ericbottard: This should actually live under a totally different scheme, sorry for not noticing earlier.
Something like `POST /tasks/composed-definitions`
The current state of this PR clashes with a task definition that would be named `compose`.

ericbottard: So this implies a new Controller class, and may even call for a new `ComposedTaskDefinitionResource`

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
ericbottard: not done?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/ComposedTaskProperties.java
ericbottard: Where are the sources for the composed task runner again?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/support/ComposedTaskValidator.java
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Identifies all task names that do not have a task definition in the
+ * task repository.
+ *
+ * @author Glenn Renfro
+ */
+public class ComposedTaskValidator extends ComposedTaskVisitor {
+
+	private TaskDefinitionRepository repository;
```
ericbottard: Not needed

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/support/ComposedTaskValidator.java
```diff
@@ -0,0 +1,67 @@
...
+	Set<String> availableDefinitionNames;
+
+	public ComposedTaskValidator(TaskDefinitionRepository repository) {
+		Assert.notNull(repository, "repository must not be null");
+		this.repository = repository;
+		availableDefinitionNames = new HashSet();
+
+		Stream<TaskDefinition> taskDefinitionStream =
+				StreamSupport.stream(repository.findAll().spliterator(), false);
+		taskDefinitionStream.forEach(taskDefinition ->
```
ericbottard: Better done with `map` then `collect`

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
```diff
@@ -113,9 +131,50 @@ public TaskDefinitionResource save(@RequestParam("name") String name,
...
+							composedTaskValidator.getInvalidDefinitions().get(s),
+							DSLMessage.CT_ELEMENT_IN_COMPOSED_DEFINITION_DOES_NOT_EXIST))
+							);
+			throw new ComposedTaskValidationException(composedTaskNode,
+					composedTaskValidationProblems);
+		}
+	}
+
+	private String getComposedArgs(String graph) {
+		return String.format("%s --graph=\"%s\"", composedTaskProperties.getTaskName(), graph);
```
ericbottard: Still need to escape the double quotes

spring-cloud-dataflow-shell-core/src/test/java/org/springframework/cloud/dataflow/shell/command/TaskCommandTemplate.java
```diff
@@ -64,6 +64,15 @@ public TaskCommandTemplate(JLineShellComponent shell) {
...
 		doCreate(taskName, taskDefinition, true, values);
 	}
+	/**
+	 * Compose a task.
+	 *
+	 * @param taskName the name of the task
+	 * @param graph the compoed task definition DSL
+	 */
+	public void compose(String taskName, String graph) {
+		doCompose(taskName, graph);
```
ericbottard: What's the point of having a `doXXX` method here?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/ComposedTaskProperties.java
cppwfs: done

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/TaskDefinition.java
```diff
@@ -38,24 +41,35 @@
 	private final String dslText;
 
 	public TaskDefinition(String name, String dsl) {
+		this(name, dsl, false);
+	}
+
+	TaskDefinition(String registeredAppName, String label, Map<String, String> properties) {
+		super(registeredAppName, label, properties);
+		this.dslText = "";
+		properties.put("spring.cloud.task.name", registeredAppName);
```
cppwfs: done.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/ComposedTaskProperties.java
cppwfs: done

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
cppwfs: Still trying to figure this out...    For example a user can embed quotes already.  i.e.
`task compose delme --graph "AAA '\"FAILED\"' -> CCC && BBB"` -or-
`task compose delme --graph "AAA ''FAILED'' -> CCC && BBB"`
this pattern follows what we already do here:
`stream create adf --definition "http --port=9700 | filter --expression=payload.matches('\"hello\"') | file"`

ericbottard: That's what I'm saying. Since this `String.format()` uses `--graph="<insert-me-here>"`, everything in `<insert-me-here>` should have its double quotes escaped

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
```diff
@@ -113,9 +131,50 @@ public TaskDefinitionResource save(@RequestParam("name") String name,
...
+		repository.save(composedTaskDefinition);
+		return taskAssembler.toResource(taskDefinition);
+	}
+
+	private void validateComposedTaskElementsExist(String name, String graph) {
+		ComposedTaskValidator composedTaskValidator =
+				new ComposedTaskValidator(repository);
+		ComposedTaskNode composedTaskNode = new ComposedTaskParser().parse(name, graph);
+		composedTaskNode.accept(composedTaskValidator);
+		if(composedTaskValidator.getInvalidDefinitions().size() > 0) {
```
ericbottard: Try this:
```
		List<ComposedTaskValidationProblem> problems = composedTaskValidator.getInvalidDefinitions().values().stream()
			.map(pos -> new ComposedTaskValidationProblem(graph, pos, DSLMessage.CT_ELEMENT_IN_COMPOSED_DEFINITION_DOES_NOT_EXIST))
		if (!problems.isEmpty()) {
			throw new ComposedTaskValidationException(composedTaskNode, problems);
		}
```

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/support/ComposedTaskValidator.java
cppwfs: done

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/support/ComposedTaskValidator.java
cppwfs: done

spring-cloud-dataflow-shell-core/src/test/java/org/springframework/cloud/dataflow/shell/command/TaskCommandTemplate.java
cppwfs: removed.  done

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
cppwfs: done

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
cppwfs: done

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/ComposedTaskProperties.java
cppwfs: https://github.com/spring-cloud-task-app-starters/composed-task-runner
Trying to get a PR out on it today / tomorrow.

ericbottard: OK, so it should get its companion metadata artifact generated for free already.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
cppwfs: Understood about the controller.  But for the resource I'm not so sure.  It is still just a task that is created.

ericbottard: It is.
But the fact that it is composed could be captured in there.
Also, but that is a purely technical excuse, having a dedicated resource class (that may be a subclass of the former that actually adds nothing) helps with `RootController`'s `entityLinks.linkToCollectionResource()`

cppwfs: gotcha

cppwfs: done

ericbottard: Should be `ComposedTaskDEFINITIONController` for consistency

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/ComposedTaskController.java
```diff
@@ -0,0 +1,167 @@
...
+		repository.save(composedTaskDefinition);
+		return taskAssembler.toResource(taskDefinition);
+	}
+
+	private void validateComposedTaskElementsExist(String name, String graph) {
+		ComposedTaskValidator composedTaskValidator =
+				new ComposedTaskValidator(repository);
+		ComposedTaskNode composedTaskNode = new ComposedTaskParser().parse(name, graph);
+		composedTaskNode.accept(composedTaskValidator);
+		if(composedTaskValidator.getInvalidDefinitions().size() > 0) {
```
ericbottard: One of those two ifs (I'd argue this one) is not needed and impairs readability

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/ComposedTaskController.java
```diff
@@ -0,0 +1,167 @@
...
+							.collect(Collectors.toList());
+			if (!composedTaskValidationProblems.isEmpty()) {
+				throw new ComposedTaskValidationException(
+						composedTaskNode, composedTaskValidationProblems);
+			}
+		}
+	}
+
+	private String getComposedArgs(String graph) {
+		return String.format("%s --graph=\"%s\"", composedTaskProperties.getTaskName(), graph.replaceAll("\"", "\\\\\""));
```
ericbottard: You don't need a regex here, `replace` is sufficient

ericbottard: Extracting a method named `escapeQuotes` or something would also clarify the intent

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/ComposedTaskController.java
```diff
@@ -0,0 +1,167 @@
...
+		this.composedTaskProperties = composedTaskProperties;
+	}
+
+	/**
+	 * Register a composed task definition for future execution.
+	 *
+	 * @param name name of the composed task
+	 * @param graph DSL definition for the composed task
+	 */
+	@RequestMapping(value = "/compose", method = RequestMethod.POST)
```
ericbottard: mapping should be on `""`, not `/compose`

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/ComposedTaskController.java
cppwfs: done

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/ComposedTaskController.java
cppwfs: done & done

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/ComposedTaskController.java
cppwfs: done

spring-cloud-dataflow-docs/src/test/java/org/springframework/cloud/dataflow/server/rest/documentation/ApiDocumentation.java
```diff
@@ -93,7 +93,7 @@ public void index() throws Exception {
 					linkWithRel("runtime/apps/instances").description("Provides the status for app instances"),
 					linkWithRel("tasks/definitions").description("Provides the task definition resource"),
 					linkWithRel("tasks/definitions/definition").description("Provides details for a specific task definition"),
-					linkWithRel("tasks/composed-definitions/compose").description("Requests the creation of a composed task definition"),
```
ericbottard: This is actually OK, it's the URL that should not have /compose

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/RootController.java
```diff
@@ -102,7 +102,7 @@ public RootResource info() {
 		if (featuresProperties.isTasksEnabled()) {
 			root.add(entityLinks.linkToCollectionResource(TaskDefinitionResource.class).withRel("tasks/definitions"));
 			root.add(unescapeTemplateVariables(entityLinks.linkToSingleResource(TaskDefinitionResource.class, "{name}").withRel("tasks/definitions/definition")));
-			root.add(entityLinks.linkToSingleResource(ComposedTaskResource.class, "/compose").withRel("tasks/composed-definitions/compose"));
+			root.add(entityLinks.linkToSingleResource(ComposedTaskResource.class, "").withRel("tasks/composed-definitions"));
```
ericbottard: should be `linkToCollectionResource`

spring-cloud-dataflow-docs/src/test/java/org/springframework/cloud/dataflow/server/rest/documentation/ApiDocumentation.java
cppwfs: Actually that will fail the test if we add the /compose.

ericbottard: you'll obviously need to change it at several places. But you need to distinguish between **the name of the relation** and the **URL**

cppwfs: done

ericbottard:
Rebased, squashed and merged as b91b5383

Issue #1214, opened 2017-03-06
Consider pre-fetching the metadata artifacts on register


ericbottard:
The metadata artifacts typically weigh ~50Kb for an OOTB app (compared to ~25Mb for the actual uberjar), but downloading them for the very first time still incurs a small delay which is noticeable when doing tab completion.

We may want to prefetch them on register, if 
1) metadata-uri is set
2) metadata-uri is different from main app uri

markfisher:
This violates the assumption that the artifact need not be available at the time of registration. Could it be optional, based on a boolean flag provided with the `app register` command? (we could support an option for eagerly downloading the uberjar as well). If we do proceed with this as a default behavior, we should at least treat failures as non-fatal (log a warning perhaps).

ericbottard:
Sorry, I should have made this more clear: of course this should be considered optional and any failure should be non fatal.

Pre-fetching the actual jars is
1) problematic as it could DL gigs of data (think CF)
2) not needed until we deploy the app now. So maybe we could do it at stream/task creation time (and obviously this only applies to runtimes that actually upload/need the main app, ie not in the pull model). Long story short: it's a bit more involved than what I consider this issue to be: a quick win ;)

markfisher:
Agreed that downloading the full JAR should be a separate issue (IF we pursue it at all), and perhaps it could be a background operation anyways?

ericbottard:
Yes, in my mind, all of this happens in the background

markfisher:
well, in your mind is one thing... I was talking about the server ;)

PR #1222, opened 2017-03-09
Preftech metadata on app registration

ericbottard:
Fixes https://github.com/spring-cloud/spring-cloud-dataflow/issues/1214

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppRegistryController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppRegistryController.java
@@ -217,7 +217,6 @@ public class AppRegistryController implements ResourceLoaderAware {
 				.filter(r -> r.getMetadataUri() != null)
 				.parallel()
 				.forEach(r -> {
-					System.out.println("Getting " + r.getMetadataUri());
 					logger.info("Eagerly fetching {}", r.getMetadataUri());
 					try {
 						r.getMetadataResource();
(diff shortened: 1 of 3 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppRegistryController.java
```diff
@@ -191,9 +203,32 @@ else if (!CollectionUtils.isEmpty(apps)) {
...
+	 * Trigger early resolution of the metadata resource of registrations that have an explicit metadata artifact.
+	 * This assumes usage of {@link org.springframework.cloud.deployer.resource.support.DelegatingResourceLoader}.
+	 */
+	private void prefetchMetadata(List<AppRegistration> appRegistrations) {
+		forkJoinPool.execute(() -> {
+			appRegistrations.stream()
+				.filter(r -> r.getMetadataUri() != null)
+				.parallel()
+				.forEach(r -> {
+					System.out.println("Getting " + r.getMetadataUri());
```
ilayaperumalg: This needs to be removed.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppRegistryController.java
```diff
@@ -191,9 +203,32 @@ else if (!CollectionUtils.isEmpty(apps)) {
...
 	}
 
+	/**
+	 * Trigger early resolution of the metadata resource of registrations that have an explicit metadata artifact.
+	 * This assumes usage of {@link org.springframework.cloud.deployer.resource.support.DelegatingResourceLoader}.
+	 */
+	private void prefetchMetadata(List<AppRegistration> appRegistrations) {
+		forkJoinPool.execute(() -> {
+			appRegistrations.stream()
+				.filter(r -> r.getMetadataUri() != null)
```
ilayaperumalg: Does this mean during the prefetch, we don't download the artifact where the app itself contains the metadata resource? ie) only prefetch if there are explicit metadata resource available for the app registration?

ericbottard: That's exactly that

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed, rebased and merged as 058de3a

Issue #1216, opened 2017-03-06
Expand REST API to Return Version Info


ghillert:
This issue is related to:

* https://github.com/spring-cloud/spring-cloud-deployer/issues/183
* https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/160

Also see:

* https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/91
* https://github.com/spring-cloud/spring-cloud-dataflow/wiki/Reporting-Issues

So far the only artifact version we ever returned to the user was the SCDF version as part of the `/management/info` rest endpoint. This is impractical, though - would need to duplicate stuff for the various SCDF Implementations. Whatever we do, we should do only once (if possible) in _SCDF Server Core_.

PR #1224, opened 2017-03-09
gh-1216 Expand REST API to Return Version Info

ghillert:
This PR unifies multiple pre-existing rest-endpoints into one:

* `/security/info`
* `/features`

* Add API documentations
* Add tests
  - Make `LocalDataflowResource` more dynamic so that features (Task & Stream) can be toggled
  - Add test suite that runs security REST tests with disabled task feature

Resolves https://github.com/spring-cloud/spring-cloud-dataflow/issues/1216

### DIFF
--- a/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/resource/about/RuntimeEnvironment.java
+++ b/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/resource/about/RuntimeEnvironment.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.dataflow.rest.resource.about;
+
+/**
+ * Provides information about the runtime environment.
+ *
+ * @author Gunnar Hillert
+ */
+public class RuntimeEnvironment {
+
+	/**
+	 * Default constructor for serialization frameworks.
+	 */
+	public RuntimeEnvironment() {
+	}
+
+	/**
+	 * The {@link RuntimeEnvironmentDetails} of the deployer.
+	 */
+	private RuntimeEnvironmentDetails deployer;
+
+	/**
+	 * The {@link RuntimeEnvironmentDetails} of the task launcher.
+	 */
+	private RuntimeEnvironmentDetails taskLauncher;
+
+	/**
+	 * @return Null, if the stream feature is disabled
+	 */
+	public RuntimeEnvironmentDetails getDeployer() {
+		return deployer;
+	}
+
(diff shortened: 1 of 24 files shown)


ghillert:
@sabbyanandan @trisberg Updated PR. I hope it is ready to go now.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AboutController.java
```diff
@@ -0,0 +1,198 @@
...
+		this.featuresProperties = featuresProperties;
+		this.versionInfoProperties = versionInfoProperties;
+		this.securityProperties = securityProperties;
+		this.authorizationConfig = authorizationConfig;
+	}
+
+	/**
+	 * Return features that are enabled/disabled on the dataflow server.
+	 */
+	@ResponseBody
```
ericbottard: I believe this can be removed given this is a `RestController`

ghillert:
@ericbottard Addressed comment

ericbottard:
This PR has a fairly old parent commit, could you rebase it @ghillert ?

spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/resource/about/RuntimeEnvironment.java
```diff
@@ -0,0 +1,64 @@
...
+	/**
+	 * Default constructor for serialization frameworks.
+	 */
+	public RuntimeEnvironment() {
+	}
+
+	/**
+	 * The {@link RuntimeEnvironmentDetails} of the deployer.
+	 */
+	private RuntimeEnvironmentDetails deployer;
```
ericbottard: Should read `appDeployer` rather than just `deployer`, to be consistent with `taskLauncher`

ericbottard:
LGTM apart from the comment about `deployer` vs `appDeployer`. Should be able to merge once rebased.

I wonder if we should consider moving the `api.revision` field to `/about` though. Thoughts?

ericbottard:
That PR should increment `api.revision` btw

ghillert:
@ericbottard Made the changes.

ericbottard:
What do you think about moving the `api.revision` inside the `/about` endpoint?

ghillert:
@ericbottard Good idea - added it with the latest commit.

trisberg:
Wonder why we need these:

```
"commitId": "55e7286c5117b309c53903bf22b6987c6824d909",
"commitTime": "2017-03-15T16:35:16.000Z",
"branch": "ghillert-gh-1216"
```

Not clear that they refer to the commit for the SCDF-core, and we don't have them for the deployer or server-impl

ghillert:
@trisberg It SHOULD be the implementation Git Hash. I still need to polish that as at the moment this is populated for Core. We need to ensure that the git properties (generated by Maven) is only generated once (for the implementation).

ericbottard:
Squashed and merged as cffd1eb5

But I reverted commit ad9229b: My initial intention was to *move* the `api.revision` number to `/about`, not to _add_ it. Then I realised this is more complicated than it seems: it is needed in early negociation with clients, and `/about` _may_ (should?) be secured when security is on.

We can always revisit

ericbottard:
Created https://github.com/spring-cloud/spring-cloud-dataflow/issues/1232 to track `/about` and security

ghillert:
@ericbottard You should have left it in there. My thinking was to use the `/about` REST endpoint for early negotiation for the UI (e.g. the UI needs to know the enabled features, needs to know about general security settings).

This, raises some interesting question: is the broadcasting of of environmental information (incl. dependency version) worthy of being secured? @sabbyanandan

ericbottard:
I hear you. As I said, we can always revisit. The current situation is the best "stable" situation IMO: no information is duplicated. We can take it from here

Issue #1228, opened 2017-03-15
Don't convert props to csv on the shell side when deploying


ericbottard:
See https://github.com/spring-cloud/spring-cloud-dataflow/pull/1184#issuecomment-283319132

The current behavior is doing conversion back and forth, using the brittle `DeploymentPropertiesUtils#parse` method more than necessary.

When dealing with a `.properties/.yml` file, that file should be sent as is to the server, leveraging escape mechanisms etc for that file format to allow the user to pass complicated values (_i.e._ ones that contain _e.g._ a comma)

PR #1229, opened 2017-03-15
Avoid back and forth conversion of deployment properties

ericbottard:
Fixes https://github.com/spring-cloud/spring-cloud-dataflow/issues/1228

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,7 +24,7 @@
 	</scm>
 	<properties>
 		<java.version>1.8</java.version>
-		<spring-cloud-dataflow-ui.version>1.2.0.M2</spring-cloud-dataflow-ui.version>
+		<spring-cloud-dataflow-ui.version>1.2.0.BUILD-SNAPSHOT</spring-cloud-dataflow-ui.version>
 
 		<spring-cloud-deployer.version>1.2.0.M3</spring-cloud-deployer.version>
 		<spring-cloud-deployer-local.version>1.2.0.M3</spring-cloud-deployer-local.version>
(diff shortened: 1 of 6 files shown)


jvalkeal:
Looks good other that we need to fix UI as well. UI simply sends
```
Request URL:http://localhost:9393/streams/deployments/foostream?properties=deployer.*.count%3D2
```
and that not gets lost. Should we wait merging this PR until we have fix for UI?

jvalkeal:
I was writing new issue for `spring-cloud-dataflow-ui` and realised that it only have one field for passing in deployment properties and cannot send those using a map as UI cannot do parsing.

ericbottard:
Would be good to find some sort of solution before RC though, IMO.
Maybe we can keep a deprecated, content-negotiated endpoint that accepts a raw unparsed string for the UI.

jvalkeal:
Yeah, I don't see any other option if UI allows to define comma-delimited property string. Thought it might be good if in UI we'd force to use separate text boxes for each property.

sabbyanandan:
UI updates will be supported via: https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/180.

ghillert:
Also linking to https://github.com/spring-cloud/spring-cloud-dataflow-ui/pull/181

See my comment regarding YAML support. I have the feeling we should also support raw YAML to be passed in for deployment properties.

Furthermore, a super-minor note: when passing key/value pairs via REST controllers, should we standardize on what method parameter type to use? E.g. in [sdfsdf](https://github.com/spring-cloud/spring-cloud-dataflow-ui/pull/181) we use `java.util.Properties` when passing key-value pairs.

ghillert:
@ericbottard Besides my comment regarding `Map<String, String> properties` versus `Properties properties`, I think this should be ok. The change for the UI is literally just 1 line and will be done once this is merged.

jvalkeal:
merged per https://github.com/spring-cloud/spring-cloud-dataflow/commit/0d30bf26da3559ea14d3567ef70fb2997b3a6ed5

PR #1231, opened 2017-03-15
Remove auto registration of composed-task-runner

### LINKED ISSUE #1230: composed-task-runner registered by default
When I deploy a server implementation that build on the most recent snapshots of scdf-core I get a new `composed-task-runner` app registered:

```
composed-task-runner	task	maven://org.springframework.cloud.task.app:composedtaskrunner-task:1.0.0.BUILD-SNAPSHOT
```

Not sure why this is, but it certainly will not be useful running with the Kubernetes or Mesos servers.

cppwfs:
resolves #1230 

Let's check with @mminella before merging.

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/ComposedTaskProperties.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/ComposedTaskProperties.java
@@ -16,9 +16,6 @@
 
 package org.springframework.cloud.dataflow.server.config.features;
 
-import java.net.URI;
-import java.net.URISyntaxException;
-
 import org.springframework.boot.context.properties.ConfigurationProperties;
 
 /**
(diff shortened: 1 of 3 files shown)


ericbottard:
I thought we discussed this already and we concluded that this is a core feature of Data Flow itself.
Specific runtimes can override the uri of the runner to get an appropriate (eg docker) version

ericbottard:
cc @markfisher ^^

markfisher:
It should not be considered a "core feature" if it doesn't work for all server implementations.

cppwfs:
But it can be configured via the dataflow-server.yml/properties/etc to
select whatever composed-task-runner is required.  The default was just
maven.

markfisher:
I don't know if a "default" really makes sense in this case. It feels like each dataflow server should handle this explicitly.

trisberg:
Rebased and merged as commit 2193d11ec0e3c4a7dc43eb35a9ec7886a4e54590

PR #1236, opened 2017-03-17
Overhaul deployment properties parsing

### LINKED ISSUE #1235: Providing long deployment property values hangs
Trying to set a property and if the value is too long it seems to hang - saw this from both shell and UI

```
dataflow:>stream deploy --name ticktock --properties "app.log.foo=FOOOOOOOOOOOBaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaar"
```

doesn't complete in a reasonable time.

jvalkeal:
- Move away from regex in favor of simple
  comma splitting.
- As comma split will produce wrong key/value
  pairs if values itself are comma delimited list,
  go through split and combine to previous candidate
  if parsed key/value doesn't look like it is
  key/value.
- Fixes #1235

### DIFF
--- a/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
+++ b/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
@@ -47,11 +47,6 @@ public final class DeploymentPropertiesUtils {
 	}
 
 	/**
-	 * Pattern used for parsing a String of comma-delimited key=value pairs.
-	 */
-	private static final Pattern DEPLOYMENT_PROPERTIES_PATTERN = Pattern.compile("(?:,{0,1}\\s*)((\\.*(\\w|\\*|-)+)+)(\\s*=\\s*)");
-
-	/**
 	 * Pattern used for parsing a String of command-line arguments.
 	 */
 	private static final Pattern DEPLOYMENT_PARAMS_PATTERN = Pattern.compile("(\\s(?=([^\\\"']*[\\\"'][^\\\"']*[\\\"'])*[^\\\"']*$))");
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
```diff
@@ -61,40 +56,40 @@ private DeploymentPropertiesUtils() {
...
-	 *    we capture those as many time as needed. We get groups 1, 2 and 3 but we only use 1.
-	 * 3. (\s*=\s*) here we find boundary of '=' which may have empty characters around it.
-	 *    This becomes capturing group 4.
-	 * 4. Now we know where is key with group 1 and start of value with group 4.
-	 * 5. Now we can start to do matching. Need to match one ahead so that we
-	 *    know where next key possible is as that is then end of previous value.
-	 * 6. Lastly we have one pair left which is a same case if we only have
-	 *    one pair to parse.
+	 * Logic of parsing key/value pairs from a string is based on few rules and assumptions
+	 * 1. keys will not have commas or equals.
```
markfisher: given the 1st assumption, would it be simpler to split on `=` and then split between value and the next key by splitting on the last comma (since no comma can exist in a key)?

jvalkeal: I don't think so as value may have `=` and we need to know boundaries of key/value pairs. Also splitting by `=` would not work as it'd be more difficult to parse ill-formed string where one key/value is without its `=<value>`.

markfisher: gotcha... I wasn't thinking of values containing `=` but that does make sense

trisberg:
LGTM, big improvement in processing time

Issue #1238, opened 2017-03-17
task destroy command should call TaskLauncher#destroy()


ericbottard:
So, we added `TaskLauncher#destroy()`, but SCDF never actually calls it!

PR #1246, opened 2017-03-20
task destroy calls TaskLauncher#destroy()

ericbottard:
Fixes https://github.com/spring-cloud/spring-cloud-dataflow/issues/1238

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
@@ -16,21 +16,11 @@
 
 package org.springframework.cloud.dataflow.server.controller;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.stream.Collectors;
-
 import org.springframework.cloud.dataflow.core.ApplicationType;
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
```diff
@@ -144,6 +134,7 @@ public void destroyTask(@PathVariable("name") String name) {
 			throw new NoSuchTaskDefinitionException(name);
 		}
 		repository.delete(name);
+		taskLauncher.destroy(name);
```
ilayaperumalg: Should the `definition` repository `delete` happen after the completion of `destroy`?

ericbottard: Would be slightly better in case something goes wrong, that's right

ilayaperumalg: ok, will make the changes and merge.

ilayaperumalg:
Rebased and merged as 92d00673b7066c873c229efd22bbc95fb894d314 with the changes to perform task definition repo deletion after task destroy e74ca03bf31ad94d594c538c6d03f78270fb9d30

PR #1252, opened 2017-03-22
Add paging to RuntimeAppController

### LINKED ISSUE #1251: RuntimeAppController doesn't use paging
Listing runtime apps doesn't use `Pageable` and thus all apps are always returned. This is a root cause for spring-cloud/spring-cloud-dataflow-ui/issues/176.

jvalkeal:
- list method now understand Pageable.
- modified stream handling to only return current page
- modified RuntimeTemplate to request 10000 items which is
  a same value what StreamTemplate is using. For some
  reason Pageable defaults to size 20 so shell need to request
  more to get past 20 items is shell doesn't use paging.
- fixes #1251
- this also is a root cause for spring-cloud/spring-cloud-dataflow-ui#176

### DIFF
--- a/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/RuntimeTemplate.java
+++ b/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/RuntimeTemplate.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2016 the original author or authors.
+ * Copyright 2015-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/RuntimeTemplate.java
```diff
@@ -50,7 +50,9 @@
 
 	@Override
 	public PagedResources<AppStatusResource> status() {
-		return restTemplate.getForObject(appStatusesUriTemplate.expand().getHref(), AppStatusResource.Page.class);
+		String uriTemplate = appStatusesUriTemplate.expand().getHref();
```
ericbottard: Not really a uri "template" anymore then...

can change on merge

jvalkeal: Yeah, I just followed how it's done in `StreamTemplate`.

ericbottard: [Whoever](https://github.com/spring-cloud/spring-cloud-dataflow/blame/master/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/StreamTemplate.java#L72) did that should be hung :)

Issue #1255, opened 2017-03-22
spring.cloud.task.executionid should be passed as a cmdLine arg


ericbottard:
Currently passed as an app property, those properties often make part of the "app identity" in the target runtime (being bound as ENV VARS on CF for example). As such, two invocations of the same task definition result in the same (first) executionid being passed.

Should be moved as a cmdLine arg.

PR #1256, opened 2017-03-22
Pass spring.cloud.task.executionid as cmdLine arg

ericbottard:
Fixes https://github.com/spring-cloud/spring-cloud-dataflow/issues/1255

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
@@ -16,6 +16,7 @@
 
 package org.springframework.cloud.dataflow.server.service.impl;
 
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -145,13 +146,15 @@ public class DefaultTaskService implements TaskService {
 		Resource metadataResource = appRegistration.getMetadataResource();
 
 		TaskExecution taskExecution = taskExecutionRepository.createTaskExecution();
-		taskDefinition = this.updateTaskProperties(taskDefinition, taskExecution);
+		taskDefinition = this.updateTaskProperties(taskDefinition);
 
 		Map<String, String> appDeploymentProperties = extractAppProperties(taskDefinition.getRegisteredAppName(), taskDeploymentProperties);
 		Map<String, String> deployerDeploymentProperties = DeploymentPropertiesUtils.extractAndQualifyDeployerProperties(taskDeploymentProperties, taskDefinition.getRegisteredAppName());
 		AppDefinition revisedDefinition = mergeAndExpandAppProperties(taskDefinition, metadataResource, appDeploymentProperties);
 
-		AppDeploymentRequest request = new AppDeploymentRequest(revisedDefinition, appResource, deployerDeploymentProperties, commandLineArgs);
+		List<String> updatedCmdLineArgs = new ArrayList<>(commandLineArgs);
+		updatedCmdLineArgs.add("--spring.cloud.task.executionid=" + taskExecution.getExecutionId());
+		AppDeploymentRequest request = new AppDeploymentRequest(revisedDefinition, appResource, deployerDeploymentProperties, updatedCmdLineArgs);
 
 		String id = this.taskLauncher.launch(request);
 		if (!StringUtils.hasText(id)) {
@@ -192,7 +195,7 @@ public class DefaultTaskService implements TaskService {
 		return new AppDefinition(original.getName(), merged);
 	}
 
-	private TaskDefinition updateTaskProperties(TaskDefinition taskDefinition, TaskExecution taskExecution) {
+	private TaskDefinition updateTaskProperties(TaskDefinition taskDefinition) {
 		TaskDefinitionBuilder builder = TaskDefinitionBuilder.from(taskDefinition);
 		builder.setProperty("spring.datasource.url", dataSourceProperties.getUrl());
 		builder.setProperty("spring.datasource.username",
@@ -205,7 +208,6 @@ public class DefaultTaskService implements TaskService {
 		builder.setProperty("spring.datasource.driverClassName",
 			dataSourceProperties.getDriverClassName());
 
-		builder.setProperty("spring.cloud.task.executionid", String.valueOf(taskExecution.getExecutionId()));
 		return builder.build();
 	}
 
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
@@ -166,8 +166,8 @@ public class DefaultTaskService implements TaskService {
 
 	private List<String> updateCommandLineArgs(List<String> commandLineArgs, TaskExecution taskExecution) {
 		return Stream.concat(
-			commandLineArgs.stream().filter(a -> !a.startsWith("--spring.cloud.task.executionid=")),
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
```diff
@@ -145,13 +146,15 @@ public long executeTask(String taskName, Map<String, String> taskDeploymentPrope
...
 		TaskExecution taskExecution = taskExecutionRepository.createTaskExecution();
-		taskDefinition = this.updateTaskProperties(taskDefinition, taskExecution);
+		taskDefinition = this.updateTaskProperties(taskDefinition);
 
 		Map<String, String> appDeploymentProperties = extractAppProperties(taskDefinition.getRegisteredAppName(), taskDeploymentProperties);
 		Map<String, String> deployerDeploymentProperties = DeploymentPropertiesUtils.extractAndQualifyDeployerProperties(taskDeploymentProperties, taskDefinition.getRegisteredAppName());
 		AppDefinition revisedDefinition = mergeAndExpandAppProperties(taskDefinition, metadataResource, appDeploymentProperties);
 
-		AppDeploymentRequest request = new AppDeploymentRequest(revisedDefinition, appResource, deployerDeploymentProperties, commandLineArgs);
+		List<String> updatedCmdLineArgs = new ArrayList<>(commandLineArgs);
```
cppwfs: Should we check to see if the spring.cloud.task.executionid is already present in the command line args.  If already present should we throw an exception, since Data Flow needs to own it (at least for CF).

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/TaskControllerTests.java
```diff
@@ -468,7 +468,7 @@ public void testLaunchWithArguments() throws Exception {
 		verify(this.taskLauncher, atLeast(1)).launch(argumentCaptor.capture());
 
 		AppDeploymentRequest request = argumentCaptor.getValue();
-		assertThat(request.getCommandlineArguments().size(), is(3));
+		assertThat(request.getCommandlineArguments().size(), is(3 + 1)); // +1 for spring.cloud.task.executionid
```
cppwfs: The comment should be enough.  No need for the addition.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
ericbottard: Maybe. But same argument applied when it was an app property...

cppwfs: correct.  So do we need to remove the others and give precedence to SCDF's executionId

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/TaskControllerTests.java
ericbottard: Then you don't understand there is an addition...

This gets inlined by the compiler, if that's your worry...

cppwfs: Just curious.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
ericbottard: IMO, that reasoning could apply to A LOT of properties. There is no reason for users to mess with those normally. I would address in another PR, if at all

cppwfs: So let me ask it this way.  Based on current boot behavior if I pass in the parameter --foo=bar --foo=baz the result would be bar,baz.

ericbottard: What I'm saying is there is no reason for a user to pass that property himself.
And IF we want to protect against this, then we want to apply protection for soooo many others IMO

cppwfs: okee dokee

cppwfs:
Found a hiccup when re-start a failed batch job.   In short the new code adds a new spring.cloud.task.executionid to the command line for each restart.  And thus the restart fails with the following exception: 
```
017-03-22 14:03:51.552  WARN 14628 --- [           main] s.c.a.AnnotationConfigApplicationContext : Exception encountered during context initialization - cancelling refresh attempt: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'job1' defined in class path resource [io/spring/configuration/JobConfiguration.class]: Initialization of bean failed; nested exception is org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'org.springframework.cloud.task.batch.configuration.TaskBatchAutoConfiguration$TaskBatchExecutionListenerAutoconfiguration': Unsatisfied dependency expressed through field 'taskProperties'; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'spring.cloud.task-org.springframework.cloud.task.configuration.TaskProperties': Could not bind properties to TaskProperties (prefix=spring.cloud.task, ignoreInvalidFields=false, ignoreUnknownFields=true, ignoreNestedProperties=false); nested exception is org.springframework.validation.BindException: org.springframework.boot.bind.RelaxedDataBinder$RelaxedBeanPropertyBindingResult: 1 errors
Field error in object 'spring.cloud.task' on field 'executionid': rejected value [435,436]; codes [typeMismatch.spring.cloud.task.executionid,typeMismatch.executionid,typeMismatch.java.lang.Long,typeMismatch]; arguments [org.springframework.context.support.DefaultMessageSourceResolvable: codes [spring.cloud.task.executionid,executionid]; arguments []; default message [executionid]]; default message [Failed to convert property value of type 'java.lang.String' to required type 'java.lang.Long' for property 'executionid'; nested exception is org.springframework.core.convert.ConverterNotFoundException: No converter found capable of converting from type [java.lang.String] to type [java.lang.Long]]
```

To reproduce the exception.  Go to the Spring Cloud Task project and under the spring-cloud-task-samples module update the batch-job project.
1)  Update the pom.xml to add a dependency for your database if not using H2.
2) At Line 54 of the JobConfiguration.java add the following:
```
							if(true)
								throw new IllegalStateException("oops");

```
3) mvn clean install
4) From DataFlow Shell register the application: `app register --name batchdemo --type task --uri maven://io.spring.cloud:batch-job:1.2.0.BUILD-SNAPSHOT`
5) Create the task definition: task create bar --definition "batchdemo"
6) Launch the task definition: task launch bar
7) From your browser : http://localhost:9393/dashboard/index.html#/jobs/executions you will see that job one failed.   
8) while on that page click the restart button for Job 1.

ericbottard:
Pushed a new commit that prevents duplicates. Please re-run your test, you'll see we now get a different error, that I can't explain; Spring Batch is complaining about having already run with those parameters, which is false

cppwfs:
Spring batch ignores the parameters that are prefixed with the -- (environment properties).  So you can add an argument like foo=bar (without the --) and it will work.

ericbottard:
OK, not sure I understand all that is going on here, but I changed to not use `--`.
Please test and confirm this is doing what you expect

cppwfs:
Regression testing now.  Things look good.  Glenn Happy :-)

cppwfs:
LGTM

cppwfs:
rebased squashed
Tested with restartable Job and non-restartable job.

PR #1263, opened 2017-03-27
Replace existing TaskParser with ComposedTaskParser and support args

aclement:
With this change the parser changes to look out for applications
with arguments to be specified in composed task definitions rather
than simple names that represent task definitions.

Rather than two parsers - one for simple definitions and one
for composed definitions - there is now just one parser called
TaskParser that can understand both. The parse result
knows if it is for a simple or composed task.

There is some new consistency in that the parser is more in line with the stream parser - using more common infrastructure methods for recognizing identifiers and it is invoked in the same way now (pass data in ctor then call parse, rather than call parse with data).


spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/TaskAppNode.java
```diff
@@ -16,22 +16,34 @@
...
 package org.springframework.cloud.dataflow.core.dsl;
 
+import java.util.Collections;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
 
 /**
- * Common AST base class for nodes representing job definitions or job references.
+ * Represents either a task application or task definition reference. Task application references
```
markfisher: is this still applicable for "task definition" references, or only "task application"?

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/TaskNode.java
```diff
@@ -0,0 +1,336 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Stack;
+
+import org.springframework.cloud.dataflow.core.dsl.graph.Graph;
+import org.springframework.util.Assert;
+
+/**
+ * The root AST node for any AST parsed from a composed task specification.
```
markfisher: I believe the comments (and some variable names) should be updated throughout this class, given that it can hold either composed OR simple tasks now.

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/TaskValidatorVisitor.java
```diff
@@ -59,65 +64,60 @@ public void reset() {
...
+			if (taskAppNamesWithoutLabels.contains(labelString)) {
+				pushProblem(taskApp.getLabel().startPos,DSLMessage.TASK_VALIDATION_LABEL_CLASHES_WITH_TASKAPP_NAME);
+			}
+		}
+		else {
+			String name = taskApp.getName();
+			if (labelsDefined.contains(name)) {
+				pushProblem(taskApp.startPos,DSLMessage.TASK_VALIDATION_APP_NAME_CLASHES_WITH_LABEL);
+			}
+			if (taskAppNamesWithoutLabels.contains(name)) {
```
cppwfs: We can have cases where a task can be used multiple times within a composedTask.   (Booo) :-(

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/TaskParser.java
```diff
@@ -16,63 +16,344 @@
...
-	 * {@code myTask = filejdbc}
+	 * Task DSL text.
+	 */
+	private final String taskDefinition;
+	
+	/**
+	 * Controls usage mode for the parser. If true then apps can take arguments, if
+	 * false then app definitions are expected which take no arguments (currently).
+	 */
+	private final boolean inAppMode;
```
markfisher: is this still relevant?

markfisher: actually... I believe non-appMode is for when this parser is used by the composed-task-runner, correct?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/support/ComposedTaskValidator.java
```diff
@@ -0,0 +1,65 @@
+/*
```
cppwfs: Not necessary anymore since the parser will create the task definitions.

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/TaskParser.java
cppwfs: correct

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/TaskAppNode.java
cppwfs: Its for both the task definitions and task applications.   Its used to identify individual tasks-apps/task-definitions in the DSL.

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/TaskValidationException.java
```diff
@@ -19,30 +19,30 @@
...
 
-	private ComposedTaskNode composedTaskNode;
+	private TaskNode taskNode;
 
-	private List<ComposedTaskValidationProblem> validationProblems;
+	private List<TaskValidationProblem> validationProblems;
 
-	public ComposedTaskValidationException(ComposedTaskNode composedTaskNode, List<ComposedTaskValidationProblem> validationProblems) {
-		this.composedTaskNode = composedTaskNode;
+	public TaskValidationException(TaskNode composedTaskNode, List<TaskValidationProblem> validationProblems) {
```
markfisher: first constructor arg is still named "composedTaskNode"

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/TaskValidationException.java
```diff
@@ -19,30 +19,30 @@
...
 	}
 
-	public List<ComposedTaskValidationProblem> getValidationProblems() {
+	public List<TaskValidationProblem> getValidationProblems() {
 		return validationProblems;
 	}
 
-	public ComposedTaskNode getComposedTaskNode() {
-		return composedTaskNode;
+	public TaskNode getComposedTaskNode() {
```
markfisher: should be `getTaskNode()`?

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/TaskValidationProblem.java
```diff
@@ -19,18 +19,18 @@
 /**
  * After parsing a composed task definition from a DSL string, the validation visitor may optionally run.
```
markfisher: does this only apply to composed task definitions?

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/TaskAppNode.java
aclement: Given that the parser can be used for both the 'initial dsl':

task create --name "someTask" --definition "fooApp && barApp --p1=v1"

and the 'internal dsl' --graph param:

_someTask.fooApp && _someTask.barApp

Then it does apply to both, representing either a task app or a task def reference.

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/TaskNode.java
aclement: yep! I was just code blind looking at this stuff so many times. All fixed up.

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/TaskParser.java
aclement: yes

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/TaskValidationException.java
aclement: fixed

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/TaskValidationException.java
aclement: indeed! fixed.

cppwfs:
Found one hiccup when trying to create a task with the name _btest.gtimestamp.   The parser threw a illegal name for a task.
org.springframework.cloud.dataflow.core.dsl.ParseException: 123E:(pos 0): illegal name for a task '_btest.gtimestamp'
_btest.gtimestamp
^

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/TaskValidationProblem.java
aclement: Currently the only things found as validation problems would occur on composed tasks but I think it can be a more generally applicable thing (to simple tasks too). So I'll change that.

spring-cloud-dataflow-core/src/test/java/org/springframework/cloud/dataflow/core/dsl/TaskParserTests.java
```diff
@@ -257,41 +275,1281 @@ public void testUnbalancedDoubleQuotes() {
...
+		            "[0-1][1-2][2-3]['retry':3-2][3-4][4-5]",spec);
+		spec = "aa 0->:two && bb 0->:three && cc;two: dd;three:ee";
+		assertGraph("[0:START][1:aa][2:bb][3:cc][4:END][11:dd][12:ee]"+
+		            "[0-1][1-2][2-3][3-4][0:1-11][11-4][0:2-12][12-4]",spec);
+		spec = "aa 'foo'->:two && bb && three: cc && dd;two: ee && ff '*'->:three";
+		assertGraph("[0:START][1:aa][2:bb][3:cc][4:dd][5:END][10:ee][11:ff]"+
+		            "[0-1][1-2][2-3][3-4][4-5]['foo':1-10][10-11][11-5]['*':11-3]",spec);
+		spec = "aa 'foo'->:two && bb && four: cc && dd;two: ee && ff '*'->:three;three: gg '*'->:four";
+		TaskNode parse = parse(spec);
+		System.out.println(parse);
```
aclement: not picky at all, done!

spring-cloud-dataflow-core/src/test/java/org/springframework/cloud/dataflow/core/dsl/TaskParserTests.java
```diff
@@ -257,41 +275,1281 @@ public void testUnbalancedDoubleQuotes() {
...
+	}
+
+	@Test
+	public void toGraph$END() {
+		TaskNode ctn = parse("foo 'oranges'->$END");
+		assertEquals("foo 'oranges'->$END", ctn.toDSL());
+
+		// Graph creation
+		assertGraph("[0:START][1:foo][2:$END][3:END][0-1]['oranges':1-2][1-3]", "foo 'oranges'->$END");
+//		assertEquals(toExpectedGraph("n:0:START,n:1:foo,n:3:END,l:0:1,l:1:3:transitionName=oranges"),
```
aclement: done

spring-cloud-dataflow-core/src/test/java/org/springframework/cloud/dataflow/core/dsl/TaskParserTests.java
```diff
@@ -257,41 +275,1281 @@ public void testUnbalancedDoubleQuotes() {
...
+
+		// Graph to DSL
+		checkDSLToGraphAndBackToDSL("foo 'oranges'->$END");
+	}
+	
+	@Test
+	public void toGraph$END2() {
+		String definition = "aaa 'foo'->$END 'B'->bbb '*'->ccc && bbb && ccc";
+		assertParseAndBackToDSL(definition);
+		// {"nodes":[{"id":"0","name":"START"},{"id":"1","name":"aaa"},{"id":"3","name":"bbb"},{"id":"4","name":"ccc"},
```
aclement: done

spring-cloud-dataflow-core/src/test/java/org/springframework/cloud/dataflow/core/dsl/TaskParserTests.java
```diff
@@ -257,41 +275,1281 @@ public void testUnbalancedDoubleQuotes() {
...
+	@Test
+	public void toGraph$END3() {
+		// The trailing 'bbb' is redundant here...
+		String spec = "aaa 'foo'->$END 'B'->bbb '*'->$END && bbb";
+
+		assertEquals(spec,parse(spec).toDSL());
+
+		// Graph creation
+		//		assertEquals(
+		//				toExpectedGraph("n:0:START,n:1:aaa,n:2:bbb,n:3:ccc,n:4:END," +
```
aclement: done

spring-cloud-dataflow-core/src/test/java/org/springframework/cloud/dataflow/core/dsl/TaskParserTests.java
```diff
@@ -257,41 +275,1281 @@ public void testUnbalancedDoubleQuotes() {
...
+	private ParseException checkForParseError(String dsl, DSLMessage msg, int pos, Object... inserts) {
 		try {
-			AppNode sn = parse(task);
-			fail("expected to fail but parsed " + sn.stringify());
+			TaskNode ctn = parse(dsl);
+			fail("expected to fail but parsed " + ctn.stringify());
+			return null;
 		}
 		catch (ParseException e) {
+			System.out.println(e);
```
aclement: done

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/TaskApp.java
```diff
@@ -0,0 +1,85 @@
...
+	private String composedTaskName;
+	
+	private String name;
+	
+	private Map<String,String> arguments;
+	
+	private String label;
+
+	TaskApp(String composedTaskName, TaskAppNode taskAppNode) {
+		this.composedTaskName = composedTaskName;
```
aclement: done

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/TaskApp.java
```diff
@@ -0,0 +1,85 @@
...
+		return arguments;
+	}
+
+	public String getLabel() {
+		return label;
+	}
+	
+	public String toString() {
+		StringBuilder s = new StringBuilder();
+		if (label!=null) {
```
aclement: done

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/TaskNode.java
```diff
@@ -0,0 +1,336 @@
...
+	 */
+	private List<LabelledTaskNode> sequences;
+
+	/**
+	 * All the apps mentioned in the composed task definition.
+	 */
+	private List<TaskApp> taskApps;
+	
+	TaskNode(String name, String composedTaskSpecification, List<LabelledTaskNode> sequences, boolean inAppMode) {
+		super((sequences.size()==0) ? 0 : sequences.get(0).getStartPos(),
```
aclement: Is that necessary when it isn't public?

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/TaskNode.java
```diff
@@ -0,0 +1,336 @@
...
+	 */
+	private List<TaskApp> taskApps;
+	
+	TaskNode(String name, String composedTaskSpecification, List<LabelledTaskNode> sequences, boolean inAppMode) {
+		super((sequences.size()==0) ? 0 : sequences.get(0).getStartPos(), 
+				(sequences.size()==0) ? 0 : sequences.get(sequences.size()-1).getEndPos());
+		this.name = name;
+		this.composedTaskText = composedTaskSpecification;
+		this.sequences = sequences;
+		// TODO use inAppMode to police what can be called on this node?
```
aclement: If I do that I get an 'unused variable' warning on it since no-one ever does anything with it (at the moment), for now I just left it in the ctor args and the comment.

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/TaskNode.java
```diff
@@ -0,0 +1,336 @@
...
+		if (sequences.size()==0) {
+			return null;
+		}
+		else {
+			return sequences.get(0);
+		}
+	}
+	
+	public List<LabelledTaskNode> getSequences() {
+		return sequences;
```
aclement: mm, ok. the whole notion of sequences may disappear depending on where we go with DSL design post next release.

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/TaskNode.java
```diff
@@ -0,0 +1,336 @@
...
+		ExecutableDSLVisitor v = new ExecutableDSLVisitor();
+		accept(v);
+		return v.getDSL();
+	}
+	
+	static class ExecutableDSLVisitor extends TaskVisitor {
+
+		private final String PREFIX = "_";
+		
+		StringBuilder dsl = new StringBuilder();
```
aclement: done

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/TaskParser.java
```diff
@@ -16,63 +16,344 @@
...
+	 * @return array of arguments or {@code null} if the next token(s) do not
+	 * contain arguments
 	 */
-	public AppNode parse() {
-		AppNode ast = eatApp();
-
-		// Check the task name, however it was specified
-		if (ast.getName() != null && !isValidName(ast.getName())) {
-			throw new ParseException(ast.getName(), 0, DSLMessage.ILLEGAL_TASK_NAME, ast.getName());
+	protected ArgumentNode[] maybeEatAppArgs() {
```
aclement: Done. Given more time should look at whether to remove and use the arg eating code in the superclass.

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/TaskValidatorVisitor.java
aclement: Yes, but for now they'll all need separate labels to keep them unique, is that going to be a problem? It does match stream from this point of view but we could have used something else to name the definitions produced by the first DSL parse (like a counter suffix to keep them unique).

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/support/ComposedTaskValidator.java
aclement: ok, i'll delete that.

aclement:
I've addressed the comments. I also did a text search for 'composed' to see if there were other places that needed updating to remove the composed/non-composed distinction and fixed a few more places.

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/TaskValidatorVisitor.java
cppwfs: Composed Task Runner does not support labels yet.  So this will be one feature (supporting apps with labels) that I'll try to squeeze in between RC and release.

markfisher: I just want to understand the case where the first DSL parsing result would contain non-unique generated definition names. If the first DSL text itself contains 2 ambiguous app references, then at least one of them *must* have a label, right? Wouldn't that same label be used to generate the definition name?

aclement: > We can have cases where a task can be used multiple times within a composedTask

What is a case of that?

What Mark says should be what happens. The uniqueness constraints enforced on first parse means the subsequent executable dsl passed to the composed task runner cant have non-unique names in it.  How/why does composed task runner need to support labels?

aclement:
What Glenn found is a bit of a hiccup, we agreed on the created reference names being of the form _TASKNAME.APPNAME but there are validation checks in some places that want it to look like a Java identifier (well a java identifier that can also include hyphens).  The 'dot' in that name is tripping the validation logic up. Rather than change what is valid, I changed it to do what glenn suggested and create names of the form _TASKNAME_APPNAME

cppwfs:
Simple Transitions are not being recognized as a Composed Task.
i.e. `_foo1_gtimestamp 'FAILED'->_foo1_taskEx '*'->_foo1_timestamp` is treated as a regular task.
while `_foo1_gtimestamp 'FAILED'->_foo1_taskEx '*'->_foo1_timestamp && _foo1_batchdemo` is treated as a composed task.

cppwfs:
When creating the following task: 
`task create foo --definition "AAA && FFF 'FAILED' -> EEE && <BBB||CCC> && DDD"`
the following is produced by toExecutableDSL
`_foo_AAA && _foo_FFF 'FAILED'->_foo_EEE<_foo_BBB || _foo_CCC> && _foo_DDD`
Notice that the && is missing between _foo_EEE and the split.

aclement:
Both those cases fixed up - sorry about all the problems! That executable dsl producer visitor is a tricky beast. I've also fixed up a few related situations you hadn't hit (yet). All pushed.

cppwfs:
We're getting close.  
The following `task create foo4 --definition "AAA 'FAILED' -> BBB * -> CCC"`
generates

AAA 'FAILED' -> BBB null -> CCC

cppwfs:
LGTM unless there are any objections.  
I will reformat the new classes on merge.

cppwfs:
Rebased, Squashed, Polished, Merged

cppwfs:
oops forgot to close

PR #1264, opened 2017-03-27
Task tools endpoint

aclement:
Needs the other PR to be in first. This adds a controller to expose back end task parsing support to tools (Flo in this case).

This depends on https://github.com/spring-cloud/spring-cloud-dataflow/pull/1263 - please do not merge it until the parent is merged.


cppwfs:
* The code makes sense to me.  
* But noticed that it wasn't added to the RootController, I think this is a requirement.  Which also means that the endpoints would need to be updated to return a resource.  
* Revision in Version.class needs to be updated.  
* Don't think we need ToolsOperations since it wouldn't be used by the shell. (unless that is a requirement)
@ericbottard What are your thoughts?

ericbottard:
Agreed with Glenn, this would be ideal.
Also, if the controller could be made more "resource oriented" in its approach (rather than RPC), this would be beneficial for consistency

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/RootController.java
```diff
@@ -134,6 +136,12 @@ public RootResource info() {
 		root.add(new Link(completionStreamTemplated).withRel("completions/stream"));
 		String completionTaskTemplated = entityLinks.linkFor(CompletionProposalsResource.class).withSelfRel().getHref() + ("/task{?start,detailLevel}");
 		root.add(new Link(completionTaskTemplated).withRel("completions/task"));
+
+		// Should this be in the isTasksEnabled() block above? Why are the completions endpoints not guarded by feature checks?
+		String taskTextToGraphTemplated = entityLinks.linkFor(TaskGraphResource.class).withSelfRel().getHref() + ("/parseTaskTextToGraph{?name,definition}");
```
aclement: I have no idea what to do here. But this doesn't work. Tests fail when you mvn build it. The test failure message is unhelpful but if I run it anyway, it fails (curling to localhost:9393):

[{"logref":"IllegalArgumentException","message":"Cannot determine link for org.springframework.cloud.dataflow.rest.resource.TaskGraphResource! No EntityLinks instance found supporting the domain type!","links":[]}]

which is probably the big clue. I patterned what I did off what it did for completions/ endpoint - per my comment there I don't know why completions/task isn't guarded on being only active if the tasks feature is on? (Mine would be the same)

spring-cloud-starter-dataflow-server-local/src/test/java/org/springframework/cloud/dataflow/server/local/security/LocalServerSecurityWithUsersFileTests.java
```diff
@@ -135,6 +135,32 @@
 			{ HttpMethod.GET, HttpStatus.BAD_REQUEST,  "/completions/task", createOnlyUser, ImmutableMap.of("start", "2", "detailLevel", "-123") },
 			{ HttpMethod.GET, HttpStatus.UNAUTHORIZED, "/completions/task", null, ImmutableMap.of("detailLevel", "2") },
 
+			/* ToolsController */
+
+			{ HttpMethod.GET, HttpStatus.FORBIDDEN,    "/tools/parseTaskTextToGraph", adminOnlyUser, null },
```
aclement: These need some work, but until the fundamental problems in the RootController are addressed I can't look into them.  If you remove these tests the existing tests still fail due to the changes in the RootController above.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/ToolsController.java
```diff
@@ -0,0 +1,140 @@
...
+	
+	/**
+	 * Parse a task definition into a graph structure. The definition map is expected to have
+	 * a 'dsl' key containing the composed task DSL and optionally a 'name' key indicating the name of the
+	 * composed task.
+	 * 
+	 * @return a map with either a 'graph' or 'error' key set 
+	 */
+	@RequestMapping(value = "/parseTaskTextToGraph", method = RequestMethod.GET)
+	public TaskGraphResource parseTaskTextToGraph(@RequestParam(value = "name", defaultValue = "__dummy") String name,
```
aclement: Is this a GET, is it a POST - input welcome...

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/ToolsController.java
```diff
@@ -0,0 +1,140 @@
...
+			error = pe.toExceptionDescriptor();
+		}
+		return taskGraphAssembler.toResource(new ParsedGraphOutput(graph,error));
+	}
+	
+	/**
+	 * Convert a graph format into DSL text format.
+	 */
+	@RequestMapping(value = "/convertTaskGraphToText", method = RequestMethod.GET)
+	public TaskDslResource convertTaskGraphToText(@RequestBody Graph graph) {
```
aclement: Similar to above, GET/POST?

aclement:
Ok, i've switched to to be 'resource based' causing me to write much more code than I had with my nice little simple ToolsController that worked and could be understood in 5 seconds....  But I'm stuck, I don't understand what it wants me to put into RootController. And until I get that right I can't test anything . Input welcome!

sabbyanandan:
@ghillert Until @ericbottard gets to it, if you have any insights to share, please do.

ghillert:
@aclement I see the issue: You have to refactor `ToolsController` and break it into 2 controllers. At the class-level you need to add `@ExposesResourceFor(TaskGraphResource.class)` and `@ExposesResourceFor(TaskDslResource.class)` respectively. Once, you do, the exception will go away.

aclement:
Thanks @ghillert instead of splitting the controller I reworked it to a single resource. Tests are passing now.

ghillert:
Merged with commit 26227d3676c819a8ece8d861925ebdef31b7fd49
Fixed trailing white-space issues and updated copyright years during merge. Furthermore, updated the `ApiDocumentation`

Issue #1273, opened 2017-03-30
StreamDefinitionResource: Make status lower-case and add statusDescription property


ghillert:
Make status lower-case and add statusDescription property.

The current statuses are camel-case and conflict with the UI expectations. Therefore, this issue will also implicitly address UI issue https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/178

In addition to that we need to add more context to the statuses and add a `statusDescription` property. The `statusDescription` will follow closely the descriptions in `org.springframework.cloud.deployer.spi.app.DeploymentState`.

PR #1274, opened 2017-03-31
 gh-1273 StreamDefinitionResource: Make status lower-case and add statusDescription property

ghillert:
resolves https://github.com/spring-cloud/spring-cloud-dataflow/issues/1273
resolves https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/178

For acceptance, please double check also: https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/178

### DIFF
--- a/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/resource/DeploymentStatus.java
+++ b/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/resource/DeploymentStatus.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.cloud.dataflow.rest.resource;
+
+import org.springframework.util.Assert;
+
+import com.fasterxml.jackson.annotation.JsonFormat;
+
+/**
+ * Provides a typed enumeration of deployment statuses for Streams.
+ *
+ * @author Gunnar Hillert
+ *
+ */
+@JsonFormat(shape = JsonFormat.Shape.OBJECT)
+public enum DeploymentStatus {
+
+	DEPLOYING("deploying", "Deploying" ,"The app or group is being deployed."),
+	DEPLOYED("deployed", "Deployed", "All apps have been successfully deployed"),
+	UNDEPLOYED("undeployed", "Undeployed", "The app or group is known to the system, but is not currently deployed"),
+	PARTIAL("partial", "Partial", "In the case of multiple apps, some have successfully deployed, while others have not"),
+	FAILED("failed", "Failed", "All apps have failed deployment"),
+	ERROR("error", "Error", "A system error occurred trying to determine deployment status"),
+	UNKNOWN("unknown", "Unknown", "The app or group deployment is not known to the system");
+
+	private final String key;
+	private final String displayName;
+	private final String description;
+
+	/**
+	 * Constructor.
+	 */
+	DeploymentStatus(final String key, final String displayName, final String description) {
+		this.key = key;
+		this.displayName = displayName;
+		this.description = description;
+	}
+
+	public String getKey() {
+		return key;
+	}
+
+	public String getDisplayName() {
+		return displayName;
+	}
+
+	public String getDescription() {
+		return description;
+	}
+
+	public static DeploymentStatus fromKey(String deploymentStatusKey) {
+
+		Assert.hasText(deploymentStatusKey, "Parameter deploymentStatusKey must not be null or empty.");
+
+		for (DeploymentStatus deploymentStatus : DeploymentStatus.values()) {
+			if (deploymentStatus.getKey().equals(deploymentStatusKey)) {
+				return deploymentStatus;
+			}
+		}
+
+		return null;
+	}
+}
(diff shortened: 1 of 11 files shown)


spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/resource/DeploymentStatus.java
```diff
@@ -0,0 +1,76 @@
...
+import com.fasterxml.jackson.annotation.JsonFormat;
+
+/**
+ * Provides a typed enumeration of deployment statuses for Streams.
+ *
+ * @author Gunnar Hillert
+ *
+ */
+@JsonFormat(shape = JsonFormat.Shape.OBJECT)
+public enum DeploymentStatus {
```
ericbottard: The choice of name `DeploymentStatus` is dangerous IMHO, as the vocabulary is already very crowded with similar terms.

If you take into account my other comment, all this does is bridge a name to a human readable description. What's the value of bringing in an enum (on which you don't even do `==`?). Also, the code should make sure (eg in a static block, but without coupling this to `DeploymentState`) that all values of `DeploymentState` are covered (in case we change/add them in the future)

spring-cloud-dataflow-rest-resource/src/test/java/org/springframework/cloud/dataflow/rest/resource/DeploymentStatusSerializationTests.java
```diff
@@ -0,0 +1,50 @@
...
+import com.jayway.jsonpath.DocumentContext;
+import com.jayway.jsonpath.JsonPath;
+
+/**
+ * @author Gunnar Hillert
+ */
+public class DeploymentStatusSerializationTests {
+
+	@Test
+	public void testSerializationOfSingleStepExecution() throws JsonProcessingException {
```
ericbottard: Why do that, (and Jackon annotate the enum) since it's never sent on the wire?

spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/resource/DeploymentStatus.java
ghillert: That is unfortunate but we already have a parallel collection of DTO objects in package `org.springframework.cloud.dataflow.rest.resource`. Unless we want to add a common suffix for those objects. Honestly, I am surprised that this has not come up earlier. I cannot suffix it with `Resource` since it is not a resource object. It is basically a direct mirror object to `DeploymentState` without having any dependencies to the Deployer.

When talking to @markpollack my main idea was to add a `DeploymentStatus` property to `StreamDefinitionResource` replacing the `status` property. Technically, it is wrong to add the `statusDescription` flatly to the `StreamDefinitionResource`. BUT, for `1.2` we did not want to break backwards-compatibility. Thus, this compromise was done but I think it is directly usable for `2.0`. Also, I had to store the Statuses somewhere, rather than using a `Utils` class, I felt to add it to the `eventually correct` spot.

That being said, I still believe that `DeploymentStatus` should eventually be its own property. To indicate that I create the class the way I did, and created it in respect to serializability as well. 

As it directly, translates to the `DeploymentState` enums, I created it as an emum as well.

spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/resource/DeploymentStatus.java
```diff
@@ -0,0 +1,76 @@
...
+
+	DEPLOYING("deploying", "Deploying" ,"The app or group is being deployed."),
+	DEPLOYED("deployed", "Deployed", "All apps have been successfully deployed"),
+	UNDEPLOYED("undeployed", "Undeployed", "The app or group is known to the system, but is not currently deployed"),
+	PARTIAL("partial", "Partial", "In the case of multiple apps, some have successfully deployed, while others have not"),
+	FAILED("failed", "Failed", "All apps have failed deployment"),
+	ERROR("error", "Error", "A system error occurred trying to determine deployment status"),
+	UNKNOWN("unknown", "Unknown", "The app or group deployment is not known to the system");
+
+	private final String key;
```
ghillert: Personally, I believe using the enum's name for **complex enumerations** is wrong. I ran into to many cases in my life where that bit me.

Also, that way, Jackson can nicely render the enum using:

```
@JsonFormat(shape = JsonFormat.Shape.OBJECT)
```

spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/resource/DeploymentStatus.java
ericbottard: OK, this all makes sense then. Still, could you
- add a comment somewhere explaining that the new field is transient
- change the name, as I'm really worried about term pollution here

spring-cloud-dataflow-rest-resource/src/test/java/org/springframework/cloud/dataflow/rest/resource/DeploymentStatusSerializationTests.java
ghillert: I created the test so that it eventually could be sent over the wire.

ghillert:
@ericbottard Made the changes. I hope `StreamDetploymentStatus` differentiates the status sufficiently. Mentioned to @sabbyanandan that eventually we should come up with a generally applicable naming strategy for classes under `spring-cloud-dataflow-rest-resource` that might conflict with other domain classes elsewhere.

ericbottard:
`StreamDetploymentStatus` is not a good choice IMO because then it can not apply to single app instances, nor can it be used in tasks context. Remember that DeploymentState is used for both instance state and grouped state

ghillert:
@markpollack Do you have better ideas on the naming? I am a bit of a loss here in regards to a name that does not clash with `DeploymentState` but basically means the same.

ericbottard:
What about `DeploymentStateResource`? Even though not 100% accurate, this could be a good compromise. Or else just go with `DeploymentState`

ghillert:
@ericbottard Rebased and changed to `DeploymentStateResource`.

ericbottard:
Merged as 8f724b72

Changed the `stream list` command to use the `statusDescription`, since this was the whole point of the problematic change that this PR tries to cure (provide more explanation for ambiguous states [error])

Issue #1272, opened 2017-03-30
Delete Child Tasks for Composed Task


cppwfs:
When executing task destroy on a composed task.  The child tasks for that task should be destroyed.

mminella:
Are we re-using definitions when a definition is parsed?  AKA, if we find a definition that matches one that already exists, do we re-use it?  If so, we wouldn't want to do this.

cppwfs:
good point.  But this means that tasks that are created as a part of the _taskname_taskApp will remain.  Not quite an orphan, but untidy.

mminella:
I'd vote for this to be an option.  Something like `task destroy --name=foo --cascade` that we could then warn the user that this may delete definitions that are in use.

markfisher:
why would you ever want to preserve the generated definitions though?

mminella:
@markfisher To my point above, if we are re-using definitions, you'd want to keep them.  If we are not, then I retract my concern (I'm asking...I honestly don't know).   @aclement would know...

cppwfs:
@markfisher To Michael's point, a user can technically create a
composed_task_runner by hand using existing tasks.  And for this we would
not want to destroy the "child" tasks.

mminella:
Wait... @cppwfs That is different than what I was implying.  I'm implying that if you had the following:
```
task create --name=foo --definition="AAA --a=1 && BBB"
task create --name=bar --definition="BBB && CCC"
```
Does the above result in three six definitions or five definitions?  Specifically, does the generated `BBB` definition get re-used between the two composed definitions?

cppwfs:
in the case above you would get 6 (and an egg roll) ;-) :
_foo_AAA
_foo_BBB
foo
_bar_BBB
_bar_CCC
bar

mminella:
If that's the case, then disregard my concern.

PR #1276, opened 2017-03-31
Task save needs to support composed task definitions returned from pa…

### LINKED ISSUE #1262: Task save needs to support composed task definitions returned from parser
The task parser will now support composed tasks as well as traditional task definitions.  It will return 2 types of results from a parsed composed task definition:  
1) The task definitions that are derived from the composed task definition.
2) The composed task definition itself.  

With that being said the task save will now need to parse the task definition to obtain the task definitions and composed task definition and save those to the repository.

cppwfs:
…rser

resolves #1262,#1272

Added ability to allow the user set the CTR app name

Added task destroy code for composed task

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/ComposedTaskProperties.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/ComposedTaskProperties.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.dataflow.server.service.impl;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * Properties used to define the behavior of composed tasks created by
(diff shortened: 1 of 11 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/ComposedTaskProperties.java
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Properties used to define the behavior of composed tasks created by
+ * Spring Cloud Data Flow.
+ *
+ * @author Glenn Renfro
+ */
+@ConfigurationProperties(prefix = ComposedTaskProperties.COMPOSED_TASK_PREFIX)
+public class ComposedTaskProperties {
+
+	public static final String COMPOSED_TASK_PREFIX = "spring.cloud.dataflow.composed.task";
```
markfisher: while this prefix flows logically, I wonder if hierarchically it would be better to have:
`spring.cloud.dataflow.task.composed`

markfisher: ...or should we just have `TaskConfigurationProperties` with a prefix: `spring.cloud.dataflow.task`

then we could have a single property for now, called `composedTaskRunnerName`

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
```diff
@@ -217,5 +231,78 @@ private TaskDefinition updateTaskProperties(TaskDefinition taskDefinition) {
...
+						}
+					}
+			);
+		}
+		//destroy normal task or composed-task
+		destroyTask(name);
+	}
+
+	private void destroyTask(String name) {
+		TaskDefinition taskDefinition = taskDefinitionRepository.findOne(name);
```
ilayaperumalg: It would be safe to throw `NoSuchTaskDefinitionException` if `taskDefinition` is null here as the `destroyTask` is called for child definitions as well.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed and merged as 519a7f4

PR #1281, opened 2017-04-04
Secure /about page

### LINKED ISSUE #1232: /about should be secured
With security on, the `/about` endpoint should not be freely accessible

ilayaperumalg:
- Use `/security/info` to retrieve the basic security info retrieval
   - With this, we don't deprecate the `/security/info` REST endpoint controller
 - Remove `/about` link from the default authenticated links
 - Define role for the `/about` GET request
 - Update tests & doc

Resolves #1232

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/configuration.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/configuration.adoc
@@ -525,6 +525,20 @@ for instance every user shall have have access to all endpoints, then you can al
 set `spring.cloud.dataflow.security.authorization.enabled=false`.
 ====
 
+If you are using basic security configuration by using security properties then it is important to set the roles for the users.
+
+For instance,
+
+[source,bash]
+----
+java -jar spring-cloud-dataflow-server-local/target/spring-cloud-dataflow-server-local-<version>.jar \
+    --security.basic.enabled=true \
+    --security.user.name=test \
+    --security.user.password=pass \
+    --security.user.role=VIEW &
+----
+
+
 [[authorization-shell-and-dashboard]]
 ==== Authorization - Shell and Dashboard Behavior
 
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/configuration.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/configuration.adoc
@@ -435,6 +435,10 @@ spring:
         authorization:
           enabled: true
           rules:
+            # About
+
+            - GET    /about                          => hasRole('ROLE_VIEW')
+
             # Metrics
 
             - GET    /metrics/**                     => hasRole('ROLE_VIEW')
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/security/BasicAuthSecurityConfiguration.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/security/BasicAuthSecurityConfiguration.java
@@ -118,7 +118,6 @@ public class BasicAuthSecurityConfiguration extends WebSecurityConfigurerAdapter
 					"/authenticate",
 					"/security/info",
 					"/dashboard",
-					"/about",
 					"/features",
 					"/assets/**").permitAll();
 
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/security/SecurityController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/security/SecurityController.java
@@ -42,7 +42,6 @@ import org.springframework.web.bind.annotation.RestController;
  * @author Gunnar Hillert
  * @author Ilayaperumal Gopinathan
  * @since 1.0
- * @deprecated Functionality now provided by {@link AboutController}
  */
 @RestController
 @RequestMapping("/security/info")
--- a/spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
+++ b/spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
@@ -13,6 +13,10 @@ spring:
         authorization:
           enabled: true
           rules:
+            # About
+
+            - GET    /about                          => hasRole('ROLE_VIEW')
+
             # Metrics
 
             - GET    /metrics/**                     => hasRole('ROLE_VIEW')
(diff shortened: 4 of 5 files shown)


ericbottard:
When I enable security and successfully `dataflow config server --username xx --password yy`, it seems `/about` is still not read properly, as I can't
* `dataflow config info`
* all other commands are disabled

ilayaperumalg:
@ericbottard have you tried this with the user assigned with `VIEW` role. Since we now have made the `/about` page secure, the `VIEW` role is also set for that link. Hence, the user **requires** the `VIEW` role. and, that is applicable for other commands as well.

ilayaperumalg:
Also, noticed that some of the changes made in `Shell` were not made as part of this PR branch. Hence, I have also rebased this PR to include `featureInfo` part of `dataflow config info` command. Is that what you think is missing?

ericbottard:
I tried this by simply enabling `--security.basic.enabled=true`, which used to work.
It seems that now, the single user that is created (`user`) doesn't have any role, so this makes it a useless configuration.

See http://docs.spring.io/spring-cloud-dataflow/docs/1.2.0.BUILD-SNAPSHOT/reference/htmlsingle/#configuration-security-basic-authentication

ilayaperumalg:
yeah, I agree. May be we can track it as a separate issue as the current master behaves the same way?

ericbottard:
Ah, did not realize master had the same issue. Created https://github.com/spring-cloud/spring-cloud-dataflow/issues/1285 to track it, will merge this

ericbottard:
The thing is, `master` **doesn't** suffer from the aforementioned problem.

I suggest changing the `/about` rule from `hasRole('ROLE_VIEW')` to `isAuthenticated()`, wdyt?

ilayaperumalg:
@ericbottard I just tried the following and it looks ok:

```
spring-cloud-dataflow (secure-about-page)]$ java -jar spring-cloud-dataflow-server-local/target/spring-cloud-dataflow-server-local-1.2.0.BUILD-SNAPSHOT.jar --security.basic.enabled=true --security.user.name=test --security.user.password=pass --security.user.role=VIEW
```

From the shell:

```
server-unknown:>dataflow config server --username test --password pass
Successfully targeted http://localhost:9393/
dataflow:>dat
dataflow    date
dataflow:>dataflow config info
╔═════════════╤════════════════════════════════════════════════════════╗
║   Target    │                 http://localhost:9393/                 ║
╟─────────────┼────────────────────────────────────────────────────────╢
║   Result    │      Successfully targeted http://localhost:9393/      ║
╟─────────────┼────────────────────────────────────────────────────────╢
║             │Analytics Enabled: true                                 ║
║  Features   │  Streams Enabled: true                                 ║
║             │    Tasks Enabled: true                                 ║
╟─────────────┼────────────────────────────────────────────────────────╢
║             │spring-cloud-dataflow-server-local: 1.2.0.BUILD-SNAPSHOT║
║  Versions   │       Spring Cloud Data Flow Core: 1.2.0.BUILD-SNAPSHOT║
║             │          Spring Cloud Dataflow UI: 1.2.0.BUILD-SNAPSHOT║
╟─────────────┼────────────────────────────────────────────────────────╢
║             │         Authenticated: true                            ║
║             │Authentication Enabled: true                            ║
║  Security   │ Authorization Enabled: true                            ║
║             │            Form Login: true                            ║
║             │              Username: test                            ║
╟┈┈┈┈┈┈┈┈┈┈┈┈┈┼┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈╢
║    Roles    │                       ROLE_ADMIN                       ║
║             │                       ROLE_VIEW                        ║
╟─────────────┼────────────────────────────────────────────────────────╢
║             │Deployer Implementation Version: 1.2.0.M3               ║
║             │                  Deployer Name: LocalAppDeployer       ║
║             │           Deployer Spi Version: 1.2.0.M3               ║
║             │                   Java Version: 1.8.0_20               ║
║App Deployer │           Platform Api Version: Mac OS X 10.9.5        ║
║             │        Platform Client Version: 10.9.5                 ║
║             │          Platform Host Version: 10.9.5                 ║
║             │                  Platform Type: Local                  ║
║             │            Spring Boot Version: 1.5.2.RELEASE          ║
║             │                 Spring Version: 4.3.7.RELEASE          ║
╟─────────────┼────────────────────────────────────────────────────────╢
║             │Deployer Implementation Version: 1.2.0.M3               ║
║             │                  Deployer Name: LocalTaskLauncher      ║
║             │           Deployer Spi Version: 1.2.0.M3               ║
║             │                   Java Version: 1.8.0_20               ║
║Task Launcher│           Platform Api Version: Mac OS X 10.9.5        ║
║             │        Platform Client Version: 10.9.5                 ║
║             │          Platform Host Version: 10.9.5                 ║
║             │                  Platform Type: Local                  ║
║             │            Spring Boot Version: 1.5.2.RELEASE          ║
║             │                 Spring Version: 4.3.7.RELEASE          ║
╚═════════════╧════════════════════════════════════════════════════════╝

dataflow:>
```

ericbottard:
OK, but then at a minimum we need to update the documentation.

Also, how come the table shows that you have `ROLE_ADMIN`?

ilayaperumalg:
I believe that `ADMIN` is not part of the data flow at all and may be comes from spring-boot by default.

ericbottard:
Hmm, then maybe we need hierarchical roles then. I would expect that ADMIN > VIEW

ilayaperumalg:
The `ADMIN` comes from [this](https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml#L3) configuration. I am not sure why it is there as default? cc @ghillert

ilayaperumalg:
This was infact added from [this](https://github.com/spring-cloud/spring-cloud-dataflow/commit/47721a719b37e76c0fd285c241b6dc9571af99b1#diff-7c04dac4148843e35372b2248cf5c344) commit. cc @trisberg

ilayaperumalg:
@ericbottard Discussed with Thomas on that. Here the `ADMIN` role refers to the actuator endpoints configuration from [here](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-1.5-Release-Notes#actuator-security) and I don't think it is something hierarchical from what we configure as roles. what do you think?

ericbottard:
What I meant by hierarchical roles is that, if you are ADMIN, then you should also automatically have VIEW role (see http://docs.spring.io/spring-security/site/docs/4.2.3.BUILD-SNAPSHOT/reference/htmlsingle/#authz-hierarchical-roles).

I understand this is somewhat unrelated to this PR, but still. If this was the case, then you wouldn't need additional config

ilayaperumalg:
This was already merged as https://github.com/spring-cloud/spring-cloud-dataflow/commit/e6dc0041096061b018f61390c1b9e5dd51073208

Issue #1279, opened 2017-04-03
Implement OAUTH2 "Client Credentials Grant" type for REST-APIs


sabbyanandan:
As a user, I'd like to interact with (OAUTH2) secured SCDF REST-APIs by passing ~~"Client ID" and "Client Secret"~~ **Access Token** as opposed to a `username/password` combination.

This change will provide an additional option for REST API users. Instead of providing a `username:password` combination via BasicAuth, users have the ability to retrieve an OAuth2 Access token from their OAuth2 provider directly and then provide the Access Token as an HTTP header as part of their Spring Cloud Data Flow REST API calls.

The to-be-provided HTTP Header will have the form:

```
Authorization: bearer 1234567890
``` 

This optional header is in **Addition** to the current security setup. Thus, no breaking change for existing users is to be expected. But if the `Authorization` header is provided it will take precedence.

_note: #1278 (spike) is a prerequisite for this feature implementation._

**Acceptance:**
- New requests with the Authorization header via `-H Authorization: bearer <token>` are automatically parsed and evaluated for validity against the configured OAUTH2 server
- REST-APIs accept "Client Credentials Grant" type to evaluate the client credentials against the configured OAUTH2 server

PR #1284, opened 2017-04-10
gh-1279 Implement OAUTH2 "Acess Token" support for REST-APIs

ghillert:
* Relax security configuration: `security.basic.enabled=true` is not required for OAuth2 any longer
* Add `OAuth2ServerResource` for creating JUnit integration tests using an embedded OAuth2 server instance
* Add tests

resolves spring-cloud/spring-cloud-dataflow#1279

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/security/BasicAuthSecurityConfiguration.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/security/BasicAuthSecurityConfiguration.java
@@ -28,6 +28,7 @@ import org.springframework.boot.autoconfigure.security.SecurityProperties;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.cloud.dataflow.core.DataFlowPropertyKeys;
 import org.springframework.cloud.dataflow.server.config.security.support.OnSecurityEnabledAndOAuth2Disabled;
+import org.springframework.cloud.dataflow.server.config.security.support.SecurityStateBean;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Conditional;
 import org.springframework.context.annotation.Configuration;
@@ -96,6 +97,9 @@ public class BasicAuthSecurityConfiguration extends WebSecurityConfigurerAdapter
 	@Autowired
 	private AuthorizationConfig authorizationConfig;
 
+	@Autowired
+	private SecurityStateBean securityStateBean;
+
 	@Override
 	protected void configure(HttpSecurity http) throws Exception {
 		final RequestMatcher textHtmlMatcher = new MediaTypeRequestMatcher(
@@ -157,6 +161,9 @@ public class BasicAuthSecurityConfiguration extends WebSecurityConfigurerAdapter
 		http.addFilterBefore(sessionRepositoryFilter,
 				ChannelProcessingFilter.class).csrf().disable();
 		http.sessionManagement().sessionCreationPolicy(SessionCreationPolicy.IF_REQUIRED);
+
+		securityStateBean.setAuthenticationEnabled(true);
+		securityStateBean.setAuthorizationEnabled(true);
 	}
 
 	/**
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/security/support/OnSecurityEnabledAndOAuth2Enabled.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/security/support/OnSecurityEnabledAndOAuth2Enabled.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 the original author or authors.
+ * Copyright 2016-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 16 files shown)


sabbyanandan:
Hi, @rwinch:  At your convenience, if you can _also_ review and give us some feedback on this PR, that'd be helpful.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/security/BasicAuthSecurityConfiguration.java
```diff
@@ -157,6 +161,9 @@ protected void configure(HttpSecurity http) throws Exception {
 		http.addFilterBefore(sessionRepositoryFilter,
 				ChannelProcessingFilter.class).csrf().disable();
 		http.sessionManagement().sessionCreationPolicy(SessionCreationPolicy.IF_REQUIRED);
+
+		securityStateBean.setAuthenticationEnabled(true);
+		securityStateBean.setAuthorizationEnabled(true);
```
jvalkeal: This is set to `false` in `OAuthSecurityConfiguration` and looks like assumption is that `OnSecurityEnabledAndOAuth2Enabled` protects that only one config is enabled. I can't wrap my head around how `OnSecurityEnabledAndOAuth2Enabled` now works that `security.basic.enabled` has been removed from it. Am I right that `BasicAuthSecurityConfiguration` is called anyway and if `security.oauth2.client.client-id` is set, also `OAuthSecurityConfiguration` is called and `SecurityStateBean` parameters are set twice. In that case we're rely on order how config classes are called?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/security/support/OnSecurityEnabledAndOAuth2Enabled.java
```diff
@@ -33,9 +33,6 @@ public OnSecurityEnabledAndOAuth2Enabled() {
 		super(ConfigurationPhase.REGISTER_BEAN);
 	}
 
-	@ConditionalOnProperty(name = "security.basic.enabled", havingValue = "true")
```
jvalkeal: javadocs still mention
```
{@link Condition} that is only valid if {@code security.basic.enabled}
```
which I assume is not valid anymore?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/security/BasicAuthSecurityConfiguration.java
jvalkeal: It'd be nice to get comment from @rwinch about a usage of `securityStateBean`. It'd be nice if that bean could be created once with correct settings from a getgo if it's intended to only get passed to security info page.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/security/support/OnSecurityEnabledAndOAuth2Enabled.java
ghillert: Correct - I removed the JavaDoc reference

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/security/BasicAuthSecurityConfiguration.java
ghillert: I agree that the Spring Boot conditional annotations are not too easy on the brain.

Coming from the old Spring XD, we had originally used `security.basic.enabled` as the main driver to enable security in SCDF. That by itself is confusing also, also compounded by the fact that our OAuth2 configuration also uses `BasicAuth` which is completely unrelated to the normal Spring Security setup we use (File-based and Ldap authentication) that also uses Basic Auth.

As a auxiliary requirement (customer - @sabbyanandan), I removed the requirement to have `security.basic.enabled` to be present for OAuth2 configuration. In order to not break backwards-compatibility, you can still leave it there.

The key point is that `security.oauth2.client.client-id` is THE property that triggers OAuth2 configuration to be enabled. Also, either `BasicAuthSecurityConfiguration` is loaded OR `OAuthSecurityConfiguration` but never both.

ghillert: @jvalkeal Regarding `securityStateBean` - I was not to keen on using that BUT somehow I have to share some state that answers the basic questions:

* Is authentication enabled?
* Is authorization enabled?

As our security enablement is `implicit` and `conditional`, it gets a bit blurry to answer that question with Spring Boot properties alone. Also, our OAuth2 setup does not support roles directly as of yet. Therefore, I felt more comfortable to answer those 2 question from within  the configuration classes `BasicAuthSecurityConfiguration` and `OAuthSecurityConfiguration` directly.

jvalkeal:
ok, gotcha, think that cleared ii. merging.

jvalkeal:
squashed/merged per https://github.com/spring-cloud/spring-cloud-dataflow/commit/32a06c7d1bf7841c45e2b0af21111cb75cc6d3c4

PR #1283, opened 2017-04-07
Add Metrics system

jvalkeal:
- Add hystrix dependency used by MetricStore
- New AppMetricResource in rest-resource
- ApplicationsMetrics is a simple domain class mapping
  collector response as I didn't want to pull any
  deps from collector into scdf for now.

- Fixes #1225
- Fixes #1260
- Fixes #1261

This is a 3 stage PR, one in each spring-cloud-dataflow, spring-cloud-deployer-local and spring-cloud-dataflow-ui. These can go in independently but nothing really works until all are merged and metrics collector is running.
https://github.com/spring-cloud/spring-cloud-dataflow-ui/pull/189
https://github.com/spring-cloud/spring-cloud-deployer-local/pull/38
```
# java -jar spring-cloud-dataflow-server-local/target/spring-cloud-dataflow-server-local-1.2.0.BUILD-SNAPSHOT.jar --spring.cloud.dataflow.metrics.collector.url=http://localhost:8080/collector/metrics

# java -jar target/metrics-collector-rabbit-1.0.0.BUILD-SNAPSHOT.jar --spring.cloud.stream.bindings.input.destination=metrics --spring.cloud.stream.default.contentType=application/json --security.basic.enabled=false --management.security.enabled=false

# shell
dataflow:>app register --name time120RS --type source --uri maven://org.springframework.cloud.stream.app:time-source-rabbit:1.2.0.BUILD-SNAPSHOT

dataflow:>app register --name log120RS --type sink --uri maven://org.springframework.cloud.stream.app:log-sink-rabbit:1.2.0.BUILD-SNAPSHOT

dataflow:>stream create --name foostream --definition "time120RS --spring.cloud.stream.bindings.applicationMetricsChannel.destination=metrics --spring.cloud.stream.metrics.properties=spring*|log120RS --spring.cloud.stream.bindings.applicationMetricsChannel.destination=metrics --spring.cloud.stream.metrics.properties=spring*"

dataflow:>stream deploy --name foostream --properties "deployer.*.count=2"
```

PR #1291, opened 2017-04-10
Add base metric support

### LINKED ISSUE #1290: Add metrics data to RuntimeAppsController's return value
The runtime apps endpoint `runtime/apps` already returns some metric information on a few platforms, for example cpu%,disk%,memory% on PCF.

Enrich the attributes of each AppInstance to contain metrics collected from the Metrics Collector for incoming and outgoing message rates.

markpollack:
Builds on PR https://github.com/spring-cloud/spring-cloud-dataflow/pull/1283

* Removes metric endpoint that took a map of deploymentId->GUID   (this may make the current PR a bit hard to review since there is addition/subtraction noise of a few classes, maybe squash first and diff to last commit on master)
* Metrics are now added to the existing `runtime apps` endpoint in `RuntimeAppsController`

When deploying apps, add the properties
* spring.cloud.stream.metrics.key as `<streamName>.<appLabel>.${spring.cloud.application.guid}`
* `spring.cloud.stream.metrics.properties to` `spring*`

Use AppInstanceStatus's attribute 'guid' to link up to metics collector stats

* Change name and formatting of incoming/outgoing rates to
** `metrics.integration.channel.input.receiveRate`
** `metrics.integration.channel.output.sendRate`
** %.2f formatting

Fixes issue #1290 and replaces PR #1290

You can run using the local deployer/server by first building the local deployer in the PR https://github.com/spring-cloud/spring-cloud-deployer-local/pull/39.

Then build this PR, and run the following commands.

```
in the metrics collector project, create the servers via `./mvnw clean install -PgenerateApps`.  Then cd to `apps/metrics-collector-rabbit` and run

# java -jar target/metrics-collector-rabbit-1.0.0.BUILD-SNAPSHOT.jar --spring.cloud.stream.bindings.input.destination=metrics --spring.cloud.stream.default.contentType=application/json --security.basic.enabled=false --management.security.enabled=false

# java -jar spring-cloud-dataflow-server-local/target/spring-cloud-dataflow-server-local-1.2.0.BUILD-SNAPSHOT.jar --spring.cloud.dataflow.metrics.collector.url=http://localhost:8080/collector/metrics

# shell
dataflow:>app register --name time120RS --type source --uri maven://org.springframework.cloud.stream.app:time-source-rabbit:1.2.0.BUILD-SNAPSHOT

dataflow:>app register --name log120RS --type sink --uri maven://org.springframework.cloud.stream.app:log-sink-rabbit:1.2.0.BUILD-SNAPSHOT

stream create --name foostream --definition "time120RS --spring.cloud.stream.bindings.applicationMetricsChannel.destination=metrics |log120RS --spring.cloud.stream.bindings.applicationMetricsChannel.destination=metrics "

*OR*

stream create --name foostream --definition "time120RS --spring.cloud.stream.bindings.applicationMetrics.destination=metrics |log120RS --spring.cloud.stream.bindings.applicationMetrics.destination=metrics "

if you are using an OOTB built after where https://github.com/spring-cloud/spring-cloud-stream/pull/905 has been merged.

dataflow:>stream deploy --name foostream --properties "deployer.*.count=2,app.*.security.basic.enabled=false,app.*.management.security.enabled=false"

dataflow:> runtime apps
```
Here is sample output for when the count=1

```dataflow:>runtime apps 
╔═════════════════════╤═════════╤══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════╗
║App Id / Instance Id │  Unit   │                                                No. of Instances / Attributes                                                 ║
║                     │ Status  │                                                                                                                              ║
╠═════════════════════╪═════════╪══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════╣
║foostream.log120RS   │deployed │                                                              1                                                               ║
╟┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┼┈┈┈┈┈┈┈┈┈┼┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈╢
║                     │         │                                         guid = 52731                                                                         ║
║                     │         │metrics.integration.channel.input.receiveRate = 1.00                                                                          ║
║                     │         │  metrics.integration.channel.output.sendRate = 0.00                                                                          ║
║                     │         │                                          pid = 17904                                                                         ║
║foostream.log120RS-0 │deployed │                                         port = 52731                                                                         ║
║                     │         │               stderr = /tmp/spring-cloud-dataflow-8493233097930055662/foostream-1491851478581/foostream.log120RS/stderr_0.log║
║                     │         │               stdout = /tmp/spring-cloud-dataflow-8493233097930055662/foostream-1491851478581/foostream.log120RS/stdout_0.log║
║                     │         │                                          url = http://127.0.1.1:52731                                                        ║
║                     │         │                       working.dir = /tmp/spring-cloud-dataflow-8493233097930055662/foostream-1491851478581/foostream.log120RS║
╟─────────────────────┼─────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╢
║foostream.time120RS  │deployed │                                                              1                                                               ║
╟┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┼┈┈┈┈┈┈┈┈┈┼┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈╢
║                     │         │                                         guid = 21432                                                                         ║
║                     │         │metrics.integration.channel.input.receiveRate = 0.00                                                                          ║
║                     │         │  metrics.integration.channel.output.sendRate = 1.00                                                                          ║
║                     │         │                                          pid = 17919                                                                         ║
║foostream.time120RS-0│deployed │                                         port = 21432                                                                         ║
║                     │         │              stderr = /tmp/spring-cloud-dataflow-8493233097930055662/foostream-1491851478746/foostream.time120RS/stderr_0.log║
║                     │         │              stdout = /tmp/spring-cloud-dataflow-8493233097930055662/foostream-1491851478746/foostream.time120RS/stdout_0.log║
║                     │         │                                          url = http://127.0.1.1:21432                                                        ║
║                     │         │                      working.dir = /tmp/spring-cloud-dataflow-8493233097930055662/foostream-1491851478746/foostream.time120RS║
╚═════════════════════╧═════════╧══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════╝

```


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/support/MetricStore.java
```diff
@@ -0,0 +1,81 @@
...
+		this.metricsProperties = metricsProperties;
+		restTemplate = new RestTemplate();
+	}
+
+	@HystrixCommand(fallbackMethod = "defaultMetrics")
+	public List<ApplicationsMetrics> getMetrics() {
+		List<ApplicationsMetrics> metrics = null;
+		if (StringUtils.hasText(metricsProperties.getCollector().getUrl())) {
+			try {
+				metrics = restTemplate.exchange(metricsProperties.getCollector().getUrl(), HttpMethod.GET, null,
```
ilayaperumalg: How are Cross Origin Resource (CORS) requests handled in this case (given the REST template is used to connect to external URL from the data flow server)?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/MetricsProperties.java
```diff
@@ -0,0 +1,52 @@
...
+	public Collector getCollector() {
+		return collector;
+	}
+
+	public void setCollector(Collector collector) {
+		this.collector = collector;
+	}
+
+	public static class Collector {
+		private String url;
```
ilayaperumalg: Can this be more specific type to denote the URL (just to have some URI scheme validation etc.,)?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/RuntimeAppsController.java
```diff
@@ -143,10 +146,39 @@ public RuntimeAppsController(StreamDefinitionRepository streamDefinitionReposito
...
+			}
+		}
+
+		for (AppStatus appStatus : statuses) {
+			Map<String, AppInstanceStatus> appInstanceStatusMap = appStatus.getInstances();
+			appInstanceStatusMap.forEach((k,appInstanceStatus) -> {
+				String trackingKey = appInstanceStatus.getAttributes().get("guid");
+				if (metricsInstanceMap.containsKey(trackingKey)) {
+					ApplicationsMetrics.Instance metricsAppInstance = metricsInstanceMap.get(trackingKey);
+					appInstanceStatus.getAttributes().put("metrics.integration.channel.input.receiveRate",
```
ilayaperumalg: The channel name can't always be `input`

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
```diff
@@ -250,6 +250,20 @@ private void deployStream(StreamDefinition stream, Map<String, String> streamDep
 			Resource appResource = registration.getResource();
 			Resource metadataResource = registration.getMetadataResource();
 
+			// add properties needed for metrics system
+			appDeployTimeProperties.put("spring.cloud.dataflow.stream.name", currentApp.getStreamName());
```
ilayaperumalg: Can we move these property names to a constant Class (something like `BindingPropertyKeys`)?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/support/MetricStore.java
jvalkeal: Mark can probably confirm but currently we don't care as collector have no authentication or authorization so we blindly trust what server asks. So this is simply between scdf server and collector.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
```diff
@@ -250,6 +250,20 @@ private void deployStream(StreamDefinition stream, Map<String, String> streamDep
...
+			appDeployTimeProperties.put("spring.cloud.dataflow.stream.app.label", currentApp.getName());
+			appDeployTimeProperties.put("spring.cloud.dataflow.stream.app.type", type.toString());
+			StringBuilder sb = new StringBuilder()
+					.append(currentApp.getStreamName()).append(".")
+					.append(currentApp.getName()).append(".")
+					.append("${spring.cloud.application.guid}");
```
ilayaperumalg: I could have missed the context here. But, can the property `${spring.cloud.application.guid}` always resolve to a specific value? I guess we probably need a value for `default` (some unique number)?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/MetricsProperties.java
jvalkeal: what u mean by more specific type?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
```diff
@@ -250,6 +250,20 @@ private void deployStream(StreamDefinition stream, Map<String, String> streamDep
...
+			appDeployTimeProperties.put("spring.cloud.stream.metrics.key", sb.toString());
```
ilayaperumalg: All the `spring.cloud.stream.metrics` properties' keys could probably be defined in `BindingPropertyKeys` Class. Perhaps, we could rename the Class `BindingPropertyKeys` to `SpringCloudStreamPropertyKeys` which would also lead a way to create `SpringCloudDataFlowPropertyKeys` for other property keys that start with prefix `spring.cloud.dataflow`.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
jvalkeal: new stuff, required to be set by deployer.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/MetricsProperties.java
ilayaperumalg: I was thinking about `java.net.URL`. If it is too aggressive, we could at least add some validation for the URL String before it is being used I guess.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
ilayaperumalg: ok, In that case, Can we can refer this from `AppDeployer` constant (once it is added).

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
jvalkeal: yeah true, I added some and mark added more. Might be good if we put those to constants somewhere.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/RuntimeAppsController.java
ilayaperumalg: Is the `input` supposed to be the `name` of the channel here? If so, the channel name varies based on Spring Cloud Stream configuration for the application.

jvalkeal:
Other than that, this doesn't need no changes to UI as things are directly shown within instance attributes.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/RuntimeAppsController.java
```diff
@@ -143,10 +146,39 @@ public RuntimeAppsController(StreamDefinitionRepository streamDefinitionReposito
...
+							String.format("%.2f", metricsAppInstance.getIncomingRate()));
+					appInstanceStatus.getAttributes().put("metrics.integration.channel.output.sendRate",
```
ilayaperumalg: Same here on the `output`.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/RuntimeAppsController.java
jvalkeal: this is just renamed from `incomingRate": 0.9993076938325146` what collector gives. collector itself have this naming issue as `input` is not universal. Issue is tracked there and we need to change things here when collector is made more generic.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/support/ApplicationsMetrics.java
```diff
@@ -0,0 +1,137 @@
...
+
+	public void setApplications(List<Application> applications) {
+		this.applications = applications;
+	}
+
+	public static class Application {
+
+		private String name;
+		private List<Instance> instances;
+		private double incomingRate;
```
ilayaperumalg: How does this `incomingRate` different from the `incomingRate` from the `Instance`. Same is for the `outgoingRate`.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/support/MetricStore.java
```diff
@@ -0,0 +1,81 @@
...
+import com.netflix.hystrix.contrib.javanica.annotation.HystrixCommand;
+
+/**
+ * Store implementation returning metrics info from a collector application.
+ * Implemented via hystrix command having a fallback to empty response.
+ *
+ * @author Janne Valkealahti
+ *
+ */
+@Component
```
ilayaperumalg: Why do we require `@Component` here as the `MetricStore` bean is created from `DataFlowControllerAutoConfiguration`?

jvalkeal: True, this is a leftover from early development.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/MetricsProperties.java
jvalkeal: We seem to use this format in other places like zuul having
```
		/**
		 * A full physical URL to map to the route. An alternative is to use a service ID
		 * and service discovery to find the physical address.
		 */
		private String url;
```

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/RuntimeAppsController.java
```diff
@@ -143,10 +146,39 @@ public RuntimeAppsController(StreamDefinitionRepository streamDefinitionReposito
...
+							String.format("%.2f", metricsAppInstance.getOutgoingRate()));
+				}
+			});
+		}
+
+	}
+
 	@RequestMapping("/{id}")
 	public AppStatusResource display(@PathVariable String id) {
```
ilayaperumalg: do you think it is worth adding a REST endpoint in `spring-cloud-dataflow-metrics-collector` that fetches the metrics for a given Stream and incorporate the status with the metrics for that stream here?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
jvalkeal: That is actually valid point indeed.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/RuntimeAppsController.java
jvalkeal: collector will have search/filter by stream names, so then it'd be doable.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
jvalkeal: created github.com/spring-cloud/spring-cloud-deployer/issues/209

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
jvalkeal: created github.com/spring-cloud/spring-cloud-dataflow/issues/1297 to address this.

jvalkeal:
Squashed/merged per https://github.com/spring-cloud/spring-cloud-dataflow/commit/088aed60633c13b80a76a820aa8ccccede14e9d1

PR #1296, opened 2017-04-11
Fix NPE at config command

### LINKED ISSUE #1295: NPE at dataflow config info when the secured server is not authenticated
When trying to run the command `dataflow config info` against the secured data flow server without the shell command authenticated, NPE is thrown.

```
server-unknown:>dataflow config info
Command failed java.lang.NullPointerException
```

ilayaperumalg:
- When security is enabled and `dataflow config info` is invoked without user credentials, the NPE is thrown
   - Handle exceptions inside `dataflow config info` command
   - Fix ConfigCommandTests class to avoid NPE
 - Fix DataFlowTemplate NPE on RootResource

Resolves #1295


trisberg:
Rebased and merged as commit 2e42b3215f4a64498be16c8035e4f5a7c932591d

PR #1299, opened 2017-04-11
Use constants for dataflow/stream property keys

### LINKED ISSUE #1297: Create constant class for stream and metric keys
`StreamDeploymentController` for metrics work in #1291 is adding keys like `spring.cloud.stream.metrics.key`. 

Comment from @ilayaperumalg related to that
```
All the spring.cloud.stream.metrics properties' keys could probably be defined in BindingPropertyKeys Class. Perhaps, we could rename the Class BindingPropertyKeys to SpringCloudStreamPropertyKeys which would also lead a way to create SpringCloudDataFlowPropertyKeys for other property keys that start with prefix spring.cloud.dataflow.
```
Creating ticket here is it's a bit complex issue to get handled in a PR itself.

ilayaperumalg:
- Define constant prefix/names wherever possible
 - Update `spring.cloud.dataflow.` prefix in other places in the codebase as well

Resolves #1297

### DIFF
--- a/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/DataFlowPropertyKeys.java
+++ b/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/DataFlowPropertyKeys.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.dataflow.core;
+
+/**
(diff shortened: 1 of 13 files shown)


spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/DataFlowPropertyKeys.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+package org.springframework.cloud.dataflow.core;
+
+/**
+ * Spring Cloud Data Flow property keys.
+ *
+ * @author Ilayaperumal Gopinathan
+ */
+public class DataFlowPropertyKeys {
```
markfisher: can the `PREFIX` vars here all be `private`?

ilayaperumalg: yeah, but we wanted to use this `PREFIX` in other places especially the `ConfigurationProperties` classes.

markfisher: but I believe `STREAM_APP_PREFIX` and `STREAM_PREFIX` can be private

ilayaperumalg: yeah, sure.

jvalkeal:
squashed/merged per https://github.com/spring-cloud/spring-cloud-dataflow/commit/9983bc7c2dc11e77b7320f7c37a59eaf5b542682

PR #1303, opened 2017-04-11
Change the EXECUTABLE_DSL_JOIN_CHAR to "-"

### LINKED ISSUE #1293: Remove underscore delimiter from the generated task definitions
If we want to support composed tasks on Kubernetes we need to make the EXECUTABLE_DSL_JOIN_CHAR configurable and also make the leading use of it configurable as well.

On Kubernetes we can't run pods named like `_foobar_foo` since underscores aren't allowed and the pod name must start with an alphanumeric character

trisberg:
- also remove the use of it as a prefix

Resolves #1293


cppwfs:
TaskParserTests.executableDsl tests need to be updated.

cppwfs:
LGTM except tests failed.  Updated the tests and those changes are in this branch https://github.com/cppwfs/spring-cloud-dataflow/tree/trisberg-gh1293-composed-task-delimiter
If you agree with them I can merge.

cppwfs:
LGTM

aclement:
Unfortunately this clashes with my PR that changes this and has some other bits in that I've been working on... I also centralized creation of the names so it isn't in two places....

PR #1304, opened 2017-04-11
Tools polish

aclement:
Key things in here:

- changed composed task 'executable dsl' task names from _taskname_appname to taskname-appname
- altered tests to cope with the change above
- altered graph to text so that it will quote any graph values that may need it when being put into the dsl: exit statuses for transitions and property values
- removed some tests that deal with move advanced DSL constructs that we don't care about yet
- altered text to graph to remove quotes so the user doesn't have to worry about them when editing graph properties
- added validation to ensure apps targeted by transitions are also unique across the composed task def


trisberg:
LGTM, can merge after resolving conflicts from earlier merge of https://github.com/spring-cloud/spring-cloud-dataflow/pull/1303

cppwfs:
Some formatting hiccups, but overall LGTM.  The conflicts are because of https://github.com/spring-cloud/spring-cloud-dataflow/pull/1303.  But that really comes down to just some duplicate work on both stories.

trisberg:
Merged as commit bc5a1881b4f956f35bd84b7cd839ba8732757ae8

PR #1305, opened 2017-04-11
Store only composed task dsl

### LINKED ISSUE #1302: CTR app is applied to CT definition at launch time
Currently the CTR App is applied to the task Composed Task definition at save time.  
This should be delayed till CT is launched.

cppwfs:
Add CTR app to definition at runtime

resolves #1302


trisberg:
My concern with this change is that the UI doesn't have anything to go by trying to determine what is a composed task and what is a regular task. It currently looks for `--graph`

trisberg:
We are adding the `isComposed` flag in #1313

trisberg:
LGTM, will be amended by https://github.com/spring-cloud/spring-cloud-dataflow/pull/1318

Rebased and merged as commit 93f43419c96bb46f465b6ce44e3e4a0fb19644dd

Issue #1285, opened 2017-04-10
Simplest secured setup is unusable


ericbottard:
Simply setting up security with `--security.basic.enabled=true` (and thus creating a default boot user `user`) is in effect unusable now: the created user has no role (not even `ROLE_VIEW`) and thus no actions can be undertaken

PR #1308, opened 2017-04-12
gh-1285 Fix: Simplest secured setup is unusable

ghillert:
* Carefully determine what a "default-user" is
* Ensure that a default user is only setup if neither Ldap- nor file-based authentication are setup
* Add new conditional annotation `OnDefaultBootUserAuthenticationEnabled`

resolves https://github.com/spring-cloud/spring-cloud-dataflow/issues/1285

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/security/support/CoreSecurityRoles.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/security/support/CoreSecurityRoles.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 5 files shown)


ilayaperumalg:
LGTM, verified using the default and non-default user as well. Merging.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/security/support/CoreSecurityRoles.java
```diff
@@ -0,0 +1,76 @@
...
+
+	public String getKey() {
+		return key;
+	}
+
+	public String getName() {
+		return name;
+	}
+
+	public static CoreSecurityRoles fromKey(String role) {
```
ilayaperumalg: Just noticed that this method is not being used. May be we can remove it?

ilayaperumalg: It doesn't harm to keep it. going ahead with merge. we can remove this later if not needed.

ilayaperumalg:
Merged as 34aaa8aa239307ae98f0a5ddda65613caa755848

PR #1312, opened 2017-04-12
Fix MetricStore for hal response

### LINKED ISSUE #1311: Fix metric collector input
Collector app now returns hal response and need to adapt to that.

### LINKED ISSUE #1317: Refactor metrics mapping for metrics instance attributes

jvalkeal:
Fix MetricStore for hal response
    
    - Change RestTemplate to understand that response
      from collector is hal json and strip needed
      info out from it to keep output from store a same.
    - Change MetricStore and its domain model to support
      new collector format.
    - Fix RuntimeAppsController to use new metrics list/map
      from collector structure now that incomingRate and
      outgoingRate are gone.
    - Fixes #1311
    - Fixes #1317


jvalkeal:
rebased and added fixes for #1317 to fix `runtime apps`.

markpollack:
Aside from the jackson mapping configuration, this all looks good to me.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/support/MetricStore.java
```diff
@@ -50,17 +57,24 @@
 	 */
 	public MetricStore(MetricsProperties metricsProperties) {
 		this.metricsProperties = metricsProperties;
-		restTemplate = new RestTemplate();
+		ObjectMapper mapper = new ObjectMapper();
```
viniciusccarvalho: I thought the hateoas-starter from boot already bootstrapped a `_halObjectMapper` bean in the context. Couldn't we reuse that?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/RuntimeAppsController.java
```diff
@@ -169,10 +171,18 @@ private void enrichWithMetrics(List<AppStatus> statuses) {
...
-					appInstanceStatus.getAttributes().put("metrics.integration.channel.input.receiveRate",
-							String.format("%.2f", metricsAppInstance.getIncomingRate()));
-					appInstanceStatus.getAttributes().put("metrics.integration.channel.output.sendRate",
-							String.format("%.2f", metricsAppInstance.getOutgoingRate()));
+					List<Metrics> metrics = metricsAppInstance.getMetrics();
+					if (metrics != null) {
+						for (Metrics m : metrics) {
+							if (ObjectUtils.nullSafeEquals("integration.channel.input.send.mean", m.getName())) {
+								appInstanceStatus.getAttributes().put("metrics.integration.channel.input.receiveRate",
+										String.format("%.2f", m.getValue()));
```
markpollack: maybe we can add `msgs/sec` here to provide units.  We have `%` added in the case of cloud foundry for the mem/cpu/disk utilization.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/support/MetricStore.java
markpollack: I will open up another issue for this, need to merge this now since current master isn't working end to end.

markpollack:
Merged as 29f6720d4c33d53984f0438810f94dae142449f3

PR #1331, opened 2017-04-14
Set dataFlowUri for composed task definition

### LINKED ISSUE #1323: When launching composed tasks we should automatically provide data-flow-uri
When launching a composed task it is necessary to specify the property `data-flow-uri` if the server is using anything but `http://localhost:9393`.

We could set this automatically based on some config property on the dataflow server. I suggest we could set `spring.cloud.dataflow.server.uri` when starting the server and if that is set provide the `--data-flow-uri` with the value when launching the composed task.

ilayaperumalg:
- Infer data flow server URI and update as default property
 - Inject this property into DefaultTaskService to set `--dataFlowUri`
 - Add test

Resolves #1323

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DefaultEnvironmentPostProcessor.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DefaultEnvironmentPostProcessor.java
@@ -16,8 +16,6 @@
 
 package org.springframework.cloud.dataflow.server.config;
 
-import java.net.InetAddress;
-import java.net.UnknownHostException;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Properties;
@@ -29,7 +27,6 @@ import org.apache.commons.logging.LogFactory;
 import org.springframework.beans.factory.config.YamlPropertiesFactoryBean;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.env.EnvironmentPostProcessor;
-import org.springframework.cloud.dataflow.core.DataFlowPropertyKeys;
 import org.springframework.core.Ordered;
 import org.springframework.core.env.ConfigurableEnvironment;
 import org.springframework.core.env.MapPropertySource;
@@ -37,7 +34,6 @@ import org.springframework.core.env.MutablePropertySources;
 import org.springframework.core.env.PropertySource;
 import org.springframework.core.io.ClassPathResource;
 import org.springframework.core.io.Resource;
-import org.springframework.util.StringUtils;
 
 /**
  * Contributes the values from {@code META-INF/dataflow-server-defaults.yml} and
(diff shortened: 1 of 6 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
```diff
@@ -285,8 +291,9 @@ private void saveStandardTaskDefinition(TaskDefinition taskDefinition) {
 	}
 
 	private String createComposedTaskDefinition(String graph) {
-		return String.format("%s --graph=\"%s\"",
-				taskConfigurationProperties.getComposedTaskRunnerName(), graph);
+		String composedTaskDefinition = StringUtils.hasText(this.dataFlowUri) ? "%s --graph=\"%s\" --dataFlowUri=%s" : "%s --graph=\"%s\"";
```
trisberg: I don't think we should set this here. It would be better to check the appDeploymentProperties and commandLineArgs for this key and if not specified add it

ilayaperumalg: yeah, right. that's a good idea.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DefaultEnvironmentPostProcessor.java
```diff
@@ -58,7 +64,17 @@ public void postProcessEnvironment(ConfigurableEnvironment environment, SpringAp
...
+		String hostName = null;
+		try {
+			hostName = InetAddress.getLocalHost().getHostAddress();
+		}
+		catch (UnknownHostException e) {
+			// ignore
+		}
+		String dataflowServerPort = environment.getProperty("server.port");
+		if (StringUtils.hasText(hostName) && StringUtils.hasText(dataflowServerPort)) {
+			defaults.put(DataFlowPropertyKeys.PREFIX + "server.uri", String.format("http://%s:%s", hostName, dataflowServerPort));
```
trisberg: I'm not sure this would be useful in a distributed environment - we need to provide the routable host/ip-address and most likely the server doesn't have access to this without some explicit configuration. I think we should check for a `spring.cloud.dataflow.server.uri` property and if it is available use it, if it's not set just rely on the CTR to use its default.

ilayaperumalg: Do you mean, user needs to set `spring.cloud.dataflow.server.uri` explicitly when starting the server in order for this to be effective?

trisberg: yes

trisberg: for k8s I'm planning on setting env var SPRING_CLOUD_DATAFLOW_SERVER_URI=$SCDF_SERVICE_HOST:$SCDF_SERVICE_PORT

trisberg: also, we need to check for overrides during launch, only set this automaticaly if the user hasn't provided this as part of the launch

ilayaperumalg: I thought you were concerned about setting this explicitly here, no?

ilayaperumalg: I thought we don't want to set this value at all and rely only based on whether user has explicitly set this, based on your comment:
>  I think we should check for a spring.cloud.dataflow.server.uri property and if it is available use it, if it's not set just rely on the CTR to use its default.

 but, we can ahead and set it in the environment post processor as well.

trisberg:
LGTM, making a few tweaks in `updateDataFlowUriIfNeeded` method

rebasing and merging ...

trisberg:
Squashed and merged as commit 41426532c49de9cd46f47579e2d80b62e57ad7ee

PR #1334, opened 2017-04-14
Make app properties for metrics configurable.

### LINKED ISSUE #1298: Make metrics properties key configurable
`StreamDeploymentController` forces adding `spring.cloud.stream.metrics.properties=spring*` for all apps.

As a user I'd like this to be configurable so that if in my app I want to expose something else than `spring*` keys, I could do it.

Also need to think how to tell user that keys prefixed with `spring` are needed for metrics to work.

markpollack:
Added values to  StreamPropertyKeys and set default values if not already present in application property request.  Added after all expansion/wildcarding taking into account.

Fixes #1298
Tested using
```
stream create --name foostream --definition "time120RS |log120RS"

stream deploy --name foostream --properties "deployer.*.count=2,app.*.security.basic.enabled=false,app.*.management.security.enabled=false,app.*.spring.cloud.stream.bindings.applicationMetrics.destination=metrics"
```

with local deployer and observed default values.

Tested by overrding via

```
stream create --name foostream --definition "time120RS --spring.cloud.stream.metrics.properties=spring* |log120RS"
```

and also
```
stream deploy --name foostream --properties "deployer.*.count=2,app.*.security.basic.enabled=false,app.*.management.security.enabled=false,app.*.spring.cloud.stream.bindings.applicationMetrics.destination=metrics,app.*.spring.cloud.stream.metrics.properties=spring*"

```

The deployed apps had the overriden property values.

### DIFF
--- a/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/StreamPropertyKeys.java
+++ b/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/StreamPropertyKeys.java
@@ -45,4 +45,11 @@ public class StreamPropertyKeys {
 	 */
 	public static final String METRICS_PROPERTIES = METRICS_PREFIX + "properties";
 
+	/**
+	 * This is the spring boot property key that Spring Cloud Stream uses to filter
+	 * the metrics to import when the specific Spring Cloud Stream "applicaiton" trigger is
+	 * fired for metrics export.
+	 */
+	public static final String METRICS_TRIGGER_INCLUDES = "spring.metrics.export.triggers.application.includes";
+
 }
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/StreamPropertyKeys.java
```diff
@@ -45,4 +45,11 @@
 	 */
 	public static final String METRICS_PROPERTIES = METRICS_PREFIX + "properties";
 
+	/**
+	 * This is the spring boot property key that Spring Cloud Stream uses to filter
+	 * the metrics to import when the specific Spring Cloud Stream "applicaiton" trigger is
+	 * fired for metrics export.
+	 */
+	public static final String METRICS_TRIGGER_INCLUDES = "spring.metrics.export.triggers.application.includes";
```
ilayaperumalg: This property could probably be moved to `StreamDeploymentController` as this doesn't belong to `StreamPropertyKeys`.

markpollack: ok, please fix it on merge.

ilayaperumalg:
Fixed the above mentioned comment, rebased and merged as ab9447249708a49cdd899d9c3b0bacb8ea4fb3cf

PR #1349, opened 2017-04-17
Fix composed task's job step executions

### LINKED ISSUE #1241: Step Execution Count incorrect when same job is executed multiple times
When executing a batch job (with same job name) multiple times the step execution count for   for all instances are the same.  i.e. if a job has 3 steps, the step execution count for the first run will be 3.  The next time its run the step count for the current run and previous run will be 6.

Steps to reproduce.  Execute a batch job(task) and then rerun it.  Then go to the shell and execute a job execution list.

ilayaperumalg:
- When the step executions are retrieved for the composed job, since we have a new step name for the steps inside the batch jobs of the composed task everytime a step is executed for the batch job, the `getStepExecutions(Long jobExecutionId)` returns all the step executions that are distinct. This causes the step executions of all the batch jobs executed under the name of the composed task.
      - To fix this, look for non-null step executions for each step execution added to the job execution

Resolves #1241


cppwfs:
LGTM

PR #1353, opened 2017-04-17
Adding docs for restarting composed tasks

cppwfs:
* Also added docs on  stopping tasks and the notes on its behavior when restarting a stopped CT
* Added note that bulk load for composed tasks is not supported

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/tasks.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/tasks.adoc
@@ -480,9 +480,9 @@ via the:
 clicking the stop button by the job execution that needs to be stopped.
 
 The composed task run will be stopped
-when the currently running child task completes.  The child task that
-was running at the time that the composed task was stopped will be marked as
-`STOPPED` as well as the composed task run.
+when the currently running child task completes.  The step associated with the
+child task that was running at the time that the composed task was stopped will
+be marked as `STOPPED` as well as the composed task job execution.
 
 ==== Restarting a Composed Task
 In cases where a composed task fails during execution and the status of the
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-docs/src/main/asciidoc/tasks.adoc
```diff
@@ -471,6 +471,32 @@ dataflow:>task list
 ╠══════════════════════════╪═══════════════════════════════════════════════════════════════
 ╚══════════════════════════╧═══════════════════════════════════════════════════════════════
 ```
+==== Stopping a Composed Task
+In cases where a composed task execution needs to be stopped.  This can be done
+via the:
+
+* RESTful API
+* Spring Cloud Data Flow Dashboard by selecting the Job's tab and then
+clicking the stop button by the job execution that needs to be stopped.
```
mminella: Does the shell not support this?

spring-cloud-dataflow-docs/src/main/asciidoc/tasks.adoc
```diff
@@ -471,6 +471,32 @@ dataflow:>task list
...
+
+The composed task run will be stopped
+when the currently running child task completes.  The child task that
```
mminella: The task itself won't be marked as stopped (since that doesn't exist).  The step that launched the task will be.

spring-cloud-dataflow-docs/src/main/asciidoc/tasks.adoc
cppwfs: Not at this time.  Added story https://github.com/spring-cloud/spring-cloud-dataflow/issues/1362

PR #1354, opened 2017-04-17
Ensure consistency with quoted wildcard handling for composed tasks

aclement:
Fixes https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/197

With this change if you quote wildcards in the DSL box they will get unquoted when the graph is built and if you leave them unquoted in the graph building, they will get quoted when the DSL is built - as expected by composed-task-runner. This will need revisiting when we think about how to handle exit codes (vs exit statuses).


cppwfs:
LGTM

Issue #1355, opened 2017-04-17
Shell: `dataflow config info` results in empty response


ghillert:
When security is enabled but no roles are assigned then `dataflow config info` results in
an empty response.

![image](https://cloud.githubusercontent.com/assets/685363/25102185/fdb12a04-2352-11e7-9a80-89f35d3da717.png)

See also: https://github.com/spring-projects/spring-shell/issues/120

Related to: https://github.com/spring-cloud/spring-cloud-dataflow-server-cloudfoundry/issues/295

PR #1356, opened 2017-04-17
gh-1355 Fix Shell: `dataflow config info` results in empty response

ghillert:
* Only show Roles when authorization is enabled
* If no roles are assigned, then add value `No roles assigned`

resolves https://github.com/spring-cloud/spring-cloud-dataflow/issues/1355

### DIFF
--- a/spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/ConfigCommands.java
+++ b/spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/ConfigCommands.java
@@ -258,12 +258,8 @@ public class ConfigCommands implements CommandMarker,
 		modelBuilder.addRow().addValue("Security").addValue(securityInfo);
 		rowIndex++;
 
-		if (securityInfo.isAuthenticated() && securityInfo.isAuthorizationEnabled()) {
-			final List<String> roles = securityInfo.getRoles();
-			if (roles.isEmpty()) {
-				roles.add("No roles assigned");
-			}
-			modelBuilder.addRow().addValue("Roles").addValue(roles);
+		if (securityInfo.isAuthorizationEnabled()) {
+			modelBuilder.addRow().addValue("Roles").addValue(securityInfo.getRoles());
 			rowsWithThinSeparators.add(rowIndex++);
 		}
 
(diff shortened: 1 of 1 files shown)


spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/ConfigCommands.java
```diff
@@ -257,8 +257,13 @@ public String target(
 		SecurityInfo securityInfo = about.getSecurityInfo();
 		modelBuilder.addRow().addValue("Security").addValue(securityInfo);
 		rowIndex++;
-		if (securityInfo.isAuthenticated()) {
-			modelBuilder.addRow().addValue("Roles").addValue(securityInfo.getRoles());
+
+		if (securityInfo.isAuthenticated() && securityInfo.isAuthorizationEnabled()) {
```
ilayaperumalg: I am not sure whether this part of the code will ever be invoked without at least a `single` role as the user requires at least `VIEW` role to access with `/about` endpoint.

trisberg: Agree, the only way to get this is if  `securityInfo.isAuthorizationEnabled()` is false

trisberg: I think it should be: https://gist.github.com/trisberg/cd5e9f1930b9446541d86d0ac57e7092

ghillert: @ilayaperumalg @trisberg Not quite correct. Our OAuth setup (currently) does not require roles and for that setup authorization is disabled / ignored`.

trisberg: If I use OAuth, I still don't see the "No roles assigned" message since you test for `(securityInfo.isAuthenticated() && securityInfo.isAuthorizationEnabled())`. The Roles part of the info display is just eliminated. With my change (https://gist.github.com/trisberg/cd5e9f1930b9446541d86d0ac57e7092) I do see that message.

ghillert: @trisberg Sorry for the confusion - You're right. I am just paranoid with every security related and rather be super-explicit - but you're right ... if authorization is enabled you should always have at least a role associated.

trisberg:
LGTM, merged as commit 32e2c7d95217706c728602739c9f547b0ecfc106

PR #1358, opened 2017-04-18
Adds missing throws javadocs to the spring cloud dataflow server core…

### LINKED ISSUE #1343: NoSuchJobExecutionException is declared on many controller methods but never thrown
Look for instantiation of NoSuchJobExecutionException and remove from all throws declarations where it is not relevant.

cppwfs:
… project

resolves #1343

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/TaskJobService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/TaskJobService.java
@@ -45,6 +45,8 @@ public interface TaskJobService {
 	 * the data with a task id.
 	 * @param pageable enumerates the data to be returned.
 	 * @return List containing {@link TaskJobExecution}s.
+	 * @throws NoSuchJobExecutionException in the event that a job execution id specified
+	 * is not present when looking up stepExecutions for the result.
 	 */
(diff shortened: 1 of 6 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/TaskJobService.java
```diff
@@ -126,8 +137,8 @@ JobInstanceExecutions getJobInstance(long id)
 	 * @see org.springframework.batch.admin.service.JobService#stop(Long)
 	 *
 	 * @param jobExecutionId The id of the {@link JobExecution} to stop
-	 * @throws NoSuchJobExecutionException
-	 * @throws JobExecutionNotRunningException
+	 * @throws NoSuchJobExecutionException if no job execution exists for the jobExecutionId.
+	 * @throws JobExecutionNotRunningException if the job execution specified by jobExecutionId can not be restarted.
```
ilayaperumalg: Instead of saying `can not be restarted`, we can say the similar way we did above: `if a stop is requested on a job that is not running.` Will fix when merging.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Merged as b0eaa778b0e80cdb03d14b6cbe79d15d22b08cb3

PR #1361, opened 2017-04-18
gh-1360 REST: Ensure actuator endpoints are secured with OAuth

ghillert:
* Add `@EnableWebSecurity` to `OAuthSecurityConfiguration`
* Add integration tests

resolves https://github.com/spring-cloud/spring-cloud-dataflow/issues/1360

### DIFF
--- a/spring-cloud-starter-dataflow-server-local/src/test/resources/org/springframework/cloud/dataflow/server/local/security/oauthConfig.yml
+++ b/spring-cloud-starter-dataflow-server-local/src/test/resources/org/springframework/cloud/dataflow/server/local/security/oauthConfig.yml
@@ -1,6 +1,3 @@
-management:
-  security:
-    enabled: true
 security:
   oauth2:
     client:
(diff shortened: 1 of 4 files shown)


spring-cloud-starter-dataflow-server-local/src/test/resources/org/springframework/cloud/dataflow/server/local/security/oauthConfig.yml
```diff
@@ -1,6 +1,3 @@
-management:
-  security:
-    enabled: true
```
ilayaperumalg: I see that this is removed as it requires `ADMIN` role (from spring-boot) to access the management actuator endpoints. I understand this is only in tests, but, what happens when `management.security.enabled=true` at the server level?

trisberg:
Not sure I see where the original issue was, could you explain what you are trying to fix?

ghillert:
@trisberg One requirement ( @sabbyanandan ) by an earlier issue was that OAuth2 shall not require the presence of `security.basic.enabled`. It turns out that we still had some Boot magic going on that was affected by this and it turned out that all of a sudden the management endpoints were unprotected. That's why I added the `@EnableWebSecurity` to have all security being driven by `OAuthSecurityConfiguration`.

spring-cloud-starter-dataflow-server-local/src/test/resources/org/springframework/cloud/dataflow/server/local/security/oauthConfig.yml
ghillert: Interesting point I will add a test suite with that enabled to have that test-covered. I will add that once I brought the kids to school.

trisberg: If `management.security.enabled=true` you'll get an error from Boot saying:
```
"Access is denied. User must have one of the these roles: ADMIN"
```
when accessing `/management` endpoints

trisberg:
Rebased and merged as commit 092a00240dfe27737177ef4f5024491eb2b11877

PR #1370, opened 2017-04-18
Changes collector URL property to URI

### LINKED ISSUE #1367: Remove path from metrics collector URL properties
Currently we are setting the property `spring.cloud.dataflow.metrics.collector.url` to have the full server path including `/collector/metrics/streams` 
We should just use the server and port and leave the resolution of the right path to the client part. 
We can break in a two fold change, first we set it on the client, but later we should upgrade to Traverson to find the proper link and follow it.

viniciusccarvalho:
Fixes #1367

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/configuration.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/configuration.adoc
@@ -801,10 +801,20 @@ Data Flow Metrics Collector, and the Data Flow server is shown below.
 image::{dataflow-asciidoc}/images/dataflow-metrics-arch.png[Spring Cloud Data Flow Metrics Architecture , scaledwidth="100%"]
 
 As with the App Starters, there is a Spring Boot uber jar artifact of the
-Metrics Collector for Kafka 0.9, Kafka 0.10, and RabbitMQ.  You can find
-more information on building and running the Metrics Collector on its
+
+Metrics Collector for all of the supported binders.  You can find
+more information on building and running the Metrics Collector on its 
+
 https://github.com/spring-cloud/spring-cloud-dataflow-metrics-collector[project
 page].
+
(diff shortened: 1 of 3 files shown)


jvalkeal:
property is `URL`, not `URI`. `URI` can also be `URN` which we never use. And besides, boot properties are usually named as `url`. Even in `RestTemplate` methods parameters are named `String url`.

spring-cloud-dataflow-docs/src/main/asciidoc/configuration.adoc
```diff
@@ -801,11 +801,21 @@ Data Flow Metrics Collector, and the Data Flow server is shown below.
...
-Metrics Collector for Kafka 0.9, Kafka 0.10, and RabbitMQ.  You can find
-more information on building and running the Metrics Collector on its
+
+Metrics Collector for all of the supported binders.  You can find
+more information on building and running the Metrics Collector on its 
+
 https://github.com/spring-cloud/spring-cloud-dataflow-metrics-collector[project
 page].
 
+The dataflow server now accepts an optional property `spring.cloud.dataflow.metrics.collector.uri`, this property should point to the URI of your deployed
```
markpollack: introduce `\n` at ~80 cols in future.

viniciusccarvalho:
@jvalkeal We discussed the naming of URI vs URL yesterday on the channel. @markpollack and @trisberg were advising. Since we already use URI for config server, and an URI can be an URL, I think it would be for the best for us to switch to URI.

The original issue however is not due naming, is to the fact that we were using a full path `/collector/metrics/streams` and we should let this be done at the client tier. We have hypermedia support and link traversal should be possible, and once we get discoverability in the server, we may not even need this property.

markpollack:
Squashed and merged as 70e9b6f1b21a6b7fae88b9861561960b73f67e02

PR #1373, opened 2017-04-19
Add support for calling secured message collector endpoint

### LINKED ISSUE #1372: Add basic security for calling metrics endpoint

markpollack:
Fixes #1372


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/MetricsProperties.java
```diff
@@ -42,12 +43,42 @@ public void setCollector(Collector collector) {
 	public static class Collector {
 		private String uri;
 
+		private String userName;
```
trisberg: this should be `username` to be consistent with usage elsewere

trisberg:
Merged as commit c1977b5057d056c54c592b2f44b9951c8a9aa31c

Made some modifications via commit 45c5bd605032b7fb269065d4f5349b65c1ca9217

PR #1383, opened 2017-04-22
Add Pageable to applicable controllers to prevent issue in PagedResourcesAssembler

### LINKED ISSUE #1382: Update to Spring Boot 1.5.4
The latest versions of `spring-data-commons` (> 1.13.3.RELEASE) which include a fix for [DATACMNS-1042](https://jira.spring.io/browse/DATACMNS-1042) (Self link created with pagination request parameters) in https://github.com/spring-projects/spring-data-commons/commit/a8ff4ea1429f8a4c42df2030d7ab9ccc0d742b12 will break certain controllers using the `PagedResourcesAssembler` in `spring-cloud-dataflow-server-core`.

For example, the following exception would be thrown (using the local server for testing) when requesting http://localhost:9393/apps:

```
2017-04-22 22:24:45.941 ERROR 2253 --- [nio-9393-exec-8] o.s.c.d.s.c.RestControllerAdvice         : Caught exception while handling a request

java.lang.IllegalArgumentException: Page size must not be less than one!
	at org.springframework.data.domain.AbstractPageRequest.<init>(AbstractPageRequest.java:47) ~[spring-data-commons-1.13.3.RELEASE.jar:na]
	at org.springframework.data.domain.PageRequest.<init>(PageRequest.java:63) ~[spring-data-commons-1.13.3.RELEASE.jar:na]
	at org.springframework.data.web.PagedResourcesAssembler.addPaginationLinks(PagedResourcesAssembler.java:222) ~[spring-data-commons-1.13.3.RELEASE.jar:na]
	at org.springframework.data.web.PagedResourcesAssembler.createResource(PagedResourcesAssembler.java:205) ~[spring-data-commons-1.13.3.RELEASE.jar:na]
	at org.springframework.data.web.PagedResourcesAssembler.toResource(PagedResourcesAssembler.java:115) ~[spring-data-commons-1.13.3.RELEASE.jar:na]
	at org.springframework.cloud.dataflow.server.controller.AppRegistryController.list(AppRegistryController.java:...) ~[classes/:na]
...
```

This issue arises when upgrading to Spring Boot 1.5.3 which includes `spring-data-commons:1.13.3.RELEASE`. While not an issue for the current release it will be an issue in future.

donovanmuller:
Fixes #1382

If Spring Boot were to be updated to 1.5.3, it would bring in `spring-data-commons:1.13.3.RELEASE` which includes a fix for [DATACMNS-1042](https://jira.spring.io/browse/DATACMNS-1042) (Self link created with pagination request parameters) in https://github.com/spring-projects/spring-data-commons/commit/a8ff4ea1429f8a4c42df2030d7ab9ccc0d742b12. This commit breaks certain controllers using the `PagedResourcesAssembler` in `spring-cloud-dataflow-server-core`.

For example, the following exception will be thrown (using the local server for testing) when requesting http://localhost:9393/apps:

```
2017-04-22 22:24:45.941 ERROR 2253 --- [nio-9393-exec-8] o.s.c.d.s.c.RestControllerAdvice         : Caught exception while handling a request

java.lang.IllegalArgumentException: Page size must not be less than one!
	at org.springframework.data.domain.AbstractPageRequest.<init>(AbstractPageRequest.java:47) ~[spring-data-commons-1.13.3.RELEASE.jar:na]
	at org.springframework.data.domain.PageRequest.<init>(PageRequest.java:63) ~[spring-data-commons-1.13.3.RELEASE.jar:na]
	at org.springframework.data.web.PagedResourcesAssembler.addPaginationLinks(PagedResourcesAssembler.java:222) ~[spring-data-commons-1.13.3.RELEASE.jar:na]
	at org.springframework.data.web.PagedResourcesAssembler.createResource(PagedResourcesAssembler.java:205) ~[spring-data-commons-1.13.3.RELEASE.jar:na]
	at org.springframework.data.web.PagedResourcesAssembler.toResource(PagedResourcesAssembler.java:115) ~[spring-data-commons-1.13.3.RELEASE.jar:na]
	at org.springframework.cloud.dataflow.server.controller.AppRegistryController.list(AppRegistryController.java:...) ~[classes/:na]
...
```
The reason this happens is because a [new `PageImpl`](https://github.com/spring-cloud/spring-cloud-dataflow/pull/1383/files#diff-f920349a50e4ef571d20402f82561fffL119), with default values of `0` for page and page size, is being shipped to the `PagedResourcesAssembler`. This (new) [line](https://github.com/spring-projects/spring-data-commons/commit/a8ff4ea1429f8a4c42df2030d7ab9ccc0d742b12#diff-2c474de1858ef16219b8a3f436402199R222) in `PagedResourcesAssembler` creates a new `PageRequest` with the passed in `Pageable` and is where the [`size < 1`](https://github.com/spring-projects/spring-data-commons/blob/1.13.x/src/main/java/org/springframework/data/domain/AbstractPageRequest.java#L46) check blows up because `size == 0`.

So we need to add a `Pageable` param (with sane defaults provided by `PageableHandlerMethodArgumentResolver`) to the affected controller methods and use that when building the `PageImpl`.

Also changed the default page size to `2000` on the REST templates to match [DEFAULT_MAX_PAGE_SIZE](https://github.com/spring-projects/spring-data-commons/blob/1.13.x/src/main/java/org/springframework/data/web/PageableHandlerMethodArgumentResolver.java#L52), since `10000` would be [converted](https://github.com/spring-projects/spring-data-commons/blob/1.13.x/src/main/java/org/springframework/data/web/PageableHandlerMethodArgumentResolver.java#L251) to `2000` anyway.

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppRegistryController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppRegistryController.java
@@ -30,7 +30,6 @@ import java.util.concurrent.ForkJoinPool;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
 import org.springframework.boot.configurationmetadata.ConfigurationMetadataProperty;
 import org.springframework.cloud.dataflow.configuration.metadata.ApplicationConfigurationMetadataResolver;
 import org.springframework.cloud.dataflow.core.ApplicationType;
@@ -44,6 +43,7 @@ import org.springframework.core.io.ByteArrayResource;
 import org.springframework.core.io.DefaultResourceLoader;
 import org.springframework.core.io.ResourceLoader;
 import org.springframework.data.domain.PageImpl;
+import org.springframework.data.domain.Pageable;
 import org.springframework.data.web.PagedResourcesAssembler;
 import org.springframework.hateoas.ExposesResourceFor;
 import org.springframework.hateoas.PagedResources;
@@ -104,6 +104,7 @@ public class AppRegistryController implements ResourceLoaderAware {
 	@RequestMapping(method = RequestMethod.GET)
 	@ResponseStatus(HttpStatus.OK)
 	public PagedResources<? extends AppRegistrationResource> list(
+			Pageable pageable,
 			PagedResourcesAssembler<AppRegistration> pagedResourcesAssembler,
 			@RequestParam(value = "type", required = false) ApplicationType type,
 			@RequestParam(value = "detailed", defaultValue = "false") boolean detailed) {
@@ -116,7 +117,8 @@ public class AppRegistryController implements ResourceLoaderAware {
 			}
 		}
 		Collections.sort(list);
-		return pagedResourcesAssembler.toResource(new PageImpl<>(list), assembler);
+		return pagedResourcesAssembler
+				.toResource(new PageImpl<>(list, pageable, appRegistry.findAll().size()), assembler);
 	}
 
 	/**
@@ -196,6 +198,7 @@ public class AppRegistryController implements ResourceLoaderAware {
 	@RequestMapping(method = RequestMethod.POST)
 	@ResponseStatus(HttpStatus.CREATED)
 	public PagedResources<? extends AppRegistrationResource> registerAll(
+			Pageable pageable,
 			PagedResourcesAssembler<AppRegistration> pagedResourcesAssembler,
 			@RequestParam(value = "uri", required = false) String uri,
 			@RequestParam(value = "apps", required = false) Properties apps,
@@ -212,7 +215,9 @@ public class AppRegistryController implements ResourceLoaderAware {
 		}
 		Collections.sort(registrations);
 		prefetchMetadata(registrations);
-		return pagedResourcesAssembler.toResource(new PageImpl<>(registrations), assembler);
+		return pagedResourcesAssembler.toResource(
+				new PageImpl<>(registrations, pageable, appRegistry.findAll().size()),
+				assembler);
 	}
 
 	/**
--- a/pom.xml
+++ b/pom.xml
@@ -31,7 +31,7 @@
 		<spring-cloud-deployer-local.version>1.2.0.RC1</spring-cloud-deployer-local.version>
 
 		<!-- Note: Change ./spring-cloud-dataflow-server-local/pom.xml to match the spring-boot.version -->
-		<spring-boot.version>1.5.3.RELEASE</spring-boot.version>
+		<spring-boot.version>1.5.2.RELEASE</spring-boot.version>
 
 		<!-- Note: Change ./spring-cloud-dataflow-server-local/pom.xml to match the spring-cloud-task.version -->
 		<spring-cloud-task.version>1.2.0.RC1</spring-cloud-task.version>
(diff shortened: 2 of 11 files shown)


donovanmuller:
A before and after for [`StreamDefinitionController#listRelated`](https://github.com/spring-cloud/spring-cloud-dataflow/pull/1383/files#diff-1b8ed5096382d31ece0a8397f1eb4cf2)

screen shot 2017-04-22 at 10 24 58 pm

screen shot 2017-04-22 at 10 38 13 pm

ilayaperumalg:
@donovanmuller I haven't been able to reproduce the issue with `spring-data-commons:1.13.3.RELEASE`.  it definitely makes sense to have `Pageable` injected into `PageImpl`.  Merging this change.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppRegistryController.java
```diff
@@ -116,7 +117,8 @@ public AppRegistryController(AppRegistry appRegistry,
 			}
 		}
 		Collections.sort(list);
-		return pagedResourcesAssembler.toResource(new PageImpl<>(list), assembler);
+		return pagedResourcesAssembler
+				.toResource(new PageImpl<>(list, pageable, appRegistry.findAll().size()), assembler);
```
ilayaperumalg: The `appRegistry.findAll()` can be reused from what we computed above in the same method.

ilayaperumalg: will fix while merging.

ilayaperumalg:
@donovanmuller Thank you for the contribution. Merged the PR as 21840685925dc9272f86c0fc054f0c63e51208c4 along with eae98f4065eca66c89482bf249761f0a0c19d8e6

sabbyanandan:
Thanks, @donovanmuller and @ilayaperumalg. Unfortunately, this commit has a side effect on Flo's pagination (spring-cloud/spring-cloud-dataflow#1397). We may have to revert the commit and investigate the applicability of PR and it's side effects post 1.2.0.

trisberg:
Revert PR: #1400 

Let's make these changes after 1.2.0.RELEASE

ilayaperumalg:
The root cause of this issue is that `AppRegistry` doesn't have `findAll(Pageable)` method in its implementation. Since `AppRegistryController` always returns PagedResources, the UI that expects the paginated resources from AppRegistryController goes into the loop.

As a follow up, I have created this story: https://github.com/spring-cloud/spring-cloud-dataflow/issues/1401

kalvarez2:
So, what version of spring-boot does work?

kalvarez2:
so... I am trying to write an implementation of spring-cloud-deployer-spi at version 1.2.0.RELEASE. It is an spring-boot app, so my parent is spring-boot-starter-parent at version "1.4.6.RELEASE", also trried at version "1.5.2.RELEASE".  The app boots just fine, but then I am getting this error when trying to use the UI... what version should I set on my pom for the parent or for -spi to get this error to go away?

vijaykalidindi:
Same problem here. Is there any workaround while the fix is being worked on?

SuperMohit:
Hi,
I am facing same issue. Is there work around? or which version should I use to make this work?

sabbyanandan:
Hi, @kalvarez2, @vijaykalidindi, @SuperMohit. The fix is underway via spring-cloud/spring-cloud-dataflow#1545. Once this is merged (hopefully today), we will update the relevant bits associated with this PR and release it in the upcoming [1.2.2.RELEASE](https://github.com/spring-cloud/spring-cloud-dataflow/milestone/23). 

Thanks for the patience.

ilayaperumalg:
This can be closed as #1545 is merged with the changes from this PR.

Issue #1322, opened 2017-04-13
Clarify the use of caching behind "app register" command


sabbyanandan:
As a user, I'm registering and re-registering apps; however, I'm unable to resolve the right artifact at runtime.

**Acceptance:**

This section in the docs could describe the use of caching behind the app-registry and what happens when `--force` is used.

> When using either app register or app import, if a stream app is already registered with the provided name and type, it will not be overridden by default. If you would like to override the pre-existing stream app, then include the --force option.

PR #1385, opened 2017-04-24
Clarify caching behavior related to app registration

ericbottard:
Fixes https://github.com/spring-cloud/spring-cloud-dataflow/issues/1322


markfisher:
should we also mention that once a stream is deployed it will not re-download (even a Maven snapshot) unless it is re-deployed?

ericbottard:
Added another commit explaining what (doesn't) happen with deployed streams

ericbottard:
Merged as squashed commit 7e5d5200

PR #1403, opened 2017-04-25
Add support for redacting passwords in task args

### LINKED ISSUE #1381: Task Executions Details exposes database password
The data shown in `/tasks/executions/{executionid}` includes the database password as part of the arguments. We should show '****' instead.

cppwfs:
when viewing.

resolves #1381

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskExecutionController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskExecutionController.java
@@ -97,7 +97,7 @@ public class TaskExecutionController {
 	@ResponseStatus(HttpStatus.OK)
 	public PagedResources<TaskExecutionResource> list(Pageable pageable,
 			PagedResourcesAssembler<TaskJobExecutionRel> assembler) {
-		Page<TaskExecution> taskExecutions = explorer.findAll(pageable);
+		Page<TaskExecution> taskExecutions = this.taskService.findAll(pageable);
 		Page<TaskJobExecutionRel> result = getPageableRelationships(taskExecutions, pageable);
 		return assembler.toResource(result, taskAssembler);
 	}
@@ -118,7 +118,7 @@ public class TaskExecutionController {
 		if (this.taskDefinitionRepository.findOne(taskName) == null) {
 			throw new NoSuchTaskDefinitionException(taskName);
 		}
-		Page<TaskExecution> taskExecutions = explorer.findTaskExecutionsByName(taskName, pageable);
+		Page<TaskExecution> taskExecutions = this.taskService.findTaskExecutionsByName(taskName, pageable);
 		Page<TaskJobExecutionRel> result = getPageableRelationships(taskExecutions, pageable);
 		return assembler.toResource(result, taskAssembler);
 	}
@@ -152,10 +152,7 @@ public class TaskExecutionController {
 	@RequestMapping(value = "/{id}", method = RequestMethod.GET)
 	@ResponseStatus(HttpStatus.OK)
 	public TaskExecutionResource view(@PathVariable("id") long id) {
-		TaskExecution taskExecution = this.explorer.getTaskExecution(id);
-		if(taskExecution == null){
-			throw new NoSuchTaskExecutionException(id);
-		}
+		TaskExecution taskExecution = this.taskService.viewTaskExecution(id);
 		TaskJobExecutionRel taskJobExecutionRel = new TaskJobExecutionRel(taskExecution,
 				new ArrayList<>(explorer.getJobExecutionIdsByTaskExecutionId(
 						taskExecution.getExecutionId())));
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskExecutionController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskExecutionController.java
@@ -19,6 +19,7 @@ package org.springframework.cloud.dataflow.server.controller;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import org.springframework.cloud.dataflow.rest.job.TaskJobExecutionRel;
 import org.springframework.cloud.dataflow.rest.resource.TaskExecutionResource;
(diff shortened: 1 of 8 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskExecutionController.java
```diff
@@ -97,7 +97,7 @@ public TaskExecutionController(TaskExplorer explorer, TaskService taskService,
 	@ResponseStatus(HttpStatus.OK)
 	public PagedResources<TaskExecutionResource> list(Pageable pageable,
 			PagedResourcesAssembler<TaskJobExecutionRel> assembler) {
-		Page<TaskExecution> taskExecutions = explorer.findAll(pageable);
+		Page<TaskExecution> taskExecutions = this.taskService.findAll(pageable);
```
ilayaperumalg: Any specific reason why this can't the `masking` be handled within this method instead of creating those methods in `TaskExplorer` into `TaskService` again?

ilayaperumalg: Perhaps we could extend `SimpleTaskExplorer` and override those two methods to mask sensitive contents.

cppwfs: 1) Good question.  Usually I leave business logic outside of a controller.  In this case the masking was put in the TaskService.  I can move the logic into the controller, but what is wrong with the methods being in the TaskService?
2) To the second question, Spring Cloud Task shouldn't have a TaskExplorer that is specific to a particular app, in this case a password redaction.  This should be handled by the app that is using itself.

ilayaperumalg: > I can move the logic into the controller, but what is wrong with the methods being in the TaskService?
The reason I asked it because of the need to duplicate those two methods (that were part of TaskExplorer) again to the TaskService. 

> Spring Cloud Task shouldn't have a TaskExplorer that is specific to a particular app, in this case a password redaction. 

But, given we are fixing this only at the SCDF controller side of things where TaskExecution details are retrieved and we always have a `TaskExplorer` I was leaning towards having a custom TaskExplorer that `extends` the `SimpleTaskExplorer` we already use in SCDF.

mminella: For the record, I'm not a fan of creating a custom extension of the `TaskExplorer` for this kind of work.  This functionality doesn't have anything to do with the `TaskExplorer` itself.  It is related to how SCDF displays data.  That's an SCDF concern and by extending `TaskExplorer` you create unneeded coupling to the `TaskExplorer`.  To be honest, I view this as a formatting issue that should be handled in the controller but the service is fine as well IMHO.

trisberg: I would have expected some sanitation to happen in the controller layer, seems to be UI related to hide certain sensitive info

cppwfs: Guess I should have waited before moving the code from TaskService to a DefaultTaskExplorer ;-) .   The general consensus is that we need to move it to the controller.   I'll do that, but seems fairly heavy wait for a controller though.

mminella: @cppwfs I view service == business logic.  Controller == back end display logic.  To be *really* honest, you could argue that this is a security concern and should be handled by a `ServletFilter` (applying this logic across all requests)...

trisberg: Would be nice if we could just run the values through the existing Sanitizer - https://github.com/spring-projects/spring-boot/blob/master/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/Sanitizer.java - but that class is package private

cppwfs: Yeah, that kinda sucked :(

mminella: Have we brought it up to the Boot team? cc: @philwebb

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/services/impl/DefaultTaskServiceTests.java
```diff
@@ -127,17 +112,6 @@
...
-	private boolean initialized = false;
-
-	private Map<Long, TaskExecution> taskExecutionsMap;
-
-	private Pageable defaultPageable;
-
-
 	@Before
 	public void setupMockMVC() {
 		this.template = new JdbcTemplate(dataSource);
```
trisberg: Why do we need this? tests run fine without it.

cppwfs: artifact from previous commit.  Will remove it.

trisberg: I can do that on merge

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskExecutionController.java
```diff
@@ -198,10 +203,20 @@ public TaskExecutionResource instantiateResource(TaskJobExecutionRel taskJobExec
...
-					new ArrayList<>(explorer.getJobExecutionIdsByTaskExecutionId(
+			taskJobExecutionRels.add( new TaskJobExecutionRel(sanitizePotentialSensitiveKeys(taskExecution),
+					new ArrayList<>(this.explorer.getJobExecutionIdsByTaskExecutionId(
 							taskExecution.getExecutionId()))));
 		}
 		return new PageImpl<>(taskJobExecutionRels,pageable,taskExecutions.getTotalElements());
 	}
+
+	private TaskExecution sanitizePotentialSensitiveKeys(TaskExecution taskExecution) {
+		ArgumentSanitizer argumentSanitizer = new ArgumentSanitizer();
```
trisberg: should we create a new one each invocation? could we reuse one created for this instance?

cppwfs: It can be re-used.

trisberg: OK, fixing these issues on merge, otherwise LGTM

trisberg:
Rebased, squashed, fixed and merged as commit 2157631534f7034883730baf5960d8ea586f8482

PR #1419, opened 2017-04-27
Update spring-analytics version

### LINKED ISSUE #1418: Update spring-analytics version to 1.2.0.RELEASE
To update the changes from https://github.com/spring-projects/spring-analytics/issues/3

We can probably update the `release` version once it is available sooner.

ilayaperumalg:
Resolves #1418

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -40,7 +40,7 @@
         <!-- Note: Change ./spring-cloud-dataflow-server-local/pom.xml to match the spring-cloud.version -->
         <spring-cloud.version>Dalston.RELEASE</spring-cloud.version>
 
-        <spring-analytics.version>1.1.1.RELEASE</spring-analytics.version>
+        <spring-analytics.version>1.2.0.BUILD-SNAPSHOT</spring-analytics.version>
         <spring-batch-admin-manager.version>1.3.1.RELEASE</spring-batch-admin-manager.version>
         <spring-shell.version>1.2.0.RELEASE</spring-shell.version>
         <spring-session.version>1.2.2.RELEASE</spring-session.version>


ericbottard:
Upgrading to this version and going to the Analytics tab while redis is not running now causes a flood of "Can't connect" exceptions, while it resulted in a "simple" NPE before.

So while the `!= null` fix was legit, the overall experience is left unchanged.

ilayaperumalg:
@ericbottard yeah, I agree on the user experience perspective (given the number of refreshes that could happen if the analytics tab is open at the data flow UI). But, still better than NPE ;)

Since this is something that Data Flow drives (invoking the `spring-analytics` REST endpoint from the UI), this is the best thing we can do at `spring-analytics` I guess.

ericbottard:
Agreed

pom.xml
```diff
@@ -40,7 +40,7 @@
         <!-- Note: Change ./spring-cloud-dataflow-server-local/pom.xml to match the spring-cloud.version -->
         <spring-cloud.version>Dalston.RELEASE</spring-cloud.version>
 
-        <spring-analytics.version>1.1.1.RELEASE</spring-analytics.version>
+        <spring-analytics.version>1.2.0.BUILD-SNAPSHOT</spring-analytics.version>
```
markpollack: Shouldn't I just be making a 1.1.2.RELEASE version?

trisberg: if it's just a small fix, 1.1.2 makes sense

markpollack:
merged as 1179bf1435076cdbfc951e6e97cb74cd352e5b0e

Issue #1319, opened 2017-04-13
Documentation for Stream's Common Application Properties is in the wrong section


markpollack:
http://docs.spring.io/spring-cloud-dataflow/docs/current-SNAPSHOT/reference/htmlsingle/#spring-cloud-dataflow-global-properties

It is listed under the chapter "32. Directed Graphs in a Stream".  This should be moved to a section under "23. Creating a Stream" and a similar one for tasks, maybe "35.4. Launching a Task" ?

markpollack:
See https://github.com/spring-cloud/spring-cloud-dataflow-server-cloudfoundry/issues/307 for some additional content that maybe included.

PR #1421, opened 2017-04-27
Move common stream properties section. Add task one

ericbottard:
Fixes https://github.com/spring-cloud/spring-cloud-dataflow/issues/1319
Fixes https://github.com/spring-cloud/spring-cloud-dataflow/issues/1386

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/tasks.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/tasks.adoc
@@ -174,6 +174,29 @@ be `TaskLauncher` implementation specific.
 dataflow:>task launch mytask --properties "deployer.timestamp.foo1=bar1,app.timestamp.foo2=bar2"
 ```
 
+==== Common application properties
+
+In addition to configuration via DSL, Spring Cloud Data Flow provides a mechanism for setting common properties to all
+the task applications that are launched by it.
+This can be done by adding properties prefixed with `spring.cloud.dataflow.applicationProperties.task` when starting the server.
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-docs/src/main/asciidoc/tasks.adoc
```diff
@@ -174,6 +174,29 @@ be `TaskLauncher` implementation specific.
...
+
+```
+--spring.cloud.dataflow.applicationProperties.task.foo=bar
+--spring.cloud.dataflow.applicationProperties.task.fizz=bar2
+```
+
+This will cause the properties `foo=bar` and `fizz=bar2` to be passed to all the launched applications.
+
+[NOTE]
+Properties configured using this mechanism have lower precedence than task deployment properties.
```
cppwfs: 'launch' properties instead of 'deployment' properties.

ericbottard: Hmmm, not sure. While I see what you mean, it is indeed deployment properties, as in [here](https://github.com/spring-cloud/spring-cloud-deployer/blob/master/spring-cloud-deployer-spi/src/main/java/org/springframework/cloud/deployer/spi/core/AppDeploymentRequest.java#L62)

cppwfs: My bad.  Please disregard.

markpollack:
merged as f0f076289a0e5e6e15c10dc8b7d8f5e9f300b333

PR #1435, opened 2017-05-01
Fix task transition symbol rendering in PDF

### LINKED ISSUE #1434: Composed task DSL syntax is wrong in the PDF
As a user, I downloaded the 1.2 RC1 PDF and I notice the inconsistency between the online docs and PDF. Specifically, the DSL syntaxt for transistions in the docs is `->`, but in the PDF it is `#` - see below.

PDF:
image

DOCS:
image

trisberg:
Resolves #1434


cppwfs:
LGTM  
Another trick they showed was to do this : `+->+`

PR #1455, opened 2017-05-03
Provide generic documentation link when server version is 'unknown'

### LINKED ISSUE #1453: Don't provide download link for shell when version is 'unknown'
Currently we display an invalid download message - there is no download link on the about page for older versions

```
server-unknown:>dataflow config server --uri http://localhost:9393
Unable to contact Data Flow Server at 'http://localhost:9393': 'java.lang.IllegalStateException: Incompatible version of Data Flow server detected.
Trying to use shell which supports revision 14, while server revision is [unknown]. Both revisions should be aligned.
Follow instructions at http://localhost:9393/dashboard#about to download a compatible version of the shell.'.
```

It would be better to tell user to consult the documentation for the server version they are using.

trisberg:
Resolves #1453

### DIFF
--- a/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/DataFlowTemplate.java
+++ b/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/DataFlowTemplate.java
@@ -163,8 +163,13 @@ public class DataFlowTemplate implements DataFlowOperations {
 		final RootResource resourceSupport = restTemplate.getForObject(baseURI, RootResource.class);
 
 		if (resourceSupport != null) {
-			String serverRevision = resourceSupport.getApiRevision() != null
-					? resourceSupport.getApiRevision().toString() : "[unknown]";
+			if (resourceSupport.getApiRevision() == null) {
+				throw new IllegalStateException("Incompatible version of Data Flow server detected.\n"
+								+ "Follow instructions in the documentation for the version of the server you are "
+								+ "using to download a compatible version of the shell.\n"
+								+ "Documentation can be accessed at http://cloud.spring.io/spring-cloud-dataflow/");
+			}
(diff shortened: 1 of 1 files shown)


spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/DataFlowTemplate.java
```diff
@@ -163,8 +163,13 @@ public DataFlowTemplate(URI baseURI, RestTemplate restTemplate) {
 		final RootResource resourceSupport = restTemplate.getForObject(baseURI, RootResource.class);
 
 		if (resourceSupport != null) {
-			String serverRevision = resourceSupport.getApiRevision() != null
-					? resourceSupport.getApiRevision().toString() : "[unknown]";
+			if (resourceSupport.getApiRevision() == null) {
+				throw new IllegalStateException("Incompatible version of Data Flow server detected.\n"
```
ilayaperumalg: Given server incompatibility is handled separately [here](https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/DataFlowTemplate.java#L171), do you think we can have even more generic documentation link at this line?

trisberg: That's what I'm trying to do, we don't have a specific download page to direct to so it has to be very generic. If we know the version it is different and we can point to the about page.

ilayaperumalg: ok, merging.

ilayaperumalg:
Rebased and merged as a413f31

PR #1457, opened 2017-05-03
Fix AppRegistry list() to return non-zero pageable list

ilayaperumalg:


ilayaperumalg:
Closing this PR as this doesn't solve the entire problem as it requires some co-ordination between the Pageable implementation and Flo.

PR #1461, opened 2017-05-03
Fix http command

trisberg:
Fix HTTP POST command

- Remove charset in http command

- Don't add charset if mediatype doesn't have one defined


cppwfs:
LGTM

cppwfs:
Rebased Merged

PR #1463, opened 2017-05-03
GH-1462 Fixed MVC exception

### LINKED ISSUE #1462: StreamDefinitionController.save(..) results in java.lang.IllegalStateException if invoked outside of MVC
```
Exception in thread "main" java.lang.IllegalStateException: Could not find current request via RequestContextHolder. Is this being called from a Spring MVC handler?
    at org.springframework.util.Assert.state(Assert.java:70)
    at org.springframework.hateoas.mvc.ControllerLinkBuilder.getCurrentRequest(ControllerLinkBuilder.java:291)
    at org.springframework.hateoas.mvc.ControllerLinkBuilder.getBuilder(ControllerLinkBuilder.java:265)
    at org.springframework.hateoas.mvc.ControllerLinkBuilder.linkTo(ControllerLinkBuilder.java:124)
    at org.springframework.hateoas.mvc.ResourceAssemblerSupport.createResourceWithId(ResourceAssemblerSupport.java:90)
    at org.springframework.hateoas.mvc.ResourceAssemblerSupport.createResourceWithId(ResourceAssemblerSupport.java:81)
    at org.springframework.cloud.dataflow.server.controller.StreamDefinitionController$Assembler.toResource(StreamDefinitionController.java:377)
    at org.springframework.cloud.dataflow.server.controller.StreamDefinitionController.save(StreamDefinitionController.java:205)
    at olegz.dataflow.SomeDemo.main(SomeDemo.java:41)
```
Since MVC controllers by design must be executable outside of any Web infrastructure (i.e., testing) their execution should not result in exceptions pointing to the fact that they were executed outside of Web infrastructure.
In this particular case, the exception is well understood, since the `save(..)` operation was indeed invoked directly. Further more it did not affect anything in the flow. Every stream and the entire flow was running successfully, so in its current state it's more of a nuisance-type bug which IMHO should be easily fixed with WARN logging, thus preserving the intent of notifying user.

olegz:
- wrapped  in try/catch, WARN logging the MVC related IllegalStateException instead of throwing it
- added test

Fixes #1462

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
@@ -80,6 +80,7 @@ import org.springframework.web.bind.annotation.RestController;
  * @author Patrick Peralta
  * @author Ilayaperumal Gopinathan
  * @author Gunnar Hillert
+ * @author Oleg Zhurakousky
  */
 @RestController
 @RequestMapping("/streams/definitions")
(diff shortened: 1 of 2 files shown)


ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Rebased and merged as 3cb4759

Issue #1464, opened 2017-05-04
OAuth2 - Fix Authorization Code Grant Type Not Working


ghillert:

ghillert:
See also: https://github.com/spring-cloud/spring-cloud-dataflow/issues/1468

PR #1465, opened 2017-05-04
gh-1464 OAuth2 - Fix Authorization Code Grant Type Not Working

ghillert:
* Minimize Boot's AutoConfiguration for Security
* Maintain full backwards compatibility - no test changes
* Trigger OAuth2 Authorization Code Grant Type only for Browsers and only if they don't send a `application/json` Accept header

resolves https://github.com/spring-cloud/spring-cloud-dataflow/issues/1464

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/security/OAuthSecurityConfiguration.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/security/OAuthSecurityConfiguration.java
@@ -15,23 +15,47 @@
  */
 package org.springframework.cloud.dataflow.server.config.security;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+import javax.servlet.Filter;
+
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.boot.autoconfigure.security.oauth2.client.EnableOAuth2Sso;
+import org.springframework.boot.autoconfigure.security.SecurityProperties;
+import org.springframework.boot.autoconfigure.security.oauth2.resource.ResourceServerProperties;
+import org.springframework.boot.autoconfigure.security.oauth2.resource.UserInfoTokenServices;
 import org.springframework.cloud.dataflow.server.config.security.support.OnSecurityEnabledAndOAuth2Enabled;
 import org.springframework.cloud.dataflow.server.config.security.support.SecurityStateBean;
 import org.springframework.cloud.dataflow.server.service.impl.ManualOAuthAuthenticationProvider;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Conditional;
 import org.springframework.context.annotation.Configuration;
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/security/OAuthSecurityConfiguration.java
```diff
@@ -41,46 +65,127 @@
...
+
+	private static class BrowserDetectingContentNegotiationStrategy extends HeaderContentNegotiationStrategy {
+
+		@Override
+		public List<MediaType> resolveMediaTypes(NativeWebRequest request)
+				throws HttpMediaTypeNotAcceptableException {
+			final List<MediaType> supportedMediaTypes = super.resolveMediaTypes(request);
+
+			final String userAgent = request.getHeader(HttpHeaders.USER_AGENT);
+			if (userAgent != null && userAgent.contains("Mozilla/5.0")
```
jvalkeal: I've understood that sniffing user-agent on server side is wrong and adds more maintenance. What about if it's something else that `Mozilla/5.0`?

markpollack:
It seems like not having a change to the test implies what we didn't have a test wrt to ensuring this functionality was working.   Since OAuth functionality is important in the PCF environment, we should test there before releasing.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/security/OAuthSecurityConfiguration.java
ghillert: For `2.0` we need to revamp our:

- url structure and
- security setup

E.g. we should really move our api under something like `/api`. That way we can make more appropriate/explicit security assumption

We should also also eliminate this whole basic authentication setup for accessing the REST API eventually.

Thus, looking at the user-agent is more of a temporary solution. Took some hues from Initializr. See also:

http://www.useragentstring.com/pages/useragentstring.php

First I looked at triggering basic authentication only if a Json accept header is being sent. But that would have broken several tests - This late in the game I did not want to cause ANY breaking changes.

ghillert:
@markpollack Added issue: https://github.com/spring-cloud/spring-cloud-dataflow/issues/1468 to add tests to cover this functionality.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/security/OAuthSecurityConfiguration.java
markpollack: Why get rid of basic auth?  We can't assume OAuth setup everywhere, especially just 'out of the box' where someone on PWS/PCF or k8s on google would expose an app to the public without any security....

Merging now though....

markpollack:
BTW - there was a failing checkstyle test...fixing.

markpollack:
merged as d82f10f55bf2fe026f2fcd3776a901972fa15c9b

PR #1467, opened 2017-05-04
Change the actuator role from ADMIN to MANAGE

### LINKED ISSUE #1466: The role used for actuator endpoints is ADMIN in Kubernetes server impl
Looks like the Kubernetes server impl gets the actuator role configured as ADMIN instead of MANAGE

trisberg:
Resolves #1466

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
+++ b/spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
@@ -1,6 +1,6 @@
 management:
   security:
-    roles: ADMIN
+    roles: MANAGE
 logging:
   level:
     org.springframework.beans: 'WARN'
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/security/DefaultBootUserAuthenticationConfiguration.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/security/DefaultBootUserAuthenticationConfiguration.java
@@ -74,8 +74,8 @@ public class DefaultBootUserAuthenticationConfiguration extends GlobalAuthentica
 
 		if (this.managementServerProperties != null
 				&& this.managementServerProperties.getSecurity().getRoles().size() == 1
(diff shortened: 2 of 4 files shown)


spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
```diff
@@ -1,6 +1,6 @@
 management:
   security:
-    roles: ADMIN
```
ghillert: @sabbyanandan @trisberg Can we just remove this property altogether? It seems to create issues when allocating roles - It looks like the management role is merged into the user roles, making it hard to assign default roles. Or even to create a user without a Manage role.

ilayaperumalg:
Please note that whatever the changes we make with this PR, we need to fix [here](https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/security/DefaultBootUserAuthenticationConfiguration.java#L75) as well.

ilayaperumalg:
The above fix is required to make the default user (with basic security enabled) access the `secured` management endpoints.

spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
ilayaperumalg: @ghillert  Removing this property will leave us to use `ACTUATOR` role which is the default. I believe we want to override this here because of the default `ACTUATOR` role.

IMO, it makes sense to change the `management.security.roles` to `MANAGE` as this will be one less role to set explicitly (as ADMIN) when `management.security.enabled` is set to `true`.

trisberg:
We mustn't have any tests that exercise this code path, will amend the PR

ilayaperumalg:
@trisberg Also, I feel that specific code path has issue when `management.security.enabled` is set to `true`. I think we don't want hard coded value of whatever role we choose. Instead, we need to retrieve from what is set on `management.security.roles` and add all those roles to the default boot user.

ilayaperumalg:
LGTM, merging. tested using basic auth with/without security enabled, management security enabled using the basic and explicit user.

ilayaperumalg:
Squashed, rebased and merged as 493662c

PR #1493, opened 2017-05-12
Update rule for wrapping lines

### LINKED ISSUE #1490: Update eclipse rules wrt to wrapping lines
We should change

```
<setting id="org.eclipse.jdt.core.formatter.join_wrapped_lines" value="true"/>
```

To false so that fluent API calls that purposely put one API call per line aren't joined. This is an issue mostly with JDK stream() and reactive code.

ilayaperumalg:
- join_wrapped_lines to `false`

Resolves #1490

### DIFF
--- a/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/StreamPropertyKeys.java
+++ b/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/StreamPropertyKeys.java
@@ -23,7 +23,7 @@ package org.springframework.cloud.dataflow.core;
  */
 public class StreamPropertyKeys {
 
-    // @formatter:off (not working)
+	// @formatter:off (not working)
     static final String PREFIX = "spring.cloud.stream.";
     /**
      * Prefix for Spring Cloud Stream Metrics.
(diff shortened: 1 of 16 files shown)


spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/StreamPropertyKeys.java
```diff
@@ -23,7 +23,7 @@
  */
 public class StreamPropertyKeys {
 
-    // @formatter:off (not working)
+	// @formatter:off (not working)
     static final String PREFIX = "spring.cloud.stream.";
```
trisberg: Looks like we don't enforce spaces vs tabs, should/could we?

ilayaperumalg: No, this is an oversight. I don't know any ways of doing it yet :(

trisberg: do we know why the formatting isn't working on this class? seems like a simple class that should be easy to format

trisberg:
Merged as commit e406a9877cd23cccd3224c3680fb529f1a2f405d

Issue #1497, opened 2017-05-16
Consider adding some URI filter/validation to AppRegistration when registering a new app


olegz:
When registering a new app, DSL allows for any valid URI. For example:
`$ app register --name time-source --type source --uri file://Users/olegz/Downloads/time-source/target/classes`
And then it displays the success message
`Successfully registered application 'source:time-source'`
But in reality the above application is doomed to fail since the 'file' based URI does not point to a boot JAR.
What makes it even worst is that in SCDF you would typically deploy several apps and then create a stream and then deploy it just to see nothing happening (good, bad or indifferent) untill one actually looks at the logs to see 
```
$ cat stderr_0.log
Error: Invalid or corrupt jarfile /Users/olegz/Downloads/time-source/target/classes
```

While at some point in the future when things settle down we should probably do full JAR validation to ensure that it is indeed a valid boot JAR, at this point I would at least validate that the 'file' based URI points to a valid resource and that resource is a file and is a JAR.
And if not notify the user in the console right away

olegz:
Based on comments from https://github.com/spring-cloud/spring-cloud-dataflow/pull/1498 the proposed approach is not feasible. Perhaps need to come up with something clever on the deployer(s) side to be able to communicate to the shell user that things didn't go right

markfisher:
We have discussed supporting an option for eagerly fetching resources (at the time of registration) which would then or course mean failing fast in case of invalid resources.

olegz:
That is yet another valid point @markfisher. Basically now I believe there are two types of enhancements here. That may require an additional flag since one should be able to choose if they want such eager/fail-fast approach (see explanation below)
1. Ability to eagerly fetch and validate resources that they are "deployable" (i.e., exist and in a proper/supported format)
2. Notifying the shell/UI user (outside of looking at the logs) that something went wrong with the particular microservice during deployment of the stream.

[...]

PR #1498, opened 2017-05-16
GH-1497 added 'file' URI validation

olegz:
- added 'file' URI validation to AppRegistry
- added initial version of AppRegistryTests
- fixed AppRegistryControllerTests which were failing due to this change

### DIFF
--- a/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/AppRegistry.java
+++ b/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/AppRegistry.java
@@ -16,6 +16,7 @@
 
 package org.springframework.cloud.dataflow.registry;
 
+import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
@@ -58,6 +59,7 @@ import org.springframework.util.StringUtils;
  * @author Gunnar Hillert
  * @author Thomas Risberg
  * @author Eric Bottard
+ * @author Oleg Zhurakousky
  */
 public class AppRegistry {
 
@@ -92,6 +94,7 @@ public class AppRegistry {
 	}
 
 	public AppRegistration save(String name, ApplicationType type, URI uri, URI metadataUri) {
+		this.validateApplicationUri(uri);
 		this.uriRegistry.register(key(name, type), uri);
 		if (metadataUri != null) {
 			this.uriRegistry.register(metadataKey(name, type), metadataUri);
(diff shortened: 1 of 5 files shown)


spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/AppRegistry.java
```diff
@@ -235,4 +238,12 @@ else if (!StringUtils.hasText(uri.getSchemeSpecificPart())) {
 		}
 		return uri;
 	}
+
+	private void validateApplicationUri(URI uri){
+		if (uri.getScheme().toLowerCase().equals("file")){
+			File file = new File(uri);
+			Assert.isTrue(file.exists() && !file.isDirectory() && file.getName().toLowerCase().endsWith(".jar"),
```
ericbottard: File *can* be a directory (we support exploded archives)

ericbottard: Moreover, file *could* be something else than a boot app (could be a python app ultimately, etc)

spring-cloud-dataflow-registry/src/test/java/org/springframework/cloud/dataflow/registry/AppRegistryTests.java
```diff
@@ -0,0 +1,87 @@
...
+		AppRegistration registration = this.appRegistry.save("foo", ApplicationType.processor, uri, null);
+		assertThat(registration != null);
+		assertEquals("foo", registration.getName());
+		assertEquals(ApplicationType.processor, registration.getType());
+		assertEquals(uri, registration.getUri());
+	}
+
+	@Test
+	public void testWithExistingJARFileURI() throws Exception {
+		File jar = new File("foo.jar");
```
ericbottard: Should reside in temporary ad hoc directory, not working directory. See JUnit rule for that purpose.

Applies to all tests in this class

spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/AppRegistry.java
olegz: Hmm, that changes things. Wasn't aware of that. Perhaps I need to revisit the approach all together to at least ensure that that user has a better way of being notified that the final stream can not be deployed due to. . . .

olegz:
@ericbottard Thanks for the pointers.  I am closing this PR for now based on the file support comment. Will submit the new one once i figure out how to make it more user friendly.

Issue #1495, opened 2017-05-16
Clean up implementations of RecoveryStrategy and ExpansionStrategy to eliminate repetative code


olegz:
While reviewing the code I've noticed nearly identical blocks of code repeated in many implementations of `RecoveryStrategy` and `ExpansionStrategy`. Specifically related to the collections of "proposals".

olegz:
Resolved/merged

PR #1501, opened 2017-05-17
GH-1495 continue general code cleanup of 'completion' module

olegz:
- continuation of general code cleanup started in a4900082 commit

### DIFF
--- a/spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/ProposalsCollectorSupportUtils.java
+++ b/spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/ProposalsCollectorSupportUtils.java
@@ -15,10 +15,13 @@
  */
 package org.springframework.cloud.dataflow.completion;
 
+import java.io.IOException;
+import java.net.URLClassLoader;
 import java.util.List;
 import java.util.Set;
 
 import org.springframework.boot.configurationmetadata.ConfigurationMetadataProperty;
+import org.springframework.boot.configurationmetadata.ValueHint;
 import org.springframework.cloud.dataflow.configuration.metadata.ApplicationConfigurationMetadataResolver;
 import org.springframework.cloud.dataflow.core.ApplicationType;
 import org.springframework.cloud.dataflow.core.dsl.CheckPointedParseException;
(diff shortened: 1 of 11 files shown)


spring-cloud-dataflow-completion/src/main/java/org/springframework/cloud/dataflow/completion/ProposalsCollectorSupportUtils.java
```diff
@@ -72,6 +74,65 @@ void doAddProposals(String text, String startsWith, AppRegistration appRegistrat
...
+				return null;
+			}
+		});
+	}
+
+	boolean addAlreadyTypedValueHintsProposals(final String text, AppRegistration appRegistration, final List<CompletionProposal> collector, final String propertyName, final ValueHintProvider[] valueHintProviders, final String alreadyTyped){
+		final Resource metadataResource = appRegistration.getMetadataResource();
+
+		final URLClassLoader classLoader = metadataResolver.createAppClassLoader(metadataResource);
+		return this.doWithClassLoader(classLoader, new Callback<Boolean>() {
```
ericbottard: Can use lambda here I think, will change on merge

olegz: @ericbottard you certainly can and that would also remove the need for all those 'finals' that clutter the code, since all of them are _effectively final_. I wasn't sure what the minimum Java level is so assumed 1.7, but good to know.

olegz: Also, it would be helpful if you were to review the names for all those new `add*` operations in `ProposalsCollectorSupportUtils`. I did my best to extract the _intention_ from the code, but could be wrong.

ericbottard: Sure.

Generally speaking, SCDF is jdk8. Spring Cloud Deployer is JDK7 though

ericbottard: (you may need to add a `.jdk8` file in a particular module if this hasn't been done, though

ericbottard:
Merged with lambdas as 95bce2e1

Issue #1294, opened 2017-04-11
Add support for OAUTH authorization in PCF


ghillert:
As a user, I'd like to authenticate and authorize SCDF's RESTful APIs against the OAuth backend in PCF.

Notes:
- Similar to what we have done in [Spring Boot](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cloudfoundry/CloudFoundrySecurityService.java#L67) for Actuator endpoint integration in Apps-Mgr, SCDF will make a Cloud Controller call ([/v2/apps/:guid/permissions](https://apidocs.cloudfoundry.org/253/apps/retrieving_permissions_on_a_app.html)) to verify the token validity and that the token has `Space Developer` access.
- Upon successful authorization, SCDF allows RESTful requests to go through.
- More background details in this deck [here](https://docs.google.com/presentation/d/1Vpp4UBclmKd5fZobIpSOedR6vpsk3eRd4CwCiuwykv8/edit#slide=id.g1d7e09bcda_0_368).

**Acceptance:**
- To prevent from having CF-specific OAuth logic/components in SCDF-core, consider the possibility of splitting it up into a separate repo 
- SCDF-CF-Server could consume the library as a dependency directly
- The token authorization against the OAuth backend workflow operates as described in the deck
- Shell, UI, and REST experience works in this setup
- Unit-tests included
- Docs updated

ghillert:
For reference - created https://github.com/pivotal-cf/spring-cloud-sso-connector/issues/7

PR #1505, opened 2017-05-18
gh-1294 Add support for OAUTH authorization

ghillert:
* Add role support to OAuth2-enabled security
  - Uses same rules (matchers) as traditional security
* Add `DefaultDataflowAuthoritiesExtractor`
  - Assigns all roles by default
  - Allows for user-customization, if providing own `AuthoritiesExtractor` bean
* Add tests

resolves https://github.com/spring-cloud/spring-cloud-dataflow/issues/1294

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/security/support/DefaultDataflowAuthoritiesExtractor.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/security/support/DefaultDataflowAuthoritiesExtractor.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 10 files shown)


jvalkeal:
I think this looks ok, I'd just wish that `AuthoritiesExtractor` could be customised via props. Maybe other issue?

ghillert:
@jvalkeal Users can add their own `AuthoritiesExtractor` bean to the context if they want (it would take precedence). As the default one is so simple I did not provide customization to the `DefaultDataflowAuthoritiesExtractor`.

ghillert:
Added documentation.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/security/support/DefaultDataflowAuthoritiesExtractor.java
```diff
@@ -0,0 +1,68 @@
...
+	@Override
+	public List<GrantedAuthority> extractAuthorities(Map<String, Object> map) {
+		Assert.notNull(map, "The map argument must not be null.");
+
+		final List<String> rolesAsStrings = new ArrayList<>();
+		final List<GrantedAuthority> grantedAuthorities =
+			Stream.of(CoreSecurityRoles.values())
+				.map(roleEnum -> {
+					final String roleName = SecurityConfigUtils.ROLE_PREFIX + roleEnum.getKey();
+					rolesAsStrings.add(roleName);
```
ericbottard: Using mutation by side-effect just to avoid 2 `stream()` operations doesn't feel very functional. I think it's ok to `stream()` twice given this is not a performance bottleneck

ilayaperumalg:
There are `checkstyle` errors on this PR. Can fix while merging.

ilayaperumalg:
LGTM, verified the security configuration with basic and OAuth. Merging.

ilayaperumalg:
Squashed, rebased, fixed the `checkstyle` issues and merged as 88a422ee0a6df4ccde80d72634883396eaee30f7

PR #1507, opened 2017-05-19
Support providing HTTP credentials from an external process

mheath:
This enables use of something like `cf oauth-token` to authenticate the
Shell to a DF server using Oauth2 tokens.

See https://www.pivotaltracker.com/story/show/138282513


ghillert:
Thanks for the contribution!!! Merged with commit e828aad05b545421213b4d38440e297b101563ac and some minor polishing with commit f6423adabaae384135d0db58437d39f993a1e515

PR #1540, opened 2017-06-21
Support newline Characters in Quoted Strings

garyrussell:
Certain properties are mapped to `Properties` objects using Spring's `PropertiesEditor`.

The sytax for such properties is `someProp='foo=bar \n baz=qux'`.

Currently the abstract tokenizer terminates a quoted token when `\n` or `\r` is found.

Change the `isArgValueIdentifierTerminator` to only terminate on such characters if we are
not parsing a quoted token.


aclement:
Merged in commit fa45d1e6ae9 and subsequent commit b97b9318b2 adds testcase. Hope that's all done correctly!

PR #1544, opened 2017-06-23
Redacts sensitive/password information when listing streams

### LINKED ISSUE #1542: Password should be hidden in UI/shell
_From @kniemann on May 11, 2017 21:1_

Passwords in the Dataflow UI should be hidden. This is probably true in the REST API/shell as well. Previously in SpringXD they were automatically ****. Especially since a VIEW permission was added in 1.2.0, it wouldn't make sense to grant it if passwords are visible in plain text.

**SpringXD:**

image

**vs Spring Cloud:**

image

_Copied from original issue: spring-cloud/spring-cloud-dataflow-ui#215_

cppwfs:
resolves #1542

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/support/ArgumentSanitizer.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/support/ArgumentSanitizer.java
@@ -18,12 +18,16 @@ package org.springframework.cloud.dataflow.server.controller.support;
 
 import java.util.regex.Pattern;
 
+import org.springframework.cloud.dataflow.core.dsl.ArgumentNode;
+
 /**
  * Sanitizes potentially sensitive keys for a specific command line arg.
  * @author Glenn Renfro
  */
 public class ArgumentSanitizer {
 	private static final String[] REGEX_PARTS = { "*", "$", "^", "+" };
+
+	private static final String REDACTION_STRING = "******";
 
 	private static final String[] KEYS_TO_SANITIZE = { "password", "secret", "key", "token", ".*credentials.*",
 			"vcap_services" };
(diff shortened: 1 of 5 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/support/ArgumentSanitizer.java
```diff
@@ -18,13 +18,17 @@
 
 import java.util.regex.Pattern;
 
+import org.springframework.cloud.dataflow.core.dsl.ArgumentNode;
+
 /**
  * Sanitizes potentially sensitive keys for a specific command line arg.
  * @author Glenn Renfro
  */
 public class ArgumentSanitizer {
```
markpollack: Do you happen to know why the boot guys didn't make this class public

https://pivotal.slack.com/archives/G0M0F86AJ/p1497984170701058

markpollack: I can't find it ATM, but I think there is a way in boot to modify which keys are considered the ones to sanitize.... @trisberg recently was asking how get boot to give up it's secrets.....   Did you find anything along that line of thought?

cppwfs: I saw their implementation.  But it is not accessible.  So this is based on their implementation.

cppwfs: Did you mean this ?
https://stackoverflow.com/questions/28300232/spring-boot-actuator-hides-property-values-in-env-endpoint

cppwfs:
Closing this PR.  Please use: https://github.com/spring-cloud/spring-cloud-dataflow/pull/1561/files

PR #1561, opened 2017-06-27
Redacts sensitive/password information when listing streams

### LINKED ISSUE #1542: Password should be hidden in UI/shell

cppwfs:
resolves #1542

Refactored where stream is sanitized

Uses regex to identify values that need to be redacted.

### DIFF
--- a/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/DestinationNode.java
+++ b/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/DestinationNode.java
@@ -56,7 +56,7 @@ public class DestinationNode extends AstNode {
 
 	@Override
 	public String toString() {
-		return getDestinationName();
+		return ":" + getDestinationName();
 	}
 
 	String getDestinationName() {
--- a/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/DestinationNode.java
+++ b/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/DestinationNode.java
@@ -56,7 +56,7 @@ public class DestinationNode extends AstNode {
 
 	@Override
 	public String toString() {
-		return ":" + getDestinationName();
+		return getDestinationName();
 	}
 
 	String getDestinationName() {
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/support/ArgumentSanitizer.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/support/ArgumentSanitizer.java
@@ -19,6 +19,7 @@ package org.springframework.cloud.dataflow.server.controller.support;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
 
 /**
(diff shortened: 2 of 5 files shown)


spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/DestinationNode.java
```diff
@@ -56,7 +56,7 @@ public String stringify(boolean includePositionalInfo) {
 
 	@Override
 	public String toString() {
-		return getDestinationName();
+		return ":" + getDestinationName();
```
ilayaperumalg: Do we really want to use `:` prefix for the destination name? Given, the `toString()` is expected to *only* return the actual destination name, we may not want to use the `:` prefix here as this could introduce some regression.

cppwfs: Wouldn't we want the toString to accurately depict what the stream DSL should look like?  Or at least separate the getStreamData to do so?   I ran this past @aclement before making the change.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/support/ArgumentSanitizer.java
```diff
@@ -16,17 +16,30 @@
...
+	private static final String REDACTION_STRING = "******";
+
+	private static final String[] KEYS_TO_SANITIZE = {"password", "secret", "key", "token", ".*credentials.*",
+			"vcap_services"};
+
+	public static Pattern passwordParameterPatternForStreams = Pattern.compile(
+			"(?i)(--[\\p{Z}]*[\\p{L}]*("
+					+ StringUtils.arrayToDelimitedString(KEYS_TO_SANITIZE, "|")
+					+ ")[\\p{L}]*[\\p{Z}]*=[\\p{Z}]*)((\"[\\p{L}|\\p{Pd}|\\p{Ps}|\\p{Pe}|\\p{Pc}|\\p{S}|\\p{N}|\\p{Z}]*\")|([\\p{N}|\\p{L}|\\p{Po}|\\p{Pc}|\\p{S}]*))",
+			Pattern.UNICODE_CASE);
```
mminella: A small comment on what that regex actually does would be helpful for "the next guy".

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/DestinationNode.java
markpollack: Destination nodes have the `:` prefix, so if we don't have this, then as glenn mentions, we don't generate the correct DSL representation.  I agree, it is the type of change that can introduce regressions, but ATM our tests pass and can't think of anything, can you?

aclement: I was happy with Glenn changing it to add the ":" - if you want just the destination name, call getDestinationName(). In general I am not a fan of relying on what toString() does. If I really wanted a way to reconstruct identical DSL I'd have a method called toDSLString() across all the nodes.

ilayaperumalg: ok, I am trying to understand what prompted this change as I don't seem to find a usage of this change with respect to this PR.

aclement: Good question Ilaya. The original approach Glenn was looking at was parsing the DSL - redacting the password in the argument value objects then converting that parsed AST back into DSL text. It became apparent we don't totally support that (never needed it before) - this was one place that needed changing but a more serious change is required in argument value processing because they lose whether they were quoted (and what quotes were used) when being parsed so converting back to DSL text isn't so easy. So we didn't do that extra work and instead Glenn went for a basic regex solution. This change to toString() remained. I don't mind it but strictly speaking it isn't necessary for the regex solution.

cppwfs: It can be removed since it doesn't technically belong here and can be submitted on a new PR.

cppwfs: I kept it in because it was a valid change.

markpollack: I think we can settle this here.  Seems like a nice little tweak but unrelated.

Removing the `:` has the newly added test `StreamParserTests.testGetStreamData` failing 

```
		node = parse (":fooStream > bar");
		assertThat( node.getStreamData(), equalTo(":fooStream > bar"));
```
```
Expected ::fooStream > bar
Actual   :fooStream > bar
```

The method `getStreamData` was created a while ago and delegates to the `toString` method for all the nodes, looks like the sort of `toDSLString` method any mentioned.  As the `getStreamData` method wasn't called at all before, I think we are safe making this change.

cppwfs:
Merged Squashed.

PR #1579, opened 2017-06-30
General cleanup of AppRegistry

olegz:


spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/AppRegistry.java
```diff
@@ -122,95 +127,24 @@ public AppRegistration save(String name, ApplicationType type, URI uri, URI meta
...
-			}
-		};
+		Set<String> registeredKeys = overwrite ? Collections.emptySet() : uriRegistry.findAll().keySet();
+		return Stream.of(resources)
+			.map(this::loadProperties)
+			.flatMap(prop -> prop.entrySet().stream()
+					.map(toStringAndUriFUNC)
+					.flatMap(kv -> toValidAppRegistration(kv, metadataUriFromProperties(kv.getKey(), prop)))
+					.filter(ar -> !registeredKeys.contains(key(ar.getName(), ar.getType())))
+					.map(ar -> save(ar.getName(), ar.getType(), ar.getUri(), ar.getMetadataUri()))
```
markpollack: I see that you are removing the iteration over `Resources` to use `Stream.of` instead, but in the original code there is a comment 

```
						.filter(ar -> !keysAlreadyThere.contains(key(ar.getName(), ar.getType())))
						.collect(Collectors.toList()) // Force eager evaluation to fail
														// early
						.forEach(ar -> apps.add(save(ar.getName(), ar.getType(), ar.getUri(), ar.getMetadataUri())));
```
I am not sure of the motivation behind forcing the list to be created and then calling save.  Maybe @ericbottard can comment once he gets back from vacation.

markpollack: i think the current behavior is to avoid saving any app registrations if a single one of them is already present.... verifying.  This PR would then change this behavior

markpollack: Nope, same behavior.  Fixed one small javadoc typo, nice job oleg!

markpollack:
merged as cf988bdc5c8ddc711c52a6aaccb2390d92bcfc47

PR #1594, opened 2017-07-10
Move RootController endpoints to /api

### LINKED ISSUE #1581: Move RootController endpoints to "/api"
As a developer, I'd like to move `RootController` endpoints to `/api`, so we could use the root location for Dashboard redirect and the APIs are under a more designated location.

**Acceptance:**
- `RootController` endpoints are moved under `/api`
- Launching the root URI redirects to `/dashboard`
- Test coverage adapted to the URI changes
- Reference guide updated

ilayaperumalg:
- Fix endpoint definitions
 - Fix configurations related to this endpoint change (security etc.,)
 - Fix tests

Resolves #1581

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/RootController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/RootController.java
@@ -62,6 +62,8 @@ import static org.springframework.hateoas.mvc.ControllerLinkBuilder.linkTo;
 @ExposesResourceFor(RootResource.class)
 public class RootController {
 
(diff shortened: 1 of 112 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/RootController.java
```diff
@@ -62,6 +62,8 @@
 @ExposesResourceFor(RootResource.class)
 public class RootController {
 
+	public static final String BASE_PATH = "/api";
```
ilayaperumalg: Maybe not in the scope of this PR. But, we want to be able to make this configurable via some SCDF base path property.

PR #1623, opened 2017-08-02
Add Shell completion for stream,task,counters, etc names.

### LINKED ISSUE #1271: Re-instanciate completion for stream/task names
XD used to have smart completion for stream/task job names in several places, even taking status into account (eg `stream deploy <undeployed-stream-name>`)

ericbottard:
Fixes #1271

### DIFF
--- a/spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/converter/EntityNameConverter.java
+++ b/spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/converter/EntityNameConverter.java
@@ -1,0 +1,122 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/converter/EntityNameConverter.java
```diff
@@ -0,0 +1,122 @@
...
+	private TaskOperations taskOperations() {
+		return dataFlowShell.getDataFlowOperations().taskOperations();
+	}
+
+	private StreamOperations streamOperations() {
+		return dataFlowShell.getDataFlowOperations().streamOperations();
+	}
+
+	private Optional<String> determineKind(String optionContext) {
+		return Arrays.stream(optionContext.split(" "))
```
olegz: Should it actually be `split("\\s+")`?

ericbottard: That's better, indeed. Although we control the source code for those commands

PR #1634, opened 2017-08-22
Fixes a bug where stepExecutionId is used instead of jobExecutionId

### LINKED ISSUE #1633: StepExecutionController returns stepExecutionID instead of JobExecutionId
The StepExecutionResourceBuilder sets the JobExecutionId to stepExecutionId.

cppwfs:
StepExecutionResourceBuilder has been updated to resolve this issue
resolves #1633

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/job/support/StepExecutionResourceBuilder.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/job/support/StepExecutionResourceBuilder.java
@@ -34,7 +34,7 @@ import org.springframework.util.Assert;
 public class StepExecutionResourceBuilder {
 
 	static public StepExecutionResource toResource(StepExecution entity) {
-		return new StepExecutionResource(entity.getId(), entity, generateStepType(entity));
+		return new StepExecutionResource(entity.getJobExecution().getId(), entity, generateStepType(entity));
 	}
 
 	private static String generateStepType(StepExecution stepExecution) {


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/job/support/StepExecutionResourceBuilder.java
```diff
@@ -34,7 +34,7 @@
 public class StepExecutionResourceBuilder {
 
 	static public StepExecutionResource toResource(StepExecution entity) {
-		return new StepExecutionResource(entity.getId(), entity, generateStepType(entity));
+		return new StepExecutionResource(entity.getJobExecution().getId(), entity, generateStepType(entity));
```
jvalkeal: Remind me, why `id` for step execution should be one from job execution and not from its own entity?

cppwfs: IIRC it's because jobExecutionId is the first path variable in the URL.

jvalkeal: what path has to do with step execution id? I don't really understand what this is fixing.

cppwfs: The jobExecutionId that is returned from StepExecutionController has the stepExecutionId.  This PR resolves that issue.

jvalkeal: Ah right, constructor assumes jobExecutionId not entity id. I'll merge this.

jvalkeal:
Merged per https://github.com/spring-cloud/spring-cloud-dataflow/commit/5e65b106876b8f59fd00a10919c26b509f01c8a9

PR #1638, opened 2017-09-08
Move common security configs into spring-common-security-config

### LINKED ISSUE #1481: Move the common security infrastructure to a separate project
As a developer, I'd to study how to split and move the common security infrastructure to a library.

1) Remove Spring Boot's security utilities
2) Move the common code to a separate project

ilayaperumalg:
- Move all the extended security configuration classes into a separate project `spring-common-security-config` so that other projects could re-use them
 - Use `spring-common-security-config-web` module for the server security configurations
 - Update `spring-common-security-config-starter-web` as a dependency
 - Update tests

Add customizable properties for authorization

 - Use the authenticated/permitted path properties in config
 - Refactor security configuration to support the existing `spring.cloud.dataflow.security` prefix for the security configuration properties
 - Update the tests

Use the actual properties/configuration classes

 - instead of subclassing

Resolves #1481


pom.xml
```diff
@@ -126,6 +128,11 @@
 			</dependency>
 			<dependency>
 				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-common-security-config-starter-web</artifactId>
```
jvalkeal: I was a bit too quick to merge that rename pr in spring-cloud-common-security-config as now that I look at this name, it's wrong. If using convention then it should be `spring-cloud-starter-common-security-config-web` or any other shorter combination which makes sense like `spring-cloud-starter-common-security`. It just needs to start with `spring-cloud-starter-`.

ilayaperumalg: @jvalkeal we can have that name change if that's the convention. The name `spring-cloud-starter-common-security` is generic and that's the reason, we added `common-security-config-web` to work on commong web security configuration. Are you ok with the name `spring-cloud-starter-common-security-config-web` (apart from it being long  ;)) ?

spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
```diff
@@ -23,6 +23,7 @@ spring:
       security:
         authorization:
           enabled: true
+          permit-all-paths: "/authenticate,/security/info,/features,/assets/**"
```
jvalkeal: Where is this used or who reads it?

ilayaperumalg: This is from the [config](https://github.com/spring-cloud/spring-cloud-common-security-config/blob/master/spring-cloud-common-security-config-web/src/main/java/org/springframework/cloud/common/security/AuthorizationProperties.java#L46) and used in the auth configurations. For instance [this](https://github.com/spring-cloud/spring-cloud-common-security-config/blob/master/spring-cloud-common-security-config-web/src/main/java/org/springframework/cloud/common/security/BasicAuthSecurityConfiguration.java#L93) is one of the instances.

jvalkeal:
I don't think I have anything else to nitpick. @markpollack u have something to say?

sabbyanandan:
@jvalkeal: We have secured 💚  from @markpollack, so please go ahead and merge if all the comments are addressed.

jvalkeal:
ok, merging

jvalkeal:
Merged per https://github.com/spring-cloud/spring-cloud-dataflow/commit/c679afe90a781f6db74d0ec59cba22891e7e8348

PR #1646, opened 2017-09-13
Add search filter to app list

### LINKED ISSUE #1617: Add search parameter to AppRegistryController.list()
Currently the `AppRegistryController` does not have the ability to filter the results. We should add the `search` parameter similar to the `StreamDefinitionController`.

jvalkeal:
- Add simple search match by apps name. Can't
  really filter by address as it would give
  unrealistic results together with name.
- Fix some issues with pagination i.e. when page
  which doesn't exist is requested. Also make paging
  work with filtered set.
- Fixes #1617

### DIFF
--- a/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/AppRegistryControllerTests.java
+++ b/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/AppRegistryControllerTests.java
@@ -154,6 +154,24 @@ public class AppRegistryControllerTests {
 	}
 
 	@Test
+	public void testListApplicationsBySearch() throws Exception {
+		mockMvc.perform(get("/apps?search=timestamp").accept(MediaType.APPLICATION_JSON)).andDo(print()).andExpect(status().isOk())
+				.andExpect(jsonPath("content", hasSize(1)));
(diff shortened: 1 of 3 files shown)


spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/AppRegistryControllerTests.java
```diff
@@ -154,6 +154,24 @@ public void testListApplicationsByType() throws Exception {
 	}
 
 	@Test
+	public void testListApplicationsBySearch() throws Exception {
+		mockMvc.perform(get("/apps?search=timestamp").accept(MediaType.APPLICATION_JSON)).andDo(print()).andExpect(status().isOk())
```
cppwfs: we can remove the print().  I can do that on merge.

jvalkeal: Indeed there is a lot of other tests to log these as well. Feel free to remove those as well as we don't need to be so noisy.

cppwfs:
LGTM

cppwfs:
Polished, rebased, squashed, merged.

PR #1651, opened 2017-09-13
Add old images back for previous doc releases.  Add versioned images …

### LINKED ISSUE #1648: 1.2.3.RELEASE Dashboard screens are replaced by 1.3.x WIP
As a user, while reviewing 1.2.3.RELEASE docs, I realized the dashboard screenshots are mixed with 1.2.x and 1.3.x WIP -- see [here](https://docs.spring.io/spring-cloud-dataflow/docs/1.2.3.RELEASE/reference/htmlsingle/#dashboard).

**Notes:**
Turns out the images aren't versioned since we are directly pulling them as raw GH links [eg: `{dataflow-asciidoc}/images/dataflow-dashboard-about.png`]. 

**Acceptance:**
Maybe we could [revert this commit](https://github.com/spring-cloud/spring-cloud-dataflow/commit/6ef976948752e098eb4512075162b221232b6c48) and add 1.3.x screenshots with a `1-3` suffix or something similar. That way, we can avoid another release of Local-server.

cppwfs:
…for current docs.

resolves #1648


jvalkeal:
I think this whole approach for images is totally wrong. I think real problem why I think images are mixed is that for snapshots both `1.2.x` and `1.3.x` simply resolve import images from a master. Now that `1.2.x` is built from branch `1.2.x`, it should resolve images from that branch, not from `master`, and `1.3.x` should continue resolve images from `master`.

It's a little different how project itself resolves images as it needs to use its own branch as tags are not yet available(tag is added after successful release). While server impl(k8s,pcf) would resolve same images using tagged versions. 

I believe correct approach is to modify `1.2.x` build scripts to resolve its images from branch `1.2.x` as that is its main source, not master.

When we originally started to import these pages from github I said that this approach is not sustainable for a long rung and now we're starting to see the impact.

jvalkeal:
So might be that in `spring-cloud-dataflow-docs/src/main/asciidoc/index.adoc` `:dataflow-asciidoc: https://raw.githubusercontent.com/spring-cloud/spring-cloud-dataflow/master/spring-cloud-dataflow-docs/src/main/asciidoc` needs to point to `2.0.x`. Then when server impl release is done, those get same images from a tagged version(which I believe already happens). If server impl releases use scdf master for anything, then those builds needs to get fixed.

cppwfs:
@jvalkeal I was thinking the same thing.  executing full build locally, then once complete I'll close this PR and open a new one the 1.2.x branch.

cppwfs:
Closing this PR and opened a new one on the 1.2.x branch here: https://github.com/spring-cloud/spring-cloud-dataflow/pull/1652

PR #1660, opened 2017-09-18
Fix access to oauth logout notify page

### LINKED ISSUE #1659: SCDF tries to authenticate logout-success page
In `1.2.3.RELEASE` I can access: `/dashboard/logout-success-oauth.html` without being authenticated. In the latest snapshot version, I was logging out and surprised that I was still logged in after logging out. It looks like `/logout` actually logout, then redirect to `/dashboard/logout-success-oauth.html` which however will redirect to the OAuth server for authentication (e.g. `/uaa/oauth/authorize`).

`1.2.3.RELEASE`:

image

`1.3.0.M2`:

image

jvalkeal:
- Add logout-success-oauth.html to permitted pages.
- Fixes #1659


ilayaperumalg:
LGTM

PR #1675, opened 2017-10-04
prevents child tasks from being created if apps do not exist

### LINKED ISSUE #1669: Subtasks are created even when overall task definition is invalid
Version: 1.2.4.RELEASE
Repro steps: 

3. In SCS: task list

cppwfs:
in the case where a user enters a valid composed task definition but contains an app that does not exist, we should not create child task definitions until we have validated that all task apps are present.  For example task create bar --definition "AAA && BBB  'COMPLETED' -> $END" should not create AAA nor BBB since $END does not exist.

resolves #1669


sabbyanandan:
@cppwfs: This PR needs to be reissued as well.

cppwfs:
@sabbyanandan Thanks !

PR #1684, opened 2017-10-09
Added --useSkipper to deploy command in the shell.

### LINKED ISSUE #1680: Support useSkipper for Stream deploy command
As a user, I'd like to specify `stream deploy <stream> --useSkipper`

cppwfs:
resolves #1680

### DIFF
--- a/spring-cloud-dataflow-shell-core/pom.xml
+++ b/spring-cloud-dataflow-shell-core/pom.xml
@@ -23,6 +23,10 @@
 			<artifactId>spring-cloud-dataflow-rest-client</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-dataflow-server-core</artifactId>
+		</dependency>
+		<dependency>
 			<groupId>org.springframework.shell</groupId>
 			<artifactId>spring-shell</artifactId>
 		</dependency>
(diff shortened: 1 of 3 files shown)


spring-cloud-dataflow-shell-core/pom.xml
```diff
@@ -23,6 +23,10 @@
 			<artifactId>spring-cloud-dataflow-rest-client</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>org.springframework.cloud</groupId>
```
ilayaperumalg: This dependency may not be needed at the shell core level as it brings in the entire server core as the dependency.

spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/StreamCommands.java
```diff
@@ -150,11 +154,19 @@ public String deployStream(
...
+				propertiesToUse = new HashMap<>(1);
+			}
+			else {
+				propertiesToUse = Collections.<String, String>emptyMap();
+			}
 			break;
 		default:
 			throw new AssertionError();
 		}
+		if(useSkipper) {
```
ilayaperumalg: While this helps to send the `actual` property as the deployment property, we could still keep the boolean flag instead of the actual property which comes from the `data flow server core`.

ilayaperumalg:
@cppwfs I would like to submit a PR with the changes suggested so that you can review it and we can take the PR from there.

cppwfs:
closing and will review Ilaya's PR.

Issue #1688, opened 2017-10-11
Support configurable prefix for release name when deploying stream using Skipper


ilayaperumalg:
Currently, we set `my` as the prefix to the `streamName` for the release name when deploying the stream using Skipper. We need a configurable property for this prefix.

ilayaperumalg:
This property should be aligned with the client side as well. For instance, `stream update` uses `releaseName` to create the StreamUpdateRequest.

markpollack:
I don't understand the previous comment.  Maybe the default should be `scdf-` ?

ilayaperumalg:
What I meant there was if we make a configuration property for the skipper release name prefix at the server side, the same needs to be brought to the shell side as well.

Now on a second thought, I think we may not need server-side property and just a prefix as a shell option. 

But, sticking to `scdf-` as the prefix would be a better option for now as well.

PR #1691, opened 2017-10-12
Record stream deployment

### LINKED ISSUE #1681: Record stream deployments
As a developer, I'd like to establish the core schema changes to persist stream deployments.

ilayaperumalg:
- Add StreamDeployment repository that keeps track stream deployments
   - Define necessary operations in StreamDeploymentRepository interface
   - Implement both RDBMS and in-memory (for test only)
   - Define DDL for supported DB schemas
 - Add StreamDeployment domain object
 - Verify StreamDeployment entry when upgrading stream
 - Verify StreamDeployment when calculating state
 - Remove usage of USE_SKIPPER as the system property
 - Add tests to verify both the upgrade and state in DefaultStreamService

Resolves #1681


ilayaperumalg:
Squashed and rebased.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultStreamService.java
```diff
@@ -416,17 +426,45 @@ private boolean isPartitionedConsumer(Map<String, String> appDeploymentPropertie
...
+			}
+			else {
+				if (streamDeployment.getDeployerName().equals(StreamDeployers.skipper.name())) {
+					skipperStreams.add(streamDefinition);
+				}
+				else if (streamDeployment.getDeployerName().equals(StreamDeployers.appdeployer.name())){
+					appDeployerStreams.add(streamDefinition);
+				}
+				else {
+					logger.error("Unknown deployer " + streamDeployment.getDeployerName() +
```
cppwfs: Just wondering if we should throw an exception here.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/stream/SkipperStreamDeployer.java
```diff
@@ -184,18 +192,23 @@ public void deployStream(StreamDeploymentRequest streamDeploymentRequest) {
...
 		packageIdentifier.setPackageVersion("1.0.0");
-		packageIdentifier.setRepositoryName("local");
+		String repoName = "local";
+		packageIdentifier.setRepositoryName(repoName);
 		installRequest.setPackageIdentifier(packageIdentifier);
 		InstallProperties installProperties = new InstallProperties();
 		installProperties.setPlatformName("default");
 		//todo: Better naming for the release name
-		installProperties.setReleaseName("my" + streamDeploymentRequest.getStreamName());
+		String releaseName = "my" + streamName;
```
cppwfs: Should `my` be something like `SCDF`?

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentControllerTests.java
```diff
@@ -81,10 +82,14 @@
 	@Mock
 	private DefaultStreamService defaultStreamService;
 
+	@Mock
+	private StreamDeploymentRepository stramDeploymentRepository;
```
cppwfs: `streamDeploymentRepository` is misspelled.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/repository/InMemoryStreamDeploymentRepository.java
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright 2015 the original author or authors.
```
cppwfs: 2017 ;-)

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/services/impl/DefaultStreamServiceTests.java
```diff
@@ -0,0 +1,174 @@
+/*
+ * Copyright 2015-2017 the original author or authors.
```
cppwfs: This is a new class so just 2017.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/services/impl/DefaultStreamServiceTests.java
```diff
@@ -0,0 +1,174 @@
...
+		StreamDeployment streamDeployment3 = new StreamDeployment(streamDefinition3.getName(),
+				StreamDeployers.skipper.name(), "pkg1", "release2", "local");
+		when(streamDeploymentRepository.findOne(streamDeployment1.getStreamName())).thenReturn(streamDeployment1);
+		when(streamDeploymentRepository.findOne(streamDeployment2.getStreamName())).thenReturn(streamDeployment2);
+		when(streamDeploymentRepository.findOne(streamDeployment3.getStreamName())).thenReturn(streamDeployment3);
+	}
+
+	@Test
+	public void verifyUpgradeStream() {
+		StreamDeploymentRepository streamDeploymentRepository = mock(StreamDeploymentRepository.class);
```
cppwfs: Why aren't we using the mocks we setup in the `@Before`?

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/services/impl/DefaultStreamServiceTests.java
```diff
@@ -0,0 +1,174 @@
...
+				null, null);
+		verify(skipperStreamDeployer, times(1)).upgradeStream(streamDeployment2.getStreamName(),
+				streamDeployment2.getReleaseName(), null, null);
+		verifyNoMoreInteractions(skipperStreamDeployer);
+		verify(appDeployerStreamDeployer, never()).deployStream(any());
+	}
+
+	@Test
+	public void verifyAppDeployerUpgrade() {
+		StreamDeploymentRepository streamDeploymentRepository = mock(StreamDeploymentRepository.class);
```
cppwfs: Why aren't we using the mocks we setup in the `@Before`?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultStreamService.java
ilayaperumalg: Currently, we have the pre-defined deployer names and it is most likely to have this path of the code to execute. Given this is for state() calculation and throwing an exception for one incorrect entry may not be necessary. Hence, just error logging.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/stream/SkipperStreamDeployer.java
ilayaperumalg: We have a story to cover this: https://github.com/spring-cloud/spring-cloud-dataflow/issues/1688

ilayaperumalg:
@cppwfs thank you for the review. Pushed the changes that address review comments.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultStreamService.java
cppwfs: Understood.   I guess the log is enough, just was concerned about leaving an entry out of the result.

cppwfs:
I'll kick the tires a bit more.   But would like @jvalkeal or @markpollack to take give it a once over before merging.

jvalkeal:
@cppwfs I think from my side this looks ok.

cppwfs:
👍

cppwfs:
LGTM

cppwfs:
Squashed Merged.

PR #1693, opened 2017-10-13
Add package name/version for stream deploy

### LINKED ISSUE #1686: Expose repository name and package version when deploying using skipper
May want to only have these options when using `stream deploy` and not `stream create` , otherwise the stream create command gets too many options related to deployment.

The hardcoded package version (1.0.0) and repository name (local) used when deploying at stream creating time can maybe used and make configurable for `stream create`

### LINKED ISSUE #1698: Add support to specify platform name for stream deploy/update
Currently, platform name for install/update requests is set to `default` which is for `local` app deployer. We need to get the `platform-name` option from the client.

ilayaperumalg:
Add package name/version/platformName/repoName for stream deploy

  - When deploying stream via Skipper, specify package name, version, platformName and repoName
        - By default, package name will be the stream name
        - package version is a mandatory property when using Skipper
     - Set package name and version as skipper deployer properties
        - Use the skipper deployer properties to set Upload/InstallRequest's package name/version,
    platform name and repo name
        - For platform name use `default` as the default if no value is specified
        - For repo name use `local` as the default
        - Make sure these properties are updated when creating the package

    - Add tests

  Resolves #1686
  Resolves #1698


ilayaperumalg:
Rebased this PR and also addressed #1698 along with this PR as we already address platorm name/version support here in this PR.

PR #1696, opened 2017-10-16
Support undeploy/destroy operation via StreamService

### LINKED ISSUE #1682: Add support to undeploy a stream deployed by skipper

### LINKED ISSUE #1683: Add support to destroy a stream deployed by skipper

ilayaperumalg:
- Have the StreamDeploymentController delegate the undeploy operation via StreamService
 - Add SPI contract `undeployStream` at StreamService and StreamDeployer
 - Implement `undeployStream` in SkipperStreamDeployer to invoke `delete` release REST endpoint after constructing the release name
 - Move AppDeployerStreamDeployer specific code
 - Add tests

Resolves #1682
Resolves #1683

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultStreamService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultStreamService.java
@@ -78,8 +78,8 @@ public class DefaultStreamService implements StreamService {
 
 	/**
 	 * This is the spring boot property key that Spring Cloud Stream uses to filter the
-	 * metrics to import when the specific Spring Cloud Stream "applicaiton" trigger is fired
-	 * for metrics export.
+	 * metrics to import when the specific Spring Cloud Stream "applicaiton" trigger is
+	 * fired for metrics export.
 	 */
 	private static final String METRICS_TRIGGER_INCLUDES = "spring.metrics.export.triggers.application.includes";
 
@@ -141,6 +141,24 @@ public class DefaultStreamService implements StreamService {
 		deployStreamWithDefinition(createStreamDefinitionForDeploy(name, deploymentProperties), deploymentProperties);
 	}
 
+	@Override
+	public void undeployStream(String streamName) {
+		StreamDeployment streamDeployment = this.streamDeploymentRepository.findOne(streamName);
+		if (streamDeployment != null) {
+			switch (StreamDeployers.valueOf(streamDeployment.getDeployerName())) {
+			case appdeployer: {
+				this.appDeployerStreamDeployer.undeployStream(streamName);
+				break;
+			}
+			case skipper: {
+				this.skipperStreamDeployer.undeployStream(streamName);
+				break;
+			}
+			}
+			this.streamDeploymentRepository.delete(streamName);
+		}
+	}
+
 	private StreamDefinition createStreamDefinitionForDeploy(String name,
 			Map<String, String> deploymentProperties) {
 		StreamDefinition streamDefinition = this.streamDefinitionRepository.findOne(name);
@@ -151,10 +169,10 @@ public class DefaultStreamService implements StreamService {
 		Map<String, String> skipperDeploymentProperties = getSkipperProperties(deploymentProperties);
 		String status;
 		if (skipperDeploymentProperties.containsKey(SkipperStreamDeployer.SKIPPER_ENABLED_PROPERTY_KEY)) {
-			status = skipperStreamDeployer.calculateStreamState(name);
+			status = this.skipperStreamDeployer.calculateStreamState(name);
 		}
 		else {
-			status = appDeployerStreamDeployer.calculateStreamState(name);
+			status = this.appDeployerStreamDeployer.calculateStreamState(name);
 		}
 
 		if (DeploymentState.deployed.equals(DeploymentState.valueOf(status))) {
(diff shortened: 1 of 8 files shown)


ilayaperumalg:
Please note that this PR is built on top of #1691

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultStreamService.java
```diff
@@ -141,6 +141,24 @@ public void deployStream(String name, Map<String, String> deploymentProperties)
 		deployStreamWithDefinition(createStreamDefinitionForDeploy(name, deploymentProperties), deploymentProperties);
 	}
 
+	@Override
+	public void undeployStream(String streamName) {
+		StreamDeployment streamDeployment = this.streamDeploymentRepository.findOne(streamName);
+		if (streamDeployment != null) {
```
cppwfs: Do we need to throw an Exception on the case where a Unknown Deployer Name is specified.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/stream/SkipperStreamDeployer.java
```diff
@@ -307,6 +307,12 @@ private Package createDependentPackage(String streamName, AppDeploymentRequest a
 		return templateList;
 	}
 
+	@Override
+	public void undeployStream(String streamName) {
+		String releaseName = "my" + streamName;
```
cppwfs: Mentioned it in previous PR.  Would `SCDF` be a better name instead of `my`?

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/services/impl/DefaultStreamServiceTests.java
```diff
@@ -126,6 +126,34 @@ public void verifyUpgradeStream() {
 	}
 
 	@Test
+	public void verifyUndeployStream() {
```
cppwfs: Can we use he mocks setup in the `@Before`?

ilayaperumalg:
@cppwfs Since this PR is on top of #1691, I have rebased this PR after addressing the review comments.

cppwfs:
Kicked the tires a little this morning and found one issue.  I may have missed something though.
1) pulled the latest Skipper, built it, and started it.
2) pulled data flow from the latest update to this PR, built it, started it
3) `stream deploy bar --useSkipper`
    a) Skipper deployed bar
    b) STREAM_DEPLOYMENTS shows the entry
    c) apps are deployed to local system
4) `stream undeploy bar`
    a) Skipper undeploys bar
    b) STREAM_DEPLOYMENTS entry is removed
    c) apps are removed from local system
5) `stream deploy bar`
    a) Skipper gives me a 500 error:
`org.springframework.cloud.skipper.SkipperException: Release with the name [mybar2] already exists.`

sabbyanandan:
I'm used to supplying `-DUSE_SKIPPER` system property while starting SCDF. I'm not sure if it is still needed, though, but the `undeploy` needs the hint to use skipper at runtime. Maybe that's the catch here.

cppwfs:
@sabbyanandan that the interesting thing here.  Skipper is getting called when we do the undeploy.   And all apps destroyed on the local system.

sabbyanandan:
Ah, it uses the [persisted deployer name](https://github.com/spring-cloud/spring-cloud-dataflow/pull/1696/files#diff-c624577d4ca1517d83a6b0075651c249R148) to dynamically route to the respective backend service. So, no need of the system property anymore.

sabbyanandan:
@ilayaperumalg: Any thoughts on the last comment?

ilayaperumalg:
@cppwfs That scenario is expected and will be fixed once we have this https://github.com/spring-cloud/spring-cloud-skipper/issues/193 merged.

@sabbyanandan Yes, that is correct. With the persistence layer added to track of stream deployments, we no longer need system property.

cppwfs:
If you could rebase this to master then I think we are set.

ilayaperumalg:
@cppwfs Rebased, thanks!

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/services/impl/DefaultStreamServiceTests.java
cppwfs: Will fix this on merge.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultStreamService.java
cppwfs: Based on code I've seen in Skipper.  The answer is no.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/services/impl/DefaultStreamServiceTests.java
ilayaperumalg: @cppwfs I thought this was fixed along with the other PR, no?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultStreamService.java
ilayaperumalg: We have restrictions when storing the StreamDeployment entry as well.

cppwfs:
LGTM.  Polishing the tests based on code review comments.

cppwfs:
Polished, Squashed, Merged.

PR #1704, opened 2017-10-20
Add package name/version/platformName/repoName for stream deploy

### LINKED ISSUE #1686: Expose repository name and package version when deploying using skipper

### LINKED ISSUE #1698: Add support to specify platform name for stream deploy/update

ilayaperumalg:
- When deploying stream via Skipper, specify package name, version, platformName and repoName
    - By default, package name will be the stream name
    - package version is a mandatory property when using Skipper
 - Set package name and version as skipper deployer properties
    - Use the skipper deployer properties to set Upload/InstallRequest's package name/version,
platform name and repo name
    - For platform name use `default` as the default if no value is specified
    - For repo name use `local` as the default
    - Make sure these properties are updated when creating the package
 - Update packageVersion and repoName options for `stream update` command

- Add tests

Resolves #1686
Resolves #1698

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/stream/SkipperStreamDeployer.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/stream/SkipperStreamDeployer.java
@@ -66,8 +66,6 @@ import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
 import org.springframework.web.client.HttpStatusCodeException;
 
-import static org.springframework.cloud.deployer.spi.app.AppDeployer.COUNT_PROPERTY_KEY;
-
 /**
  * Delegates to Skipper to deploy the stream.
  *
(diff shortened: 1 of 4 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/stream/SkipperStreamDeployer.java
```diff
@@ -172,34 +178,40 @@ public String calculateStreamState(String streamName) {
...
 		logger.info("Deploying Stream " + streamDeploymentRequest.getStreamName() + " using skipper.");
+		Map<String, String> streamDeployerProperties = streamDeploymentRequest.getStreamDeployerProperties();
+		String repoName = streamDeployerProperties.get(SKIPPER_PACKAGE_REPO_NAME);
+		repoName = (StringUtils.hasText(repoName)) ? (repoName) : "local";
+		String platformName = streamDeployerProperties.get(SKIPPER_PLATFORM_NAME);
+		platformName = (StringUtils.hasText(platformName)) ? platformName : "default";
+		String packageName = streamDeployerProperties.get(SKIPPER_PACKAGE_NAME);
+		packageName = (StringUtils.hasText(packageName)) ? packageName : streamDeploymentRequest.getStreamName();
+		String packageVersion = streamDeployerProperties.get(SKIPPER_PACKAGE_VERSION);
+		packageVersion = (StringUtils.hasText(packageVersion)) ? packageVersion : "1.0.0";
```
jvalkeal: This default `1.0.0` looks a bit wrong here.

jvalkeal:
Merged with fixing conflict per https://github.com/spring-cloud/spring-cloud-dataflow/commit/eaf275b7e05e5310b76df3a18f51d90179125f31

PR #1705, opened 2017-10-20
prevents child tasks from being created if apps dont exist

### LINKED ISSUE #1669: Subtasks are created even when overall task definition is invalid

cppwfs:
in the case where a user enters a valid composed task definition but contains an app that does not exist, we should not create child task definitions until we have validated that all task apps are present.  For example task create bar --definition "AAA && BBB  'COMPLETED' -> $END" should not create AAA nor BBB since $END does not exist.

resolves #1669

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
@@ -275,6 +275,10 @@ public class DefaultTaskService implements TaskService {
 		TaskParser taskParser = new TaskParser(name, dsl, true, true);
 		TaskNode taskNode = taskParser.parse();
 		if (taskNode.isComposed()) {
+			// verify that the task apps required for composed task exist
+			taskNode.getTaskApps().stream().forEach(task -> {
+				validateAppExists(task.getName());
+			});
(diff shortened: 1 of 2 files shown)


mminella:
Am I missing something or how are we handling the concurrent aspect of the server?  I don't see any synchronization around this operation via a transaction/etc.

cppwfs:
Not sure Data Flow handles concurrent issues as of yet.  I'm not seeing `@Transactional` annotations for example.

mminella:
But this type of things is exactly where that arises.  We probably have a similar issue when deploying streams.  If a stream of 5 apps starts to deploy and then an app is deleted while it's happening, do we undeploy the others?  This issue is about the creation of an atomic unit of work.  This fix doesn't do that IMHO.

cppwfs:
I think what you are describing is valid however, this is not what this PR is meant to address per se'.   The key here is that someone creates a CTR with an invalid task.   for example `AAA && BBB && BAD`.  Without this bug fix `FOO-AAA` and `FOO-BBB` are created, but then Data Flow fails trying to create 'FOO-BAD`.  i.e. we should check to see if AAA, BBB and BAD are there first before even beginning the process.

cppwfs:
Closing this PR as it does not resolve all the scenarios.  Look at https://github.com/spring-projects/spring-retry

mminella:
I'm closing this until @cppwfs finishes his work with Spring Retry.

PR #1722, opened 2017-10-25
Added DSL package and base classes

### LINKED ISSUE #1620: Move Java DSL bits from acceptance-tests project to SCDF
Currently acceptance-tests project provides a Java DSL capable of constructing and executing the flow. The DSL itself may be useful to an everyday developer, since it would allow them to create, modify and execute flows without leaving their IDE. Further more it could also provide for a versioned artifact that represents executable flow as well as an artifact that could allow deployment options (local vs PCF) to be managed via configuration.

viniciusccarvalho:
Fixes #1620 

This adds a Fluent DSL around the `DataFlowTemplate`.  

Users can use a fluent style using `from`, `via`, `to` methods to chain a builder, or simply override it using `definition(dslDefinition)`

To build (but not create on the server) a Stream definition:

```java
DataFlowTemplate template = ...;
StreamApplication time = new StreamApplication("time")
    .addProperty("fixed-interval", 5000);

Stream ticktock = Stream.Builder(template)
    .name("ticktock")
    .from(time)
    .to(new StreamApplication("log")
    .build();

//create:

stream.create();
stream.deploy();

//deploying:
String status = stream.getStatus();

...
//invokes server to fetch latest status
stream.refreshStatus();

stream.undeploy();

stream.destroy();

```

### DIFF
--- a/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/dsl/Stream.java
+++ b/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/dsl/Stream.java
@@ -1,0 +1,323 @@
+package org.springframework.cloud.dataflow.rest.client.dsl;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+
+import org.springframework.cloud.dataflow.rest.client.DataFlowOperations;
+import org.springframework.cloud.dataflow.rest.resource.StreamDefinitionResource;
+import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
+
+/**
+ * @author Vinicius Carvalho
+ */
+public class Stream {
+
+	private String name;
+
+	private List<StreamApplication> applications = new LinkedList<>();
+
+	private String definition;
+
+	private DataFlowOperations client;
+
+	private Stream(String name, List<StreamApplication> applications, String definition,
+			DataFlowOperations client) {
+		this.name = name;
+		this.applications = applications;
+		this.definition = definition;
+		this.client = client;
+	}
+
+	public static Builder builder(DataFlowOperations client) {
+		return new Builder(client);
+	}
+
+	String getDefinition() {
+		return this.definition;
+	}
+
+	public StreamBuilder undeploy() {
+		client.streamOperations().undeploy(this.name);
+		return new StreamBuilder(this.name, this.client, this.definition,
+				this.applications);
+	}
+
+	public void destroy() {
+		client.streamOperations().destroy(this.name);
+	}
+
+	public String getStatus() {
+		StreamDefinitionResource resource = client.streamOperations()
+				.getStreamDefinition(this.name);
+		return resource.getStatus();
+	}
+
+	public static class Builder {
+
+		private DataFlowOperations client;
+
+		private Builder(DataFlowOperations client) {
+			this.client = client;
+		}
+
+		public StreamNameBuilder name(String name) {
+			return new StreamNameBuilder(name, client);
+		}
+
+	}
+
+	public static class StreamNameBuilder {
+
+		private String name;
+
+		private List<StreamApplication> applications = new LinkedList<>();
+
+		private DataFlowOperations client;
+
+		private String definition;
+
+		private StreamNameBuilder(String name, DataFlowOperations client) {
+			this.client = client;
+			Assert.hasLength(name, "Stream name can't be empty");
+			this.name = name;
+		}
+
+		public SourceBuilder source(StreamApplication source) {
+			Assert.notNull(source, "Source application can't be null");
+			return new SourceBuilder(
+					source.type(StreamApplication.ApplicationType.SOURCE), this);
+		}
+
+		public SourceBuilder source(String name) {
+			Assert.isTrue(StringUtils.hasLength(name),
+					"Source application can't be null");
+			StreamApplication source = new StreamApplication(name);
+			return new SourceBuilder(
+					source.type(StreamApplication.ApplicationType.SOURCE), this);
+		}
+
+		/**
+		 * Creates a Stream bypassing the fluent API and just using the provided
+		 * definition
+		 * @param definiton the Stream definition to use
+		 * @return A {@link Stream} object
+		 */
+		public StreamDefinitionBuilder definition(String definiton) {
+			Assert.hasLength(name, "Stream definition can't be empty");
+			this.definition = definiton;
+			return new StreamDefinitionBuilder(this.name, this.client, this.definition);
+		}
+
+		private StreamBuilder create() {
+			return new StreamBuilder(this.name, this.client, this.definition,
+					this.applications);
+		}
+
+		private void addApplication(StreamApplication application) {
+			if (contains(application)) {
+				throw new IllegalStateException(
+						"There's already an application with the same definition in this stream");
+			}
+			this.applications.add(application);
+		}
+
+		private boolean contains(StreamApplication application) {
+			for (StreamApplication app : this.applications) {
+				if (app.getType().equals(application.getType())
+						&& app.getIdentity().equals(application.getIdentity())) {
+					return true;
+				}
+			}
+			return false;
+		}
+	}
+
+	public static class StreamDefinitionBuilder {
+
+		private String name;
+
+		private DataFlowOperations client;
+
+		private String definition;
+
+		private StreamDefinitionBuilder(String name, DataFlowOperations client,
+				String definition) {
+			this.name = name;
+			this.client = client;
+			this.definition = definition;
+		}
+
+		public StreamBuilder create() {
+			return new StreamBuilder(this.name, this.client, this.definition,
+					Collections.emptyList());
+		}
+	}
+
+	public static class StreamBuilder {
+
+		private String name;
+
+		private DataFlowOperations client;
+
+		private String definition;
+
+		private List<StreamApplication> applications = new LinkedList<>();
+
+		private StreamBuilder(String name, DataFlowOperations client, String definition,
+				List<StreamApplication> applications) {
+			this.name = name;
+			this.client = client;
+			this.definition = definition;
+			this.applications = applications;
+			if (StringUtils.isEmpty(definition)) {
+				createStreamDefinition();
+			}
+			this.client.streamOperations().createStream(this.name, this.definition,
+					false);
+		}
+
+		public void destroy() {
+			this.client.streamOperations().destroy(this.name);
+		}
+
+		public Stream deploy(Map<String, String> deploymentProperties) {
+			Map<String, String> resolvedProperties = resolveDeploymentProperties(
+					deploymentProperties);
+			client.streamOperations().deploy(this.name, resolvedProperties);
+			return new Stream(this.name, this.applications, this.definition, this.client);
+		}
+
+		public Stream deploy() {
+			return deploy(null);
+		}
+
+		private void createStreamDefinition() {
+			StringBuilder buffer = new StringBuilder();
+			this.definition = StringUtils.collectionToDelimitedString(applications,
+					" | ");
+		}
+
+		/**
+		 * Concatenates any deployment properties from the apps with a given map used
+		 * during {@link StreamBuilder#deploy(Map)}
+		 * @return
+		 */
+		private Map<String, String> resolveDeploymentProperties(
+				Map<String, String> deploymentProperties) {
+			Map<String, String> properties = new HashMap<>();
+			if (deploymentProperties != null) {
+				properties.putAll(deploymentProperties);
+			}
+			for (StreamApplication app : this.applications) {
+				for (Map.Entry<String, Object> entry : app.getDeploymentProperties()
+						.entrySet()) {
+					properties.put(entry.getKey(), entry.getValue().toString());
+				}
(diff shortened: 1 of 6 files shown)


sabbyanandan:
Hi, @kbastani. I shared your [sample](https://github.com/kbastani/event-stream-processing-microservices/blob/master/platform-services/data-flow-server/src/main/java/demo/ImportResources.java) as a reference many times. It'd be great to hear your thoughts on this improvement. 😄

jvalkeal:
I've wondered that why the dsl in tests used `from/via/to`, instead of `source/processor/sink`? Think I chatted with Marius about this a long time ago but don't remember if there were a specific reasoning.
Also having a need to pass `StreamApplication` to those methods kinda feels redundant. It'd be nice if one could write dsl without explicitly defining `StreamApplication` as I'd assume builder could create it behind a scenes.

spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/dsl/Stream.java
```diff
@@ -0,0 +1,250 @@
...
+
+	public Stream create() {
+		StreamDefinitionResource definition = this.client.streamOperations().createStream(this.name, getDefinition(), false);
+		this.status = definition.getStatus();
+		return this;
+	}
+
+	public Stream deploy(Map<String, String> deploymentProperties){
+		Map<String, String> resolvedProperties = resolveDeploymentProperties(deploymentProperties);
+		if(this.status == null){
```
jvalkeal: It might be better not to keep any state in this class as it will always be out of sync even if you call `refreshStatus`. For example, create to exactly same dsl's using different builder instances, one of those is always out of sync. You'll end up chasing your own tail.

viniciusccarvalho: Yes keeping in sync will be a challenge but the state is there for a few reasons:

1. We need to know whether we are deploying using a definition bypassing the fluent api, that was a request from Mark
2. In order to support a deploy/create style, for instance to call `create(name, definition, true)` we need to see if there is already a status from a create call, or not.

markpollack: In the sample above
```
Stream ticktock = Stream.Builder(template)
    .name("ticktock")
    .from(time)
    .to(new StreamApplication("log")
    .build();
```
It looks silly to do `new StreamApplication("log")` because no other properties on the StreamApplication were set.  This example would be better represented using the `definition` method on the build

```
		Stream stream = Stream.Builder(client)
				.name("ticktock")
				.definition("time | log")
				.deploy(Collections.singletonMap("deployer.log.count","2"));

```
The idea suggested by @mminella was that we have have the apps declared as `@Bean` in an `@Configuration` class and then the DSL is mostly concerned about wiring them up together and providing deployment properties.  This makes it much easier to reuse the StreamApplication instances across different stream definitions and to more easily see what the stream is doing vs. the DSL style where each individual app was created+properties set+stitched together with from/via/to.

markpollack: I am not remembering the discussion regarding storing the status.   
```
We need to know whether we are deploying using a definition bypassing the fluent api, that was a request from Mark
```
If I was in the shell and deployed 'ticktock' and then executed an application using the DSL which also tried to create `ticktock`, I should get `DuplicateStreamDefinitionException` thrown back to the client.  Can you remind me what I was saying.... ;)  I can certainly see the argument for not storing the state.

```
In order to support a deploy/create style, for instance to call create(name, definition, true) we need to see if there is already a status from a create call, or not.
```
It would be possible to work around this by requiring the users to write

```
		Stream stream = Stream.Builder(client)
				.name("ticktock")
				.definition("time | log")
                                .create()
				.deploy(Collections.singletonMap("deployer.log.count","2"));
```
That is, require the `.create()` method to be called.  This whole shortcut of `--deploy` on create has kind of been a PITA when it came to Skipper integration.  We used to have `stream create BLAH --deploy --useSkipper` as a shortcut but it started to get complex/confusing.  So now, we decided to have a new command `stream skipper deploy` to cleanly separate out the skipper variant of deploying.  To deploy using skipper, you always have to create the stream and then deploy it in two steps.   Following the same approach here would seem to work well, contact is 'clean' and would avoid the need for state to be stored.

Which reminds me... we will have to adopt the DSL to take into account skipper...

markpollack:
It feels a little bit inconvenient to have to bootstrap the builder with `Stream stream = Stream.Builder(dataflowOperations)` which raises the question, where do we get the instance of `dataflowOperations`?  It seems that in the shell we are creating the underlying RestTemplate passed to `new DataFlowTemplate` here - https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/ConfigCommands.java#L192

We don't want to have the user duplicate all that code in order to make an authorized request to the server from this app.  Perhaps DataFlowTemplate should have some autoconfigurion?

Main point being, from an easy of use perspective, having a builder autowired would help to a user's code would look like.

```
@Autowired
public MyCoolApp(StreamDefinitionBuilder streamDefinitionBulder)
```

Even if we don't go as far as to autowire the builder, we need to provide some way to get a properly configured instance of DataFlowTemplate...

Thoughts?

viniciusccarvalho:
Mark I thought we discussed the need for having the dataflowOperations on the DSL over slack. And I thought we decided that instead of having the builder to create a new template everytime, we would require one to be passed from the user.

Having an autoconfiguration does not exclude that, we can provide an auto configured builder so users can inject, but still it would require a dataflowOperations that can be also auto configured using `ConditionalOnMissing`, I think that would solve both problems. Users can inject autoconfigured version of the builder, or manually instantiate one passing a configured dataflow operations.

markpollack:
Wasn't suggesting to drop using dataflowOperations, just to not put the burden on the user to create it.

spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/dsl/Stream.java
```diff
@@ -0,0 +1,262 @@
...
+
+import org.springframework.cloud.dataflow.rest.client.DataFlowOperations;
+import org.springframework.cloud.dataflow.rest.resource.StreamDefinitionResource;
+import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
+
+/**
+ * @author Vinicius Carvalho
+ */
+public class Stream {
```
olegz: Wondering about the name and if it would create confusion and/or inconvenience for cases where the code also uses 'java.util.Stream'

olegz:
Any idea why we decided against simple String representation of stream definition the same way we do in Shell `foo | bar | baz`?

viniciusccarvalho:
We haven't, you can use `definition()` method

spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/dsl/StreamApplication.java
```diff
@@ -0,0 +1,100 @@
...
+
+	private final String name;
+
+	private String label;
+
+	private Map<String, Object> properties = new HashMap<>();
+
+	private Map<String, Object> deploymentProperties = new HashMap<>();
+
+	private StreamType type;
```
ilayaperumalg: This can be `ApplicationType` instead of `StreamType`.

spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/dsl/Stream.java
```diff
@@ -0,0 +1,290 @@
...
+		}
+
+		public ProcessorBuilder processor(StreamApplication processor) {
+			Assert.notNull(processor, "Processor application can't be null");
+			return new ProcessorBuilder(
+					processor.type(StreamApplication.StreamType.PROCESSOR), this.parent);
+		}
+
+		public ProcessorBuilder processor(String name) {
+			Assert.hasLength(name, "Stream definition can't be empty");
```
ilayaperumalg: Can we say, `Application name` instead of `Stream definition`?

ilayaperumalg: It looks like the same message is used in other places where the application name is used. If there is no specific reason for it, we can fix them as well.

spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/dsl/Stream.java
```diff
@@ -0,0 +1,290 @@
...
+		public SinkBuilder sink(String name){
+			Assert.hasLength(name, "Stream definition can't be empty");
+			StreamApplication sink = new StreamApplication(name);
+			return new SinkBuilder(sink.type(StreamApplication.StreamType.SINK),
+					this.parent);
+		}
+
+	}
+
+	public static class ProcessorBuilder extends BaseBuilder {
```
ilayaperumalg: We could add some check in the processor builder if the same name is repeated for the application name for the given Stream and suggest using a label instead. This way, we can fail fast on the client side.

markpollack:
```@Autowired
private Stream.StreamNameBuilder builder;```

This is too long of a name to access the builder.  Assuming that Stream gets autoconfigured with a client.  Shouldn't this be `Stream.StreamBuilder` or better yet, shorten it to `Stream.Builder`

File: spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/dsl/Stream.java [hunk 6]
```diff
@@ -0,0 +1,313 @@
...
+	private String name;
+
+	private List<StreamApplication> applications = new LinkedList<>();
+
+	private String definition;
+
+	private DataFlowOperations client;
+
+	private Stream(String name, List<StreamApplication> applications, String definition,
+			DataFlowOperations client) {
```
markpollack: rename method argument to 'dataFlowOperations' here and elsewhere

viniciusccarvalho:
@markpollack I think there's a classical chicken and egg problem entangling all the DSL stories. Docs are missing auto-configuration because the autoconfiguration is still not approved. I think if we can get the DSL merged, then work on the auto configuration story that will provide the auto configured builder, and finally check the docs.

viniciusccarvalho:
@markpollack The latest commit fixed the name from `StreamNameBuilder` to `StreamBuilder`

markpollack:
It isn't chicken an egg.  If autoconfig is for RC1, then the docs for M3 needs to work with M3.

markpollack:
When creating the stream, after `.create()` I was expecting only to see the option to deploy.  Not to be able to `destroy`, `undeploy`, `getStatus`, or `getDefinition`.    Not sure what `namedChannelFor` is about in this context as well.  Attached screenshot.

screenshot-dsl

spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/dsl/Stream.java
```diff
@@ -0,0 +1,313 @@
...
+		this.name = name;
+		this.applications = applications;
+		this.definition = definition;
+		this.client = client;
+	}
+
+	public static Builder Builder(DataFlowOperations client) {
```
markpollack: should be lowercase `builder`

spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/dsl/Stream.java
```diff
@@ -0,0 +1,323 @@
...
+import java.util.List;
+import java.util.Map;
+
+import org.springframework.cloud.dataflow.rest.client.DataFlowOperations;
+import org.springframework.cloud.dataflow.rest.resource.StreamDefinitionResource;
+import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
+
+/**
+ * @author Vinicius Carvalho
```
markpollack: needs javadoc

spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/dsl/Stream.java
```diff
@@ -0,0 +1,323 @@
...
+			this.definition = definition;
+		}
+
+		public StreamBuilder create() {
+			return new StreamBuilder(this.name, this.client, this.definition,
+					Collections.emptyList());
+		}
+	}
+
+	public static class StreamBuilder {
```
markpollack: This needs to be a top level class, the usage ```Stream.StreamBuilder streamBuilder = Stream.builder(dataFlowOperations).name("foo").definition("time | log").create();`` looks bad.

markpollack:
merged as 37f7df4751312b20f18472c983a21406e3596c09

PR #1723, opened 2017-10-27
Rollsback CTR artifacts if CTR creation fails.

### LINKED ISSUE #1669: Subtasks are created even when overall task definition is invalid

cppwfs:
resolves #1669

Added tests

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/TaskConfiguration.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/TaskConfiguration.java
@@ -57,6 +57,7 @@ import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.core.io.ResourceLoader;
 import org.springframework.jdbc.datasource.DataSourceTransactionManager;
+import org.springframework.transaction.annotation.EnableTransactionManagement;
 
 /**
  * @author Thomas Risberg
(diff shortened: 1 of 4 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/TaskConfiguration.java
```diff
@@ -57,6 +57,7 @@
 import org.springframework.context.annotation.Configuration;
 import org.springframework.core.io.ResourceLoader;
 import org.springframework.jdbc.datasource.DataSourceTransactionManager;
+import org.springframework.transaction.annotation.EnableTransactionManagement;
```
mminella: Is this really not already somewhere in our configuration?

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/services/impl/DefaultTaskServiceTests.java
```diff
@@ -317,11 +326,83 @@ public void verifyDataFlowUriProperty() throws Exception {
...
 		assertThat(taskDefinition.getName(), is(equalTo(taskName)));
 		assertThat(taskDefinition.getDslText(), is(equalTo(dsl)));
 	}
+
+	private void initializeSuccessfulRegistry() {
+		when(this.appRegistry.find(anyString(), any(ApplicationType.class)))
+				.thenReturn(new AppRegistration("some-name", task, URI.create("http://helloworld"), resourceLoader));
+	}
+
+	private void initializeFailRegistry() throws IllegalArgumentException{
```
mminella: This fails on the first insert, doesn't it?  Which would mean that it's not really testing the rollback (since no inserts were successful in the first place).

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/TaskConfiguration.java
cppwfs: no.  This is the only location.

mminella:
Squashed and merged as 5f9a8ec453ec63a1cfb88a85b1734251b279c81b.

PR #1735, opened 2017-10-31
Skipper integration enhancement/bug fixes

### LINKED ISSUE #1688: Support configurable prefix for release name when deploying stream using Skipper

### LINKED ISSUE #1707: Wrong stream status after a successful deployment to CF using Skipper
As a user, I deployed a stream via Skipper successfully. However, I'm seeing an inconistent stream status.

**stream:**
```
dataflow:>stream create --name bazticktock --definition "time | log" --deploy --useSkipper
```

**stream list:**
```
dataflow:>stream list
╔═══════════╤═════════════════╤══════════════════════════════════════════════════════════════════════╗
║Stream Name│Stream Definition│                                Status                                ║
╠═══════════╪═════════════════╪══════════════════════════════════════════════════════════════════════╣
║bazticktock│time | log       │The app or group is known to the system, but is not currently deployed║
╚═══════════╧═════════════════╧══════════════════════════════════════════════════════════════════════╝
```

**runtime apps:**
```
dataflow:>runtime apps
╔════════════════════╤═══════════╤═════════════════════════════╗
║App Id / Instance Id│Unit Status│No. of Instances / Attributes║
╚════════════════════╧═══════════╧═════════════════════════════╝
```

**cf apps:**
```
name                         requested state   instances   memory   disk   urls
bazticktock-log-v1           started           1/1         1G       1G     bazticktock-log-v1.cfapps.io
bazticktock-time-v1          started           1/1         1G       1G     bazticktock-time-v1.cfapps.io
```

### LINKED ISSUE #1730: Stream destroyAll doesn't work when using Skipper

### LINKED ISSUE #1731: Refactor StreamDefinition/DeploymentController
Given the `StreamService` delegates the deployment/status operations now, the Stream Definition controller doesn't require AppDeployer, StreamDeploymentController etc.,

Currently, the `StreamDeploymentController` has dependencies that are not used and there are some methods like `mergeAndExpandAppProperties` can be moved to `DefaultStreamService` or some possible Util class.

### LINKED ISSUE #1733: Package version defaults to 1.0.0 during stream deploy if not specified
We need to assert if the package version is specified when using Skipper stream deployment.

### LINKED ISSUE #1734: Skipper stream deploy fails when the package-name is not set via skipper properties
The uploadRequest is set with `null` package name when it is not set via Skipper deployer properties

ilayaperumalg:
- Add release name prefix as `scdf_` for the releases created from stream deploy
   - This prefix is added to the stream name in scope
 - Add a common REST resource class SkipperStream to have all the constants used by both SCDF server/client
   - Move all the skipper related properties in there
Resolves #1688

 - Make package-version a mandator property when using Skipper at the shell
  - Update Shell command and the template methods
Resolves #1733

 - Refactor StreamDefinition/Deployment controllers
   - Remove unused dependencies
 - Move all the stream deployment properties handler methods into `StreamDeploymentPropertiesUtils`
 - Add StreamUpdateService interface
 - Have StreamService extend StreamUpdateService
 - Use interface type `StreamService` in the dependencies instead of strong type
 - Avoid using `useSkipper` as the deployment property and this will just be a shell option
Resolves #1731

 - When using destroyAll() make sure to call the `undeploy` via stream service
Resolves #1730

 - Fix setting UploadRequest package name
Resolves #1734

 - RuntimeApps controller now uses SkipperClient to get the status of streams that are deployed via Skipper
   - Check the StreamDeploymentRepository to retrive the status from appDeployer/Skipper
   - Update test
   Resolves #1707

### DIFF
--- a/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/SkipperStream.java
+++ b/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/SkipperStream.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.cloud.dataflow.rest;
+
+/**
+ *
+ * @author Ilayaperumal Gopinathan
+ */
+public class SkipperStream {
+
+	public static final String SKIPPER_STREAM_PREFIX = "scdf_";
+
+	public static final String USE_SKIPPER = "useSkipper";
+
+	public static final String SKIPPER_KEY_PREFIX = "spring.cloud.dataflow.skipper";
+
+	public static final String SKIPPER_PACKAGE_NAME = SKIPPER_KEY_PREFIX + ".packageName";
+
+	public static final String SKIPPER_PACKAGE_VERSION = SKIPPER_KEY_PREFIX + ".packageVersion";
+
+	public static final String SKIPPER_REPO_NAME = SKIPPER_KEY_PREFIX + ".repoName";
+
+	public static final String SKIPPER_PLATFORM_NAME = SKIPPER_KEY_PREFIX + ".platformName";
+
+	public static final String SKIPPER_ENABLED_PROPERTY_KEY = SKIPPER_KEY_PREFIX + ".enabled";
+
+	public static final String SKIPPER_DEFAULT_API_VERSION = "skipper/v1";
+
+	public static final String SKIPPER_DEFAULT_KIND = "SpringBootApp";
+
+	public static final String SKIPPER_DEFAULT_MAINTAINER = "dataflow";
+}
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/StreamUpdateService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/StreamUpdateService.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.cloud.dataflow.server.service;
+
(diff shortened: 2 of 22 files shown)


jvalkeal:
I played with this and looks good. Hopefully I didn't miss anything.

spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/SkipperStream.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+package org.springframework.cloud.dataflow.rest;
+
+/**
+ *
+ * @author Ilayaperumal Gopinathan
+ */
+public class SkipperStream {
+
+	public static final String SKIPPER_STREAM_PREFIX = "scdf_";
```
markpollack: What about if we create a `DataFlowSkipperProperties` `@ConfigurationClass`  (prefix spring.cloud.dataflow.skipper) and allow this to be configurable?

markpollack: Actually, upon discussion, seems the principle of least surprise is to not have a prefix at all.  This also simplifies the status lookup delegated to skipper from scdf shell.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/StreamDeploymentPropertiesUtils.java
```diff
@@ -0,0 +1,306 @@
...
+
+/**
+ * @author Eric Bottard
+ * @author Mark Fisher
+ * @author Patrick Peralta
+ * @author Ilayaperumal Gopinathan
+ * @author Marius Bogoevici
+ * @author Janne Valkealahti
+ */
+public class StreamDeploymentPropertiesUtils {
```
markpollack: Rename to StreamDeploymentRequestCreator ?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/StreamDeploymentPropertiesUtils.java
```diff
@@ -0,0 +1,306 @@
...
+			ApplicationConfigurationMetadataResolver metadataResolver) {
+		Assert.notNull(appRegistry, "AppRegistry must not be null");
+		Assert.notNull(commonApplicationProperties, "CommonApplicationProperties must not be null");
+		Assert.notNull(metadataResolver, "MetadataResolver must not be null");
+		this.appRegistry = appRegistry;
+		this.commonApplicationProperties = commonApplicationProperties;
+		this.whitelistProperties = new WhitelistProperties(metadataResolver);
+	}
+
+	List<AppDeploymentRequest> createRequests(StreamDefinition stream,
```
markpollack: make public

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/StreamUpdateService.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+public interface StreamUpdateService {
+
+	/**
+	 * Update the stream using the UpdateStreamRequest.
+	 *
+	 * @param streamName the name of the stream to update
+	 * @param updateStreamRequest the UpdateStreamRequest to use during the update
+	 */
+	void updateStream(String streamName, UpdateStreamRequest updateStreamRequest);
```
markpollack: We can move this into StreamService since it is a new interface in 1.3.x....

ilayaperumalg:
Rebased.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/StreamDeploymentPropertiesUtils.java
markpollack: Made the private method package protected so as to be able to write finer grained tests.

PR #1765, opened 2017-11-06
Use SkipperClient from skipper

### LINKED ISSUE #1713: Invalid deployer configuration throws "500 null"
As a user, while deploying a stream via skipper, I see the following error when invalid deployer configuration is provided. 

**server logs:**
```
2017-10-23 22:59:41.298 ERROR 1 --- [p-nio-80-exec-4] o.s.c.d.s.c.RestControllerAdvice         : Caught exception while handling a request

org.springframework.web.client.HttpServerErrorException: 500 null
	at org.springframework.web.client.DefaultResponseErrorHandler.handleError(DefaultResponseErrorHandler.java:66) ~[spring-web-4.3.11.RELEASE.jar!/:4.3.11.RELEASE]
```

**skipper logs:**
```
2017-10-23 22:59:41.173 ERROR 1 --- [nio-7577-exec-2] o.a.c.c.C.[.[.[/].[dispatcherServlet]    : Servlet.service() for servlet [dispatcherServlet] in context with path [] threw exception [Request processing failed; nested exception is org.springframework.cloud.skipper.SkipperException: Can not install, no deployer named 'dev'] with root cause

org.springframework.cloud.skipper.SkipperException: Can not install, no deployer named 'dev'
```

It'd be better to show a description of the error instead of "500 null".

### LINKED ISSUE #1764: Ignore ssl tests
Adding this ticket as a reference point.

With #1713 we started to use `skipperClient` from skipper, which effectively uses `RestTemplateBuilder` from boot. On default it causes init of `org.springframework.http.client.HttpComponentsClientHttpRequestFactory` which via apache rest classes call `SSLSocketFactory.getDefault()`. `SSLSocketFactory` seem to cache something in its private static fields and thus caching something. Our ssl tests sets ssl related system props manually in tests, thus theory is that DS Server/Client libs then gets broken(effectively ignoring our self-signed test certs).

Should not happen in a real env where system props are set with a jvm. Ignoring tests for now in case someone needs/wants to run those manually.

jvalkeal:
- In favour of letting skipper to create its client
  modify DataFlowControllerAutoConfiguration to import
  SkipperClientConfiguration and move skipperStreamDeployer
  into nested config class.
- With this mod we get same RestTemplate as used in
  skipper shell which works better for exception handling
  related to errors thrown via rest layer.
- Modify some existing tests and add new one to ensure
  SkipperClientProperties are processed into new
  SkipperClient.
- Also ignore ssl tests which turned out to be unreliable.
- Fixes #1713
- Fixes #1764

### DIFF
--- a/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/support/TestUtils.java
+++ b/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/support/TestUtils.java
@@ -1,0 +1,113 @@
+/*
+ * Copyright 2016-2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.dataflow.server.support;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.stream.IntStream;
+
+import org.springframework.util.ReflectionUtils;
+
+/**
+ * Utils for tests.
+ *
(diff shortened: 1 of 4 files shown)


ilayaperumalg:
LGTM, merging.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/support/TestUtils.java
```diff
@@ -0,0 +1,113 @@
...
+
+import org.springframework.util.ReflectionUtils;
+
+/**
+ * Utils for tests.
+ *
+ * @author Janne Valkealahti
+ * @author Soby Chacko
+ */
+public class TestUtils {
```
ilayaperumalg: This is nice! :-)

PR #1767, opened 2017-11-06
Fix status msg description

### LINKED ISSUE #1754: Wrong message for status when deploying via skipper
```
║ticktock   │time | log       │The app or group is being deployed.║
```
there is never a possibility in SCDF to deploy only an app.  should say "The Stream is being deploy"

ilayaperumalg:
- Given the DeploymentStateResource is only used by the Streams, changed the description to reflect the `Stream` instead of `all the apps`

Resolves #1754

### DIFF
--- a/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/resource/DeploymentStateResource.java
+++ b/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/resource/DeploymentStateResource.java
@@ -30,7 +30,7 @@ import org.springframework.util.Assert;
 public enum DeploymentStateResource {
 
 	DEPLOYING("deploying", "Deploying", "The app or group is being deployed."),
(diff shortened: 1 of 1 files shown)


spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/resource/DeploymentStateResource.java
```diff
@@ -30,7 +30,7 @@
 public enum DeploymentStateResource {
 
 	DEPLOYING("deploying", "Deploying", "The app or group is being deployed."),
```
cppwfs: to handle the first part of the story should we also say, "The stream is being deployed"  If you agree with me I can fix on merge.

ilayaperumalg: Sure, I think that looks better. Thanks @cppwfs

cppwfs:
Added updates on merge, squashed merged.

PR #1774, opened 2017-11-08
Support downloading http based resources

### LINKED ISSUE #1762: Can't deploy `http` based resources using skipper
When registering an app using http, the DelegatingResourceLoader will cache the file and we lose the version information.

Downloading `http://repo.spring.io/libs-release/org/springframework/cloud/stream/app/log-sink-rabbit/1.1.1.RELEASE/log-sink-rabbit-1.1.1.RELEASE.jar` gives a local file name of

```
file:/tmp/deployer-resource-cache1752520177092508778/http-5659221efdecc880dc84ca0874b5735ae026fec2
```

And the exception thrown is 
```

java.lang.StringIndexOutOfBoundsException: String index out of range: -59

	at java.lang.String.substring(String.java:1967)
	at org.springframework.cloud.dataflow.server.stream.SkipperStreamDeployer.extractGenericVersion(SkipperStreamDeployer.java:123)
```

DelegatingResourceLoader needs to change, I think to avoid caching any http based resources since the feedback is that 'the principle of least surprise' should be to redownload the same file over and over again.  Users use http endpoints for apps as a poor-mans maven snapshot.

The local file name should be preserved, only the directory should be a uuid.

### LINKED ISSUE #1773: Handling ResourceMetadata URI when not set
When resource metadata URI is not set, we don't want to set it to be the resource URI as we expect the metadata URI to be explicit since 1.2.x

ilayaperumalg:
-In SkipperStreamDeployer, return the URI string when setting the spec.resource
which will make sure the the resource location is appropriately set

Resolves #1762

Return null when metadataUri is not set for AppRegistration

 - This avoids downloading the resource when metadataUri is not found
 - It is also expected that the metadata be set only via metadataUri

Resolves #1773

Update Spring Cloud Deployer to 1.3.0.BUILD-SNAPSHOT

### DIFF
--- a/spring-cloud-dataflow-registry/src/test/java/org/springframework/cloud/dataflow/registry/AppRegistrationTests.java
+++ b/spring-cloud-dataflow-registry/src/test/java/org/springframework/cloud/dataflow/registry/AppRegistrationTests.java
@@ -43,7 +43,7 @@ public class AppRegistrationTests {
 	@Test
 	public void testMetadata() throws IOException {
 		AppRegistration registration = new AppRegistration("foo", task, URI.create("file:///foobar"),
-				new DefaultResourceLoader());
+				URI.create("file:///foobar"), new DefaultResourceLoader());
 		assertThat(registration.getMetadataResource().getFile()).hasName("foobar");
 
 		registration = new AppRegistration("foo", task, URI.create("file:///foobar"),
--- a/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/stream/SkipperStreamDeployerTests.java
+++ b/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/stream/SkipperStreamDeployerTests.java
@@ -106,19 +106,19 @@ public class SkipperStreamDeployerTests {
 	@Test
 	public void testFileResourceProcessing() throws MalformedURLException{
 		Resource resource = new UrlResource("file:/springcloudstream/file-source-kafka-10-1.2.0.RELEASE.jar");
-		assertThat(SkipperStreamDeployer.getResourceWithoutVersion(resource)).isEqualTo("file:/springcloudstream/file-source-kafka-10.jar");
+		assertThat(SkipperStreamDeployer.getResourceWithoutVersion(resource)).isEqualTo("file:/springcloudstream/file-source-kafka-10-1.2.0.RELEASE.jar");
 		assertThat(SkipperStreamDeployer.getResourceVersion(resource)).isEqualTo("1.2.0.RELEASE");
 
 		resource = new UrlResource("file:/springcloudstream/file-source-kafka-10-1.2.0.BUILD-SNAPSHOT.jar");
-		assertThat(SkipperStreamDeployer.getResourceWithoutVersion(resource)).isEqualTo("file:/springcloudstream/file-source-kafka-10.jar");
+		assertThat(SkipperStreamDeployer.getResourceWithoutVersion(resource)).isEqualTo("file:/springcloudstream/file-source-kafka-10-1.2.0.BUILD-SNAPSHOT.jar");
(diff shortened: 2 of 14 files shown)


markpollack:
There is a test failure 
```
testMetadata(org.springframework.cloud.dataflow.registry.AppRegistrationTests)  Time elapsed: 0.008 sec  <<< ERROR!
java.lang.NullPointerException
	at org.springframework.cloud.dataflow.registry.AppRegistrationTests.testMetadata(AppRegistrationTests.java:47)```

File: spring-cloud-dataflow-registry/src/test/java/org/springframework/cloud/dataflow/registry/AppRegistrationTests.java [hunk 1]
```diff
@@ -43,7 +43,7 @@ public void testResource() {
 	@Test
 	public void testMetadata() throws IOException {
 		AppRegistration registration = new AppRegistration("foo", task, URI.create("file:///foobar"),
-				new DefaultResourceLoader());
+				URI.create("file:///foobar"), new DefaultResourceLoader());
```
markpollack: can also assert that getMetadataResource is null if not supplied.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/stream/SkipperStreamDeployerTests.java
```diff
@@ -106,19 +106,19 @@ public void testInstallUploadProperties() {
 	@Test
 	public void testFileResourceProcessing() throws MalformedURLException{
 		Resource resource = new UrlResource("file:/springcloudstream/file-source-kafka-10-1.2.0.RELEASE.jar");
-		assertThat(SkipperStreamDeployer.getResourceWithoutVersion(resource)).isEqualTo("file:/springcloudstream/file-source-kafka-10.jar");
+		assertThat(SkipperStreamDeployer.getResourceWithoutVersion(resource)).isEqualTo("file:/springcloudstream/file-source-kafka-10-1.2.0.RELEASE.jar");
```
markpollack: the method is getResourceWithoutVersion but yet it will return with the version for the 'generic' http resource.  Will look to change some names around.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/stream/SkipperStreamDeployerTests.java
```diff
@@ -106,19 +106,19 @@ public void testInstallUploadProperties() {
...
 		assertThat(SkipperStreamDeployer.getResourceVersion(resource)).isEqualTo("1.2.0.BUILD-SNAPSHOT");
 
 		resource = new UrlResource("http://springcloudstream/file-source-kafka-10-1.2.0.RELEASE.jar");
-		assertThat(SkipperStreamDeployer.getResourceWithoutVersion(resource)).isEqualTo("http://springcloudstream/file-source-kafka-10.jar");
+		assertThat(SkipperStreamDeployer.getResourceWithoutVersion(resource)).isEqualTo("http://springcloudstream/file-source-kafka-10-1.2.0.RELEASE.jar");
 		assertThat(SkipperStreamDeployer.getResourceVersion(resource)).isEqualTo("1.2.0.RELEASE");
 
 		resource = new UrlResource("http://springcloudstream/file-source-kafka-crap.jar");
-		assertThat(SkipperStreamDeployer.getResourceWithoutVersion(resource)).isEqualTo("http://springcloudstream/file-source-kafka.jar");
+		assertThat(SkipperStreamDeployer.getResourceWithoutVersion(resource)).isEqualTo("http://springcloudstream/file-source-kafka-crap.jar");
```
markpollack: If the jar doesn't have any `-` in it, then getResourceVersion returns the full name....

ilayaperumalg:
This was merged as 3db69a39e1d9f7f2fb9964995b634abde66d7ee2

PR #1781, opened 2017-11-13
Added docs on how to launch CTR from a stream

### LINKED ISSUE #1780: Document launching a composed-task via a stream
As a user, I'd like to refer to the docs on how a composed-task can be launched via a stream. 

Refer to the [SO thread](https://stackoverflow.com/questions/47250280/launching-composed-task-built-by-dsl-from-stream-application) for more details.

**Acceptance:**
- Include a new sub-section under `Launching Tasks from a Stream`
- Move the `Launching Tasks from a Stream` section after the `Composed Tasks DSL` 
- List an example via `time` or `trigger` be the source that launches the composed-task graph on a `fixed-delay`

This relates to spring-cloud/spring-cloud-dataflow#1752 from datasource standpoint.

cppwfs:
resolves #1780

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/tasks.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/tasks.adoc
@@ -283,75 +283,6 @@ The default Task/Batch event and destination names on the broker are enumerated 
 |Skip events|`skip-events`
 |===
 
-[[spring-cloud-dataflow-launch-tasks-from-stream]]
-== Launching Tasks from a Stream
-
-You can launch a task from a stream by using one of the available `task-launcher` sinks. Currently the platforms supported
-via the `task-launcher` sinks are
-https://github.com/spring-cloud-stream-app-starters/tasklauncher-local[local],
-https://github.com/spring-cloud-stream-app-starters/tasklauncher-cloudfoundry[Cloud Foundry], and
-https://github.com/spring-cloud-stream-app-starters/tasklauncher-yarn[Yarn].
-
-NOTE: `task-launcher-local` is meant for development purposes only.
-
-A `task-launcher` sink expects a message containing a https://github.com/spring-cloud/spring-cloud-task/blob/master/spring-cloud-task-stream/src/main/java/org/springframework/cloud/task/launcher/TaskLaunchRequest.java[TaskLaunchRequest] object in its payload. From the `TaskLaunchRequest` object the `task-launcher` will obtain the URI of the artifact to be launched as well as the environment properties, command line arguments, deployment properties and application name to be used by the task.
-
-The https://github.com/spring-cloud-stream-app-starters/tasklauncher-local/blob/v1.2.0.RELEASE/spring-cloud-starter-stream-sink-task-launcher-local/README.adoc[task-launcher-local] can be added to the available sinks by executing the app register command as follows (for the Rabbit Binder):
-
-```
-app register --name task-launcher-local --type sink --uri maven://org.springframework.cloud.stream.app:task-launcher-local-sink-rabbit:jar:1.2.0.RELEASE
-```
-
-In the case of a maven based task that is to be launched, the `task-launcher` application is responsible for downloading the artifact.  You *must* configure the `task-launcher` with the appropriate configuration of https://github.com/spring-cloud/spring-cloud-deployer/blob/master/spring-cloud-deployer-resource-maven/src/main/java/org/springframework/cloud/deployer/resource/maven/MavenProperties.java[Maven Properties] such as `--maven.remote-repositories.repo1.url=http://repo.spring.io/libs-milestone"` to resolve artifacts, in this case against a milestone repo.  Note that this repo can be different than the one used to register the `task-launcher` application itself.
-
-=== TriggerTask
-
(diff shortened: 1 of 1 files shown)


spring-cloud-dataflow-docs/src/main/asciidoc/tasks.adoc
```diff
@@ -529,6 +529,46 @@ Spring Cloud Data Flow Dashboard or RESTful API), will relaunch  the
 `STOPPED` child task, and then launch the remaining (unlaunched) child tasks
 in the specified order.
 
+==== Launching a Composed Task From a Stream
```
sabbyanandan: What if we move this under "Launching Tasks from a Stream" section? That way, all the stream launching task examples could be all together. _(I had a reordering suggestion in the acceptance)_

spring-cloud-dataflow-docs/src/main/asciidoc/tasks.adoc
```diff
@@ -351,6 +351,44 @@ For example:
...
 ```
+=== Launching a Composed Task From a Stream
+A composed task can be launched using one of the `task-launcher` sinks as discussed
+<<spring-cloud-dataflow-launch-tasks-from-stream, here>>.  Since we will be using
+the `ComposedTaskRunner` directly we will need to setup the task definitions it
+will use prior to the creation of the composed task launching stream.  So let's
+say that we wanted to create the following composed task definition `AAA && BBB`.
+The first step would be to create the task definitions.  For example:
+```
+task create CTR-AAA --definition "AAA"
```
sabbyanandan: Why can't we simply create regular task definitions and combine the definitions in the composed-task graph DSL?

```
task create AAA --definition "timestamp"
task create BBB --definition "timestamp"
task create CTR --definition "AAA && BBB"
```

This would follow the general process of a composed-task orchestration in SCDF.

spring-cloud-dataflow-docs/src/main/asciidoc/tasks.adoc
```diff
@@ -351,6 +351,44 @@ For example:
 ```
 stream create task-stream --definition "http --port=9000 | tasklaunchrequest-transform --uri=maven://org.springframework.cloud.task.app:timestamp-task:jar:1.2.0.RELEASE | task-launcher-local --maven.remote-repositories.repo1.url=http://repo.spring.io/libs-release"
 ```
+=== Launching a Composed Task From a Stream
```
sabbyanandan: The composed-task and the feature description are listed after this section. I think it would have a better flow if we move the entire section of "stream launching tasks" after the composed-task (_see suggestions in the acceptance criteria_).

spring-cloud-dataflow-docs/src/main/asciidoc/tasks.adoc
```diff
@@ -351,6 +351,44 @@ For example:
...
+task create CTR-BBB --definition "BBB"
+```
+Now that the task definitions we need for composed task definition are ready, we
+need to create a stream that will launch `ComposedTaskRunner`.
+So in this case we will create a stream that has a trigger that will emit a message
+once every 30 seconds, a transformer that will create a `TaskLaunchRequest` for
+each message received, and a `task-launcher-local` sink that will launch a
+the `ComposedTaskRunner` on our local machine. The stream should look something like this:
+```
+stream create ctr-stream --definition "time --fixed-delay=30 | tasklaunchrequest-transform --uri=maven://org.springframework.cloud.task.app:composedtaskrunner-task:<current release> --command-line-arguments='--graph=taskfun-AAA&&taskfun-BBB --increment-instance-enabled=true --spring.datasource.url=...' | task-launcher-local"
```
sabbyanandan: Maybe a link to CTR releases tab in GH would be useful.

spring-cloud-dataflow-docs/src/main/asciidoc/tasks.adoc
```diff
@@ -351,6 +351,44 @@ For example:
...
+`ComposedTaskRunner`:
+
+* *graph* - this is the graph that is to be executed by the `ComposedTaskRunner`.
+In this case it is `AAA&&BBB`
+* *increment-instance-enabled* - this allows each execution of `ComposedTaskRunner`
+to be unique.  `ComposedTaskRunner` is built using http://projects.spring.io/spring-batch/[Spring Batch], and thus each
+we will want a new Job Instance for each launch of the `ComposedTaskRunner`.  To
+do this we set the increment-instance-enabled to be `true`.
+* *spring.datasource.** - establishes the datasource that is used by Spring Cloud
+Data Flow.  This is so that the `ComposedTaskRunner` can update the state of
```
sabbyanandan: Maybe also indicate how SCDF will use it?

spring-cloud-dataflow-docs/src/main/asciidoc/tasks.adoc
cppwfs: If I understand your question correctly, DataFlow Launch command looks at the definition and determines whether the definition is a regular task or a CTR.  If it determines that it is a CTR it then builds up the graph based on the definition above (adding the prefixes) and establishes the app to use as CTR.  When using a stream the TLR-Transformer doesn't have the ability to determine if the definition is a CTR or not, so it just passes the definition as is to the TaskLaunchRequest.  So this is the reason, we used the definition as below.

cppwfs: I will update the sample to use timestamp instead of just `AAA`  & `BBB` as the apps for the definitions.

cppwfs:
Rebased, Squashed Merged

PR #1811, opened 2017-11-27
Adding Autoconfiguration to DSL client

### LINKED ISSUE #1768: Create autoconfiguration for DataFlowTemplate
This will make the use of the Java DSL easier on the user. This story will cover the basic-auth, we will have OAUTH in a different PR.

viniciusccarvalho:
Fixes #1768

### DIFF
--- a/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/config/DataFlowClientAutoConfiguration.java
+++ b/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/config/DataFlowClientAutoConfiguration.java
@@ -4,9 +4,7 @@ import java.net.URI;
 
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
-import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
-import org.springframework.cloud.dataflow.core.DataFlowPropertyKeys;
 import org.springframework.cloud.dataflow.rest.client.DataFlowOperations;
 import org.springframework.cloud.dataflow.rest.client.DataFlowTemplate;
 import org.springframework.cloud.dataflow.rest.client.dsl.Stream;
@@ -23,7 +21,6 @@ import org.springframework.web.client.RestTemplate;
  */
 @Configuration
 @EnableConfigurationProperties(DataFlowClientProperties.class)
-@ConditionalOnProperty(prefix = DataFlowPropertyKeys.PREFIX + "client", name = "enableDsl")
 public class DataFlowClientAutoConfiguration {
 
 	@Autowired
@@ -37,14 +34,14 @@ public class DataFlowClientAutoConfiguration {
 	public DataFlowOperations dataFlowOperations() throws Exception{
 		RestTemplate template = DataFlowTemplate.prepareRestTemplate(restTemplate);
 		final HttpClientConfigurer httpClientConfigurer = HttpClientConfigurer.create()
-				.targetHost(new URI(properties.getTarget()))
+				.targetHost(new URI(properties.getUri()))
 				.skipTlsCertificateVerification(properties.isSkipSslValidation());
 		if(!StringUtils.isEmpty(properties.getSecurity().getUsername()) &&
 				!StringUtils.isEmpty(properties.getSecurity().getPassword())){
 			httpClientConfigurer.basicAuthCredentials(properties.getSecurity().getUsername(), properties.getSecurity().getPassword());
 			template.setRequestFactory(httpClientConfigurer.buildClientHttpRequestFactory());
 		}
-		return new DataFlowTemplate(new URI(properties.getTarget()), template);
+		return new DataFlowTemplate(new URI(properties.getUri()), template);
 	}
 
 	@Bean
--- a/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/config/DataFlowClientAutoConfiguration.java
+++ b/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/config/DataFlowClientAutoConfiguration.java
@@ -36,9 +36,9 @@ public class DataFlowClientAutoConfiguration {
 		final HttpClientConfigurer httpClientConfigurer = HttpClientConfigurer.create()
 				.targetHost(new URI(properties.getUri()))
 				.skipTlsCertificateVerification(properties.isSkipSslValidation());
-		if(!StringUtils.isEmpty(properties.getSecurity().getUsername()) &&
-				!StringUtils.isEmpty(properties.getSecurity().getPassword())){
-			httpClientConfigurer.basicAuthCredentials(properties.getSecurity().getUsername(), properties.getSecurity().getPassword());
+		if(!StringUtils.isEmpty(properties.getAuthentication().getBasic().getUsername()) &&
+				!StringUtils.isEmpty(properties.getAuthentication().getBasic().getPassword())){
+			httpClientConfigurer.basicAuthCredentials(properties.getAuthentication().getBasic().getUsername(), properties.getAuthentication().getBasic().getPassword());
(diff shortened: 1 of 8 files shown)


spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/config/DataFlowClientAutoConfiguration.java
```diff
@@ -0,0 +1,56 @@
...
+import org.springframework.util.StringUtils;
+import org.springframework.web.client.RestTemplate;
+
+/**
+ * Configuration class to provide default beans for {@link DataFlowOperations} and {@link org.springframework.cloud.dataflow.rest.client.dsl.StreamBuilder} instances.
+ * @author Vinicius Carvalho
+ */
+@Configuration
+@EnableConfigurationProperties(DataFlowClientProperties.class)
+@ConditionalOnProperty(prefix = DataFlowPropertyKeys.PREFIX + "client", name = "enableDsl")
```
markpollack: If you have the client library on your classpath, then you should get the `DataFlowTemplate` and `StreamBuilder`.  This is similar to how other `xyzTemplate` classes are created via autoconfiguration, e.g. RabbitTemplate, RedisTemplate etc.

spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/config/DataFlowClientProperties.java
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Configuration properties used in {@link DataFlowClientAutoConfiguration}
+ *
+ * @author Vinicius Carvalho
+ */
+@ConfigurationProperties(prefix = DataFlowPropertyKeys.PREFIX + "client")
+public class DataFlowClientProperties {
+
+	private String target = "http://localhost:9393";
```
markpollack: The property name should be `uri` as that matches better with the existing usage in the shell, e.g. 

```
$ java -jar spring-cloud-dataflow-shell-1.3.0.M3.jar --help
Data Flow Options:
  --dataflow.uri=<uri>                              Address of the Data Flow Server [default: http://localhost:9393].
  --dataflow.username=<USER>                        Username of the Data Flow Server [no default].
  --dataflow.password=<PASSWORD>                    Password of the Data Flow Server [no default].
  --dataflow.credentials-provider-command=<COMMAND> Executes an external command which must return an OAuth Access Token [no default].
  --dataflow.skip-ssl-validation=<true|false>       Accept any SSL certificate (even self-signed) [default: no].$ java -jar spring-cloud-dataflow-shell-1.3.0.M3.jar --help
Data Flow Options:
  --dataflow.uri=<uri>                              Address of the Data Flow Server [default: http://localhost:9393].
  --dataflow.username=<USER>                        Username of the Data Flow Server [no default].
  --dataflow.password=<PASSWORD>                    Password of the Data Flow Server [no default].
  --dataflow.credentials-provider-command=<COMMAND> Executes an external command which must return an OAuth Access Token [no default].
  --dataflow.skip-ssl-validation=<true|false>       Accept any SSL certificate (even self-signed) [default: no].
```

spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/config/DataFlowClientProperties.java
```diff
@@ -0,0 +1,76 @@
...
+
+	public Security getSecurity() {
+		return security;
+	}
+
+	public void setSecurity(Security security) {
+		this.security = security;
+	}
+
+	public static class Security {
```
markpollack: This is too generic a name for basic auth.  There was some careful thought into redoing spring boot security autoconfiguration, we can follow their lead and use `User` as the class to specify username and password for basic auth.

https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/SecurityProperties.java#L107

@ghillert @ilayaperumalg What do you think about including ROLES in the User class?

spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/config/DataFlowClientAutoConfiguration.java
viniciusccarvalho: We discussed this before. This is due shell, it causes conflict with it's created template. I believe you and I get in an agreement that we should use conditional properties because of that (instead of changing the shell code to figure out which template to use), but if that has changed I can work on changing the shell classes instead.

spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/config/DataFlowClientProperties.java
viniciusccarvalho: Wasn't `security` the name of the property from boot? I have no problem in changing that, but I used boot as a reference (at least 1.x branch that is)

ilayaperumalg: Yes, along with the username/password, we need the option to specify `role` as a list of accepted values (VIEW, CREATE, MANAGE). This is in line with what Boot does [here](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/SecurityProperties.java#L122)

viniciusccarvalho:
@ilayaperumalg but roles are supposed to be declared on server side only. Clients should not declare roles as this is a server responsibility.

ilayaperumalg:
@viniciusccarvalho yeah, sorry I missed that one. Thanks for clarifying!

spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/config/DataFlowClientAutoConfiguration.java
markpollack: Changing the shell class to adapt to autoconfig would be the way to go.

spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/config/DataFlowClientProperties.java
markpollack: Spring Boot 1.5.x has this https://github.com/spring-projects/spring-boot/blob/1.5.x/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/SecurityProperties.java

The reference docs have examples like you coded it, but they were just examples.  Note that the 1.5.x branch doesn't have the roles.  Seems like moving to the way that boot 2 does things will help.

spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/config/DataFlowClientProperties.java
```diff
@@ -0,0 +1,76 @@
...
+
+	public boolean isSkipSslValidation() {
+		return skipSslValidation;
+	}
+
+	public void setSkipSslValidation(boolean skipSslValidation) {
+		this.skipSslValidation = skipSslValidation;
+	}
+
+	public String getTarget() {
```
markpollack: In a similar discussion on skipper, this property should be renamed to `serverUri`.

spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/config/DataFlowClientProperties.java
markpollack: A nested class 'Basic' inside Security would make room for oauth2 etc.

Also your suggestion to rename 'security' to authentication is a good one.

markpollack:
squashed and merged as 0f82d44df98407cea0e543fa226c81cc59955679

PR #1812, opened 2017-11-27
Skipper manifest/platform-list/history commands

### LINKED ISSUE #1803: Add command to query skipper for available platforms/deployers
As a user I would like to use only the Data Flow shell to discover the valid values to pass to the `--platformName` command argument when deploying a stream using skipper.

There should be a new  endpoint in the dataflow server that serves up this data,As a first step, it can live under the current `StreamsController` endpoint.

The commands can have the `stream skipper` prefix.

The shell should not make direct calls to skipper.

### LINKED ISSUE #1804: Add command to query skipper for package history of a stream
As a user I would like to use only the Data Flow shell to discover the valid values to pass to the `--releaseVersion` command argument when rolling back a stream in Data Flow using  skipper.

There should be a new  endpoint in the dataflow server that serves up this data, As a first step, it can live under the current `StreamsController` endpoint.

The commands can have the `stream skipper` prefix.

The shell should not make direct calls to skipper.

### LINKED ISSUE #1805: Add command to query skipper for package manifest of a stream
As a user I would like to use only the Data Flow shell to show the underlying application manifest for a stream when deploying a stream using skipper

There should be an endpoint in the dataflow server that serves up this data, the shell should not make direct calls to the skipper server.  As a first step, it can live under the current `StreamsController` endpoint.

The commands can have the `stream skipper` prefix

The shell should not make direct calls to skipper.

ilayaperumalg:
Resolves #1803
Resolves #1804
Resolves #1805

### DIFF
--- a/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/StreamOperations.java
+++ b/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/StreamOperations.java
@@ -20,6 +20,7 @@ import java.util.Collection;
 import java.util.Map;
 
 import org.springframework.cloud.dataflow.rest.resource.StreamDefinitionResource;
+import org.springframework.cloud.skipper.domain.Deployer;
 import org.springframework.cloud.skipper.domain.PackageIdentifier;
 import org.springframework.cloud.skipper.domain.Release;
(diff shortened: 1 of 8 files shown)


sabbyanandan:
This is merged in: https://github.com/spring-cloud/spring-cloud-dataflow/commit/253d9c613f04cc9d18b470ef526d1bf2ed0ce061.

PR #1836, opened 2017-12-14
Updated the configs for AboutController.

### LINKED ISSUE #1829: Resurface the checksum addition to AboutController
As a developer, I'd like to recommit the addition of checksum to `AboutController` that was merged via: https://github.com/spring-cloud/spring-cloud-dataflow/commit/49de3abaffd7566aa72bb5a99d9d8dfdfc4f2831.

This commit https://github.com/spring-cloud/spring-cloud-dataflow/commit/b4c3835c77b88d2d04df49c019a08413ef556747 removed it due to the issues encountered at the time of 1.3 M3 release.

cppwfs:
* Establishes shell as the only dependency to return checksum
* Establishes enable disable feature for showing checksums.  Default is false
* Establshes that all data returned for the dependencies can be updated via properties
* Sets all results to be consistently named
* Added Tests to the About controller
* Updated dataflow-server-defaults.yml with appropriate defaults

resolves #1829

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/configuration.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/configuration.adoc
@@ -1167,14 +1167,60 @@ then add to the generated Maven pom file the MetricWriter implementation you
 want to use.  The documentation on the Data Flow Metrics project pages provides
 the additional information you need to get started.
 
+== About Configuration
+The Spring Cloud Data Flow About Restful API result contains a display name,
+version, and if specified a url for each of the major dependencies that
+comprise Spring Cloud Data Flow.  The result (if enabled) also contains the
+sha1 and or sha256 checksum values for the shell dependency. The information
+that is returned for each of the dependencies is configurable via the following
+properties:
 
+* spring.cloud.dataflow.version-info.spring-cloud-dataflow-core.name - the
+name to be used for the core.
+* spring.cloud.dataflow.version-info.spring-cloud-dataflow-core.version -
+the version to be used for the core.
+* spring.cloud.dataflow.version-info.spring-cloud-dataflow-dashboard.name - the
+name to be used for the dashboard.
+* spring.cloud.dataflow.version-info.spring-cloud-dataflow-dashboard.version -
+the version to be used for the dashboard.
+* spring.cloud.dataflow.version-info.spring-cloud-dataflow-implementation.name - the
+name to be used for the implementation.
+* spring.cloud.dataflow.version-info.spring-cloud-dataflow-implementation.version -
+the version to be used for the implementation.
+* spring.cloud.dataflow.version-info.spring-cloud-dataflow-shell.name - the
+name to be used for the shell.
+* spring.cloud.dataflow.version-info.spring-cloud-dataflow-shell.version -
+the version to be used for the shell.
+* spring.cloud.dataflow.version-info.spring-cloud-dataflow-shell.url -
+the url to be used for downloading the shell dependency.
+* spring.cloud.dataflow.version-info.spring-cloud-dataflow-shell.checksum-sha1 - the
+checksum1 value that will be returned with the shell dependency info.
+* spring.cloud.dataflow.version-info.spring-cloud-dataflow-shell.checksum-sha256 -
+the checksum256 value that will be returned with the shell dependency info.
+* spring.cloud.dataflow.version-info.spring-cloud-dataflow-shell.checksum-sha1-url -
+if the `spring.cloud.dataflow.version-info.spring-cloud-dataflow-shell.checksum-sha1`
+is not specified then SCDF will use contents of the file specified at this URL for the checksum.
+* spring.cloud.dataflow.version-info.spring-cloud-dataflow-shell.checksum-sha256-url -
+if the `spring.cloud.dataflow.version-info.spring-cloud-dataflow-shell.checksum-sha256`
+is not specified then SCDF will use contents of the file specified at this URL for the checksum.
 
+=== Enabling Shell Checksum values
+By default checksum values will not be displayed for the shell dependency if
+you need this feature enabled then set the
+`spring.cloud.dataflow.version-info.dependency-fetch.enabled` property to true.
 
+=== Reserved values for urls
+There are reserved values (surrounded by curly braces) that you can insert into
+the url that will make sure that the links are up to date:
 
+* repository - if using a build-snapshot, milestone or release candidate of
+Data Flow, the repository will refer to repo-spring-io repository else it will
+refer to Maven Central.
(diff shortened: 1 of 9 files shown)


jvalkeal:
I think you slightly misunderstood how I wanted this to work. Point was to make it fully customisable so that we would never need to touch this code again whatever user wants to do.

1. `versionInfo` deps are now hardcoded and we've forced to have `implementation`, `core`, `dashboard` and `shell`. I wanted `spring.cloud.dataflow.version-info.dependencies` to be a map so that user can add their own entries or easily customise existing ones(if we force some and we want to add at least one for shell as that's what tile team care about). Also if we want to add/force some other deps, we just need to modify yml.

Essentially if user adds:
```
spring:
  cloud:
    dataflow:
      version-info:
        dependencies:
          my-other-dep:
            name: My Other Dep
            version: 1.1.1
```
Payload would automatically have
```
versionInfo": {
 "my-other-dep": {
   "name": "My Other Dep",
   "version": "1.1.1"
 }
}
```

2. We talked about having either `spring.cloud.dataflow.version-info.dependency-fetch.enabled` or `spring.cloud.dataflow.version-info.dependency.enabled`. Trouble with `dependency-fetch.enabled` is that option is a little awkward as you don't immediately understand what it does as it kinda becomes conditional whether `checksum-sha1-url` exists and `checksum-sha1` is not present.

For example with
```
--spring.cloud.dataflow.version-info.dependencies.spring-cloud-dataflow-shell.checksum-sha1=sha1xxx
```
sha1 is already given and there's nothing to fetch but json don't show it.
```
"shell": {
  "name": "Spring Cloud Data Flow Shell",
  "version": "1.3.0.BUILD-SNAPSHOT",
  "url": "https://repo.spring.io/libs-snapshot/org/springframework/cloud/spring-cloud-dataflow-shell/1.3.0.BUILD-SNAPSHOT/spring-cloud-dataflow-shell-1.3.0.BUILD-SNAPSHOT.jar"
}
```

You need to enable fetch
```
--spring.cloud.dataflow.version-info.dependencies.spring-cloud-dataflow-shell.checksum-sha1=sha1xxx
--spring.cloud.dataflow.version-info.dependency-fetch.enabled=true
```
Which then actually doesn't fetch anything as sha1 is already given and defaults to that.
```
"shell": {
  "name": "Spring Cloud Data Flow Shell",
  "version": "1.3.0.BUILD-SNAPSHOT",
  "checksumSha1": "sha1xxx",
  "url": "https://repo.spring.io/libs-snapshot/org/springframework/cloud/spring-cloud-dataflow-shell/1.3.0.BUILD-SNAPSHOT/spring-cloud-dataflow-shell-1.3.0.BUILD-SNAPSHOT.jar"
}
```
It's more clear that `spring.cloud.dataflow.version-info.dependency.enabled` simply enables whole `versionInfo` in json and on default we don't have anything there as flag is `false` by default.

3. I think we should simple default to(thought I'm not against if we add other defaults, I just don't want them to cause additional http hits):
```
spring:
  cloud:
    dataflow:
      version-info:
        dependency:
          enabled: false
        dependencies:
          spring-cloud-dataflow-shell:
            name: Spring Cloud Data Flow Shell
            version: "@project.version@"
            url: "{repository}/org/springframework/cloud/spring-cloud-dataflow-shell/{version}/spring-cloud-dataflow-shell-{version}.jar"
            checksum-sha1-url: "{repository}/org/springframework/cloud/spring-cloud-dataflow-shell/{version}/spring-cloud-dataflow-shell-{version}.jar.sha1"
```

This way when tile team adds `spring.cloud.dataflow.version-info.dependency.enabled=true`, they get sha1 from artifactory for shell, every hit to `/about` does only one fetch(sha256 is not there so we don't want to configure it).

Whether we use `dependencies.spring-cloud-dataflow-shell` or `dependencies.shell`, that should be reflected to a json as is.
```
"spring-cloud-dataflow-shell": {}
```
or
```
"shell": {}
```

cppwfs:
1) I disagree.   The version info should be fixed because we only want to display those dependencies that we are supporting.   We don't want users adding other dependencies to the about (how albeit it only being for their own deployment) result, that could cause problems with support.  Meaning support gets a list of dependencies they don't understand.  
2) We always should return about information.  It is a requirement for the UI and for the shell.   We can change the `spring.cloud.dataflow.version-info.dependency-checksum-fetch.enabled=true`. Or like you said don't even configure sha1 or sha256 and drop the `enabled` property all together.

jvalkeal:
Right, ok. I didn't realise those are needed. I just feel everything other than `implementation` being pointless(we already know what's in a fatjar which is only source of truth) together with shell version.

Even a shell dep is bogus as it really doesn't have anything to do with server version. You can have completely different shell(like what I did with yarn impl) what is build with server core. That's why I was driving this to be more like user(or the ones maintaining custom builds) configurable feature.

cppwfs:
I'll update the PR to remove the `enabled` property from the code and the checksum defaults from the yml file.  Thus the user can choose to add a checksum-sha1-url, checksum-sha1,... if they want to obtain that information.  I'll also update the docs accordingly.

jvalkeal:
I'm not going to fight over this. Maybe in point 2, sha could be shown always if it's already given.

cppwfs:
I agree.  Sorry I did not communicate that clearly.  I agree, if they have shaX or a shaX_URL it should be shown.

jvalkeal:
As much as I hate it, tile team wanted to have that default checksum and url for a shell. Otherwise they need to keep it up-to-date in tile build every time they change a dataflow version. Hate it but kinda reasonable request.

cppwfs:
understood.  I'll keep the defaults.

jvalkeal:
Why did you enable these on default? Now if default url are not routable, about waits timeout. I changed that to ip I know will timeout and `curl http://localhost:9393/about` will just hang. This would be same regression if our repos are not accessible behind firewall.

Also `dataflow config info` in shell hangs and there's now way normal user is able to figure out what happens.

```
java -jar spring-cloud-dataflow-server-local/target/spring-cloud-dataflow-server-local-1.3.0.BUILD-SNAPSHOT.jar --spring.cloud.dataflow.version-info.spring-cloud-dataflow-shell.checksum-sha256-url=http://1.2.3.4/hello
```

All json objects under `versionInfo` should have same consistent model, other than shell only `name` and `version` are used. Thus I can't use
```
java -jar spring-cloud-dataflow-server-local/target/spring-cloud-dataflow-server-local-1.3.0.BUILD-SNAPSHOT.jar --spring.cloud.dataflow.version-info.spring-cloud-dataflow-core.checksum-sha1=sha1xxx
```

jvalkeal:
Merged per https://github.com/spring-cloud/spring-cloud-dataflow/commit/06adc4697fdcdaa6a7f89c960a714ab57ecaf263

PR #1837, opened 2017-12-15
Fixes #1736 : Add versioning support for app registry

### LINKED ISSUE #1736: Add versioning support for app registry
As a user, I'd like to register multiple versions of the app, so I can pick and choose what goes as the default.

**Design notes:**

1) Include a new boolean column (`useInDSL?`; values = Y or N) in the app-registry table.

2) Allow registration of "n" number of versions for the same app with same 'GAV' coordinates.

_examples:_
```
app register --name log --uri <PATH-TO-File>/log:1.1.0
app register --name log --uri <PATH-TO-File>/log:1.2.0
app register --name log --uri <PATH-TO-File>/log:1.2.1
app register --name log --uri <PATH-TO-File>/log:1.3.0
app register --name log --uri <PATH-TO-File>/log:1.4.0
```

3) Upon usage of a specific version in the DSL, we will update the `useInDSL` column for that record to "Y".

4) We will add a new command to change the default version.

> app default sink:log --version 1.4.0

5) The `app unregister` command can accept a `--version` property. If a version is unregistered, we could choose the next default.

6) The `app info` command will _require_ the version to be supplied (should also enable suggestion for all the registered versions)

7) Stream create workflow will use the version that is set to the default.
> stream create foo --definition "time | log" <- it could default to 1.4.0 for example

8) Stream update workflow only accepts a registered app version.
```
stream skipper update foo --properties version.log=1.2.0,app.log.logging.level=DEBUG <- this will work
stream skipper update foo --properties version.log=1.7.0,app.log.logging.level=DEBUG <- this will fail with validation error
```

tzolov:
## Server Side
* Add AppRegistrationRepository, a JPA based App repository (the old UriRegistry is still used for the non app-registry related use cases) 
* Refactor AppRegistration into a JPA DO. (version and isDefault fields added). resource loading responsibility is moved to the AppRegistryService
* new VersionedAppRegistryController that exposes current API + the version specific extensions
* new AppRegistryService between the controller and the new repository
* SCDF server can be run old (`classic`) or the new (`skipper`) mode. Classic is by default. Use `--spring.cloud.dataflow.features.skipper-enabled=true` to enable Skipper mode. 
## Client Side (Shell + Rest)
* Add support for app version parameters where approriate
* add `app default --id --version` to set the desired version
* `app list --id <type:name> will list all version for the given app.

Resolves spring-cloud/spring-cloud-dataflow#1837


ilayaperumalg:
Review was done as a collaborative work on a separate branch. The plan is to get the changes for supporting app versioning including Skipper vs Classic mode changes iteratively.

LGTM, merging this PR. Created some follow up stories and the PRs to follow for the TODO items.

ilayaperumalg:
Squashed and merged as 44fddefa172d63e16bd5f089ec41227aea9ac8b0 along with 395a6876ea1d34ca77df9e872ba8ff8fefc0b44a

Issue #1850, opened 2017-12-20
Graph-to-text fails with 403 when security is enabled


sabbyanandan:
_From @sabbyanandan on December 7, 2017 15:6_

As a user, I'm unable to use composed-task canvas when security is enabled. Both the visual and textual representation errors. 

```
Request URL:http://HOST/tools/convertTaskGraphToText
Request Method:POST
Status Code:403 
Remote Address:HOST
Referrer Policy:no-referrer-when-downgrade
```

![image](https://user-images.githubusercontent.com/1562654/33721794-e6428e1c-db1c-11e7-8a31-9cd00c302bdd.png)

_Copied from original issue: spring-cloud/spring-cloud-dataflow-ui#565_

sabbyanandan:
@ghillert: When you get a chance, please have a look at this.

sabbyanandan:
It might be a server-side problem; we can move the issue to SCDF backlog then.

PR #1851, opened 2017-12-20
Map CT parser endpoints with CREATE-only role

sabbyanandan:
To reproduce the problem, start the server with following security configuration.

```
security.basic.enabled=true
security.user.name=test
security.user.password=pass
security.user.role=VIEW,CREATE
```

Once after you login to the dashboard, go to CT's Flo designer and start typing the DSL; you will see 403 errors in the browser console. 

This PR addresses this by adding the role-mapping for CT parser endpoints. No changes to the UI is necessary.

Resolves spring-cloud/spring-cloud-dataflow#1850

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
+++ b/spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
@@ -121,4 +121,4 @@ spring:
             # Tools
 
             - POST   /tools/**                       => hasRole('ROLE_CREATE')
-            - POST   /tools/**                       => hasRole('ROLE_CREATE')
+
--- a/spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
+++ b/spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
@@ -117,3 +117,8 @@ spring:
             - GET    /tasks/executions/*             => hasRole('ROLE_VIEW')
             - POST   /tasks/executions               => hasRole('ROLE_CREATE')
             - DELETE /tasks/executions/*             => hasRole('ROLE_CREATE')
+
+            # Tools
+
+            - POST   /tools/**                       => hasRole('ROLE_CREATE')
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
```diff
@@ -117,3 +117,8 @@ spring:
             - GET    /tasks/executions/*             => hasRole('ROLE_VIEW')
             - POST   /tasks/executions               => hasRole('ROLE_CREATE')
             - DELETE /tasks/executions/*             => hasRole('ROLE_CREATE')
+
+            # Tools
+
+            - POST   /tools/**                       => hasRole('ROLE_CREATE')
+            - POST   /tools/**                       => hasRole('ROLE_CREATE')
```
ghillert: Return missing

spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
```diff
@@ -117,3 +117,8 @@ spring:
             - GET    /tasks/executions/*             => hasRole('ROLE_VIEW')
             - POST   /tasks/executions               => hasRole('ROLE_CREATE')
             - DELETE /tasks/executions/*             => hasRole('ROLE_CREATE')
+
+            # Tools
+
+            - POST   /tools/**                       => hasRole('ROLE_CREATE')
```
ghillert: This line can be deleted, right? `/tools/**` covers all tools endpoints from the `ToolsController`.

ghillert:
@sabbyanandan We should also add a test case to `LocalServerSecurityWithSingleUserTests` in the `spring-cloud-starter-dataflow-server-local` module. Let me know if you feel brave enough or if I shall add it. Otherwise this looks good.

ghillert:
Created https://github.com/spring-cloud/spring-cloud-dataflow/issues/1864

ghillert:
Merged with commit 58bbc9b4d5dd777c881566c4c370a3b9fe40585d

Issue #1840, opened 2017-12-18
Add skipper to acceptance tests


sabbyanandan:
Add skipper into acceptance tests https://github.com/spring-cloud/spring-cloud-dataflow-acceptance-tests/

This involves creating new CI plans.

tzolov:
- I've pushed the https://github.com/spring-cloud/spring-cloud-dataflow-acceptance-tests/pull/139
- Another PR will address the test failing in skipper mode due to app names check in different modes

ilayaperumalg:
CI Plan is triggered as a child plan from the classic mode local acceptance test: https://build.spring.io/browse/SCD-ALSK

sabbyanandan:
Local and CF are already addressed. We have a diff issue for K8S.

PR #1860, opened 2017-12-21
Stream updates should reflect the state in SCDF correctly

### LINKED ISSUE #1739: Stream updates should reflect the state in SCDF correctly
As a user, when updating a stream with overrides to app-level properties or deployment properties at runtime, I'd also like to synchronize the state in SCDF.

For a stream,
> stream create foo --definition "time | log" 
> stream skipper deploy foo
> stream skipper update foo --properties "log.spec.applicationProperties.log.level=WARN"

```
dataflow:>stream list
╔═══════════╤═════════════════╤═══════════════════════════════════╗
║Stream Name│Stream Definition│              Status               ║
╠═══════════╪═════════════════╪═══════════════════════════════════╣
║fooz       │time | log       │The app or group is being deployed.║
╚═══════════╧═════════════════╧═══════════════════════════════════╝
```

Though we have upgraded the log-sink with the new property override in a rolling-upgrade manner, the stream DSL still doesn't reflect the override - we have to make it adapt to the change.

**Design notes:**
1) We will reverse engineer stream definition and its application/deployment properties based on the manifest associated with the stream/release name. 
2) Likewise, we will also auto-register any changes to the application versions.

### LINKED ISSUE #1859: Divide the DefaultStreamService logic into SkipperStreamService and SimpleStreamService
Currently `DefaultStreamService` is handling both the Simple (aka Classic) and Skipper deployment functionality. It is hard to follow the maze of Ifs and it is difficult to test. 
DefaultStreamService has to be split into an abstract, common parent and dedicated `Simple` and `Skipper` subclasses all implementing the `StreamService` interface. At run time only one `StreamService` will be started depends on the `skipper-enabled` flag.

tzolov:
Resolves #1739 - Stream updates should reflect the state in SCDF correctly
Resolves #1859 - Divide the DefaultStreamService logic into SkipperStreamService and SimpleStreamService


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/SimpleStreamService.java
```diff
@@ -0,0 +1,121 @@
...
+import static org.springframework.cloud.dataflow.rest.SkipperStream.SKIPPER_KEY_PREFIX;
+
+/**
+ * {@link AppDeployerStreamDeployer} specific {@link AbstractStreamService}.
+ *
+ * @author Mark Pollack
+ * @author Ilayaperumal Gopinathan
+ * @author Christian Tzolov
+ */
+@Service
```
ilayaperumalg: Any specific reason why this is marked with `@Service`?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/SimpleStreamService.java
```diff
@@ -0,0 +1,121 @@
...
+public class SimpleStreamService extends AbstractStreamService {
```
ilayaperumalg: Can we call this call something like `AppDeployerStreamService`?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/stream/manifest/SpringCloudDeployerApplicationManifest.java
```diff
@@ -0,0 +1,112 @@
...
+ *  log.expression: hellobaby
+ * deploymentProperties:
+ *   memory: 1024
+ *   disk: 2
+ * }
+ *
+ * @author Mark Pollack
+ * @author Ilayaperumal Gopinathan
+ */
+public class SpringCloudDeployerApplicationManifest {
```
ilayaperumalg: Instead of duplicating these classes here, Can we move these manifest reader classes in Skipper into `spring-cloud-skipper` module so that we can directly re-use them here in SCDF?

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/service/impl/SkipperStreamServiceIntegrationTests.java
```diff
@@ -87,8 +89,8 @@
 
 	@After
 	public void destroyStream() {
-		appRegistry.delete("log", ApplicationType.sink);
-		appRegistry.delete("time", ApplicationType.source);
+//		appRegistry.delete("log", ApplicationType.sink);
```
ilayaperumalg: did you intend to delete these?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/SimpleStreamService.java
tzolov: ah, the original DefaultStreamService as annotated and i may have missed to clean it while converting it.  will check the abstract and the skipper classes too

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/SimpleStreamService.java
tzolov: I've been trying to avoid names like 'Classic'XXX. The AppDeployerXXX is good because will be consistent with the naming with some of the other non-skipper-path classes. 
My only concern with this name (here and in the other places) is that it is ambiguous given that the Skipper deployer is conceptually an app deployer as well. 
Still I will rename it to AppDeployerStreamService for now and we can think latter  fro better names for all AppDeployerXxxx names

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/stream/manifest/SpringCloudDeployerApplicationManifest.java
tzolov: I intended to make this in a separate ticket, but i guess you a right to fix it now before introducing the duplication. Will fix it

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/service/impl/SkipperStreamServiceIntegrationTests.java
tzolov: ah, i've missed this

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/SkipperStreamService.java
```diff
@@ -0,0 +1,259 @@
...
+
+	private void updateStreamDefinitionFromReleaseManifest(String streamName, String releaseManifest) {
+
+		List<SpringCloudDeployerApplicationManifest> appManifests =
+				new SpringCloudDeployerApplicationManifestReader().read(releaseManifest);
+
+		Map<String, SpringCloudDeployerApplicationManifest> appManifestMap = new HashMap<>();
+
+		for (SpringCloudDeployerApplicationManifest am : appManifests) {
+			String name = am.getSpec().getApplicationProperties().get("spring.cloud.dataflow.stream.app.label");
```
ilayaperumalg: Can refer constants instead of string here.

ilayaperumalg: We can do the same for other places where the string values are used to refer the Data Flow property keys.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/SkipperStreamService.java
```diff
@@ -0,0 +1,259 @@
...
+
+	private void updateAppVersionIfChanged(StreamAppDefinition appDefinition,
+			SpringCloudDeployerApplicationManifest appManifest) {
+
+		String version = appManifest.getSpec().getVersion();
+		//String name = appManifest.getSpec().getApplicationProperties().get("spring.cloud.dataflow.stream.app.label");
+		String name = appDefinition.getRegisteredAppName();
+		ApplicationType type = ApplicationType.valueOf(
+				appManifest.getSpec().getApplicationProperties().get("spring.cloud.dataflow.stream.app.type"));
+		String resource = appManifest.getSpec().getApplicationProperties().get("spring.cloud.dataflow.stream.app.type");
```
ilayaperumalg: This needs to be `appManifest.getSpec().getResource()`

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/SkipperStreamService.java
```diff
@@ -0,0 +1,259 @@
...
+
+		if (!this.appRegistryService.appExist(name, type, version)) {
```
ilayaperumalg: We probably need some tests (maybe as part of this PR or as a separate PR) to check the flow here.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/SkipperStreamService.java
```diff
@@ -0,0 +1,259 @@
...
+			this.appRegistryService.save(name, type, version, URI.create(resource), metadataUri);
+		}
+	}
+
+	@Override
+	public void doRollbackStream(String streamName, int releaseVersion) {
+		this.skipperStreamDeployer.rollbackStream(streamName, releaseVersion);
+	}
+
+	public String convertPropertiesToSkipperYaml(String streamName, Map<String, String> updateProperties) {
```
ilayaperumalg: Can we move this into somewhere into skipper domain (as it involves reading values from Skipper defined spec) and use it here? Maybe having it in ManifestUtils of Skipper would be better.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/configuration/TestDependencies.java
```diff
@@ -166,52 +171,62 @@ public FeaturesProperties featuresProperties() {
...
-				featuresProperties);
+				appDeploymentRequestCreator);
 	}
 
 	@Bean
-	AppDeploymentRequestCreator streamDeploymentPropertiesUtils(AppRegistryCommon appRegistry,
-																CommonApplicationProperties commonApplicationProperties,
-																ApplicationConfigurationMetadataResolver applicationConfigurationMetadataResolver) {
-		return new AppDeploymentRequestCreator(appRegistry,
+	@ConditionalOnSkipperDisabled
```
ilayaperumalg: This is nice!

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/service/impl/SkipperStreamServiceTests.java
```diff
@@ -123,83 +107,95 @@ public void setupMock() {
...
 		skipperDeployerStates.put(this.streamDefinition2, DeploymentState.undeployed);
 		skipperDeployerStates.put(this.streamDefinition3, DeploymentState.failed);
+		skipperDeployerStates.put(this.streamDefinition4, DeploymentState.deployed);
+
 		when(this.skipperStreamDeployer.state(this.skipperStreamDefinitions)).thenReturn(skipperDeployerStates);
-		Map<StreamDefinition, DeploymentState> states = this.defaultStreamService.state(this.streamDefinitionList);
+
+		Map<StreamDefinition, DeploymentState> states = this.skipperStreamService.state(this.streamDefinitionList);
+
+		System.out.println(states.size());
```
ilayaperumalg: I believe this can be removed.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/AbstractStreamService.java
```diff
@@ -0,0 +1,185 @@
...
+import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
+
+/**
+ * Performs manipulation on application and deployment properties, expanding shorthand
+ * application property values, resolving wildcard deployment properties, and creates a
+ * {@link StreamDeploymentRequest}.
+ * </p>
+ * The {@link AbstractStreamService} deployer is agnostic service. For deploying streams on
+ * Skipper use the {@link SkipperStreamService} and for the Simple mode stream deployment use
```
ilayaperumalg: We can probably say deployer via `AppDeployer` implementation as saying simple mode could be confusing.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/AbstractStreamService.java
```diff
@@ -0,0 +1,185 @@
...
+		for (StreamDefinition streamDefinition : streamDefinitions) {
+			StreamDeployment streamDeployment = this.streamDeploymentRepository.findOne(streamDefinition.getName());
+			if (streamDeployment == null) {
+				states.put(streamDefinition, DeploymentState.unknown);
+			}
+			else if (this.streamDeployer == StreamDeployers.valueOf(streamDeployment.getDeployerName())) {
+				deployerSpecificStreamDefinitions.add(streamDefinition);
+			}
+			else {
+				logger.warn("Deployer incompatible stream definition:" + streamDefinition);
```
ilayaperumalg: I think we need to throw an exception here as it is not expected to have an entry of deployer other than `stream` and `skipper`.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/AppDeployerStreamService.java
```diff
@@ -0,0 +1,119 @@
...
+import org.springframework.cloud.dataflow.server.stream.StreamDeploymentRequest;
+import org.springframework.cloud.deployer.spi.app.DeploymentState;
+import org.springframework.cloud.deployer.spi.core.AppDeploymentRequest;
+import org.springframework.cloud.skipper.domain.PackageIdentifier;
+import org.springframework.util.Assert;
+
+import static org.springframework.cloud.dataflow.rest.SkipperStream.SKIPPER_KEY_PREFIX;
+
+/**
+ * {@link AppDeployerStreamDeployer} specific {@link AbstractStreamService}.
```
ilayaperumalg: We could refer `{@link AppDeployer}`

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/AbstractStreamService.java
tzolov: agree

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/SkipperStreamService.java
tzolov: addressed

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/SkipperStreamService.java
tzolov: ah, this is big one. good spot

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/SkipperStreamService.java
tzolov: You have a point i was thinking to add tests in the context of #1840, but i would try to add some unit tests for the logic here.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/SkipperStreamService.java
tzolov: This is a good idea, but since this code is not created by this PR and furthermore this suggestion would require and depend on committing code in another project (e.g. skipper)
So i would suggest to do this in separate ticket.
Btw the ManifestUtils is not accessible form scdf and will need to be moved as well.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/AbstractStreamService.java
tzolov: right

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/AppDeployerStreamService.java
tzolov: Not  sure in what context should i refer it?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/SkipperStreamService.java
ilayaperumalg: > This is a good idea, but since this code is not created by this PR and furthermore this suggestion would require and depend on committing code in another project (e.g. skipper)

+1

> ManifestUtils is not accessible form scdf and will need to be moved as well.

Yes, similar to what we did for other Manifest reader.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/AppDeployerStreamService.java
ilayaperumalg: I meant it to be something like `StreamService` that uses AppDeployer specific implementation`. But, it is nice to have as the current Javadoc inherently refers it.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed and merged as c6d5e59d475d2190803e1b8d33ebcd57cdb650b9

PR #1861, opened 2017-12-22
Retrieve app version without downloading

### LINKED ISSUE #1843: Retrieve the app version from a Resource without downloading entire maven artifact
Currently the LRUCleaningResourceLoader overrides the delegated resource loader and causes full artifact download on getResource(uri) method call

ilayaperumalg:
- Instead of invoking `getResource` via the underlying LRUCleaningResourceLoader, construct the resource based on the schemes used in the URI
  - For maven resource, parse the co-ordinates and construct MavenResource with the appropriate MavenProperties (configured in the server)
   - Make sure to inject the maven properties

 - Remove `getResource` from AppRegistryService as we no longer need to download the resource at the registry
 - Use DefaultResourceLoader in AppRegistry for the use case where the `--uri` option is used to download the http resource when `registerAll`
 - The delegating resource loader is now used only when downloading the metadata resource at the AppRegistry

 - Add tests for resource version retrieval

Resolves #1843

### DIFF
--- a/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/AbstractAppRegistryCommon.java
+++ b/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/AbstractAppRegistryCommon.java
@@ -17,6 +17,7 @@ import org.slf4j.LoggerFactory;
 import org.springframework.cloud.dataflow.core.ApplicationType;
 import org.springframework.cloud.dataflow.registry.domain.AppRegistration;
 import org.springframework.cloud.dataflow.registry.support.ResourceUtils;
+import org.springframework.cloud.deployer.resource.maven.MavenProperties;
 import org.springframework.core.io.Resource;
 import org.springframework.core.io.ResourceLoader;
 import org.springframework.core.io.support.PropertiesLoaderUtils;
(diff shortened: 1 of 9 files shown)


tzolov:
@ilayaperumalg , perhaps we can encapsulate the `ResourceUtils` internals behind a `ResourceServer` which is auto-wired with all necessary dependencies (such as MavenProperties) without exposing them.  
I've experimented with the idea here and IMO it look good: https://github.com/tzolov/spring-cloud-dataflow/commit/f9900b309e6a983d5f5500cefe32585bfb1863a4
Perhaps we can even try to encapsulate the access to the ResourceLoader instances in the same services?

tzolov:
@ilayaperumalg I've also squashed an encapsulation of the ResourceLoader access through the same ResourceService: https://github.com/tzolov/spring-cloud-dataflow/commit/7990f93af3c025f1e187158e80369072c826aa07
The ResourceUtils access is restricted to package level to prevent unintentional direct access to this utility. If the above approach is ok, then we can completely merge the ResourceUtils code into the ResourceService and get ride of the static methods.

tzolov:
As agreed we will proceed with current ResourceUtils implementation and will tackle the suggested encapsulation improvements in https://github.com/spring-cloud/spring-cloud-dataflow/issues/1873

Current code LGTM and i'm about to merge it.

PR #1879, opened 2018-01-02
Split the Skipper and Non-Skipper configurations and functionality

### LINKED ISSUE #1869: Move the SkipperClient from the StreamDeploymentController&RuntimeAppsController to services down in the stack
Currently the StreamDeploymentController uses to channels to interact with Skipper: (1) StreamService and (2) directly via SkipperClient. 
The SkipperClient needs to be moved under the StreamService instead.
Similarly `RuntimeAppsController` has direct SkipperClient access and tries to mange skipper and non-skipper modes at runtime.

### LINKED ISSUE #1872: Uncluttered the Skipper and Non-Skipper Configurations. Use dedicated configuration  classes for  both modes
Currently all Skipper and non-Skipper @Beans are created inside the same `DataFlowControllerAutoConfiguration`,  using a conditional annotations for each bean to control when the skipper and non-skipper modes are enabled. 
Create instead an dedicated Skipper and NonSkipper Configuration classes (inline or external) and annotate the entire class with the conditional annotation. 

uncluttered

tzolov:
Resolves #1869 
Resolves #1872

 - Use StreamDeployer interface to abstract Skipper and AppDeployer functionality from top-level Controllers.
 - Move SkipperClient from RuntimeAppsController to SkipperStreamDeployer
 - Move SkipperClient from StreamDeploymentController to SkipperStreamDeployer
 - Refactor DataFlowControllerAutoConfiguration to have clear wiring for Skipper and Non-Skipper beans
 - Move AppDeployer from AboutController to StreamDeployer
 - Move AppDeployer from RuntimeController to StreamDeployer
 - Resolve missing support for App state and instance state in  RuntimeController

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/StreamService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/StreamService.java
@@ -15,18 +15,24 @@
  */
 package org.springframework.cloud.dataflow.server.service;
 
+import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 
 import org.springframework.cloud.dataflow.core.StreamDefinition;
 import org.springframework.cloud.dataflow.rest.UpdateStreamRequest;
+import org.springframework.cloud.dataflow.server.repository.IncompatibleStreamDeployerException;
+import org.springframework.cloud.dataflow.server.stream.StreamDeployers;
 import org.springframework.cloud.deployer.spi.app.DeploymentState;
+import org.springframework.cloud.skipper.domain.Deployer;
+import org.springframework.cloud.skipper.domain.Release;
 
 /**
  * Specify the supported operations on the stream.
  *
  * @author Mark Pollack
  * @author Ilayaperumal Gopinathan
+ * @author Christian Tzolov
  */
 public interface StreamService {
 	/**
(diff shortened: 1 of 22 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/StreamService.java
```diff
@@ -51,7 +65,9 @@
 	 * @param streamName the name of the stream to update
 	 * @param updateStreamRequest the UpdateStreamRequest to use during the update
 	 */
-	void updateStream(String streamName, UpdateStreamRequest updateStreamRequest);
+	default void updateStream(String streamName, UpdateStreamRequest updateStreamRequest) {
+		throw new IncompatibleStreamDeployerException(StreamDeployers.appdeployer.toString());
```
ilayaperumalg: Do we really want to refer `appDeployer` here? Maybe it is better to keep this generic and refer `appDeployer` in its own StreamService.

ilayaperumalg:
LGTM except those default interface methods using `appDeployer` reference.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/StreamService.java
tzolov: Ok, i will remove all defaults from the `StreamDeployer` and `StreamService` interfaces and move the exception throwing to the implementing classes.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed and merged as 3c88714f4295ac8fe7c0c34cc79d310820a84533

PR #1882, opened 2018-01-04
Store stream deployment properties

### LINKED ISSUE #1738: Include deployment properties in the stream repository
As a developer, I'd like to also capture and update the stream definition with the deployment properties. 

**Design notes:**
We will reverse engineer stream definition and its deployment properties based on the latest  "version" of the manifest from Skipper. In other words, this is strictly only applicable to Skipper driven workflows.

### LINKED ISSUE #1848: Add "stream info" command
As a developer, I'd like to add a new `stream info --name <STREAM_NAME>` command, so it can be useful to review the details of the stream. This approach is cleaner as opposed to shoving all the information in `stream list`, which is already a noisy table listing.

**Design notes:**
1) `stream info --name <STREAM_NAME>` command shows the details of the stream including apps, its version, and the app/deployment properties associated with it.

2) The backing REST endpoint will be exposed in the root-controller.

3) The backing REST endpoint will be mapped to security roles; specifically for `GET` invocations.

4) Include the backing endpoint in the RESTDocs infrastructure.

ilayaperumalg:
- When deploying stream, save the deployment properties into STREAM_DEPLOYMENTS table
  - Currently, the deployment properties are updated for both AppDeployer/Skipper modes
 - Along with deployment properties, store the app versions as the default version for the app could be changed by the user after the stream is deployed.

Resolves #1738

 - Add REST endpoint to GET stream info in StreamDeploymentController
  - This controller is chosen in order to retrieve both the StreamDefinition as well as the deployment information from the corresponding repositories
  - Inject StreamDeploymentRepository as a dependency to StreamDeploymentController
  - Add HATEOAS resource assembler that constructs the StreamDeploymentResource
  - Add tests

 - Update StreamDeploymentResource extend from StreamDefinitionResource
  - This helps reusing stream DSL and status

 - Add `stream info` Shell command
  - Retrieves StreamDeploymentResource and constructs the shell output like this:

```
dataflow:>stream info a1
╔══════════════════════════════╤══════════════════════════════╤══════════════════════════════╗
║             Name             │             DSL              │            Status            ║
╠══════════════════════════════╪══════════════════════════════╪══════════════════════════════╣
║a1                            │time | log                    │deploying                     ║
╚══════════════════════════════╧══════════════════════════════╧══════════════════════════════╝

Deployed App versions: {
  "log" : "1.3.0.RELEASE",
  "time" : "1.3.0.RELEASE"
}
Stream Deployment properties: {
  "log" : {
    "spring.cloud.deployer.indexed" : "true",
    "spring.cloud.deployer.group" : "a1"
  },
  "time" : {
    "spring.cloud.deployer.group" : "a1"
  }
}

```

Resolves #1848


ilayaperumalg:
Updated the PR based on the redesign.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
```diff
@@ -152,4 +174,45 @@ public String manifest(@PathVariable("name") String name, @PathVariable("version
...
+
+	/**
+	 * {@link org.springframework.hateoas.ResourceAssembler} implementation that converts
+	 * {@link StreamDeployment}s to {@link StreamDeploymentResource}s.
+	 */
+	class Assembler extends ResourceAssemblerSupport<StreamDeployment, StreamDeploymentResource> {
+
+		private final StreamDefinition streamDefinition;
+
+		public Assembler(StreamDefinition streamDefinition) {
```
tzolov: I feel like, we should be passing here only the pre-computed properties rather than an entire domain class?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
```diff
@@ -152,4 +174,45 @@ public String manifest(@PathVariable("name") String name, @PathVariable("version
...
+			}
+			return null;
+		}
+
+		@Override
+		public StreamDeploymentResource instantiateResource(StreamDeployment stream) {
+			final StreamDeploymentResource resource = new StreamDeploymentResource(this.streamDefinition.getName(),
+					ArgumentSanitizer.sanitizeStream(this.streamDefinition.getDslText()), stream.getDeploymentProperties()
+			);
+			Map<StreamDefinition, DeploymentState> streamDeploymentStates = StreamDeploymentController.this.streamService.state(Arrays.asList(this.streamDefinition));
```
tzolov: As with the previous comment we should free the Assembler of an business logic. E.g. The the Assembler should not be aware of any business logic or domain class relationships.  It's sole purpose is to turn a single domain class (StreamDeployment in this case) into a Resource

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/AbstractStreamService.java
```diff
@@ -86,14 +83,7 @@ public void deployStream(String name, Map<String, String> deploymentProperties)
...
 	public void undeployStream(String streamName) {
-		StreamDeployment streamDeployment = this.streamDeploymentRepository.findOne(streamName);
-		if (streamDeployment != null) {
-			if (this.streamDeployer != StreamDeployers.valueOf(streamDeployment.getDeployerName())) {
-				throw new IncompatibleStreamDeployerException(streamDeployer.name());
-			}
-			doUndeployStream(streamName);
-			this.streamDeploymentRepository.delete(streamName);
-		}
+		doUndeployStream(streamName);
```
tzolov: There is not point in a separate doUndeployStream method anymore as its calling method (undeployStream) is empty. We can get rid of the doUndeployStream and have the child classes override the undepoyStream instead.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/AbstractStreamService.java
```diff
@@ -121,25 +111,7 @@ else if (DeploymentState.deploying.equals(DeploymentState.valueOf(status))) {
...
-			else {
-				logger.error("Invalid deployer:" + streamDefinition.getName() + ":" + streamDeployment.getDeployerName());
-				throw new IncompatibleStreamDeployerException(this.streamDeployer.name());
-			}
-		}
-		if (!deployerSpecificStreamDefinitions.isEmpty()) {
-			states.putAll(this.doState(deployerSpecificStreamDefinitions));
-		}
-		return states;
+		return this.doState(streamDefinitions);
```
tzolov: similar as the doUndeployStream we can remove the redundant doState and use the state method instead.

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/StreamDeployment.java
```diff
@@ -31,59 +32,26 @@
 	private final String streamName;
 
 	/**
-	 * The deployer used for the stream deployment. The deployer could be app deployer or Skipper.
+	 * The JSON String value of the deployment properties Map<String, String> values.
```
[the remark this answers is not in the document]
ilayaperumalg: JSON String turned out to be just the choice when storing the deployment properties map as String.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
ilayaperumalg: Yeah, this technical debt is a long due as you would notice similar pattern in some other resource assemblers as well. I will try to fix it for this assembler.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/AbstractStreamService.java
ilayaperumalg: +1

spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/resource/StreamDeploymentResource.java
```diff
@@ -17,14 +17,48 @@
...
-import org.springframework.hateoas.ResourceSupport;
 
 /**
  * A HATEOAS representation of a stream deployment.
  *
  * @author Eric Bottard
+ * @author Ilayaperumal Gopinathan
  */
-public class StreamDeploymentResource extends ResourceSupport {
+public class StreamDeploymentResource extends StreamDefinitionResource {
```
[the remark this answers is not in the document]
ilayaperumalg: yeah, I thought through that before adding the StreamDefinitionResource as the parent one. Went with this way just to avail those properties. But, I agree from the object representation. I guess we can simply replicate the required properties into `StreamDeploymentResource`.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/AbstractStreamService.java
tzolov: we can remove the `doState` and use directly the `state` method instead

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentControllerTests.java
```diff
@@ -59,11 +69,19 @@
 	@Mock
 	private StreamService defaultStreamService;
 
+	@Mock
```
tzolov: this is not being used

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentControllerTests.java
```diff
@@ -59,11 +69,19 @@
...
+	private StreamDeploymentRepository streamDeploymentRepository;
+
 	@Mock
 	private Deployer deployer;
 
+	@Mock
+	private ServletRequestAttributes attrs;
```
tzolov: this is not being used

tzolov:
Squashed, rebased and merged as ff54bef595b055eaadfc27c4bdc884e32e9a8bb5

PR #1884, opened 2018-01-04
Handle missing RuntimeEnvironmentInfo in AboutController

### LINKED ISSUE #1883: AboutController to handle gracefully missing RuntimeEnvironmentInfo
Currently the RuntimeEnvironmentInfo is not available in Skipper mode (due to https://github.com/spring-cloud/spring-cloud-skipper/issues/392)
Instead an exception is thrown that propagates through AboutController#getAboutResource method

tzolov:
Resolves #1883

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AboutController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AboutController.java
@@ -168,7 +168,7 @@ public class AboutController {
 
 				runtimeEnvironment.setAppDeployer(deployerInfo);
 			} catch (UnsupportedOperationException uoe) {
-				logger.debug("expected in skipper mode due to #392");
+				logger.debug("RuntimeEnvironmentInfo is not supported!");
 			}
 		}
 
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AboutController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AboutController.java
@@ -149,23 +149,27 @@ public class AboutController {
 		final RuntimeEnvironment runtimeEnvironment = new RuntimeEnvironment();
 
 		if (this.streamDeployer != null) {
-			final RuntimeEnvironmentInfo deployerEnvironmentInfo = this.streamDeployer.environmentInfo();
-			final RuntimeEnvironmentDetails deployerInfo = new RuntimeEnvironmentDetails();
+			try {
+				final RuntimeEnvironmentInfo deployerEnvironmentInfo = this.streamDeployer.environmentInfo();
+				final RuntimeEnvironmentDetails deployerInfo = new RuntimeEnvironmentDetails();
 
-			deployerInfo.setDeployerImplementationVersion(deployerEnvironmentInfo.getImplementationVersion());
-			deployerInfo.setDeployerName(deployerEnvironmentInfo.getImplementationName());
-			deployerInfo.setDeployerSpiVersion(deployerEnvironmentInfo.getSpiVersion());
-			deployerInfo.setJavaVersion(deployerEnvironmentInfo.getJavaVersion());
-			deployerInfo.setPlatformApiVersion(deployerEnvironmentInfo.getPlatformApiVersion());
-			deployerInfo.setPlatformClientVersion(deployerEnvironmentInfo.getPlatformClientVersion());
-			deployerInfo.setPlatformHostVersion(deployerEnvironmentInfo.getPlatformHostVersion());
-			deployerInfo.setPlatformSpecificInfo(deployerEnvironmentInfo.getPlatformSpecificInfo());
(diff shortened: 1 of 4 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AboutController.java
```diff
@@ -149,23 +149,27 @@ public AboutResource getAboutResource() {
...
+				deployerInfo.setPlatformHostVersion(deployerEnvironmentInfo.getPlatformHostVersion());
+				deployerInfo.setPlatformSpecificInfo(deployerEnvironmentInfo.getPlatformSpecificInfo());
+				deployerInfo.setPlatformHostVersion(deployerEnvironmentInfo.getPlatformHostVersion());
+				deployerInfo.setPlatformType(deployerEnvironmentInfo.getPlatformType());
+				deployerInfo.setSpringBootVersion(deployerEnvironmentInfo.getSpringBootVersion());
+				deployerInfo.setSpringVersion(deployerEnvironmentInfo.getSpringVersion());
+
+				runtimeEnvironment.setAppDeployer(deployerInfo);
+			} catch (UnsupportedOperationException uoe) {
+				logger.debug("expected in skipper mode due to #392");
```
ilayaperumalg: This message can be generic as there could be other modes in the future. Maybe we don't need to specify Skipper's 392 here.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/stream/SkipperStreamDeployer.java
```diff
@@ -352,7 +351,7 @@ public AppStatus getAppStatus(String id) {
 	@Override
 	public RuntimeEnvironmentInfo environmentInfo() {
 		// TODO To fix when https://github.com/spring-cloud/spring-cloud-skipper/issues/392
-		throw new IncompatibleStreamDeployerException("AppDeployer");
+		throw new UnsupportedOperationException("EnvironmentInfo is not supported for Skipper yet (issue#392)");
```
ilayaperumalg: Having Skipper in the message is good. I think `(issue#392)` can be removed as the comment there is sufficient.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/AboutControllerTests.java
```diff
@@ -159,4 +161,55 @@ public void testChecksumNoDefaults() throws Exception {
 					.andExpect(jsonPath("$.versionInfo.shell.checksumSha256").doesNotExist());
 		}
 	}
+
+	@RunWith(SpringRunner.class)
+	@SpringBootTest(classes = TestDependencies.class, properties = {"spring.cloud.dataflow.features.skipper-enabled=true"})
```
ilayaperumalg: I am not sure why does this property use `true` for skipper-enabled.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AboutController.java
tzolov: +1

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/AboutControllerTests.java
tzolov: Because we want to test that the the about controller handles the  `UnsupportedOperationException` thrown when in skipper mode.  
The TestDependecy#aboutController bean creation uses the `skipper-enabled` condition to throw to throw the exception for `streamDeployer.environmentInfo()`

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/stream/SkipperStreamDeployer.java
tzolov: +1

ghillert:
It works but the REST response looks a bit funky:

```
...
  "runtimeEnvironment": {
    "appDeployer": {
      "deployerImplementationVersion": null,
      "deployerName": null,
      "deployerSpiVersion": null,
      "javaVersion": null,
      "platformApiVersion": null,
      "platformClientVersion": null,
      "platformHostVersion": null,
      "platformSpecificInfo": {},
      "platformType": null,
      "springBootVersion": null,
      "springVersion": null
    },
...
```

ghillert:
@sabbyanandan On the UI it will look like this:

image

tzolov:
@ghillert this is a temporal workaround for until the  spring-cloud/spring-cloud-skipper#392 is resolved. 
When 392 is resolved, SCDF will be able to populate with the Runtime Env info retrieved from Skipper's deployers.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed, rebased and merged as 101d45ec

PR #1897, opened 2018-01-08
Resolve #1868 :  Shell commands compatible to the DataFlow mode

### LINKED ISSUE #1868: SCDF shell should enable Skipper commands only for Feature: skipper-enabled=true
The skipper and classic modes on the server are controlled via the skipper-enabled feature.  The feature status is exposed through the about entry point and can be used to allow or disallow skipper related commands on the client side. Similar approach  is used for Task commands

tzolov:
* SCDF shell loads only shell commands (and beans) applicable for the DataFlow mode of the server it connects to. Supported modes are: `classic` [*default*] or `skipper`.
* No need off mode specific, command prefixes (e.g. stream skipper xxx) in the commands' signatures.
* Shell is started with a `--dataflow.mode=` to set the shell's mode to correspond to the server mode. Missing argument starts the shell in classic mode.
* If shell's mode is different from the server such (verified during dataflow config serverinitialization) and error message is printed with instructions to use the correct mode.
* Retain backward compatibility (when possible) for the entry points backing the commands from all modes. (API evolution)

Changes:
- Add DataFlowMode enum and IncompatibleDataFlowMode exception.
- ConfigCommands#target is extended to verify that the shell is started in the same mode as the server.
- Use the AboutResource#FeaturesInfo#Skipper-Enabled to determine the server mode.
- Move the common for all modes shell commands from the o.s.c.d.shell.command into o.s.c.d.shell.command.common package.
- Create o.s.c.d.shell.command.skipper package to hold the skipper specific commands.
- Create o.s.c.d.shell.command.classic package to hold the classic specific commands.
- Refactor o.s.c.d.s.c.common.AppRegistryCommands into o.s.c.d.shell.command.common.AbstractAppRegistryCommands, o.s.c.d.s.c.classic.ClassicAppRegistryCommands and o.s.c.d.s.c.skipper.SkipperAppRegistryCommands to split the common and mode-specific app commands.
- For backward compatibility all XXXAppRegistryCommands use the same AppCommandsOperations.
- Refactor o.s.c.d.s.c.StreamCommands into o.s.c.d.s.c.common.AbstractStreamCommands, o.s.c.d.s.c.classic.ClassicStreamCommands and o.s.c.d.s.c.skipper.SkipperStreamCommands to split the common and mode-specific stream commands.
- Remove the `skipper` prefixes for the stream commands in skipper mode.
- For backward compatibility all XXXStreamCommands use the same StreamOperations.
- Refactor the o.s.c.d.s.a.BaseShellAutoConfiguration#RegisterInteranlCommands bean into conditional ClassicRegisterInternalCommands (scans the common + classic command pkgs) and SkipperRegisterInternalCommands ( scans the common + skipper command pkgs).
- Use the @ConditionalOnProperty(name = "dataflow.mode"...) condition
- Add dataflow.mode to the usage.txt

### DIFF
--- a/spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/autoconfigure/BaseShellAutoConfiguration.java
+++ b/spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/autoconfigure/BaseShellAutoConfiguration.java
@@ -24,6 +24,7 @@ import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.ApplicationArguments;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.cloud.dataflow.shell.ShellCommandLineParser;
 import org.springframework.cloud.dataflow.shell.ShellProperties;
 import org.springframework.cloud.dataflow.shell.TargetHolder;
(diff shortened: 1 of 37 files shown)


spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/autoconfigure/BaseShellAutoConfiguration.java
```diff
@@ -91,14 +92,33 @@ public void log() {
 	}
 
 	@Configuration
-	@ComponentScan({ "org.springframework.shell.commands", "org.springframework.cloud.dataflow.shell.command",
+	@ComponentScan({ "org.springframework.shell.commands", "org.springframework.cloud.dataflow.shell.command.common",
+			"org.springframework.cloud.dataflow.shell.command.classic",
 			"org.springframework.cloud.dataflow.shell.converter", "org.springframework.cloud.dataflow.shell.config" })
-	public static class RegisterInternalCommands {
+	@ConditionalOnProperty(name = "dataflow.mode", havingValue = "classic", matchIfMissing = true)
+	public static class ClassicRegisterInternalCommands {
```
ilayaperumalg: I understand this separation of packages gives a clear organization of classic/skipper commands which is nice. 

I am just wondering if we still need to support the case where some users could have already extended these commands and hence do we need to keep the existing commands (the to-be classic commands) in their current package itself. Also, I am sure this situation is not common and users would still be using our `DataFlowTemplate` for REST client extension, making it in the current package could be safer if we decide to support this case.

tzolov: Here is my opinion on this:
- In my understanding the shell commands are internal components rather a publicly supported API? 
- Furthermore even if we retain the common commands in the original o.s.c.d.s.command package we still will have to remove the existing AppRegistryCommand and StreamCommands from the common 'commands' package so we can control what gets loaded. 
- As pointed out this PR doesn't impact neither the REST APIs nor the XxxOptions templates

ilayaperumalg:
LGTM, merging. Lets create a different story if the comment https://github.com/spring-cloud/spring-cloud-dataflow/pull/1897#discussion_r160171021 has any impact.

PR #1901, opened 2018-01-09
Skipper package - app version inside specMap

### LINKED ISSUE #1900: Skipper package creation should use app version inside spec map
With the changes from https://github.com/spring-cloud/spring-cloud-skipper/issues/407. we need to use app version inside spec map

ilayaperumalg:
- Set `version` as one of the properties inside SpecMap

Resolves #1900


ilayaperumalg:
This depends on https://github.com/spring-cloud/spring-cloud-skipper/pull/411

tzolov:
Squashed and merged at 22fab2f8

PR #1903, opened 2018-01-09
Fix Skipper manifest command

### LINKED ISSUE #1902: Skipper manifest cmd doesn't work when version not specified
The skipper manifest command should use both the `/name` and `/name/version` REST endpoints at Skipper server to serve the requests that come with/without version in it.

Currently, the following error occurs:
```
dataflow:>stream skipper manifest a1
Command failed org.springframework.cloud.dataflow.rest.client.DataFlowClientException: Release with the name [a1] doesn't exist

```

ilayaperumalg:
- Use Skipper server controller's `/manifest/name/version` only when a valid version int is used. Otherwise, use `/manifest/name` REST endpoint
 - Update test

Resolves #1902


tzolov:
Rebased and merged as 10231c24cae8bf7660aa63954d7d97294f702e39

PR #1905, opened 2018-01-09
Deploy/Update Streams with registered apps

### LINKED ISSUE #1881: Stream update accepts only registered applications
As a continuation of spring-cloud/spring-cloud-dataflow#1736 (refer to **8.**), as a developer, I'd like to add validation support to fail-fast if the version used in the stream-skipper update command is not already a registered entry in the App Registry.

**Acceptance:**
- Remove the automatic [app-registration](b650b9/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/SkipperStreamService.java#L198) routine - it is not needed anymore.
- Stream update command must validate to make sure the supplied version is already a registered artifact.
- Upon validation failure, show the error message (_"The version you're trying to update is not available in the App Registry. Please register the app and the companion metadata for its properties to proceed with the updates."_) as part of the REST response and in Shell.

tzolov:
Resolves #1881

- Add app registration validation during skipper stream deploy
- Add app registration validation during skipper stream update
- Add tests

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/SkipperStreamService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/SkipperStreamService.java
@@ -15,7 +15,6 @@
  */
 package org.springframework.cloud.dataflow.server.service.impl;
 
-import java.net.URI;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.LinkedList;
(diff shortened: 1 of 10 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/SkipperStreamService.java
```diff
@@ -161,7 +152,7 @@ private void updateStreamDefinitionFromReleaseManifest(String streamName, String
 			updatedStreamAppDefinitions.addLast(appDefinitionBuilder.build(streamDefinition.getName()));
 
 			//Check and register any missing app version
-			updateAppVersionIfChanged(appDefinition, applicationManifest);
+			//this.skipperStreamDeployer.updateAppVersionIfChanged(appDefinition, applicationManifest);
```
ilayaperumalg: I guess we can remove this.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/SkipperStreamService.java
```diff
@@ -243,7 +221,9 @@ public String convertPropertiesToSkipperYaml(String streamName, Map<String, Stri
 				appMap.put(SpringCloudDeployerApplicationManifest.SPEC_STRING, specMap);
 			}
 			if (appDeploymentRequest.getCommandlineArguments().size() == 1) {
-				appMap.put("version", appDeploymentRequest.getCommandlineArguments().get(0));
+				String version = appDeploymentRequest.getCommandlineArguments().get(0);
```
ilayaperumalg: Just a note to make the changes (either here or in the PR where we changed `spec.version`) to put `version` inside specMap.

ilayaperumalg:
LGTM, merging along with the minor fixes for the above comments.

ilayaperumalg:
Rebased and merged as 395a318945b71c3632c30c67c94034415a2dc7d4 along with 8f47c2e9ba080ab430045597e472154d65425326

PR #1925, opened 2018-01-11
Add '--deploy' option for 'stream create' when in Skipper mode

### LINKED ISSUE #1907: Add `--deploy` option for stream create when on Skipper mode

tzolov:
Resolves #1907

 - When the SKIPPER_PACKAGE_VERSION deployment property is not provided the SkipperStreamService#doDeployStream will use the default DEFAULT_SKIPPER_PACKAGE_VERSION = "1.0.0"
 - In AbstractStreamCommmands (e.g. all modes) create a stream-create command implementation that supports an optional '--deploy' argument.
 - Remove the mode specific stream-create command implementations.
 - Use of the DeploymentState enum instead of string state values. Change StreamDeployer#calculateStreamState return type
 - Implement the SkipperStreamDeployer##calculateStreamState(stream)
 - Add tests for deploying streams with pre-defined and default package versions


ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Rebased and merged as 7aa39e05

PR #1931, opened 2018-01-11
Added exclusion for jackson in the common-security-config-web depende…

### LINKED ISSUE #1930: Exclude Jackson from Spring Cloud Common Security dependency
Due to `Spring Cloud Common Security` we pull in an old `jackson-core-asl` `1.9.13`. We should exclude that.

see also https://github.com/spring-cloud/spring-cloud-dataflow/issues/1922

cppwfs:
…ncy.

resolves #1930

### DIFF
--- a/spring-cloud-dataflow-server-core/pom.xml
+++ b/spring-cloud-dataflow-server-core/pom.xml
@@ -52,6 +52,12 @@
 		<dependency>
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-starter-common-security-config-web</artifactId>
+			<exclusions>
+				<exclusion>
+					<groupId>org.codehaus.jackson</groupId>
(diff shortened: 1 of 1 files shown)


ilayaperumalg:
@cppwfs @ghillert Could you share the issue you ran into when using this dependency?

cppwfs:
@ilayaperumalg I didn't see any issues with it other than it was an older version of Jackson in the dependency tree.  @ghillert may have seen an issue, since he reported it.

ghillert:
@ilayaperumalg No functional issue ... no reason to include an old version of Jackson though. The problem is that due to the different artifact id, but dependencies get included in the resulting JAR.

ghillert:
LGTM - merged with commit 1e41f6e8dce5da12e13af17f72ece28f7842e96a

PR #1948, opened 2018-01-17
Sync Skipper client changes

### LINKED ISSUE #1947: Update Skipper client delete request
We need to update the changes from https://github.com/spring-cloud/spring-cloud-skipper/pull/475

ilayaperumalg:
- Update release delete request changes
  - Use boolean path variable instead of DeleteProperties
  - Update tests
 - Remove release history --max option
  - Remove the references of max revisions from history command, client and controller side
 - Fix Release.setManifest changes

Resolves #1947

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
@@ -36,6 +36,7 @@ import org.springframework.cloud.dataflow.server.service.StreamService;
 import org.springframework.cloud.deployer.spi.app.AppDeployer;
 import org.springframework.cloud.deployer.spi.app.DeploymentState;
 import org.springframework.cloud.skipper.domain.Deployer;
+import org.springframework.cloud.skipper.domain.Release;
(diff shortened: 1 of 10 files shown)


ilayaperumalg:
Rebased.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
```diff
@@ -171,13 +170,6 @@ public String manifest(@PathVariable("name") String name, @PathVariable("version
 		return this.streamService.manifest(name, version);
 	}
 
-	@RequestMapping(path = "/history/{name}/{max}", method = RequestMethod.GET)
```
tzolov: why is the history entry point removed?

ilayaperumalg: sorry, this was my mistake. In Skipper's ReleaseController we got rid of the entire REST endpoint as it is served from the repository. But, here we use an explicit REST endpoint and it is required. will re-add.

tzolov:
squashed and merged at 6e2f0aaa

PR #1952, opened 2018-01-17
Documentation on server modes

### LINKED ISSUE #1932: Document Skipper and Classic modes
As a developer, I'd like to document Skipper and Classic modes and the differences.

**Acceptance:**
- Remove the existing `stream skipper ... ` [references](https://docs.spring.io/spring-cloud-dataflow/docs/1.3.0.M3/reference/htmlsingle/#spring-cloud-dataflow-streams-skipper) from the docs.
- Update "Streams with Skipper" section with the cosolidated commands and the prerequisites. 
- Consider adding a table walking through a use-case workflow. Perhaps a simple stream's create, deploy, and undeploy experience on one side and then the stream create, deploy, update, undeploy workflow on the other. As far as we are able to differentiate the use and the benefits, it'd be very useful.

ilayaperumalg:
- Add doc for using Data Flow server/shell in classic vs skipper mode
  - Update getting started doc
 - Fix stream commands to remove the `skipper` prefix

Resolves #1932

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/configuration.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/configuration.adoc
@@ -219,7 +219,7 @@ To use features such as Stream update and rollback, the Data Flow Server delegat
 +
 [source,bash,subs=attributes]
 ----
-$ java -jar -jar spring-cloud-dataflow-server-local-{project-version}.jar --spring.cloud.config.skipper.client.uri=http://192.51.100.1:7577/api
+$ java -jar spring-cloud-dataflow-server-local-{project-version}.jar --spring.cloud.skipper.client.serverUri=http://192.51.100.1:7577/api --spring.cloud.dataflow.features.skipper-enabled=true
 ----
 
 [[configuration-security]]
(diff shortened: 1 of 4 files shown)


tzolov:
1. Perhaps we should document the `stream info` command too? imho it provides a valuable feedback about the stream's status
2. (minor) Section: `27. Stream Lifecycle with Skipper` seems to expose some skipper implementation details that a irrelevant for scdf.  From scdf standpoint we are not interested of templates and values. It is important that our streams along with their apps are bundled as packages and their application and deployment properties can be (re)configured. The Skipper link provided allows to digg for skipper details interested
3. (minor) There is an ambiguity with the usage of the term `application`. In scdf, application usually stands for `stream app`, while in skipper it can mean any (sub)package including the scdf `stream`.

ilayaperumalg:
@tzolov We have a story to address `stream info` documentation and I can bring it here.

ilayaperumalg:
Updated the PR with the documentation for `stream info`. 

I have created https://github.com/spring-cloud/spring-cloud-dataflow/issues/1953 to address other comments

viniciusccarvalho:
Merged via dd5028d5887735d38f28afa35e5f136eb2cc2a4c

PR #1960, opened 2018-01-19
Fix versioned app registry update

### LINKED ISSUE #1955: Re-registering apps results in NonUniqueResultException
As a user, I re-registered OOTB apps from docker -> maven; however, I'm unable to use the apps since it results in `NonUniqueResultException`. This happens in H2 and MySQL, so it is not database specific.

**steps:**
1) dataflow:>app import --uri http://bit.ly/Celsius-SR1-stream-applications-rabbit-docker --force (*works*)
2) dataflow:>app info source:time (*works*)
3) dataflow:>app import --uri http://bit.ly/Celsius-SR1-stream-applications-rabbit-maven --force (*works*)
4) dataflow:>app info source:time (**fails**)
5) dataflow:>stream create foo --definition "time | log" --deploy (**fails**)

**error:**
```
2018-01-18 13:45:44.447 ERROR 8157 --- [io-9393-exec-10] o.s.c.d.s.c.RestControllerAdvice         : Caught exception while handling a request

org.springframework.dao.IncorrectResultSizeDataAccessException: result returns more than one elements; nested exception is javax.persistence.NonUniqueResultException: result returns more than one elements
	at org.springframework.orm.jpa.EntityManagerFactoryUtils.convertJpaAccessExceptionIfPossible(EntityManagerFactoryUtils.java:395) ~[spring-orm-4.3.13.RELEASE.jar!/:4.3.13.RELEASE]
```

### LINKED ISSUE #1959: Can't register apps on Skipper mode
```
dataflow:>app import --uri http://bit.ly/Celsius-SR1-stream-applications-rabbit-maven 
Error when registering applications from http://bit.ly/Celsius-SR1-stream-applications-rabbit-maven: could not execute statement; SQL [n/a]; constraint [null]; nested exception is org.hibernate.exception.ConstraintViolationException: could not execute statement

```

From the server side:
```
2018-01-19 10:45:29.146 ERROR 94890 --- [nio-9393-exec-6] o.h.engine.jdbc.spi.SqlExceptionHelper   : NULL not allowed for column "ID"; SQL statement:
insert into app_registration (id, object_version, default_version, metadata_uri, name, type, uri, version) values (null, ?, ?, ?, ?, ?, ?, ?) [23502-196]
2018-01-19 10:45:29.150 ERROR 94890 --- [nio-9393-exec-6] o.s.c.d.s.c.RestControllerAdvice         : Caught exception while handling a request

org.springframework.dao.DataIntegrityViolationException: could not execute statement; SQL [n/a]; constraint [null]; nested exception is org.hibernate.exception.ConstraintViolationException: could not execute statement
        at org.springframework.orm.jpa.vendor.HibernateJpaDialect.convertHibernateAccessException(HibernateJpaDialect.java:278) ~[spring-orm-4.3.13.RELEASE.jar!/:4.3.13.RELEASE]
        at org.springframework.orm.jpa.vendor.HibernateJpaDialect.translateExceptionIfPossible(HibernateJpaDialect.java:244) ~[spring-orm-4.3.13.RELEASE.jar!/:4.3.13.RELEASE]
        at org.springframework.orm.jpa.AbstractEntityManagerFactoryBean.translateExceptionIfPossible(AbstractEntityManagerFactoryBean.java:503) ~[spring-orm-4.3.13.RELEASE.jar!/:4.3.13.RELEASE]
        at org.springframework.dao.support.ChainedPersistenceExceptionTranslator.translateExceptionIfPossible(ChainedPersistenceExceptionTranslator.java:59)
```

ilayaperumalg:
- When re-register the app using `--force` option, the existing app registrations need to be updated instead of creating new app registration entries
   - This will make sure there will be unique entry for the app by the given name, type and version.
   - Update test

Resolves #1955

Fix schema issue in local server

 - Remove `spring.jpa.generate-ddl:false` and set `spring.jpa.hibernate.ddl-auto:create`

Resolves #1959

### DIFF
--- a/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/service/DefaultAppRegistryService.java
+++ b/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/service/DefaultAppRegistryService.java
@@ -143,15 +143,15 @@ public class DefaultAppRegistryService extends AbstractAppRegistryCommon impleme
 	public AppRegistration save(AppRegistration app) {
 		AppRegistration appRegistration = this.appRegistrationRepository.findAppRegistrationByNameAndTypeAndVersion(
 				app.getName(), app.getType(), app.getVersion());
-		if (getDefaultApp(app.getName(), app.getType()) == null) {
-			app.setDefaultVersion(true);
-		}
 		if (appRegistration != null) {
 			appRegistration.setUri(app.getUri());
 			appRegistration.setMetadataUri(app.getMetadataUri());
 			return this.appRegistrationRepository.save(appRegistration);
 		}
 		else {
+			if (getDefaultApp(app.getName(), app.getType()) == null) {
+				app.setDefaultVersion(true);
+			}
 			return this.appRegistrationRepository.save(app);
 		}
 	}
--- a/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/service/DefaultAppRegistryService.java
+++ b/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/service/DefaultAppRegistryService.java
@@ -141,10 +141,19 @@ public class DefaultAppRegistryService extends AbstractAppRegistryCommon impleme
 
 	@Override
 	public AppRegistration save(AppRegistration app) {
+		AppRegistration appRegistration = this.appRegistrationRepository.findAppRegistrationByNameAndTypeAndVersion(
+				app.getName(), app.getType(), app.getVersion());
 		if (getDefaultApp(app.getName(), app.getType()) == null) {
 			app.setDefaultVersion(true);
 		}
-		return this.appRegistrationRepository.save(app);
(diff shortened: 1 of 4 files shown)


spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/service/DefaultAppRegistryService.java
```diff
@@ -141,10 +141,19 @@ public AppRegistration save(String name, ApplicationType type, String version, U
 
 	@Override
 	public AppRegistration save(AppRegistration app) {
+		AppRegistration appRegistration = this.appRegistrationRepository.findAppRegistrationByNameAndTypeAndVersion(
+				app.getName(), app.getType(), app.getVersion());
 		if (getDefaultApp(app.getName(), app.getType()) == null) {
```
tzolov: I guess setting a default on first registration can go inside the `else` block as it affects only new application registrations

tzolov:
LGTM

viniciusccarvalho:
Merged via accc028

PR #1963, opened 2018-01-19
Mode split and Test coverage for StreamService and StreamDeployer

### LINKED ISSUE #1760: Improve test coverage of StreamService

### LINKED ISSUE #1918: Split the StreamService and StreamDeploymentController into Classic and Skipper components
Similar to the AppRegistration handling have dedicated Controllers and Services for each mode

tzolov:
Resolves #1918

 - Split stream and deployment services into classic and skipper parts.
 - Move skipper related operations from StreamService into UpdatableStreamService. Later extends from StreamService.
   - AppDeployerStream service implements StreamService while SkipperStreamService implements the UpdatableStreamService
 - Move the skipper specific functionality from StreamDeploymentController into a new  UpdatableStreamDeployerController
   - StreamDeploymentController uses the classic-mode StreamService while UpdatableStreamDeploymentController is configured with the UpdatableStreamService
 - The auto configuration bootstraps the StreamDeploymentController for classic mode and UpdatableDeploymentController for skipper mode (both are mutually exclusive)
 - Update tests

Resolves #1760

 - 100% method test coverage for SkipperStreamService and AppDeployerStreamService
 - 100% method test coverage for SkipperStreamDeployer and AppDeployerStreamDeployer
 - Factor out the ResourceUtils tests into its own ResourceUtilsTests


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/UpdatableStreamService.java
```diff
@@ -0,0 +1,71 @@
...
+
+	/**
+	 * Return a manifest info of a release version. For packages with dependencies, the
+	 * manifest includes the contents of those dependencies.
+	 *
+	 * @param releaseName the release name
+	 * @param releaseVersion the release version
+	 * @return the manifest info of a release
+	 */
+	String manifest(String releaseName, int releaseVersion);
```
ilayaperumalg: While `updateStream` and `rollbackStream` methods sound generic enough, the other methods can possibly moved into another interface that extends `UpdatableStreamService`. This way, if we have another mode that can just do `update` and `rollback` those implementations could use `UpdatableStreamService` directly.

tzolov: This is a good point and i completely agree. 
As it stands at the moment though to do this we need also to change the updateStream method signature to replace the skipper specific UpdateRequest with something generic. 
Perhaps we can postpone this fine tuning once we add a new mode to integrate?

ilayaperumalg:
@tzolov Could you rebase this PR please?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/UpdatableStreamDeploymentController.java
```diff
@@ -0,0 +1,105 @@
...
+@RequestMapping("/streams/deployments")
+@ExposesResourceFor(StreamDeploymentResource.class)
+public class UpdatableStreamDeploymentController extends StreamDeploymentController {
+
+	private static final Logger logger = LoggerFactory.getLogger(UpdatableStreamDeploymentController.class);
+
+	private final UpdatableStreamService streamService;
+
+	/**
+	 * Extends the {@code StreamDeploymentController} with support for stream update, rollback, update history.
```
markpollack: This comment is better suited for the class level javadoc.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/UpdatableStreamDeploymentController.java
```diff
@@ -0,0 +1,105 @@
...
+ * @author Christian Tzolov
+ */
+@RestController
+@RequestMapping("/streams/deployments")
+@ExposesResourceFor(StreamDeploymentResource.class)
+public class UpdatableStreamDeploymentController extends StreamDeploymentController {
+
+	private static final Logger logger = LoggerFactory.getLogger(UpdatableStreamDeploymentController.class);
+
+	private final UpdatableStreamService streamService;
```
markpollack: align variable name with class name, that is `private final UpdatableStreamService updatableStreamService;

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/UpdatableStreamService.java
markpollack: It isn't clear if the other methods such as `manifest` and `platformList` would apply to anything else other than when delegating to a skipper backend.  Perhaps a better name for this interface is `SkipperStreamService` and the current implementation - `public class SkipperStreamService extends AbstractStreamService implements UpdatableStreamService` changes to

`public class DefaultSkipperStreamService extends AbstractStreamService implements SkipperStreamService`

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/stream/AppDeployerStreamDeployer.java
```diff
@@ -182,18 +181,18 @@ public DeploymentState calculateStreamState(String streamName) {
...
-	public Map<StreamDefinition, DeploymentState> state(List<StreamDefinition> streamDefinitions) {
+	public Map<StreamDefinition, DeploymentState> streamsStates(List<StreamDefinition> streamDefinitions) {
 		Map<StreamDefinition, List<String>> deploymentIdsPerStream = streamDefinitions.stream()
 				.collect(Collectors.toMap(Function.identity(),
 						sd -> sd.getAppDefinitions().stream().map(
 								sad -> deploymentIdRepository.findOne(DeploymentKey.forStreamAppDefinition(sad)))
 								.collect(Collectors.toList())));
 
-		// Map from app deployment id to state
+		// Map from app deployment id to streamsStates
```
markpollack: to `DeploymentState` would be better - though this change is a result of a search/replace operation....as well as a few other spots.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/UpdatableStreamService.java
tzolov: @markpollack , sure this will make the naming more consistent. Will rename the interfaces and classes and will update the PR.   
Btw what do you think about this package structure ?
- org.springframework.cloud.dataflow.server.service.*
- org.springframework.cloud.dataflow.server.service.impl.*
- org.springframework.cloud.dataflow.server.stream.*

Why the `StreamService` is in the `...server.service` while the `StreamDeployer` is inside the `...server.stream` packages?

markpollack: The structure isn't bad from the DSM/structure101 perspective, not really and tangles, the `server.stream` package sits at the bottom of the pile.  The name of the package would be better off as `server.deployer` instead of `server.stream` but we should handle this as part of a general tech-debt story and take a look at it inside structure101 to get a complete picture.  I'll create an issue if we don't already have one on this topic.

markpollack: Created https://github.com/spring-cloud/spring-cloud-dataflow/issues/2054

markpollack:
made some class name changes.  merged as 7b8a0b0c884c71d86430b8f609eb1adfe05228ca

PR #1969, opened 2018-01-22
Change from tocify to tocbot

Buzzardo:
Spring's other projects are headed in the direction of using Tocbot, which improves on tocify by needing fewer libraries to work. (In fact, it has no dependencies.) It also works with our other features, such as content toggling.


ghillert:
@Buzzardo Thanks for fixing the Table of Contents header. Looks much better now. Merging. @sabbyanandan Maybe we should create issue for the other issues as you see fit.

ghillert:
Merged with commit 20c324f3749fedbf27282abd05dbc1a3bf3dc3e7

PR #1971, opened 2018-01-23
Fix info.app.name in implementation name

### LINKED ISSUE #1950: About controller is not setting the server-name override
As a user, I'm hitting the `/about` endpoint to retrieve the customized (SCDF) server's name (via: `--info.app.name=foo` or similar); however, I'm getting the project "artifactId" instead.

**Observation:**
The name override has been removed in https://github.com/spring-cloud/spring-cloud-dataflow/commit/06adc4697fdcdaa6a7f89c960a714ab57ecaf263#diff-330c29d6d867cdeefd8e20e03eea0934L100 commit; we have also refactored the domain object from `FeatureInfo` to `RuntimeEnvironment`; probably something missed-out in the transition.

1) Boot's Info endpoint (http://localhost:9393/management/info) point to:

image

2) The `/about` endpoint shows:

image

3) Dashboard shows:

image

ilayaperumalg:
- Add info.app.name/version properties in dataflow-server defaults

Resolves #1950

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
+++ b/spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
@@ -33,8 +33,8 @@ spring:
             name: Spring Cloud Data Flow Core
             version: "@project.version@"
           spring-cloud-dataflow-implementation:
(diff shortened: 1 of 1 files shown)


ghillert:
Merged with commit 177ffeff50dea059d1d5db8901678763d59903fd

PR #1988, opened 2018-01-24
Fix app info not showing resource metadata

### LINKED ISSUE #1983: app info is not showing resource metadata
```dataflow:>app info source:time
Information about source application 'time':
Version: '1.2.0.RELEASE':
Default application version: 'true':
Resource URI: maven://org.springframework.cloud.stream.app:time-source-rabbit:1.2.0.RELEASE
╔══════════════════════════════╤══════════════════════════════╤══════════════════════════════╤══════════════════════════════╗
║         Option Name          │         Description          │           Default            │             Type             ║
╚══════════════════════════════╧══════════════════════════════╧══════════════════════════════╧══════════════════════════════╝```

### PR DESCRIPTION
**Pull request by sobychacko**
- Fix the regression issue where `app info` is not showing the metadata information
 - Add missing copyright header to `AbstractAppRegistryCommon`
 - Add tests

Resolves #1983

### DIFF
--- a/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/AbstractAppRegistryCommon.java
+++ b/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/AbstractAppRegistryCommon.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2017-2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package org.springframework.cloud.dataflow.registry;
 
 import java.io.IOException;
@@ -7,6 +23,8 @@ import java.util.AbstractMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -29,6 +47,7 @@ import org.springframework.util.StringUtils;
  * 
  * @author Christian Tzolov
  * @author Ilayaperumal Gopinathan
+ * @author Soby Chacko
  */
 public abstract class AbstractAppRegistryCommon implements AppRegistryCommon {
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/AbstractAppRegistryCommon.java [hunk 1]
```diff
@@ -56,8 +77,14 @@ public Resource getAppResource(AppRegistration appRegistration) {
 
 	@Override
 	public Resource getAppMetadataResource(AppRegistration appRegistration) {
-		return appRegistration.getMetadataUri() != null ? this.resourceLoader.getResource(
-				appRegistration.getMetadataUri().toString()) : null;
+		if (appRegistration.getMetadataUri() != null) {
+			return this.resourceLoader.getResource(appRegistration.getMetadataUri().toString());
+		}
+		else {
+			appRegistrationResourceCache.putIfAbsent(appRegistration.getUri(),
```
ilayaperumalg: I think we want to return `null` in this case where the metadata URI is not set for the app registration.

For instance, from what is reported in #1983 the metadata should have been registered as part of [this](https://github.com/spring-cloud-stream-app-starters/app-starters-release/blob/master/spring-cloud-stream-app-descriptor/src/main/resources/META-INF/rabbit-apps-maven.properties#L90)

Maybe in this case of #1983 the app was registered with `app register` and the metadata wasn't registered at all?

On a related note, this fix would cause the actual app resource to be downloaded(via `getAppResource`) which we want to avoid doing at the server level.

markpollack: The original issue didn't show the registration command.  It did not explicitly list `--metadata-uri`.

This PR shows the same behavior as 1.2.3

```dataflow:>app register --name log --type sink --uri maven://org.springframework.cloud.stream.app:log-sink-rabbit:1.2.0.RELEASE --force
Successfully registered application 'sink:log'
dataflow:>app info sink:log
Information about sink application 'log':
Version: '1.2.0.RELEASE':
Default application version: 'true':
Resource URI: maven://org.springframework.cloud.stream.app:log-sink-rabbit:1.2.0.RELEASE
╔══════════════════════════════╤══════════════════════════════╤══════════════════════════════╤══════════════════════════════╗
║         Option Name          │         Description          │           Default            │             Type             ║
╠══════════════════════════════╪══════════════════════════════╪══════════════════════════════╪══════════════════════════════╣
║log.name                      │The name of the logger to use.│<none>                        │java.lang.String              ║
║log.level                     │The level at which to log     │<none>                        │org.springframework.integratio║
║                              │messages.                     │                              │n.handler.LoggingHandler$Level║
║log.expression                │A SpEL expression (against the│payload                       │java.lang.String              ║
║                              │incoming message) to evaluate │                              │                              ║
║                              │as the logged message.        │                              │                              ║
╚══════════════════════════════╧══════════════════════════════╧══════════════════════════════╧══════════════════════════════╝
```

ilayaperumalg:
We are OK with the downloading of app resource when `app info` is invoked for the app.

LGTM, merging.

sabbyanandan:
The PR is merged in https://github.com/spring-cloud/spring-cloud-dataflow/commit/c58ff63e076462ed105077d869edc44bb8a1257d.

PR #1991, opened 2018-01-25
Skipper reverse engineered  DSL can skip implicit data flow properties

### LINKED ISSUE #1982: Skipper reverse engineered stream definition DSL can skip implicit dataflow properties
When the stream definition DSL is reverse engineered by parsing through the Skipper release manifest, we can filter out the implicit Data Flow properties such as the SCSt binding properties, metrics properties and some of the data flow properties from stream application properties list.

markpollack:
* Remove some duplication of property constants

Fixes #1982


tzolov:
Rebased and merged at #062239a4

PR #1996, opened 2018-01-26
Fix partial aggregate deployment state

### LINKED ISSUE #1974: When one app fails to deploy in a stream, the stream status is "all apps failed to deploy"
This is an incorrect message.
This is what was observed when testing with skipper and investigating the issue with escaping quote characters.  The transformer app failed to start, the runtime apps command was accurate reporting what was running and what had failed.

ilayaperumalg:
- If the stream has at least one of the apps successfully deployed but at least one of the apps has failed, then the aggregate state needs to be set to `partital` instead of `failed`
  - This state will show the appropriate status description as:
`In the case of multiple apps, some have successfully deployed, while others have not`

Resolves #1974


tzolov:
LGTM

PR #2006, opened 2018-01-27
AboutController returns correct repo for version RC1

### LINKED ISSUE #2005: Repository selection for about URL's is incorrect
Specifically the if the version of the application is RC it should return a repo of libs-milestone but instead it uses a repo of libs-release.

cppwfs:
Originally the repo was libs-release for RC1 for any About URL that used the {repository} tag.
This has been updated to set the repo to libs-milestone.
resolves #2005

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AboutController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AboutController.java
@@ -271,6 +271,8 @@ public class AboutController {
 		final String REPO_SNAPSHOT_ROOT = "https://repo.spring.io/libs-snapshot";
 		final String REPO_MILESTONE_ROOT = "https://repo.spring.io/libs-milestone";
 		final String MAVEN_ROOT = "https://repo1.maven.org/maven2";
+		final String REPO_RELEASE_ROOT = "https://repo.spring.io/libs-release";
+
 		String result = MAVEN_ROOT;
 		if (version.endsWith(BUILD_SNAPSHOT_CRITERIA)) {
 			result = REPO_SNAPSHOT_ROOT;
@@ -280,6 +282,9 @@ public class AboutController {
 		}
 		else if (version.contains(".RC")) {
 			result = REPO_MILESTONE_ROOT;
+		}
+		else if (version.contains(".RELEASE")) {
+			result = REPO_RELEASE_ROOT;
 		}
 		return result;
 	}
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AboutController.java
```diff
@@ -270,7 +270,6 @@ private String repoSelector(String version) {
 		final String BUILD_SNAPSHOT_CRITERIA = "BUILD-SNAPSHOT";
 		final String REPO_SNAPSHOT_ROOT = "https://repo.spring.io/libs-snapshot";
 		final String REPO_MILESTONE_ROOT = "https://repo.spring.io/libs-milestone";
-		final String REPO_RELEASE_ROOT = "https://repo.spring.io/libs-release";
```
markpollack: We should keep the `REPO_RELEASE_ROOT` and add the logic in the selector so that if the version end in `.RELEASE`, use the release repository.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/AboutControllerTests.java
```diff
@@ -128,6 +128,106 @@ public void testChecksumDisabled() throws Exception {
...
+					.andExpect(jsonPath("$.versionInfo.shell.checksumSha1").doesNotExist())
+					.andExpect(jsonPath("$.versionInfo.shell.checksumSha256").doesNotExist());
+		}
+	}
+
+	@RunWith(SpringRunner.class)
+	@SpringBootTest(classes = TestDependencies.class)
+	@DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_EACH_TEST_METHOD)
+	@TestPropertySource(properties = {
+			"spring.cloud.dataflow.version-info.dependencies.spring-cloud-dataflow-shell.version=1.2.3.GA1",
```
markpollack: We don't use the `GA` suffix in releases, so perhaps this test is asserting the wrong "facts" :)  ?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AboutController.java
cppwfs: Updated.  Thanks for the review!

markpollack:
squashed and rebased.  Made minor changes - 6fc2ae8e5b622a7f38d07af7bb25fba9d1c59e03

PR #2016, opened 2018-01-31
removed (MAX) from varchar on deployment_props on stream deployments

### LINKED ISSUE #2015: Postgresql streams.sql and migration-postgresql.sql are invalid
**VARCHAR(MAX)** is invalid postgresql syntax, should be **VARCHAR**
Please fix both in `migration` and `schemas` scripts
````
CREATE TABLE STREAM_DEPLOYMENTS  (
	STREAM_NAME VARCHAR(255) NOT NULL PRIMARY KEY,
	DEPLOYMENT_PROPS VARCHAR(MAX)
);
````

cppwfs:
resolves #2015


jvalkeal:
lgtm. not sure why we used `max` as docs say
> If you desire to store long strings with no specific upper limit, use text or character varying without a length specifier, rather than making up an arbitrary length limit.

cppwfs:
I'm not sure why they used MAX (unless it was a C/P hiccup) .  It certainly is illegal.  And what @jvalkeal said above is what I read as well.

markpollack:
It should be of type `text` since deployment props can get quite long.

PR #2024, opened 2018-01-31
Upgrade to Hibernate 5.2.12, use @Lob for URIs in AppRegistration

### LINKED ISSUE #2018: Increase size of string field used to store URIs in AppRegistration
The current value of 255 is too small.

### LINKED ISSUE #2020: Upgrade to Hibernate 5.2.12

markpollack:
Fixes #2020
Fixes #2018

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -143,6 +143,23 @@
 				<type>pom</type>
 				<scope>import</scope>
 			</dependency>
+			<dependency>
+				<groupId>org.hibernate</groupId>
+				<artifactId>hibernate-core</artifactId>
+				<version>5.2.12.Final</version>
+			</dependency>
+			<dependency>
+				<groupId>org.hibernate</groupId>
+				<artifactId>hibernate-entitymanager</artifactId>
+				<version>5.2.12.Final</version>
+			</dependency>
+			<dependency>
+				<groupId>org.springframework.data</groupId>
+				<artifactId>spring-data-releasetrain</artifactId>
+				<version>Ingalls-SR10</version>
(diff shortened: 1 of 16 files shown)


pom.xml
```diff
@@ -143,6 +143,23 @@
 				<type>pom</type>
 				<scope>import</scope>
 			</dependency>
+			<dependency>
+				<groupId>org.hibernate</groupId>
+				<artifactId>hibernate-core</artifactId>
+				<version>5.2.12.Final</version>
```
markpollack: It maybe possible to extract this and use `<hibernate.version>` but since I was having so much trouble overriding, I hardcoded it everywhere and just left it when it worked.

spring-cloud-dataflow-server-core/src/main/resources/schemas/db2/jpa.sql
```diff
@@ -1,13 +1,13 @@
     create sequence hibernate_sequence start with 1 increment by 1;
 
     create table APP_REGISTRATION (
-        id bigint not null,
-        object_version bigint,
-        default_version smallint,
-        metadata_uri varchar(255),
+       id bigint not null,
+        object_Version bigint,
```
viniciusccarvalho: Are we changing the case here for any reason? I thought that the naming strategy would convert camelCase to underscore notation on db. mysql could barf on this much like it did with table names.

viniciusccarvalho:
Merged via 144c4f4465c5b85904c88b4c4b59c7f64e0558f5

PR #2025, opened 2018-01-31
Increase size of deployment properties varchar for mysql

### LINKED ISSUE #2019: Increase size of deployment properties varchar for mysql
```
CREATE TABLE STREAM_DEPLOYMENTS  (
	STREAM_NAME VARCHAR(255) NOT NULL PRIMARY KEY,
	DEPLOYMENT_PROPS VARCHAR(1000)
) ENGINE=InnoDB;;
```

`Values in VARCHAR columns are variable-length strings. The length can be specified as a value from 0 to 65,535. The effective maximum length of a VARCHAR is subject to the maximum row size (65,535 bytes, which is shared among all columns) and the character set used` from https://dev.mysql.com/doc/refman/5.7/en/char.html

set to 16kb

markpollack:
Fixes #2019


markpollack:
merged as 6fa4f903813cfba2ef1bb1d12ec8af176fcac3db

PR #2043, opened 2018-02-12
Set INDEX_PROPERTY_KEY only when upstream app is partitioned

### LINKED ISSUE #1876: Only set indexed property when deploying a partitioned stream
By setting this even for regular stream deployments, as a side effect, we are spawning StatefulSet pods in K8S, which is unnecessary. The requirement is to make sure it is applied only when a partitioned streaming characteristics are implied.

markpollack:
Also backport to 1.3.x branch

Fixes #1876


tzolov:
@markpollack the `SkipperStreamServiceIntegrationTests#testInstallVersionOverride` is failing. 
Perhaps we should drop the `spring.cloud.deployer.indexed: 'true'` line from the `SkipperStreamServiceIntegrationTests-install.yml` test feature?

tzolov:
Rebased and merged at d75f841e

tzolov:
Rebased and merged at d75f841

markpollack:
also needed to apply to 1.3.x branch

PR #2049, opened 2018-02-17
Prevent un-registration of Apps used in streams

### LINKED ISSUE #2013: Prevent un-registration of an app-version if it is already in use
As a user, I'd like for the un-registration routine to applies on unused App-versions. If the app-version is in use, the system should prevent me from deleting it.

**Notes:**
```
1) List versions:

dataflow:>app list --id processor:xfmr
╔══════╤══════════════════════════════════════╤════╤════╗
║source│              processor               │sink│task║
╠══════╪══════════════════════════════════════╪════╪════╣
║      │> xfmr-0.0.1.BUILD-20180129.011003-8 <│    │    ║
║      │xfmr-0.0.1.BUILD-20180129.015528-12   │    │    ║
║      │xfmr-0.0.1.BUILD-20180129.021451-13   │    │    ║
║      │xfmr-0.0.1.BUILD-20180129.200519-1    │    │    ║
╚══════╧══════════════════════════════════════╧════╧════╝

2) Create a stream with default version:

dataflow:>stream create foo --definition "time | xfmr | log"
Created new stream 'foo'

3) Unregister the default version:

dataflow:>app unregister --name xfmr --type processor
Successfully unregistered application 'xfmr' with type processor

4) List versions:

dataflow:>app list --id processor:xfmr
╔══════╤═══════════════════════════════════╤════╤════╗
║source│             processor             │sink│task║
╠══════╪═══════════════════════════════════╪════╪════╣
║      │xfmr-0.0.1.BUILD-20180129.015528-12│    │    ║
║      │xfmr-0.0.1.BUILD-20180129.021451-13│    │    ║
║      │xfmr-0.0.1.BUILD-20180129.200519-1 │    │    ║
╚══════╧═══════════════════════════════════╧════╧════╝
```

**Acceptance:**
- No-3 from above should not be allowed since it is already in use in the `foo` stream
- Return a user-friendly message (maybe: `The app you're trying to unregister is already in use on "Foo" stream`?) 
- Shell/UI should use the returned message to notify it to the user

tzolov:
Resolves #2013

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowControllerAutoConfiguration.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowControllerAutoConfiguration.java
@@ -16,11 +16,11 @@
 
 package org.springframework.cloud.dataflow.server.config;
 
-import javax.sql.DataSource;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.ForkJoinPool;
+import javax.sql.DataSource;
 
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -223,9 +223,11 @@ public class DataFlowControllerAutoConfiguration {
 		}
 
 		@Bean
-		public VersionedAppRegistryController appRegistryController2(AppRegistryService appRegistry,
-				ApplicationConfigurationMetadataResolver metadataResolver, ForkJoinPool appRegistryFJPFB, MavenProperties mavenProperties) {
-			return new VersionedAppRegistryController(appRegistry, metadataResolver, appRegistryFJPFB, mavenProperties);
+		public VersionedAppRegistryController appRegistryController2(StreamDefinitionRepository streamDefinitionRepository,
+				AppRegistryService appRegistry, ApplicationConfigurationMetadataResolver metadataResolver,
+				ForkJoinPool appRegistryFJPFB, MavenProperties mavenProperties) {
+			return new VersionedAppRegistryController(streamDefinitionRepository, appRegistry,
+					metadataResolver, appRegistryFJPFB, mavenProperties);
 		}
 	}
 
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppUsedInStreamException.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppUsedInStreamException.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright 2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.cloud.dataflow.server.controller;
+
+/**
+ * @author Christian Tzolov
+ */
+public class AppUsedInStreamException extends RuntimeException {
+
+	private static final long serialVersionUID = 1L;
+
+	public AppUsedInStreamException(String appId) {
+		super(String.format("The '%s' application could not be found.", appId));
+	}
+}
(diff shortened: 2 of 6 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowControllerAutoConfiguration.java
```diff
@@ -223,9 +223,11 @@ public AppRegistryService appRegistryService(AppRegistrationRepository appRegist
 		}
 
 		@Bean
-		public VersionedAppRegistryController appRegistryController2(AppRegistryService appRegistry,
-				ApplicationConfigurationMetadataResolver metadataResolver, ForkJoinPool appRegistryFJPFB, MavenProperties mavenProperties) {
-			return new VersionedAppRegistryController(appRegistry, metadataResolver, appRegistryFJPFB, mavenProperties);
+		public VersionedAppRegistryController appRegistryController2(StreamDefinitionRepository streamDefinitionRepository,
```
markpollack: method name should be `versionedAppRegistryController`

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppUsedInStreamException.java
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.cloud.dataflow.server.controller;
+
+/**
+ * @author Christian Tzolov
+ */
+public class AppUsedInStreamException extends RuntimeException {
```
markpollack: The name of this exception doesn't quite correspond to the message.  `ApplicationNotFoundException` would be better, though it feels like we have too many fine grained exceptions, maybe look to reuse another one since nothing is catching this specific exception.

markpollack: Turns out we already have a `NoSuchAppException` that is identical to this one, will use that instead.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/VersionedAppRegistryController.java
```diff
@@ -230,9 +239,41 @@ public void makeDefault(@PathVariable("type") ApplicationType type, @PathVariabl
 	@ResponseStatus(HttpStatus.OK)
 	public void unregister(@PathVariable("type") ApplicationType type, @PathVariable("name") String name,
 			@PathVariable("version") String version) {
+
+		AppRegistration defaultApp = appRegistryService.getDefaultApp(name, type);
+		if (defaultApp != null && defaultApp.getVersion().equals(version)) {
+			String streamWithApp = findStreamContainingAppOf(type, name);
+			if (streamWithApp != null) {
+				throw new AppUsedInStreamException(String.format("The app [%s:%s:%s] you're trying to unregister is " +
```
markpollack: This will result in a confusing error message as 

```
	public AppUsedInStreamException(String appId) {
		super(String.format("The '%s' application could not be found.", appId));
	}````

File: spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/VersionedAppRegistryController.java [hunk 4]
```diff
@@ -230,9 +239,41 @@ public void makeDefault(@PathVariable("type") ApplicationType type, @PathVariabl
 	@ResponseStatus(HttpStatus.OK)
 	public void unregister(@PathVariable("type") ApplicationType type, @PathVariable("name") String name,
 			@PathVariable("version") String version) {
+
+		AppRegistration defaultApp = appRegistryService.getDefaultApp(name, type);
+		if (defaultApp != null && defaultApp.getVersion().equals(version)) {
```
markpollack: should also include a test if the app is not of type `task`

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/VersionedAppRegistryController.java
```diff
@@ -230,9 +239,41 @@ public void makeDefault(@PathVariable("type") ApplicationType type, @PathVariabl
...
+						"already in use on %s stream", name, type, version, streamWithApp));
+			}
+		}
+
 		appRegistryService.delete(name, type, version);
 	}
 
+	private String findStreamContainingAppOf(ApplicationType type, String name) {
+		Iterable<StreamDefinition> streams = streamDefinitionRepository.findAll();
+		for (StreamDefinition stream : streams) {
```
markpollack: I do not think this is the correct approach, checking stream definitions for the `type` and `name`.  Consider this case in simplified syntax

```
app reg time 1.0 default
app reg time 2.0
stream create ticktock --def "time | log"
stream deploy ticktock --properties version.time=2.0
app unreg time 2.0
```
The last line should fail since we should not be unregistering apps that are currently deployed and running.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/VersionedAppRegistryController.java
```diff
@@ -230,9 +239,41 @@ public void makeDefault(@PathVariable("type") ApplicationType type, @PathVariabl
...
+			for (StreamAppDefinition streamAppDefinition : stream.getAppDefinitions()) {
+				final String streamAppName = streamAppDefinition.getRegisteredAppName();
+				final ApplicationType streamAppType;
+				try {
+					streamAppType = DataFlowServerUtil.determineApplicationType(streamAppDefinition);
+				}
+				catch (CannotDetermineApplicationTypeException e) {
+					//skip
+					continue;
```
markpollack: always log exception, with a guard to check for excluding the app type task, it is hard to imagine this exception getting thrown.

markpollack:
Reworked and issued a new PR https://github.com/spring-cloud/spring-cloud-dataflow/pull/2081

PR #2050, opened 2018-02-17
Add version warning on app un-registration if un-registering default app/

### LINKED ISSUE #1871: In Skipper mode, when a default AppRegistration is deleted pick a new default
Currently if the default app-registration version for given <name, type> pair is deleted, no new default version will be set even if there are other registered versions for the same <name, type>.  
Currently an explicit `app default set` would be required to select the new default version.
We can implement a strategy to elect automatically the new default version instead.

tzolov:
Resolves #1871


markpollack:
Fixed formatting.

Merged as 590e7cfbca76a16ba9feaf0c6b53255a32e3c783

Issue #2040, opened 2018-02-12
Add docker compose-up for Local-server


sabbyanandan:
As a user, I'd like to get started with the latest release of Local-server quickly.

**Acceptance:**
A single `docker compose-up` command from the cloned project could automate the following steps.
- Pull the latest GA release of Local-server
- Pull the latest GA release of Skipper-server
- Pull a Kafka image (use the same image used in k8s)
- Pull a Redis image (use the same image used in k8s)
- Pull a MySQL image (use the same image used in k8s)
- Run the Local and Skipper servers
- Register the latest GA release of Stream (kafka) / Task apps
- Once everything is up and running, run a `ticktock` stream from the Dashboard to verify the deployment
- Document how to access the logs of `log-sink` application
- Update Quick Start section in the project site
- Add a note re: CPU/Memory requirements for the Docker VM

In summary, we would want to automate the [Quick-Start](http://cloud.spring.io/spring-cloud-dataflow/#quick-start) experience in a single command. This also aligns with the helm-chart for K8S and the upcoming SCDF-tile for PCF.

chrisjs:
per discussion, removing skipper and mysql for now

chrisjs:
* https://github.com/spring-cloud/spring-cloud-dataflow/pull/2052 - provides capability to build and push local server to docker hub. we probably need to get CI building/pushing the images as well

* https://github.com/spring-cloud/spring-cloud-dataflow/pull/2053/files - has the compose definition for the issue requirements. currently "latest" is being used for local server and built from master for now. we can pin this to an existing GA if you would like tho

when the above get merged i think all that would be left is updating the docs (and CI) unless you want to make changes to the workflow i showed you earlier?

PR #2052, opened 2018-02-20
Add docker image support to local server

chrisjs:
Resolves spring-cloud/spring-cloud-dataflow#2040

Provides ability to build/push dockerized local server image for #2040


chrisjs:
merged

PR #2062, opened 2018-02-23
App bulk registration controller

### LINKED ISSUE #2031: Bulk app un-registration hangs
As a user, I'm trying to un-register 10 apps from the Dashboard, but the operation never successfully completes. While this operation is in progress, any query-operation on the shell also hangs.

**Steps:**
- Select 10 apps from using the checkboxes
- Click `Unregister 10 selected apps`
- The operation hangs and it never returns
- While it is in progress, invoke `app list` on shell to observe the behavior

viniciusccarvalho:
Fixes #2031 
This PR intends to reduce the number of calls from the client to the server when doing bulk unregister. Internally we will still (for now) make several calls to the Resource Loaders and Repositories.

There's a pending PR to validate that if an app is being used by a Stream we should throw an exception. The code here relies on `unregister(...)` to throw such exception if we are to return a list of failed deletions to the client. For now, all deletes would complete succesfully

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppRegistryController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppRegistryController.java
@@ -39,6 +39,9 @@ import org.springframework.cloud.dataflow.registry.domain.AppRegistration;
 import org.springframework.cloud.dataflow.registry.support.NoSuchAppRegistrationException;
 import org.springframework.cloud.dataflow.rest.resource.AppRegistrationResource;
 import org.springframework.cloud.dataflow.rest.resource.DetailedAppRegistrationResource;
+import org.springframework.cloud.dataflow.server.controller.support.BulkAppRegisterRequest;
+import org.springframework.cloud.dataflow.server.controller.support.BulkAppRegisterResponse;
+import org.springframework.cloud.dataflow.server.controller.support.UnregisterAppResponse;
 import org.springframework.context.ResourceLoaderAware;
 import org.springframework.core.io.ByteArrayResource;
 import org.springframework.core.io.DefaultResourceLoader;
@@ -55,6 +58,7 @@ import org.springframework.http.HttpStatus;
 import org.springframework.util.CollectionUtils;
 import org.springframework.util.StringUtils;
 import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.RequestParam;
@@ -71,6 +75,7 @@ import org.springframework.web.bind.annotation.RestController;
  * @author Gary Russell
  * @author Patrick Peralta
  * @author Thomas Risberg
+ * @author Vinicius Carvalho
  */
 @RestController
 @RequestMapping("/apps")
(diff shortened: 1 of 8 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/VersionedAppRegistryController.java
```diff
@@ -284,6 +289,26 @@ else if (!CollectionUtils.isEmpty(apps)) {
...
 	}
 
+	/**
+	 * Unregisters the list of applications provided via request parameter.
+	 * @param request - Contains a list of applications to be removed
+	 * @return A response with the status of each requested application with either success or error
+	 */
+	@RequestMapping(method = RequestMethod.POST, value = "/bulk/unregister")
+	@ResponseStatus(HttpStatus.OK)
+	public BulkAppRegisterResponse bulkUnregister(@RequestBody BulkAppRegisterRequest request) {
```
markpollack: Looks like this method could go into a common controller and/or service layer method that is shared between skipper and classic modes to reduce code duplication.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/AppRegistryControllerTests.java
```diff
@@ -280,4 +285,21 @@ public void testPagination() throws Exception {
 				.andExpect(jsonPath("page.number", is(0)));
 	}
 
+	@Test
+	public void testBulkUnregister() throws Exception {
```
markpollack: similar comment wrt to code duplication

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/support/RegisteredApp.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+package org.springframework.cloud.dataflow.server.controller.support;
+
+import org.springframework.cloud.dataflow.core.ApplicationType;
+
+/**
+ * @author Vinicius Carvalho
+ *
+ */
+public class RegisteredApp {
```
markpollack: We have `AppRegistrationResource` which sits in the `spring-cloud-dataflow-rest-resource` package.  

Need to consider how this new bulk unregister operation will be included in `AppRegistryOperations`.   These various support classes should move out of the `server.controller.support` package and into the `rest-client`.

markpollack:
Need additional stories for the UI and shell to invoke this endpoint.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppRegistryController.java
```diff
@@ -252,6 +257,26 @@ else if (!CollectionUtils.isEmpty(apps)) {
 				assembler);
 	}
 
+	/**
+	 * Unregisters the list of applications provided via request parameter.
+	 * @param request - Contains a list of applications to be removed
+	 * @return A response with the status of each requested application with either success or error
+	 */
+	@RequestMapping(method = RequestMethod.POST, value = "/bulk/unregister")
+	@ResponseStatus(HttpStatus.OK)
```
markpollack: maybe only 200 if all the unregister requests are successful and 207 otherwise?
  https://restpatterns.mindtouch.us/HTTP_Status_Codes/207_-_Multi-Status

sabbyanandan:
We have decided to push this improvement down to 1.5 and in the meantime, for 1.4, we will just add transaction management support (via spring-cloud/spring-cloud-dataflow#2086).

PR #2069, opened 2018-02-28
Fix yaml file handling with tasks

### LINKED ISSUE #1878: Tasks throws exception when using a .yml file as a properties file
For example:
`dataflow:>task launch mytask --arguments "--server.port=8080 --foo=bar" --propertiesFile "/Users/glennrenfro/tmp/myprops.yml"`
throws the following Illegal argument exception: `Command failed java.lang.IllegalArgumentException: Only deployment property keys starting with 'app.' or 'deployer.' allowed, got 'app'
d`
with the file content of 
```
app:
  mytask:
    AAA:
      keys: a,b,c
```

jvalkeal:
- Extract common deployment properties handling
  from streams and use it in a same way with tasks.
- Move commons-io and commons-lang dep versions to parent
  as need commons-io in rest-resource.
- Generic polish for files touched with this commit, ie
  remove dead code.
- Now shared code is in
  DeploymentPropertiesUtils.parseDeploymentProperties()
- We now get proper error if yaml is not expected like:
  Command failed java.lang.IllegalArgumentException:
  Only deployment property keys starting with 'app.' or 'deployer.' allowed, got 'appx.footask.AAA.keys'
- Fixes #1878

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -45,6 +45,8 @@
 		<spring-batch-admin-manager.version>1.3.1.RELEASE</spring-batch-admin-manager.version>
 		<spring-shell.version>1.2.0.RELEASE</spring-shell.version>
 		<spring-session.version>1.2.2.RELEASE</spring-session.version>
+		<commons-io.version>2.4</commons-io.version>
+		<commons-lang.version>2.4</commons-lang.version>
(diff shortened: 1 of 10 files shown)


cppwfs:
Aside from some small polishing that I will do on merge LGTM

cppwfs:
Rebased, polished, squashed, merged

PR #2070, opened 2018-02-28
Implement global app props for tasks

### LINKED ISSUE #1531: Common Application Properties for tasks does not work
In this case when attempting to set the dataflowServerUri.
Works using properties i.e. task launch foo --properties "app.composed-task-runner.dataflowServerUri=http://bar" 
Works using properties.
But setting it with java -jar spring-cloud-dataflow-server-local/target/spring-cloud-dataflow-server-local-1.2.2.BUILD-SNAPSHOT.jar --spring.cloud.dataflow.applicationProperties.task.dataflowServerUri=http://foo.bar does not work.

jvalkeal:
- It turned out that CommonApplicationProperties only
  have key for streams while documentation gives example
  for tasks as well.
- Add space for tasks and modify all needed tests.
- DefaultTaskServiceTests have modified test showing global
  pring.cloud.dataflow.applicationProperties.task.app.composed-task-runner.globalkey=globalvalue
  added to context.
- DefaultTaskService.extractAppProperties() now gets entries from global space as well
  and same filtering per task app name which is same logic as how stream apps use.
- i.e. --spring.cloud.dataflow.applicationProperties.task.app.timestamp.format=yyyy
  would work with timestamp task.
- Fixes #1531

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/apps/CommonApplicationProperties.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/apps/CommonApplicationProperties.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2017 the original author or authors.
+ * Copyright 2016-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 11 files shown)


cppwfs:
LGTM  I'll polish on merge.

cppwfs:
Rebased, Polished, Squashed, Merged

PR #2079, opened 2018-03-01
Scdf 1410 a

cppwfs:


cppwfs:
re-open with correct commit comment

PR #2081, opened 2018-03-02
Prevent un-registration of Apps used in deployed streams

### LINKED ISSUE #2013: Prevent un-registration of an app-version if it is already in use

markpollack:
Resolves #2013

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppUsedInStreamException.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppUsedInStreamException.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright 2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.cloud.dataflow.server.controller;
+
+/**
+ * @author Christian Tzolov
+ */
+public class AppUsedInStreamException extends RuntimeException {
+
+	private static final long serialVersionUID = 1L;
+
+	public AppUsedInStreamException(String appId) {
+		super(String.format("The '%s' application could not be found.", appId));
+	}
+}
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/SkipperAppRegistryController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/SkipperAppRegistryController.java
@@ -23,7 +23,6 @@ import java.net.URISyntaxException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
@@ -71,6 +70,8 @@ import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.ResponseStatus;
 import org.springframework.web.bind.annotation.RestController;
+
+import static org.springframework.cloud.dataflow.rest.SkipperStream.SKIPPER_SPEC_VERSION;
 
 /**
  * Handles all {@link DefaultAppRegistryService} related interactions.
@@ -250,23 +251,31 @@ public class SkipperAppRegistryController {
 	public void unregister(@PathVariable("type") ApplicationType type, @PathVariable("name") String name,
 			@PathVariable("version") String version) {
 
-		// AppRegistration defaultApp = appRegistryService.getDefaultApp(name, type);
-		if (type != ApplicationType.task) { // && defaultApp != null && defaultApp.getVersion().equals(version)) {
+		if (type != ApplicationType.task) {
 			String streamWithApp = findStreamContainingAppOf(type, name, version);
 			if (streamWithApp != null) {
-				throw new AppUsedInStreamException(String.format("The app [%s:%s:%s] you're trying to unregister is " +
-						"already in use on %s stream", name, type, version, streamWithApp));
+				throw new UnregisterAppException(String.format("The app [%s:%s:%s] you're trying to unregister is " +
+						"currently used in stream '%s'.", name, type, version, streamWithApp));
 			}
 		}
 
 		appRegistryService.delete(name, type, version);
 	}
 
+	/**
+	 * Given the application type, name, and version, determine if it is being used in a deployed stream definition.
+	 *
+	 * @param appType the application type
+	 * @param appName the application name
+	 * @param appVersion application version
+	 * @return the name of the deployed stream where the app is being used.  If the app is not deployed in a stream,
+	 * return {@code null}.
+	 */
 	private String findStreamContainingAppOf(ApplicationType appType, String appName, String appVersion) {
-		Iterable<StreamDefinition> streams = streamDefinitionRepository.findAll();
-		for (StreamDefinition stream : streams) {
-			StreamDeployment streamDeployment = this.streamService.info(stream.getName());
-			for (StreamAppDefinition streamAppDefinition : stream.getAppDefinitions()) {
+		Iterable<StreamDefinition> streamDefinitions = streamDefinitionRepository.findAll();
+		for (StreamDefinition streamDefinition : streamDefinitions) {
+			StreamDeployment streamDeployment = this.streamService.info(streamDefinition.getName());
+			for (StreamAppDefinition streamAppDefinition : streamDefinition.getAppDefinitions()) {
 				final String streamAppName = streamAppDefinition.getRegisteredAppName();
 				final ApplicationType streamAppType;
 				try {
@@ -277,9 +286,9 @@ public class SkipperAppRegistryController {
 					continue;
 				}
 				if (appType != streamAppType) {
-					break;
+					continue;
 				}
-				Map<String, Map<String, String>> streamDeploymentPropertiesMap = new HashMap<>();
+				Map<String, Map<String, String>> streamDeploymentPropertiesMap;
 				String streamDeploymentPropertiesString = streamDeployment.getDeploymentProperties();
 				ObjectMapper objectMapper = new ObjectMapper();
 				try {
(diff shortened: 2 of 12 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppUsedInStreamException.java
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.cloud.dataflow.server.controller;
+
+/**
+ * @author Christian Tzolov
+ */
+public class AppUsedInStreamException extends RuntimeException {
```
markpollack: I forgot to remove this and/or rename it.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/SkipperAppRegistryController.java
```diff
@@ -230,9 +249,58 @@ public void makeDefault(@PathVariable("type") ApplicationType type, @PathVariabl
 	@ResponseStatus(HttpStatus.OK)
 	public void unregister(@PathVariable("type") ApplicationType type, @PathVariable("name") String name,
 			@PathVariable("version") String version) {
+
+		// AppRegistration defaultApp = appRegistryService.getDefaultApp(name, type);
+		if (type != ApplicationType.task) { // && defaultApp != null && defaultApp.getVersion().equals(version)) {
+			String streamWithApp = findStreamContainingAppOf(type, name, version);
+			if (streamWithApp != null) {
+				throw new AppUsedInStreamException(String.format("The app [%s:%s:%s] you're trying to unregister is " +
```
markpollack: This needs to be fixed as it will create a confusing message.

jvalkeal:
Looks like this is not actually working
```
dataflow:>app register --name log --type sink --uri maven://org.springframework.cloud.stream.app:log-sink-rabbit:1.1.0.RELEASE
dataflow:>app register --name log --type sink --uri maven://org.springframework.cloud.stream.app:log-sink-rabbit:1.1.1.RELEASE
dataflow:>app register --name time --type source --uri maven://org.springframework.cloud.stream.app:time-source-rabbit:1.1.1.RELEASE
dataflow:>stream create --name foostream --definition "time|log"
dataflow:>stream deploy --name foostream


dataflow:>app list --id sink:log 
╔══════╤═════════╤═════════════════════╤════╗
║source│processor│        sink         │task║
╠══════╪═════════╪═════════════════════╪════╣
║      │         │> log-1.1.0.RELEASE <│    ║
║      │         │log-1.1.1.RELEASE    │    ║
╚══════╧═════════╧═════════════════════╧════╝

dataflow:>app unregister --name log --type sink 
Successfully unregistered application 'log' with type 'sink'. Please select new default version from: (1.1.1.RELEASE)

dataflow:>app list --id sink:log 
╔══════╤═════════╤═════════════════╤════╗
║source│processor│      sink       │task║
╠══════╪═════════╪═════════════════╪════╣
║      │         │log-1.1.1.RELEASE│    ║
╚══════╧═════════╧═════════════════╧════╝

$ jps|grep rabbit
22200 log-sink-rabbit-1.1.0.RELEASE.jar
22203 time-source-rabbit-1.1.1.RELEASE.jar

```

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/SkipperAppRegistryController.java
```diff
@@ -230,9 +249,58 @@ public void makeDefault(@PathVariable("type") ApplicationType type, @PathVariabl
...
+				try {
+					streamDeploymentPropertiesMap = objectMapper.readValue(streamDeploymentPropertiesString,
+							new TypeReference<Map<String, Map<String, String>>>() {
+							});
+				}
+				catch (IOException e) {
+					throw new RuntimeException("Can not deserialize string " + streamDeploymentPropertiesString);
+				}
+				if (streamDeploymentPropertiesMap.containsKey(appName)) {
+					Map<String, String> appDeploymentProperties = streamDeploymentPropertiesMap.get(streamAppName);
```
markpollack: We should change the key-value pairs that are placed in the map so that there are keys named `version` and `resource` instead of the `key` being the resource and the `value` being the version, e.g. 

not
```
  key: maven://org.springframework.cloud.stream.app:time-source-rabbit
  value: 1.2.0.RELEASE

but
```
  key: resource
  value: maven://org.springframework.cloud.stream.app:time-source-rabbit

  key: version
  value: 1.2.0.RELEASE
```
I believe this is a change in skipper.

File: spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/SkipperAppRegistryController.java [hunk 4]
```diff
@@ -230,9 +249,58 @@ public void makeDefault(@PathVariable("type") ApplicationType type, @PathVariabl
...
+				final ApplicationType streamAppType;
+				try {
+					streamAppType = DataFlowServerUtil.determineApplicationType(streamAppDefinition);
+				}
+				catch (CannotDetermineApplicationTypeException e) {
+					logger.warn("Can not determine ApplicationType for " + streamAppDefinition);
+					continue;
+				}
+				if (appType != streamAppType) {
+					break;
```
jvalkeal: I think this is wrong. Did some debugging with `time|log` when deleting `log`. It first sees time(source) which is different than log(sink) and breaks from for loop too early. Maybe this should be continue instead of break?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/SkipperAppRegistryController.java
```diff
@@ -230,9 +249,58 @@ public void makeDefault(@PathVariable("type") ApplicationType type, @PathVariabl
...
+					streamDeploymentPropertiesMap = objectMapper.readValue(streamDeploymentPropertiesString,
+							new TypeReference<Map<String, Map<String, String>>>() {
+							});
+				}
+				catch (IOException e) {
+					throw new RuntimeException("Can not deserialize string " + streamDeploymentPropertiesString);
+				}
+				if (streamDeploymentPropertiesMap.containsKey(appName)) {
+					Map<String, String> appDeploymentProperties = streamDeploymentPropertiesMap.get(streamAppName);
+					if (appDeploymentProperties.containsValue(appVersion)) {
```
markpollack: Looking in only values is because they key to find the version is the resource string.  As an example of the current deployment properties format

```
Stream Deployment properties: {
  "log" : {
    "spring.cloud.deployer.group" : "httptest",
    "maven://org.springframework.cloud.stream.app:log-sink-rabbit" : "1.1.0.RELEASE"
  },
  "http" : {
    "spring.cloud.deployer.group" : "httptest",
    "maven://org.springframework.cloud.stream.app:http-source-rabbit" : "1.2.0.RELEASE"
  }
}
```

This should instead be 

```
Stream Deployment properties: {
  "log" : {
    "spring.cloud.deployer.group" : "httptest",
    "resource" :  "maven://org.springframework.cloud.stream.app:log-sink-rabbit", 
    "version" : "1.1.0.RELEASE"
  },
  "http" : {
    "spring.cloud.deployer.group" : "httptest",
    "resource" : "maven://org.springframework.cloud.stream.app:http-source-rabbit",
    "version" : "1.2.0.RELEASE"
  }
}
```

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppUsedInStreamException.java
markpollack: renamed to UnregisterAppException and just pass through a string message.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/SkipperAppRegistryController.java
```diff
@@ -230,9 +250,70 @@ public void makeDefault(@PathVariable("type") ApplicationType type, @PathVariabl
...
+				}
+				catch (CannotDetermineApplicationTypeException e) {
+					logger.warn("Can not determine ApplicationType for " + streamAppDefinition);
+					continue;
+				}
+				if (appType != streamAppType) {
+					continue;
+				}
+				Map<String, Map<String, String>> streamDeploymentPropertiesMap;
+				String streamDeploymentPropertiesString = streamDeployment.getDeploymentProperties();
```
ilayaperumalg: If the stream is not yet deployed but the definition exists, then the app would still get unregistered. Because, the value for `streamDeploymentPropertiesString` would be `null` as the stream deployment properties is calculated from the Skipper manifest (from the deployed stream).

ilayaperumalg: As the commit message says, this PR addresses the prevention of app unregistration only for the `deployed` stream. So, in that perspective, this flow requires a `null` check on `streamDeploymentPropertiesString` to avoid exception processing it further.

jvalkeal: Yeah, noticed this as well. If starting skipper and dataflow from scratch, created a stream without deploying it, you get:
```
dataflow:>app unregister --name log --type sink 
Command failed org.springframework.cloud.dataflow.rest.client.DataFlowClientException: Can not deserialize Stream Deployment Properties JSON ''
```

ilayaperumalg: Once we have a `null` check, I guess the above error would go away as well.

jvalkeal: I'm adding a new test and:
```
if (!StringUtils.hasText(streamDeploymentPropertiesString)) {
	continue;
}
```

ilayaperumalg: I guess #2013 can still be kept open until we confirm on supporting the prevention of unregistration of the app for the stream that is **not** deployed yet.

jvalkeal:
Fix pr per eccc8e22dc83daaa576393abf7b649b0c34db8fe and additional fix per 533b753f0d26bdb8978e395d093435643cee2aa4

PR #2085, opened 2018-03-05
Add available apps table to the reference guide

sabbyanandan:
The goal of this proposed addition is to, 

1) Fix the "Applications" section of the docs since it simply points to "how to register apps"
2) Provide a quick view of all the available apps in a table format, so users can quickly jump to the respective app docs for experimentation
3) Link to the respective "generated docs" instead of direct includes (intention is not to couple the apps to scdf)

Built and generated the docs via:

> ./mvnw clean -DskipTests -Pfull package -pl spring-cloud-dataflow-docs

(a few app doc-links are missing in the Celsius guide - we are aware of it  and we will fix it)

image


ilayaperumalg:
LGTM, verified the link updates after generating the doc locally. Merging.

ilayaperumalg:
Rebased and merged as 8bd77ad5

PR #2095, opened 2018-03-06
Disable LRUResourceLoader caching on local server

### LINKED ISSUE #2087: Disable LRUResourceLoader caching on local server
Should not use cache eviction on local server.

jvalkeal:
- Overriding freeDiskSpacePercentage value in a local
  server application.yml.
- Fixes #2087

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/configuration.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/configuration.adoc
@@ -164,6 +164,13 @@ spring.cloud.deployer.local.javaOpts= # The Java options to pass to the JVM
 spring.cloud.deployer.local.freeDiskSpacePercentage=5 # The target percentage of free disk space to always aim for when cleaning downloaded resources (typically via the local maven repository). Specify as an integer greater than zero and less than 100. Default is 5.
 ----
 
+[NOTE]
+====
+Data Flow Local server itself overrides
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-docs/src/main/asciidoc/configuration.adoc
```diff
@@ -164,6 +164,13 @@ spring.cloud.deployer.local.javaOpts= # The Java options to pass to the JVM
 spring.cloud.deployer.local.freeDiskSpacePercentage=5 # The target percentage of free disk space to always aim for when cleaning downloaded resources (typically via the local maven repository). Specify as an integer greater than zero and less than 100. Default is 5.
 ----
 
+[NOTE]
+====
+Data Flow Local server itself overrides
+`spring.cloud.deployer.local.freeDiskSpacePercentage` to `0` from its
```
ilayaperumalg: I see this setting would disable deletion of the entries from the cache, I thought we want to get rid of the LRU cleaning loader completely from the local deployer?

jvalkeal: For now I didn't want to remove that as it's there and there's no alternative. This is probably safe bet for now as user can re-enable it.

ilayaperumalg: ok, makes sense. will merge as is for now.

ilayaperumalg:
Rebased and merged as ca848afc

PR #2099, opened 2018-03-08
StreamDefinitionToDslConverter produces invalid DSL

### LINKED ISSUE #2033: StreamDefinitionToDslConverter may create unparsable dsl from valid definition
I have encountered an Exception on deploying a stream containing a jdbc source.

The stream definition looks similar to this:
 
jdbc-mssql --query='UPDATE top (100) ASSURANCE SET assurance_flag = 1 OUTPUT Inserted.* WHERE assurance_flag IS NULL' --password='******' --username='******' --url='jdbc:sqlserver://db:1433;databaseName=Spring' --cron='*/10 * * * * *' --max-messages=-1 | cust-processor | router --default-output-channel=out

On deployment the following exception is thrown: 

org.springframework.cloud.dataflow.core.dsl.CheckPointedParseException: 100E:(pos x): Found unexpected data after stream definition: ';'

It appears it does not like the ';' symbol from the jdbc URL. 

I am using Spring Cloud Dataflow v1.3.0.RELEASE with Skipper v1.0.0.RELEASE. The same definitions worked fine in v1.2.0.RELEASE

tzolov:
Resolves #2033

### DIFF
--- a/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/StreamDefinitionToDslConverter.java
+++ b/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/StreamDefinitionToDslConverter.java
@@ -112,8 +112,10 @@ public class StreamDefinitionToDslConverter {
 	}
 
 	private String autoQuotes(String propertyValue) {
-		if (propertyValue.contains(" ") && !propertyValue.contains("'")) {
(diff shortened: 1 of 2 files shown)


jvalkeal:
lgtm, error is gone.

sabbyanandan:
@jvalkeal: We also need to backport this fix to 1.3.x.

PR #2103, opened 2018-03-12
Parameterize wget command in docker-compose.yml

### LINKED ISSUE #2102: Parameterize wget command to include version in Docker Compose file
As a developer, I'd like to parameterize to download the correct version of `docker-compose.yml`.

**Acceptance:**
Follow the same pattern used in [other wget commands](https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-docs/src/main/asciidoc/getting-started.adoc#getting-started-with-manual-installation).

chrisjs:
resolves #2102


chrisjs:
seems like this works. i was able to build the docs on master and have the replace tag resolve `master` and then checkout the tag `v1.4.0.RC1` and the replace tag resolved to for example: https://raw.githubusercontent.com/spring-cloud/spring-cloud-dataflow/v1.4.0.RC1/spring-cloud-dataflow-server-local/docker-compose.yml

jvalkeal:
lgtm, would not work with maintenance branches having snapshots but I think we only care about main snapshots against master.

PR #2123, opened 2018-03-28
Update to local deployer 1.3.4.BUILD-SNAPSHOT

jvalkeal:
- Needed to get metrics working with local server
  as there are a bending change in local deployer.


jvalkeal:
Also spring-cloud/spring-cloud-deployer-local#110 is needed for its fix to take effect.

PR #2126, opened 2018-04-03
Stream DSL - sanitize sensitive properties

### LINKED ISSUE #1985: On stream DSL reverse engineering hide sensitive properties
When the stream definition DSL is reverse engineered by parsing through the Skipper release manifest, filter out the sensitive properties  than contain labels such as `tokens`, `security`, `key` in their names.

tzolov:
Resolves #1985

- Extend ArgumentSanitizer to handle key,value properties
- Add tests for the ArgumentSanitizer and for the DefaultSkipperStreamService
- Resolution is applied to Skipper mode only.


ilayaperumalg:
LGTM, merging.

PR #2127, opened 2018-04-03
Move Stream creation logic from Controller to Service

### LINKED ISSUE #1924: Factor out the StreamDefinitionController#save() implementation into StreamService/AbstractStreamService
Stream definition management is owned by the `StreamService` the controller shouldn't be allowed to mutate the stream definition state. 

- Also move also the `AppRegistryCommon` form the controller to the StreamService.

tzolov:
Resolves #1924

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/StreamService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/StreamService.java
@@ -30,6 +30,20 @@ import org.springframework.cloud.deployer.spi.app.DeploymentState;
  * @author Christian Tzolov
  */
 public interface StreamService {
+
+	/**
+	 * Create a new stream.
+	 *
+	 * @param streamName stream name
+	 * @param dsl DSL definition for stream
+	 * @param deploy if {@code true}, the stream is deployed upon creation (default is
+	 * {@code false})
+	 * @return the created stream definition already exists
+	 * @throws InvalidStreamDefinitionException if there errors in parsing the stream DSL,
+	 * resolving the name, or type of applications in the stream
+	 */
+	StreamDefinition createStream(String streamName, String dsl, boolean deploy);
+
 	/**
 	 * Deploys the stream with the user provided deployment properties.
 	 * Implementations are responsible for expanding deployment wildcard expressions.
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/StreamService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/StreamService.java
@@ -20,6 +20,7 @@ import java.util.Map;
 
 import org.springframework.cloud.dataflow.core.StreamDefinition;
 import org.springframework.cloud.dataflow.core.StreamDeployment;
+import org.springframework.cloud.dataflow.server.controller.support.InvalidStreamDefinitionException;
 import org.springframework.cloud.deployer.spi.app.DeploymentState;
 
 /**
@@ -39,7 +40,7 @@ public interface StreamService {
 	 * @param deploy if {@code true}, the stream is deployed upon creation (default is
 	 * {@code false})
 	 * @return the created stream definition already exists
-	 * @throws InvalidStreamDefinitionException if there errors in parsing the stream DSL,
+	 * @throws InvalidStreamDefinitionException if there are errors in parsing the stream DSL,
 	 * resolving the name, or type of applications in the stream
 	 */
 	StreamDefinition createStream(String streamName, String dsl, boolean deploy);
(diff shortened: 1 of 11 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/StreamService.java
```diff
@@ -30,6 +30,20 @@
...
+
+	/**
+	 * Create a new stream.
+	 *
+	 * @param streamName stream name
+	 * @param dsl DSL definition for stream
+	 * @param deploy if {@code true}, the stream is deployed upon creation (default is
+	 * {@code false})
+	 * @return the created stream definition already exists
+	 * @throws InvalidStreamDefinitionException if there errors in parsing the stream DSL,
```
ilayaperumalg: Need to import this exception class.

ilayaperumalg: `if there` -> `if there are`

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/StreamService.java
```diff
@@ -30,6 +30,20 @@
...
+	 * resolving the name, or type of applications in the stream
+	 */
+	StreamDefinition createStream(String streamName, String dsl, boolean deploy);
```
ilayaperumalg: Should we add `throws InvalidStreamDefinitionException` for this method definition?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/AbstractStreamService.java
```diff
@@ -54,9 +64,60 @@
...
+
+	public AbstractStreamService(StreamDefinitionRepository streamDefinitionRepository, AppRegistryCommon appRegistry) {
 		Assert.notNull(streamDefinitionRepository, "StreamDefinitionRepository must not be null");
+		Assert.notNull(appRegistry, "AppRegistryCommon must not be null");
 		this.streamDefinitionRepository = streamDefinitionRepository;
+		this.appRegistry = appRegistry;
+	}
+
+	public StreamDefinition createStream(String streamName, String dsl, boolean deploy) {
+
```
ilayaperumalg: As a general practice, we can avoid adding extra line (though they make it look better) between the statements.

ilayaperumalg:
@tzolov LGTM.  I can add address the changes required for the above comments and merge if you are ok with them.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/StreamService.java
tzolov: addressed!

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/StreamService.java
tzolov: Given that `InvalidStreamDefinitionException` is a Runtime exception i believe we should not add it explicitly?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/AbstractStreamService.java
tzolov: which extra line are you referring to? It seems i've added more then one (extra) lines :)

mminella: Why would we discourage this?  @ilayaperumalg You even point out it makes it look better (more readable).  I know Spring Framework's code style only requires a blank line after method signatures that are multiline, but I'm a fan of it in all cases personally (https://github.com/spring-projects/spring-framework/wiki/Code-Style#blank-lines).

ilayaperumalg: @tzolov @mminella I personally like having blank lines when it makes the code look better. I guess I just got accustomed to not having blank lines between the statements inside the method after all those code review comments I received since Spring XD days ;)

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/StreamService.java
ilayaperumalg: I agree and having it in the Javadoc is sufficient. We can just fix the import part of this exception in this class file.

tzolov:
@ilayaperumalg, just pushed a commit that aims to address the review request. 
Please feel free to amend if i've missed something.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed and merged as 5dc68af6

PR #2129, opened 2018-04-03
Template local server docker image tag name

### LINKED ISSUE #2113: Apply Maven filtering on docker-compose file to be consistent with build
As a developer, I'd like to apply Maven filtering on the `docker-compose.yml` file, so we could make sure the release version is consistent with what's produced by the build.

chrisjs:
fixes #2113


ilayaperumalg:
LGTM

PR #2133, opened 2018-04-08
Pass all common-applicaton-properties to launching tasks

### LINKED ISSUE #2124: CommonApplicationProperties are not passed as parameter to starting tasks
According to the [documentation (chapter 46.4.1)](https://docs.spring.io/spring-cloud-dataflow/docs/current/reference/htmlsingle/#_common_application_properties) common application properties should be passed to the launching tasks.

For exmaple the following definition should pass the kafka broker to all tasks:
`--spring.cloud.dataflow.applicationProperties.task.spring.cloud.stream.kafka.binder.brokers=kafka:9092
`

but the [DefaultTaskService](https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java#L287) filters the properties with a prefix ("app." + registeredAppName), meaning the properties are not passed to the starting tasks.

This works different for CcommonApplicationProperties for Streams. Is this the expected behavior?

I tested it with DataFlowServerLocal Version 1.3.0.

michael-wirth:
Fixes #2124 and #1531 

According to the documentation the common-application-properties must not be prefixed with the `app.taskname`

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
@@ -68,6 +68,7 @@ import org.springframework.util.StringUtils;
  * @author Gunnar Hillert
  * @author Thomas Risberg
  * @author Ilayaperumal Gopinathan
+ * @author Michael Wirth
  */
 public class DefaultTaskService implements TaskService {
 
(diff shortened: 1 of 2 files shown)


cppwfs:
LGTM.  I added an additional test on merge.

cppwfs:
Added Test, Rebased, Squashed Merged.

cppwfs:
Thanks for the contribution!!!

PR #2134, opened 2018-04-09
Improve composed task graph to text algorithm for nested splits

### LINKED ISSUE #712: Remove stream app-starters dependency
As a developer, I'd like to remove `spring-cloud-stream-app-starters` dependency from the core project, so we don't depend on that project's releases anymore. Currently, the `app-starters-common-analytics` sub-project is directly used at the core controllers, which needs to go.

**Acceptance:**
- `app-starters-common-analytics` from app-starter's project is not a dependency anymore
- analytics APIs are consumed from a different standalone project/library 
- tests run successfully after the refactoring 

This issue carries a subset of scope from what was originally proposed in: spring-cloud/spring-cloud-dataflow#598

aclement:
A nested split is when a split includes a further split as one of
its branches. Without these changes 3 links from a node would
simply be considered as a 3 way split, but in fact it can be
a 2 way split with a nested split. For example: <<AA || BB> && CC || DD>

Fixes spring-cloud/spring-cloud-dataflow-ui#712
Fixes spring-cloud/spring-cloud-dataflow#2144

### DIFF
--- a/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/graph/Graph.java
+++ b/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/graph/Graph.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 the original author or authors.
+ * Copyright 2017-2018 the original author or authors.
  *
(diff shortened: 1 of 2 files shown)


ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Rebased and merged as c0e1bb9

Issue #2130, opened 2018-04-04
SCDF shell  authentication does not work when accessing  SCDF  behind the proxy.


jvmlet:
Hello
We have SCDF server installed with basic auth enabled. 
When accessing it from behind the proxy  with below arguments:
```
-jar spring-cloud-dataflow-shell-1.4.0.RELEASE.jar ^
--dataflow.uri=http://data-flow-server.xxxx.com ^
--dataflow.username=xxxx ^
--dataflow.password=xxxx ^
--dataflow.proxy.uri=http://185.xxx.xxx.xxx:9480 ^
```

the SCDF server credentials are not being propagated by http client because of mismatch in AuthScope, it tries to match the scope of proxy server against the scope of SCDF server. Because of mismatch, the `null` credentials are returned.
![image](https://user-images.githubusercontent.com/6664858/38296085-f08ebe5c-37f8-11e8-9262-3d0eb385689d.png)

The proxy itself  doesn't use  authentication.

ilayaperumalg:
@jvmlet Thanks for reporting the issue. This looks like a bug.

The credential provider for proxy Auth overrides the one that is set for basic auth (even when the proxy username/password are not provided - it is probably because they are considered non-null for the default empty string values).

The shell command configuration option for data flow proxy username looks incorrect as well.

PR #2136, opened 2018-04-10
gh-2130 Fix SCDF shell authentication does not work when accessing SCDF behind the proxy

ghillert:
* Add tests

resolves https://github.com/spring-cloud/spring-cloud-dataflow/issues/2130


spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/common/ConfigCommands.java
```diff
@@ -216,8 +216,7 @@ public String target(
 		try {
 			this.targetHolder.setTarget(new Target(targetUriString, targetUsername, targetPassword, skipSslValidation));
 
-			final HttpClientConfigurer httpClientConfigurer = HttpClientConfigurer.create()
```
ilayaperumalg: I think we have an issue with the `proxy-username` option configuration here. 
```
 			@CliOption(mandatory = false, key = {
 					"proxy-username" }, help = "the username for authenticated access to the secured proxy server (valid only with a "
 					+ "username)", specifiedDefaultValue = Target.DEFAULT_SPECIFIED_PASSWORD, unspecifiedDefaultValue = Target.DEFAULT_UNSPECIFIED_PASSWORD) String proxyUsername,
```

ghillert: Made the help wording better

ghillert:
Rebased with master + addressed comment

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Merged as d0dba6f1

PR #2162, opened 2018-04-25
CTR can be deleted with incomplete set of child tasks

### LINKED ISSUE #2161: Destroying CTR fails if one of child tasks is not present
In cases where a user does a bulk delete of tasks the child tasks of a CTR can be deleted before the parent CTR.  In these cases when the CTR is deleted it fails because a `NoSuchTaskDefinitionException` is thrown.  A CTR should be able to be deleted even if child tasks are missing.

cppwfs:
resolves #2161


ilayaperumalg:
Rebased and merged as 9c50e843

PR #2166, opened 2018-04-26
Update controller->service->repository layering

### LINKED ISSUE #1892: Review layering of Controller's calls into repos+services
Currently some controller methods make several calls into repository methods that would be better off being in a service layer to ensure proper transaction semantics.

markpollack:
Fixes #1892

### DIFF
--- a/spring-cloud-dataflow-docs/src/test/java/org/springframework/cloud/dataflow/server/rest/documentation/StreamDefinitionsDocumentation.java
+++ b/spring-cloud-dataflow-docs/src/test/java/org/springframework/cloud/dataflow/server/rest/documentation/StreamDefinitionsDocumentation.java
@@ -107,7 +107,7 @@ public class StreamDefinitionsDocumentation extends BaseDocumentation {
 			.andDo(this.documentationHandler.document(
 					requestParameters(
 							parameterWithName("nested")
-								.description("Should we recursively search for related stream definitions (optional)")),
+								.description("Should we recursively findByNameLike for related stream definitions (optional)")),
 					pathParameters(parameterWithName("name")
 						.description("The name of an existing stream definition (required)"))
 				));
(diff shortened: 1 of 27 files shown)


ilayaperumalg:
@markpollack I believe we wanted to refactor `StreamDeploymentController` and task related controllers in a similar way. Are we planning to do it as part of this PR?

sabbyanandan:
@ilayaperumalg: Given the number of files in this PR, we have decided to start with this first and follow-up with the remaining via different stories. This PR is good to go as-is.

ilayaperumalg:
LGTM

tzolov:
@markpollack, for consistency/completeness i we can move the `StreamDefinitionRepository` out of the [StreamDeploymentController](https://github.com/markpollack/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java#L71) into the `StreamService`.  Actually later already has the `StreamDefinitionRepository` dependency, So we:

* Need to replace the following [snippet](https://github.com/markpollack/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java#L99-L102) (which occurs [twice](https://github.com/markpollack/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java#L124-L127)) with a new `StreamService#isStreamExists(streamName)` service method. 
* And move the [undeployAll](https://github.com/markpollack/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java#L112) logic to the `StreamService` as we did for the `deleteAll` logic.

What do you think?

PR #2177, opened 2018-05-02
Add @Transactional to Job related services

markpollack:


markpollack:
looks to be failed based on failure we fixed in another PR... will remove this one and make a new PR based on latest master.

PR #2180, opened 2018-05-03
Update to use hikari connection pool

ghillert:
- Fix SkipperStreamService Tests
- Remove extraneous `system.out.println` from debugging connection pool issues

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -152,6 +152,11 @@
 				<scope>import</scope>
 			</dependency>
 			<dependency>
+				<groupId>com.zaxxer</groupId>
+				<artifactId>HikariCP</artifactId>
+				<version>3.1.0</version>
+			</dependency>
+			<dependency>
 				<groupId>org.hibernate</groupId>
 				<artifactId>hibernate-core</artifactId>
 				<version>5.2.12.Final</version>
--- a/spring-cloud-dataflow-server-core/pom.xml
+++ b/spring-cloud-dataflow-server-core/pom.xml
@@ -10,6 +10,11 @@
 	</parent>
 	<dependencies>
 		<dependency>
+			<groupId>com.zaxxer</groupId>
+			<artifactId>HikariCP</artifactId>
+			<version>3.1.0</version>
+		</dependency>
+		<dependency>
 			<groupId>org.hibernate</groupId>
 			<artifactId>hibernate-core</artifactId>
 			<version>5.2.12.Final</version>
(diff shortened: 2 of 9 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/SkipperStreamService.java
```diff
@@ -68,4 +70,6 @@
 	 * @return list of supported deployment platforms
 	 */
 	Collection<Deployer> platformList();
+
+	String convertPropertiesToSkipperYaml(StreamDefinition streamDefinition, Map<String, String> updateProperties);
```
ilayaperumalg: Should this be an interface method? If so, please add javadoc as to why we need it as a contract.

I believe this method can just be part of implementation as it sounds like a helper method.

pom.xml
```diff
@@ -151,6 +151,11 @@
 				<type>pom</type>
 				<scope>import</scope>
 			</dependency>
+			<dependency>
+				<groupId>com.zaxxer</groupId>
+				<artifactId>HikariCP</artifactId>
+				<version>3.1.0</version>
```
ilayaperumalg: Can we parameterize this `version` value so that we can just update in one place if needed?

spring-cloud-dataflow-server-core/pom.xml
```diff
@@ -96,6 +101,12 @@
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter-jdbc</artifactId>
+			<exclusions>
```
ilayaperumalg: Have we run into any issue without this exclusion?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/SkipperStreamService.java
ghillert: It is used by a test. I CANNOT use an implementation for dependency injection as was done by the original test or I get a proxy error. Either I add this method to the interface or we remove the test. I did not see how to solve the test error otherwise. @markpollack please advise.

ghillert:
@ilayaperumalg Removed the version as it is managed by Boot

spring-cloud-dataflow-server-core/pom.xml
markpollack: Not really, but having two connection pools in the classpath isn't a good idea.

pom.xml
```diff
@@ -151,11 +151,6 @@
 				<type>pom</type>
 				<scope>import</scope>
 			</dependency>
-			<dependency>
```
markpollack: I don't see why we have to live with the Boot managed version since it is a very old one, 2.5.1, from Sep 2016.  We should move to the latest

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/SkipperStreamService.java
markpollack: I will create the service 'by hand' to test the method and make it packag protected.

Issue #2160, opened 2018-04-24
Revisit Dashboard section in the reference guide


sabbyanandan:

sabbyanandan:
I'm going to wait it out until after the RC1. A few more UI/UX adjustments are in the pipeline; it'd be better to update screenshots when it is all done.

PR #2205, opened 2018-05-11
Update dashboard screenshots

sabbyanandan:
In this proposal, I've changed most of the Dashboard screenshots. The bulk-import for Tasks is gone, so I removed it from the docs, too. 

Unfortunately, though, I cannot test it with a build (via: `./mvnw package -DskipTests=true -P full -pl spring-cloud-dataflow-docs -am`) because the images are loaded from GitHub URI through `{dataflow-asciidoc}` parameterization. 

Resolves spring-cloud/spring-cloud-dataflow#2160


markpollack:
merged as cdce86791233842a6d5b601207f940a32093f8a8

PR #2226, opened 2018-05-22
Delete skipper package upon deployment failure

### LINKED ISSUE #2225: Delete Skipper package if deployment isn't successful
In case of errors during Skipper stream deploy, delete the package that is uploaded as part of the deployment.

ilayaperumalg:
Resolves #2225


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/stream/SkipperStreamDeployer.java
```diff
@@ -220,7 +221,14 @@ public Release deployStream(StreamDeploymentRequest streamDeploymentRequest) {
 		installProperties.setReleaseName(streamName);
 		installProperties.setConfigValues(new ConfigValues());
 		installRequest.setInstallProperties(installProperties);
-		Release release = skipperClient.install(installRequest);
+		Release release = null;
+		try {
+			release = this.skipperClient.install(installRequest);
+		}
+		catch (Exception e) {
+			this.skipperClient.packageDelete(packageName);
```
jvalkeal: Should we try/catch packageDelete as well? Then it's a choice of logging a warning or choosing which error to wrap in `SkipperException`?

jvalkeal:
lgtm, merging

ilayaperumalg:
@jvalkeal I think we need to back port this change to 1.5.x ?

markpollack:
This has been superseded by https://github.com/spring-cloud/spring-cloud-dataflow/pull/2240

PR #2237, opened 2018-05-30
Clarify docker compose syntax and add Windows example

### LINKED ISSUE #2214: Clarify local server docker compose DF version syntax
To allow dynamic version references in spring-cloud-dataflow/spring-cloud-dataflow-server-local/docker-compose.yml in our docs we state to do:

`DATAFLOW_VERSION=1.5.0.RELEASE docker-compose up`

This may be confusing if the user is not familiar with this style of passing environment variables or if for example the user may be using another operating system such as Windows that may not set env vars in the same way. To clarify this we could provide the same example as above with some additional information on what is being done (ie: set env var for process) or break it up into two statements, one being the `export` the second being `docker-compose up` or whatnot. Additionally, we may want to provide an example of how to set this variable on Windows as the existing syntax works on mac/unix variants

chrisjs:
Resolves #2214

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/getting-started.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/getting-started.adoc
@@ -36,16 +36,36 @@ Docker Compose is required and it's recommended to use the link:https://docs.doc
 +
 [source,bash,subs=attributes]
 ----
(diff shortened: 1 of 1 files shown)


ghillert:
LGTM - minor polishing - merged with commit 12ae4ce0c28fb5a119983bd0ce01ecf14e567b30

PR #2240, opened 2018-05-31
Fix package deletion upon stream undeploy

### LINKED ISSUE #2225: Delete Skipper package if deployment isn't successful

ilayaperumalg:
- When the stream is undeployed in skipper mode, the corresponding package gets deleted (if it was uploaded for this stream) irrespective of whether the release for the stream exists or not

Resolves #2225


tzolov:
Practically this change the implementation to `forcefully` try delete the deployment, irrelevant of the current status.
I guess this doesn't break the deployment flow semantics and IMO is good to merge

tzolov:
Actually there are broken tests. I can't merge this before those are resolved

tzolov:
I've mistakenly merged this first int 1.5.x at 6e92971 and the fort-ported it to 1.6.x (master) at bbde7cde

PR #2250, opened 2018-06-04
Fix spring session dep

markpollack:
We were managing the dependency to 1.2.2 when it should be whatever spring cloud common security config requires.

Note, the version of spring-session pulled in for the local server is managed by spring-boot-parent. The 1.2.2 dep was getting pulled into the final jar on the PCF/k8s servers.


markpollack:
wrong branch

PR #2251, opened 2018-06-04
Remove explicit mgmt of spring-session

markpollack:
**PORT ONTO MASTER BRANCH AS WELL**
We were managing the dependency to 1.2.2 when it should be whatever spring cloud common security config requires.

Note, the version of spring-session pulled in for the local server is managed by spring-boot-parent. The 1.2.2 dep was getting pulled into the final jar on the PCF/k8s servers.


ghillert:
LGTM - merged to 1.5.x and also ported to MASTER with commit 962ac880b791714a681477f51cb1737c98bd90a2

PR #2252, opened 2018-06-04
Add (exclude = SessionAutoConfiguration.class) to main class

markpollack:
Observed that exclude in yaml was not always respected


ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Merged as ce35df43 into master and back ported into 1.5.x as c3bb0659

Issue #2254, opened 2018-06-05
Task Definition status returns `unknown`


ghillert:
When retrieving task definitions, the `status` property is alway `unknown`. 

http://localhost:9393/tasks/definitions

```json
{
  "_embedded": {
    "taskDefinitionResourceList": [{
      "name": "foo",
      "dslText": "timestamp-batch",
      "composed": false,
      "status": "unknown",
      "_links": {
        "self": {
          "href": "http://localhost:9393/tasks/definitions/foo"
        }
      }
    }]
  },
  "_links": {
    "self": {
      "href": "http://localhost:9393/tasks/definitions?page=0&size=20"
    }
  },
  "page": {
    "size": 20,
    "totalElements": 1,
    "totalPages": 1,
    "number": 0
  }
}
```

It turns out the *deployment key/id* is never persisted. Therefore, in `TaskDefinitionController` when executing:

```java
String id = deploymentIdRepository.findOne(key);
```

a `null` value is always returned.

Is this `status` for Tasks still relevant? Can it be removed? Shall we instead show/return the execution status of the last Task execution?

mminella:
Status for tasks is not revelvant and should be removed.  The "status" of a task is dictated by the `TaskExecution` which is via the `TaskRepository` and have nothing related to the underlying platform.

ghillert:
The status should be replaced with a property representing the last TaskExecution (if available) for the given TaskDefinition. The `TaskExecution` itself will then include also a `status` property.

We should also define a status for task executions: 

- 0 == SUCCESS
- !0 == FAILED
- no end time == RUNNING?

Requires: https://github.com/spring-cloud/spring-cloud-task/issues/424

PR #2268, opened 2018-06-15
gh-2254 Fix Task Definition status returns `unknown`

ghillert:
- Retrieve the latest `TaskExecution` (if available) associated with the `TaskDefinition`
- Add the concept of a TaskExecution status
- Use this status to populate the status property of the `TaskDefinitionResource`
- As an enhancement also return the latest `TaskExecution` (if available) with the `TaskDefinitionResource`
- Add Tests

resolves https://github.com/spring-cloud/spring-cloud-dataflow/issues/2254

requires https://github.com/spring-cloud/spring-cloud-task/pull/425

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2017 the original author or authors.
+ * Copyright 2016-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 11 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
```diff
@@ -161,39 +162,68 @@ public TaskDefinitionResource display(@PathVariable("name") String name) {
...
+		@Override
+		public TaskExecutionAwareTaskDefinition convert(TaskDefinition source) {
+
+			TaskExecution lastTaskExecution = null;
+
+			if (taskExecutions != null) {
+				lastTaskExecution = taskExecutions.get(source.getName());
+			}
+
+			if (lastTaskExecution != null) {
```
cppwfs: A status of UKNOWN is not correct when working with a task that has not task-execution.  It should say "Not Executed"   Thoughts?

ghillert: Should probably be a follow up issue - Right now the statuses are mapped to what the UI supports.

cppwfs:
Added issues https://github.com/spring-cloud/spring-cloud-dataflow/issues/2271 & https://github.com/spring-cloud/spring-cloud-dataflow/issues/2272

cppwfs:
LGTM Polished, Squashed Merged

PR #2289, opened 2018-06-28
Modify app info cmd options

### LINKED ISSUE #1906: Shell UX: `app info` and `app unregister` should follow the same syntax
Currently we do:

`app info source:http`

and

`app unregister --name http --type source`

That seems inconsistent - Should we possibly support:

`app info --name http --type source`

ilayaperumalg:
- Use `app info --name <> --type <>` to be in sync with other app commands
 - Fix both the skipper/classic `app info` commands
 - Fix test
 - Update documenation references for `app info`

Resolves #1906


spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/classic/ClassicAppRegistryCommands.java
```diff
@@ -84,15 +84,17 @@ public boolean availableWithCreateRole() {
 
 	@CliCommand(value = APPLICATION_INFO, help = "Get information about an application")
 	public List<Object> info(
```
markpollack: I believe we wanted to make the `--name` and `--type` an additional option to avoid breaking things....

markpollack:
merged as b2de7a00c4355100931521b3cfadf0225fbfaac3

Issue #2292, opened 2018-06-28
Don't query for latest task executions when no task definitions are present


ghillert:
Depended on by https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/833

PR #2293, opened 2018-06-28
gh-2292 Fix: Don't query for latest task executions when no task definitions are present

ghillert:
resolves https://github.com/spring-cloud/spring-cloud-dataflow/issues/2292


cppwfs:
LGTM

PR #2302, opened 2018-07-02
Supports a max num of schedules to return in a single query

### LINKED ISSUE #2299: Limit Scheduler Result to 10000
Currently the scheduler will return the full result set of ScheduleInfo's back to dataflow.   But the max to be displayed should be 10000 or less.  Also the default page size will be 10000.

### LINKED ISSUE #2301: Simplify Scheduler props to follow dataflow standards
Currently to set the expression for a task scheduler at creation time a user has to enter `scheduler.<appname>.spring.cloud.scheduler.key` as the key.  
First off a expression should not be bound to a task but rather to the schedule itself.   

The correct property should for dataflow should just be `scheduler.key` (i.e. `scheduler.cron.expression` ).

cppwfs:
Spring Cloud Data Flow Scheduler Service should return no more than 10,000 per request.
The user can change this by setting the spring.cloud.dataflow.scheduler.service property to the desired max

Updated to standardize scheduler properties
This allows a user to set the expression for specific schedule by using the following key:
scheduler.cron.expression=<whatever>

You will note is this PR that the extraction of scheduler properties  was removed from the deployer utils.
This was because, the original method of defining scheduler properties matched deployment properties.  Now this is no longer the case.

resolves #2299
resolves #2301

### DIFF
--- a/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
+++ b/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
@@ -164,8 +164,8 @@ public final class DeploymentPropertiesUtils {
 	}
 
 	/**
-	 * Ensure that deployment properties only have keys starting with {@code app.} or
-	 * {@code deployer.}. In case non supported key is found {@link IllegalArgumentException}
+	 * Ensure that deployment properties only have keys starting with {@code app.},
+	 * {@code deployer.} or, {@code spring.cloud.scheduler.}. In case non supported key is found {@link IllegalArgumentException}
 	 * is thrown.
 	 *
 	 * @param properties the properties to check
@@ -176,9 +176,10 @@ public final class DeploymentPropertiesUtils {
 		}
 		for (Entry<String, String> property : properties.entrySet()) {
 			String key = property.getKey();
-			if (!key.startsWith("app.") && !key.startsWith("deployer.")) {
+			if (!key.startsWith("app.") && !key.startsWith("deployer.")
+					&& !key.startsWith("spring.cloud.scheduler.")) {
 				throw new IllegalArgumentException(
-						"Only deployment property keys starting with 'app.', 'deployer.' allowed, got '" + key + "'");
+						"Only deployment property keys starting with 'app.', 'deployer.' or, 'spring.cloud.scheduler.' allowed, got '" + key + "'");
 			}
 		}
 	}
--- a/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
+++ b/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
@@ -177,9 +177,9 @@ public final class DeploymentPropertiesUtils {
 		for (Entry<String, String> property : properties.entrySet()) {
 			String key = property.getKey();
 			if (!key.startsWith("app.") && !key.startsWith("deployer.")
-					&& !key.startsWith("spring.cloud.scheduler.")) {
+					&& !key.startsWith("scheduler.")) {
 				throw new IllegalArgumentException(
-						"Only deployment property keys starting with 'app.', 'deployer.' or, 'spring.cloud.scheduler.' allowed, got '" + key + "'");
+						"Only deployment property keys starting with 'app.', 'deployer.' or, 'scheduler.' allowed, got '" + key + "'");
 			}
 		}
 	}
--- a/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
+++ b/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
@@ -176,9 +176,9 @@ public final class DeploymentPropertiesUtils {
 		}
 		for (Entry<String, String> property : properties.entrySet()) {
 			String key = property.getKey();
-			if (!key.startsWith("app.") && !key.startsWith("deployer.") && !key.startsWith("scheduler.")) {
+			if (!key.startsWith("app.") && !key.startsWith("deployer.")) {
 				throw new IllegalArgumentException(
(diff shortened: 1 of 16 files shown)


sabbyanandan:
@tzolov / @ilayaperumalg: We would love to hear your feedback.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/SchedulerServiceProperties.java
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Specifies the properties required to configure how data flow handles schedules.
+ *
+ * @author Glenn Renfro
+ */
+@ConfigurationProperties(prefix = SchedulerServiceProperties.SCHEDULER_SERVICE_PREFIX)
+public class SchedulerServiceProperties {
+	public static final String SCHEDULER_SERVICE_PREFIX = DataFlowPropertyKeys.PREFIX + "scheduler.service";
+
+	public static final int SCHEDULER_MAX_RETURNED_NUMBER = 10000;
```
ilayaperumalg: Could you tell the reason why we set the max to `10000` which looks reasonably high?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskSchedulerController.java
```diff
@@ -119,7 +120,6 @@ public void save(@RequestParam("scheduleName") String scheduleName,
 			@RequestParam String properties,
 			@RequestParam(required = false) List<String> arguments) {
 		Map<String, String> propertiesToUse = DeploymentPropertiesUtils.parse(properties);
-		DeploymentPropertiesUtils.validateDeploymentProperties(propertiesToUse);
```
ilayaperumalg: I believe we would still want to validate the deployment properties (may be specific to scheduler properties to validate against their prefix) to make sure they only allow scheduler specific properties.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskSchedulerController.java
```diff
@@ -69,16 +70,16 @@ public TaskSchedulerController(SchedulerService schedulerService) {
...
 	 * @param assembler assembler for the {@link ScheduleInfo}
 	 * @return a list of Schedules
 	 */
 	@RequestMapping(value = "", method = RequestMethod.GET)
 	@ResponseStatus(HttpStatus.OK)
-	public PagedResources<ScheduleInfoResource> list(Pageable pageable,
+	public PagedResources<ScheduleInfoResource> list(
 			PagedResourcesAssembler<ScheduleInfo> assembler) {
+		List<ScheduleInfo> result = this.schedulerService.list();
+		Pageable pageable = new PageRequest(0, result.size());
```
ilayaperumalg: This way we enforce the pageable content size based on the result. I think this has to come from the client; otherwise, the client implementations (ex: web UI) wouldn't be able to request the desired page.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/SchedulerServiceProperties.java
cppwfs: Great question!   the list request returns all schedules.   In some cases that could be numbers greater than 10,000.   This number was selected by @markpollack .

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskSchedulerController.java
cppwfs: Updated

spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
```diff
@@ -176,9 +176,10 @@ public static void validateDeploymentProperties(Map<String, String> properties)
 		}
 		for (Entry<String, String> property : properties.entrySet()) {
 			String key = property.getKey();
-			if (!key.startsWith("app.") && !key.startsWith("deployer.")) {
+			if (!key.startsWith("app.") && !key.startsWith("deployer.")
+					&& !key.startsWith("spring.cloud.scheduler.")) {
```
ilayaperumalg: Wouldn't this be still `schedule` as  a prefix?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskSchedulerController.java
```diff
@@ -119,7 +124,6 @@ public void save(@RequestParam("scheduleName") String scheduleName,
 			@RequestParam String properties,
 			@RequestParam(required = false) List<String> arguments) {
 		Map<String, String> propertiesToUse = DeploymentPropertiesUtils.parse(properties);
-		DeploymentPropertiesUtils.validateDeploymentProperties(propertiesToUse);
```
ilayaperumalg: We still need to uncomment this line to have validation here.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultSchedulerService.java
```diff
@@ -137,14 +146,52 @@ public void unschedule(String scheduleName) {
...
+	/**
+	 * Retain only properties that are meant for the <em>scheduler</em> of a given task(those
+	 * that start with {@code scheduler.}and qualify all
+	 * property values with the {@code spring.cloud.scheduler.} prefix.
+	 *
+	 * @param input the scheduler properties
+	 * @return scheduler properties for the task
+	 */
+	private static Map<String, String> extractAndQualifySchedulerProperties(Map<String, String> input) {
+		final String prefix = "scheduler.";
```
ilayaperumalg: This is the prefix that needs to be validated during the validation not the `spring.cloud.scheduler` one.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed, rebased and merged as f3e39f4b

PR #2305, opened 2018-07-06
Allows a user to retrieve a single schedule via the RESTful api.

### LINKED ISSUE #2247: List Schedule elements have href to self that gives 404
We need to create a get method to support it.

cppwfs:
resolves #2247

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskSchedulerController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskSchedulerController.java
@@ -88,6 +88,19 @@ public class TaskSchedulerController {
 	}
 
 	/**
+	 * Return a {@link ScheduleInfo} for a specific Schedule.
+	 *
+	 * @param scheduleName assembler for the {@link ScheduleInfo}
+	 * @return a {@link ScheduleInfoResource} instance for the scheduleName specified.
+	 */
+	@RequestMapping(value = "/{name}", method = RequestMethod.GET)
+	@ResponseStatus(HttpStatus.OK)
+	public ScheduleInfoResource getSchedule(
+			@PathVariable("name") String scheduleName) {
+		return taskAssembler.toResource(schedulerService.getSchedule(scheduleName));
+	}
+
+	/**
 	 * Return a page-able list of {@link ScheduleInfo}s for a specific
 	 * {@link org.springframework.cloud.dataflow.core.TaskDefinition} name.
 	 *
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskSchedulerController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskSchedulerController.java
@@ -97,7 +97,7 @@ public class TaskSchedulerController {
 	@ResponseStatus(HttpStatus.OK)
 	public ScheduleInfoResource getSchedule(
 			@PathVariable("name") String scheduleName) {
-		return taskAssembler.toResource(schedulerService.getSchedule(scheduleName));
+		return taskAssembler.toResource(this.schedulerService.getSchedule(scheduleName));
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskSchedulerController.java
```diff
@@ -87,6 +87,19 @@ public TaskSchedulerController(SchedulerService schedulerService) {
...
+	 * Return a {@link ScheduleInfo} for a specific Schedule.
+	 *
+	 * @param scheduleName assembler for the {@link ScheduleInfo}
+	 * @return a {@link ScheduleInfoResource} instance for the scheduleName specified.
+	 */
+	@RequestMapping(value = "/{name}", method = RequestMethod.GET)
+	@ResponseStatus(HttpStatus.OK)
+	public ScheduleInfoResource getSchedule(
+			@PathVariable("name") String scheduleName) {
+		return taskAssembler.toResource(schedulerService.getSchedule(scheduleName));
```
mminella: `this.schedulerService`

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultSchedulerService.java
```diff
@@ -166,6 +166,15 @@ public void unschedule(String scheduleName) {
 				this.schedulerServiceProperties.getMaxSchedulesReturned());
 	}
 
+	@Override
+	public ScheduleInfo getSchedule(String scheduleName) {
+		List<ScheduleInfo> result = list().stream()
+				.filter(scheduleInfo -> scheduleInfo.getScheduleName().equals(scheduleName))
+				.collect(Collectors.toList());
+		Assert.isTrue(!(result.size() > 1), "more than one schedule was returned for scheduleName, should only be one");
+		return result.size() > 0 ? result.get(0) : null;
```
mminella: If you have 2 in the result, are you sure the first one is the right one?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultSchedulerService.java
```diff
@@ -166,6 +166,15 @@ public void unschedule(String scheduleName) {
 				this.schedulerServiceProperties.getMaxSchedulesReturned());
 	}
 
+	@Override
+	public ScheduleInfo getSchedule(String scheduleName) {
+		List<ScheduleInfo> result = list().stream()
```
ilayaperumalg: Instead of using `list`, can we have a find() method to just fetch the schedule by the given name from the corresponding repository? Having `list` and then iterate over it could be an overkill.

cppwfs: Understood.  We don't have a find method available in the SPI.  I can add a story to support this.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultSchedulerService.java
cppwfs: hmm...   Doesn't the assert handle this case?   if the result set has more than one value it should throw an assert.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultSchedulerService.java
cppwfs: https://github.com/spring-cloud/spring-cloud-scheduler/issues/17

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultSchedulerService.java
mminella: Ah.  I missed that.

ilayaperumalg:
LGTM

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultSchedulerService.java
cppwfs: We will need to chat with @sabbyanandan to see when we can work on that story.

ilayaperumalg:
Merging.

sabbyanandan:
Merged in https://github.com/spring-cloud/spring-cloud-dataflow/commit/a63bea15b0ec85218c070f1a23e4dd79a09746fa.

PR #2314, opened 2018-07-10
Added docs for CTR thread count req for splits

### LINKED ISSUE #2313: Add thread-count content in the Split Execution section of CTR in Ref Doc
Need to add a brief discussion on split-thread-core-pool-size within the split execution section of the doc.  Currently it is only in the configuration options.

cppwfs:
This was added as a part of configuration options, but also really needs to be in the split execution section as well.

resolves #2313


ilayaperumalg:
LGTM, merging.

PR #2318, opened 2018-07-12
Work around warnings about missing attributes in path parameters snippets

wilkinsona:
The docs build currently generates warnings like the following:

```
WARNING: skipping reference to missing attribute: task-definition-name
```

They are caused by the title of the table in the REST Docs-generated path parameters snippet having content that Asciidoctor interprets as an attribute reference. This will be fixed by https://github.com/spring-projects/spring-restdocs/issues/527. In the meantime, it can be worked around with a custom path parameters snippet template as proposed here.


cppwfs:
Thanks @wilkinsona for this contribution.   That resolved a ton of our warnings!   I included your fix as a part of PR https://github.com/spring-cloud/spring-cloud-dataflow/pull/2319.  I'll close this PR because the work is part of a larger project.   Again thanks!!!!

PR #2320, opened 2018-07-13
Implement concurrent task execution limit

### LINKED ISSUE #2140: Add support to configure a limit for concurrent Task Launches
As a user, I'd like to configure a maximum limit for concurrent Task launches, so I can limit the number of Task launches and by the virtue of it I don't run the risk of saturating the IaaS/hardware resource limits.

**Notes:**
- The user should be able to configure a "maximum task launches at a time" as a server-level setting. Perhaps call the new property `spring-cloud-dataflow-concurrent-task-limit`?
- The new property will be treated similarly to feature-toggle flags such as `spring.cloud.dataflow.features.skipper-enabled`.
- At runtime, upon a new launch request, the new property would be used to compare against the actual number of running Tasks to make decisions

dturanski:
Fixes #2140

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,5 +30,5 @@ spring-*/src/main/java/META-INF/MANIFEST.MF
 *.iml
 *.ipr
 *.iws
-.idea/*
+.idea/
 rebel.xml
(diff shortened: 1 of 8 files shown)


dturanski:
Redoing to fix conflict

PR #2321, opened 2018-07-13
Implement concurrent task execution limit

### LINKED ISSUE #2140: Add support to configure a limit for concurrent Task Launches

dturanski:
Fixes #2140

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/TaskConfigurationProperties.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/TaskConfigurationProperties.java
@@ -16,14 +16,22 @@
 
 package org.springframework.cloud.dataflow.server.service.impl;
 
+import javax.validation.constraints.Min;
+
+import org.hibernate.validator.constraints.NotBlank;
+
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.cloud.dataflow.core.DataFlowPropertyKeys;
+import org.springframework.validation.annotation.Validated;
+
 
 /**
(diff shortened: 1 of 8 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/TaskConfigurationProperties.java
```diff
@@ -32,8 +40,12 @@
 	/**
 	 * The task application name to be used for the composed task runner.
 	 */
+	@NotBlank
 	private String composedTaskRunnerName = "composed-task-runner";
 
+	@Min(1)
+	private long maximumConcurrentTasks = Long.MAX_VALUE;
```
markpollack: I think this is the wrong default, while consistent with the behavior of the past servers, we are letting people accidentally get into a situation where they can exhaust all their AIs.  Picking a value such as 20 is probably a reasonable compromise.

dturanski: Updated in the subsequent PR

dturanski:
Closing since this branch is merged with #2323

PR #2330, opened 2018-07-19
Add Batch and Tasks DB indexes

### LINKED ISSUE #2306: Review DB schema for optimizations
As a developer, I'd like to review the DB schema for optimizations. In particular, we would like to review the correlation/join queries to improve the performance where applicable. 

The community shared a noticeable improvement to Task/Batch + SCDF integration - see [here](https://gitter.im/spring-cloud/spring-cloud-dataflow?at=5b295ec2ce3b0f268d2ab919). Likewise, it'd be good to review the overall schema structuring and tune it where appropriate.

Notes:

- It'd be good to load 100s-1000s of rows as part of the load simulation and analysis.
- Given the historical nature to Task/Batch execution tracking in the database, adding indexes for improvement would generally be a good candidate (see https://github.com/spring-cloud/spring-cloud-dataflow/issues/2290#issuecomment-402504948 for data points).
- If we were to apply optimizations, we would have to consider the case for all the supported databases.
- Document caveats if any.

tzolov:
- Add batch_task_inexes.sql scripts for the supported DB that adds some index optimization
 - Scripts are initialized only if the Task feature is enabled
 - Scripts are initialized always after the Batch Tasks DB initialization
 - Only MySQL and PostgreSql indexes are supported at the moment

 Resolves #2306


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/TaskConfiguration.java
```diff
@@ -188,6 +191,23 @@ public TaskRepositoryInitializer taskRepositoryInitializerForDB(DataSource dataS
 			return taskRepositoryInitializer;
 		}
 
+		@Bean
+		@DependsOn({"batchRepositoryInitializerForDefaultDB", "taskRepositoryInitializerForDB"})
+		public AbstractDatabaseInitializer taskBatchRdbmsIndexInitializer(DataSource dataSource, ResourceLoader resourceLoader) {
```
markpollack: It is not an abstract class, so the name should be different.  Also, not quite a fan of defining these classes inside an `@Bean` definition, too much coupling.  I'll rename and move to a top level class.

PR #2333, opened 2018-07-19
Add configuration for RedisTemplate

### LINKED ISSUE #2332: Add configuration for RedisTemplate
In the case of the PCF server, we plan to turn off Boot's RedisAutoConfiguration, so still need an explicit bean definition for RedisTemplate.

markpollack:
The code was copied out Boot's RedisAutoConfiguration.

Resolves #2332

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/AnalyticsConfiguration.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/AnalyticsConfiguration.java
@@ -15,6 +15,8 @@
  */
 package org.springframework.cloud.dataflow.server.config.features;
 
+import java.net.UnknownHostException;
+
 import org.springframework.analytics.metrics.AggregateCounterRepository;
 import org.springframework.analytics.metrics.FieldValueCounterRepository;
 import org.springframework.analytics.metrics.redis.RedisAggregateCounterRepository;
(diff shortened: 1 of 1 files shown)


markpollack:
Thought this was needed to solve https://github.com/spring-cloud/spring-cloud-dataflow-server-cloudfoundry/issues/436  but it isn't.  closing

sabbyanandan:
Looks like we need this after all. Without it, I see the following error.

```
2018-07-19T20:52:09.48-0700 [APP/PROC/WEB/0] OUT ***************************
   2018-07-19T20:52:09.48-0700 [APP/PROC/WEB/0] OUT APPLICATION FAILED TO START
   2018-07-19T20:52:09.48-0700 [APP/PROC/WEB/0] OUT ***************************
   2018-07-19T20:52:09.48-0700 [APP/PROC/WEB/0] OUT Description:
   2018-07-19T20:52:09.48-0700 [APP/PROC/WEB/0] OUT A component required a bean named 'redisTemplate' that could not be found.
   2018-07-19T20:52:09.48-0700 [APP/PROC/WEB/0] OUT Action:
```

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/AnalyticsConfiguration.java
```diff
@@ -52,4 +56,24 @@ public FieldValueCounterRepository fieldValueCounterReader(RedisConnectionFactor
 	public AggregateCounterRepository aggregateCounterReader(RedisConnectionFactory redisConnectionFactory) {
 		return new RedisAggregateCounterRepository(redisConnectionFactory, new RetryTemplate());
 	}
+
+	@Bean
```
jvalkeal: I don't understand what these changes do as it looks to be exactly same as what boot auto-config would do?

jvalkeal: Right, if only reason to reintroduce these here is to assume that CF server will disable redis auto-config, it kinda pointless. Then these should be added to CF server as just looking dataflow core these are really pointless here.

sabbyanandan:
All the required changes are specific to CF-Server and it is addressed via spring-cloud/spring-cloud-dataflow-server-cloudfoundry#437. This is not required anymore.

PR #2334, opened 2018-07-21
Document how to access host folders from docker-compose

### LINKED ISSUE #2286: Add docker-compose support to point to host's Maven repository
As a user, I'm attempting to register locally developed custom apps from my local repo and as well from the file system; however, the Docker daemon that's running SCDF/UI doesn't have access to the host directories and likewise the `/m2` folder in the host, too.

More details in SO thread [here](https://stackoverflow.com/questions/51058729/how-to-register-applications). 

**Acceptance:**
- Add support to mount the host directory (`./m2`) by default (refer to a solution [here](https://stackoverflow.com/questions/36234088/how-do-i-point-a-docker-image-to-my-m2-directory-for-running-maven-in-docker-on))
- Consider the mounting ability for both windows/linux users
- Explore options to mount other host directories (as a collection, so user's can register apps via `file://` protocol)
- Update getting-started guide

tzolov:
- Add guidelines for mounting host folders and registering app-starters from them

 Resolves #2286

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/getting-started.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/getting-started.adoc
@@ -239,6 +239,74 @@ Then add the following entries to the `environment` block of the `dataflow-serve
 ----
 +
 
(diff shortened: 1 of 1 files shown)


chrisjs:
merged in 0c55bf

PR #2351, opened 2018-08-03
Use registry name for the app when deploying in skipper mode

### LINKED ISSUE #2350: Stream with App alias fails to deploy
As a user, I'm attempting to deploy a stream with Apps' including aliases, but they fail to deploy with the attached error. 

It appears we are literally using the alias string to resolve the applications from the registry; obviously there won't be any Apps with that name.

**streams:**
image

**errors:**
```
2018-08-02 09:20:52.283 ERROR 27874 --- [io-9393-exec-10] o.s.c.d.s.c.RestControllerAdvice         : Caught exception while handling a request

java.lang.IllegalStateException: The fromPayment:source:1.3.1.RELEASE app is not registered!
	at org.springframework.cloud.dataflow.server.stream.SkipperStreamDeployer.validateAppVersionIsRegistered(SkipperStreamDeployer.java:278) ~[spring-cloud-dataflow-server-core-1.7.0.BUILD-SNAPSHOT.jar!/:1.7.0.BUILD-SNAPSHOT]

....
....

2018-08-02 09:20:52.381 ERROR 27874 --- [nio-9393-exec-5] o.s.c.d.s.c.RestControllerAdvice         : Caught exception while handling a request

java.lang.IllegalStateException: The fromCustomer:source:1.3.1.RELEASE app is not registered!
....
....

2018-08-02 09:20:52.280 ERROR 27874 --- [nio-9393-exec-8] o.s.c.d.s.c.RestControllerAdvice         : Caught exception while handling a request

java.lang.IllegalStateException: The fromOrders:source:1.3.1.RELEASE app is not registered!
```

ilayaperumalg:
- When the stream definition uses `label` for the app definitions, then when deploying the stream always use the `registry` name so that the app registry has the correct app info.
 - Update tests

Resolves #2350

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/stream/SkipperStreamDeployer.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/stream/SkipperStreamDeployer.java
@@ -43,6 +43,7 @@ import org.yaml.snakeyaml.Yaml;
 
 import org.springframework.cloud.dataflow.core.ApplicationType;
 import org.springframework.cloud.dataflow.core.DataFlowPropertyKeys;
+import org.springframework.cloud.dataflow.core.StreamAppDefinition;
 import org.springframework.cloud.dataflow.core.StreamDefinition;
 import org.springframework.cloud.dataflow.core.StreamDeployment;
 import org.springframework.cloud.dataflow.registry.service.AppRegistryService;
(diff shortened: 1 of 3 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/stream/SkipperStreamDeployer.java
```diff
@@ -263,20 +264,40 @@ private String determinePlatformName(final String platformName) {
...
+		String appTypeString = appDeploymentRequest.getDefinition().getProperties()
+				.get(DataFlowPropertyKeys.STREAM_APP_TYPE);
+		ApplicationType applicationType = ApplicationType.valueOf(appTypeString);
+		if (!this.appRegistryService.appExist(registeredAppName, applicationType, appVersion)) {
+			throw new IllegalStateException(String.format("The %s:%s:%s app is not registered!",
+					registeredAppName, appTypeString, appVersion));
+		}
+	}
+
+	public void validateAppVersionIsRegistered(String registeredAppName, AppDeploymentRequest appDeploymentRequest, String appVersion) {
```
tzolov: This method is only used by the private `validateAllAppsRegistered` and should be private as well. Furthermore it can be reused by the `public void validateAppVersionIsRegistered(StreamDefinition streamDefinition, AppDeploymentRequest appDeploymentRequest, String appVersion)`

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/stream/SkipperStreamDeployer.java
```diff
@@ -263,20 +264,40 @@ private String determinePlatformName(final String platformName) {
...
+	private String getRegisteredName(StreamDefinition streamDefinition, String adrAppName) {
+		for (StreamAppDefinition appDefinition: streamDefinition.getAppDefinitions()) {
+			if (appDefinition.getName().equals(adrAppName)) {
+				return appDefinition.getRegisteredAppName();
+			}
+		}
+		return adrAppName;
+	}
+
+	public void validateAppVersionIsRegistered(StreamDefinition streamDefinition, AppDeploymentRequest appDeploymentRequest, String appVersion) {
```
tzolov: Except the first line this method is identical to https://github.com/spring-cloud/spring-cloud-dataflow/pull/2351/files#diff-50f573b367c4c71c1abdc05c961b88f9R294 and can use it internally

tzolov:
Cohesion improved and merged at 042f28df7d5caf291bfb1831215f36398a7719b3

PR #2359, opened 2018-08-08
Re-add testCurrentRunningTasks test to TestCommands

### LINKED ISSUE #2349: TestCommandTests.currentExecution test fails in CI
While this works locally it looks as if it fails during CI build because previous tests are leaving execution data in the table (with now start nor end time) which shows up as a test failure.   We need to either cleanup this data or make sure  it is populated correctly.  https://build.spring.io/browse/SCD-BMASTER-2145
The test is marked for ignore now, but needs to be reactivated.

cppwfs:
Basically make sure that previous tests cleanup after themselves.

resolves #2349


ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed, rebased and merged as a4c2c584

PR #2364, opened 2018-08-09
Changed the task-definition type to LONGVARCHAR

### LINKED ISSUE #2349: TestCommandTests.currentExecution test fails in CI

cppwfs:
Originally the type was a CLOB but when saving over 4000 char Postgres would fail via exception because it does not support clob.
It supports only LONGVARCHAR.   I regression tested this change against:
Oracle, SQLServer, DB2, H2, MYSql  for cases of under 4000 and over 4000.
They all work with the exception of Oracle which is a VARCHAR(4000) when testing over 4000 char.   I'm trying to remember why we used VARCHAR(4000) instead of a CLOB, but there was a reason. So this is expected behavior in this case.

resolves #2349


ilayaperumalg:
The changes from this PR were already merged into master as a4c2c58

PR #2368, opened 2018-08-13
Move stream aggregate state logic into Util

### LINKED ISSUE #2367: Move stream aggregate state calculation logic in Util class
Currently, the stream aggregate state calculation logic is in the `StreamDefinitionController` and this can be in a separate Util class.

ilayaperumalg:
- Refactoring updates

Resolves #2367


jvalkeal:
merged per 16edf43c5a674e2a4f59b015af8387997231db14

Issue #2374, opened 2018-08-14
Design DB schema to support audit-trails


sabbyanandan:
As a developer, I'd like to design, review, and finalize the DB schema to host the audit-trails.

Proposed Audit table includes the following columns:

```
Timestamp
SCDF Server Name
UserID
Action
ActionStatus
correlationId  (something for easy lookup, e.g. stream name, jobName, registered app name)
Action Context - JSON
Error Context - JSON
Anything else we want to pull out of the JSON context..
```

Child of spring-cloud/spring-cloud-dataflow#2369

Issue #2373, opened 2018-08-14
Implement audit-trail advice for StreamService


sabbyanandan:
As a developer, I'd like to add the advice to automatically track and persist who did, what, and when around the `StreamService` APIs.

Child of spring-cloud/spring-cloud-dataflow#2369

PR #2390, opened 2018-08-24
gh-2373 Implement audit-trail feature for StreamService

ghillert:
* gh-2374 Design DB schema to support audit-trails

resolves https://github.com/spring-cloud/spring-cloud-dataflow/issues/2374
resolves https://github.com/spring-cloud/spring-cloud-dataflow/issues/2373


spring-cloud-dataflow-dependencies/pom.xml
```diff
@@ -24,12 +24,12 @@
 			<dependency>
 				<groupId>org.hibernate</groupId>
 				<artifactId>hibernate-core</artifactId>
-				<version>5.2.12.Final</version>
+				<version>5.2.17.Final</version>
```
jvalkeal: This is immediate question of why from 12 to 17?

ghillert:
Rebased

ilayaperumalg:
@ghillert Could you add some descriptive message as commit description so as to help the reviewers (and for future references) to understand what changes are involved etc.,

Thanks!

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/audit/service/DefaultAuditRecordService.java
```diff
@@ -0,0 +1,87 @@
...
+ *
+ */
+public class DefaultAuditRecordService implements AuditRecordService {
+
+	private final AuditRecordRepository auditRecordRepository;
+
+	private final ObjectMapper objectMapper;
+
+	public DefaultAuditRecordService(AuditRecordRepository auditRecordRepository, ObjectMapper objectMapper) {
+		super();
```
ilayaperumalg: Is there a specific reason for doing this?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/audit/service/DefaultAuditRecordService.java
```diff
@@ -0,0 +1,87 @@
...
+
+		final AuditRecord auditRecord = new AuditRecord();
+		auditRecord.setAuditAction(auditActionType);;
+		auditRecord.setAuditOperation(auditOperationType);;
+		auditRecord.setCorrelationId(correlationId);
+
+		try {
+			auditRecord.setAuditData(objectMapper.writeValueAsString(data));
+		} catch (JsonProcessingException e) {
+			throw new IllegalStateException("Error serializing audit record data.", e);
```
ilayaperumalg: Should we proceed with the ongoing operation by logging the error when the audit save encounters error instead of throwing the exception here?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/AbstractStreamService.java
```diff
@@ -154,6 +168,14 @@ else if (DeploymentState.deploying == status) {
 			throw new StreamAlreadyDeployingException(name);
 		}
 		doDeployStream(streamDefinition, deploymentProperties);
+
+		final Map<String, Object> auditedData = new HashMap<>(2);
+		auditedData.put("streamDefinitionDslText", streamDefinition.getDslText());
```
ilayaperumalg: We can use constant names for these fields as they could be useful for any other systems that are built to use the audit data.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/audit/domain/AuditOperationType.java
```diff
@@ -0,0 +1,66 @@
...
+ * @author Gunnar Hillert
+ *
+ */
+public enum AuditOperationType {
+
+	STREAM_DEFINITIONS(100L, "Stream Definitions"),
+	TASK_DEFINITIONS(  200L, "Task Definitions"),
+	APP_REGISTRATIONS( 300L, "App Registrations"),
+	SCHEDULES(         400L, "Schedules");
+
```
markpollack: `stream', `task`, 'app_registration` (or maybe just `registration`), and `schedule` would be more descriptive/appropriate - reads better

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/audit/domain/AuditRecord.java
```diff
@@ -0,0 +1,168 @@
...
+
+	/**
+	 * Automatically populate the server host for auditing purposes.
+	 */
+	@PrePersist
+	@PreUpdate
+	@PreRemove
+	public void populateServerHost() {
+		try {
+			this.serverHost = InetAddress.getLocalHost().getHostName();
```
markpollack: We should discuss the server host field.  I had originally thought of it to be recording the dataflow server name, something that would be easily connected to the users/admin of the system.  Might still be useful to have the hostname, but in the cloud environment, that is often not so useful.  Other fields to consider would be something about the cloud provider, in the case of cloud foundry, the org/space or namespace in k8s.  For skipper deployments we would record the `account` that was being targeted.

markpollack: If we want a host, using the Boot property `spring.cloud.client.hostname` is a bit more robust than the code here.  See https://github.com/spring-cloud/spring-cloud-commons/blob/master/spring-cloud-commons/src/main/java/org/springframework/cloud/client/HostInfoEnvironmentPostProcessor.java

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/audit/repository/AuditRecordRepositoryCustom.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+package org.springframework.cloud.dataflow.server.audit.repository;
+
+import org.springframework.data.repository.NoRepositoryBean;
+
+/**
+ * @author Gunnar Hillert
+ */
+@NoRepositoryBean
+public interface AuditRecordRepositoryCustom {
```
markpollack: this class can be removed, it isn't doing anything

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/audit/repository/AuditRecordRepositoryImpl.java
```diff
@@ -0,0 +1,34 @@
...
+
+import org.springframework.stereotype.Repository;
+
+/**
+*
+* @author Gunnar Hillert
+*
+*/
+@Repository
+public class AuditRecordRepositoryImpl implements AuditRecordRepositoryCustom {
```
markpollack: same here - can be removed.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/audit/service/DefaultAuditRecordService.java
markpollack: Not quite sure what ObjectMapper instance we are getting in the configuration.  In skipper we had cases where we added `objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);` and also `objectMapper.configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false);`  Maybe just have an ObjectMapper instance local to this class?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/audit/service/DefaultAuditRecordService.java
markpollack: Yea, we prob should write the record with the information as best we have it and some sort of entry for the data indicating it could not be saved.  Also some logging message.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/hibernate/GenericEnumUserType.java
```diff
@@ -0,0 +1,165 @@
...
+import java.util.Properties;
+
+import org.hibernate.HibernateException;
+import org.hibernate.engine.spi.SharedSessionContractImplementor;
+import org.hibernate.type.AbstractSingleColumnStandardBasicType;
+import org.hibernate.type.TypeResolver;
+import org.hibernate.usertype.ParameterizedType;
+import org.hibernate.usertype.UserType;
+
+public class GenericEnumUserType implements UserType, ParameterizedType {
```
markpollack: Using a converter to map the enum value would be an easier approach - see https://docs.jboss.org/hibernate/orm/5.0/mappingGuide/en-US/html_single/#d5e678

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/audit/service/DefaultAuditRecordService.java
```diff
@@ -0,0 +1,87 @@
...
+	public DefaultAuditRecordService(AuditRecordRepository auditRecordRepository, ObjectMapper objectMapper) {
+		super();
+		Assert.notNull(auditRecordRepository, "auditRecordRepository must not be null.");
+		Assert.notNull(auditRecordRepository, "objectMapper must not be null.");
+		this.auditRecordRepository = auditRecordRepository;
+		this.objectMapper = objectMapper;
+	}
+
+	@Override
+	public void populateAndSaveAuditRecord(AuditActionType auditActionType, AuditOperationType auditOperationType,
```
markpollack: Small point, but having the operation type first, then the action type reads a bit better to me.

Instead of 
`AuditActionType.CREATE, AuditOperationType.STREAM`

use

`AuditOperationType.STREAM, AuditActionType.CREATE`

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/audit/domain/AuditOperationType.java
ghillert: Done with the latest commit

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/hibernate/GenericEnumUserType.java
ghillert: Done in latest commit

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/AbstractStreamService.java
```diff
@@ -85,6 +85,9 @@
 
 	protected final AuditRecordService auditRecordService;
 
+	public static final String STREAM_DEFINITION_DSL_TEXT = "streamDefinitionDslText";
+	public static final String DEPLOYMENT_PROPERTIES = "streamDefinitionDslText";
```
markpollack: looks to be cut-n-paste error

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/AbstractStreamService.java
markpollack: :+1:

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/audit/domain/AuditRecord.java
markpollack: that code turns out to not be so useful in the context of a domain class.

ghillert:
Merged by @markpollack:

- removed not null checks on data and correlationId
- service now return AuditRecord for use with 'parent-child' entries, such as delete-all and app import.

merged as a61933b23e651099874037222dc207427b3a0e44

PR #2402, opened 2018-08-30
gh-2375 Add a controller to facilitate audit queries

### LINKED ISSUE #2375: Add a controller to facilitate audit queries
As a developer, I'd like to add the controller (maybe: `AuditController`?) to facilitate the audit related queries.

Child of spring-cloud/spring-cloud-dataflow#2369

ghillert:
requires #2373

resolves #2375


ilayaperumalg:
Are we planning to have a feature flag to enable/disable audit service?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AuditController.java
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ * Controller for retrieving {@link AuditRecord}s.
+ *
+ * @author Gunnar Hillert
+ */
+@RestController
+@RequestMapping("/audit")
+@ExposesResourceFor(AuditRecordResource.class)
+public class AuditController {
```
ilayaperumalg: We could probably align with the corresponding service/repo names that start with `AuditRecord` to say,  `AuditRecordController`

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AuditController.java
```diff
@@ -0,0 +1,123 @@
...
+	 */
+	public AuditController(AuditRecordService auditRecordService) {
+		Assert.notNull(auditRecordService, "AuditRecordService must not be null");
+		this.auditRecordService = auditRecordService;
+	}
+
+	/**
+	 * Return a page-able list of {@link AuditRecordResource}s.
+	 *
+	 * @param pageable Pagination information
```
ilayaperumalg: Params not in order

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AuditController.java
```diff
@@ -0,0 +1,123 @@
...
+	 * @param assembler assembler for {@link AuditRecord}
+	 * @param action Optional. For which {@link AuditActionType} do you want to retrieve {@link AuditRecord}s
+	 * @param operation Optional. For which {@link AuditOperationType} do you want to retrieve {@link AuditRecord}s
+	 * @return list of audit records
```
ilayaperumalg: Would be better if we mention as, list of `Pageable` audit records.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AuditController.java
```diff
@@ -0,0 +1,123 @@
...
+			super(AuditController.class, AuditRecordResource.class);
+		}
+
+		@Override
+		public AuditRecordResource toResource(AuditRecord auditRecord) {
+			try {
+				return createResourceWithId(auditRecord.getId(), auditRecord);
+			}
+			catch (IllegalStateException e) {
+				logger.warn("Failed to create StreamDefinitionResource. " + e.getMessage());
```
ilayaperumalg: StreamDefinitionResource -> AuditRecordResource

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/hibernate/GenericEnumUserType.java
```diff
@@ -0,0 +1,165 @@
...
+import java.util.Properties;
+
+import org.hibernate.HibernateException;
+import org.hibernate.engine.spi.SharedSessionContractImplementor;
+import org.hibernate.type.AbstractSingleColumnStandardBasicType;
+import org.hibernate.type.TypeResolver;
+import org.hibernate.usertype.ParameterizedType;
+import org.hibernate.usertype.UserType;
+
+public class GenericEnumUserType implements UserType, ParameterizedType {
```
ilayaperumalg: Is this our own class or from a library? We need to give `author` credit if it is from outside :-)

markpollack:
We should use Spring Data REST instead of writing this boilerplate.  There are a few examples in skipper, eg. to not expose Create, Update, Delete methods, and only finders.

markpollack:
I take it back, the controller as it is now would be what is needed to help support a UI where there are multiple drop down boxes, the UI just passes the criteria off to the server and the server figures it out.

markpollack:
removed not null checks on data and correlationId
service now return AuditRecord for use with 'parent-child' entries, such as delete-all and app import.

merged as a61933b23e651099874037222dc207427b3a0e44

markpollack:
removed not null checks on data and correlationId
service now return AuditRecord for use with 'parent-child' entries, such as delete-all and app import.

merged as a61933b23e651099874037222dc207427b3a0e44

ghillert:
@sabbyanandan @markpollack Updated PR:

- Add ability to query for audit records using `AuditActionType` and `AuditOperationType`. Both parameters are optional. Multiple comma-separated values can be used
- Improve Rest Docs

PR #2410, opened 2018-09-04
Add security to schedule restful endpoints

### LINKED ISSUE #2404: Update dataflow-server-defaults.yml to include /tasks/schedules
Looks like `dataflow-server-defaults.yml` needs to be updated with the following:

```
            # Task Schedules

            - GET    /tasks/schedules                => hasRole('ROLE_VIEW')
            - GET    /tasks/schedules/*              => hasRole('ROLE_VIEW')
            - POST   /tasks/schedules                => hasRole('ROLE_CREATE')
            - DELETE /tasks/schedules/*              => hasRole('ROLE_CREATE')
```

and any corresponding tests. Otherwise when logged into the dashboard, a 403 error occurs when clicking on the schedules tab

cppwfs:
resolves #2404

Tested against SCDF-CF-Server

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/configuration.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/configuration.adoc
@@ -803,6 +803,13 @@ spring:
             - GET    /tasks/executions/*             => hasRole('ROLE_VIEW')
             - POST   /tasks/executions               => hasRole('ROLE_CREATE')
             - DELETE /tasks/executions/*             => hasRole('ROLE_CREATE')
+
+            # Task Schedules
+
+            - GET    /tasks/schedules                => hasRole('ROLE_VIEW')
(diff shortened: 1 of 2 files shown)


chrisjs:
two things:

* in `configuration.adoc` we have a section of `Customizing Authentication` that indicates `The default rules are as follows:` - for consistency, seems that section should be updated as well with these paths

* i see there are tests in the local server security package for example `LocalServerSecurityRootWithUsersFileTestsWithTasksAndWithoutStreamsMetricsAndSchedules` - might it be worth doing the same but to assert on schedule path auth? i didn't see it covered elsewhere

cppwfs:
updated the configuration.adoc as requested.  Also created story #2417 to add tests for Scheduler Security endpoints when a LocalScheduler IMPL is available.

chrisjs:
merged in 45372e9

chrisjs:
this has also been back ported to 1.6.x per the issue description and label added here

PR #2415, opened 2018-09-05
Add Skipper stream force update

### LINKED ISSUE #2412: Add support for force stream update
With Skipper introducing https://github.com/spring-cloud/spring-cloud-skipper/issues/714 we need to have stream update operation with the `force` operation.

### LINKED ISSUE #2414: Upgrade Skipper to 1.1.0 snapshot for SCDF 1.7.x

ilayaperumalg:
- Use skipper `force` update functionality
 - Set upgrade request with `force` and `appNames` options
 - Update test

Resolves #2412
Resolves #2414

### DIFF
--- a/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/UpdateStreamRequest.java
+++ b/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/UpdateStreamRequest.java
@@ -21,7 +21,9 @@ import org.springframework.cloud.skipper.domain.PackageIdentifier;
 
 /**
  * Captures the required data for updating a stream using Skipper.
+ *
  * @author Mark Pollack
+ * @author Ilayaperumal Gopinathan
  */
 public class UpdateStreamRequest {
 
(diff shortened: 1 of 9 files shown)


spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/UpdateStreamRequest.java
```diff
@@ -31,15 +33,26 @@
 
 	private Map<String, String> updateProperties;
 
+	private boolean force;
+
+	private String[] appNames;
```
markpollack: Switch to List instead?

spring-cloud-dataflow-skipper-docs/pom.xml
```diff
@@ -39,6 +39,12 @@
 			<artifactId>json-path</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>io.projectreactor</groupId>
+			<artifactId>reactor-core</artifactId>
+			<version>3.1.8.RELEASE</version>
+			<scope>test</scope>
+		</dependency>
```
markpollack: why is this here?

markpollack:
rebase, changed String[] to List<String>

merged as 8cdc072373368a52c5d78bf0c6eae02da983916c

spring-cloud-dataflow-skipper-docs/pom.xml
ilayaperumalg: This was needed to have Skipper related doc tests using the specific reactor version that is used by Skipper 1.1.0.

PR #2416, opened 2018-09-05
Client Evaluates EXECUTIONS_CURRENT_RELATION if version 1.7 and above

cppwfs:
TaskTemplate (client) extracts the numeric portion of the version number as reported by the AboutOperation
if the version is 1.7 or above then the EXECUTIONS_CURRENT_RELATION will be tested, else it will be skipped.

resolves SCDF-2401

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -54,6 +54,7 @@
 		<jacoco-maven-plugin.version>0.7.9</jacoco-maven-plugin.version>
 		<sonar-maven-plugin.version>3.0.2</sonar-maven-plugin.version>
 		<checkstyle.config.location>src/checkstyle/checkstyle.xml</checkstyle.config.location>
+		<semver.version>2.2.0</semver.version>
 	</properties>
 	<modules>
 		<module>spring-cloud-dataflow-configuration-metadata</module>
--- a/spring-cloud-dataflow-rest-client/pom.xml
+++ b/spring-cloud-dataflow-rest-client/pom.xml
@@ -53,5 +53,10 @@
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-dataflow-rest-resource</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>com.vdurmont</groupId>
+			<artifactId>semver4j</artifactId>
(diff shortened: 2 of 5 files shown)


markpollack:
There is a very small library, which the cf-java-client uses, that can remove the code which is parsing out the version numbers and doing the comparison.

https://github.com/zafarkhaja/jsemver#comparing-versions

```
        <dependency>
            <groupId>com.github.zafarkhaja</groupId>
            <artifactId>java-semver</artifactId>
            <version>0.9.0</version>
        </dependency>
```

Switching to use that library means we don't have to really go crazy testing the logic added in this commit and also allows us to reuse the testing of versions in the future for other parts of the API that may require such tests (ATM it is all bundled up inside TaskTemplate. 

Otherwise looks good.

markpollack:
Though, after trying on `1.7.0.RELEASE` it barfs, so one would have to strip the last part if it exists.  There is this issue that hasn't been addressed. https://github.com/zafarkhaja/jsemver/issues/40

markpollack:
https://github.com/vdurmont/semver4j is more active.  I had a chat a while back with Ben Hale about his choice in library (when doing skipper) since the other library seems abandoned.  it was active at the time and apparently there was not need to change.  The semver4j works well with code like this. 

```
	@Test
	public void testSemVer() {
		Semver minRequiredVersion = new Semver("1.7.0");
		Semver currentVersion = new Semver("1.6.0");
		Assert.assertTrue(currentVersion.isLowerThan(minRequiredVersion));
		currentVersion = new Semver("1.7.1");
		Assert.assertTrue(currentVersion.isGreaterThan(minRequiredVersion));
	}
```
Either way, I think we will have to strip off the 4th place in the version to do the comparison due to 'build-snapshot' and 'release' not really being something these libraries understand.  Initializr has a library that does understand those spring conventions, but it pulls in too much.

cppwfs:
Updated based on code review & rebased

markpollack:
Added a VersionUtils method for reuse in other cases and added tests.

merged as 91dd1d1d04989a3920a60880b930eba74ba2049a

PR #2421, opened 2018-09-07
Remove download log from AppDeploymentRequestCreator

### LINKED ISSUE #2409: Investigate whether the Task is downloaded more than once at the time of launch

ilayaperumalg:
- The retrieval of app resource isn't downloading the resource and needs to be removed.

Resolves #2409


markpollack:
I don't think this is related to the behavior in the issue.  This classes is only used by DefaultSkipperStreamService.  The issue with double downloads was observed in tasks.

ilayaperumalg:
@markpollack Yes, I couldn't reproduce this for the `task` as we don't log anything for task service in terms of downloading and the only place we have such a logging is in this Skipper service. Please see my earlier comment [here](https://github.com/spring-cloud/spring-cloud-dataflow/pull/2421#issuecomment-419534914)

tzolov:
rebased and merged at 87548508b348dece30058cc020ecacbe4d16847f

PR #2428, opened 2018-09-11
Retrieves only white-listed properties by default

### LINKED ISSUE #2424: Add support to retrieve only the whitelisted properties
As a continuation of spring-cloud/spring-cloud-dataflow-ui#879, we would want to improve Core to only return whitelisted properties. 

If there isn't a metadata artifact for an App, we won't go through and populate all the 1000+ Boot properties in the REST response. 

It will improve the performance, too.

tzolov:
Make the apps info return only the white listed properties by default. If no white listed properties are found an empty list is returned.
To return all (Spring Boot) properties set the exhaustive property to true.

  - Make MetadataResolver to always obey the exhaustive property.
  - Add an optional 'exhaustive' request parameter to the [Skipper]AppRegistryController#info (e.g /apps/{type}/{name}?exhaustive= )
  - Add an optional 'exhaustive' parameter to the SCDF shell app info command (both classic and skipper modes). Defaults to false.
  - Add tests to the MetadataResolver, [Skipper]AppRegistryControlle and Shell
  - Update the AppRegistry get info REST docs.

  Resolves #2424


ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Rebased and merged as 4eca3a50

PR #2437, opened 2018-09-14
Make client validation ops optional when connecting to older servers

### LINKED ISSUE #2436: Make validation links option in DataFlowTemplate
Throw UnsupportedOperationException if API is called.

markpollack:
Fixes #2436


markpollack:
I am self merging this so that the ATs can pass.

PR #2439, opened 2018-09-18
Command line args not passed to ScheduleRequest

### LINKED ISSUE #2405: Command line arguments not passed to ScheduleRequest
When on the "Create Schedules" page in the dashboard, I can add a key and value to the "Task Arguments" but that value is not passed to the ScheduleRequest.commandLineArguments.

But If you create the task with say `timestamp --timestamp.format=YYYY` it will be mapped as a property in the AppDefinition so that is working, just a note..

chrisjs:
Resolves #2405


cppwfs:
LGTM Merged.

PR #2465, opened 2018-10-03
Encapsulate the Resource loading and related states.

### LINKED ISSUE #1873: Encapsulate the ResourceUtils internals into a clean, stateful ResourceResolver class
The `ResourceUtils` static utils (after #1843) exposing internals such as MavenProperties and ResourceLoaders instances through all applications layers. Later are used only to be passed as arguments to the static ResourceUtils methods. 
Instead a ResourceService @service should encapsulate the resource loading access and hide all the internals

depends on #1843

tzolov:
- Introduce a stateful AppResourceCommon that includes the ResourceUtils methods along with MavenProperties and ResourceLoader instances.
   - Remove the direct usage of MavenProperties and ResourceLoader across the classes and make use of the AppResourceCommon instead.
   - Update the bean configurations to instantiate AppResourceCommon bean and inject it in all beans that operate with Resources.
   - Remove ResourceUtils.
   - Fix affected tests.

  Resolves #1873

### DIFF
--- a/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/AppRegistryCommon.java
+++ b/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/AppRegistryCommon.java
@@ -95,4 +95,25 @@ public interface AppRegistryCommon {
 	 * @return the saved appRegistration
 	 */
 	AppRegistration save(AppRegistration app);
+
+	/**
+	 * @param resource
+	 * @return Returns the version for the provided resource
+	 */
+	String getResourceVersion(Resource resource);
+
+	/**
+	 * TODO
+	 * @param resource
+	 * @return
+	 */
+	String getResourceWithoutVersion(Resource resource);
+
+	/**
+	 * Returns the version for the given resource URI string.
+	 *
+	 * @param uriString String representation of the resource URI
+	 * @return the resource version
+	 */
+	String getResourceVersion(String uriString);
 }
--- a/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/AppRegistryCommon.java
+++ b/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/AppRegistryCommon.java
@@ -97,21 +97,20 @@ public interface AppRegistryCommon {
 	AppRegistration save(AppRegistration app);
 
 	/**
-	 * @param resource
+	 * @param resource to retrieve the version for
 	 * @return Returns the version for the provided resource
 	 */
 	String getResourceVersion(Resource resource);
 
 	/**
-	 * TODO
-	 * @param resource
-	 * @return
+	 * Returns a string representing the resource with version subtracted
+	 * @param resource to be represented as string.
+	 * @return String representation of the resource.
 	 */
 	String getResourceWithoutVersion(Resource resource);
 
 	/**
 	 * Returns the version for the given resource URI string.
-	 *
 	 * @param uriString String representation of the resource URI
 	 * @return the resource version
 	 */
--- a/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/support/AppResourceCommon.java
+++ b/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/support/AppResourceCommon.java
@@ -45,14 +45,15 @@ public class AppResourceCommon {
 	private MavenProperties mavenProperties;
 
 	/**
-	 *
+	 * Delegated resource loader for resolving metadata from the metadata URI
 	 */
-	private ResourceLoader resourceLoader;
+	private ResourceLoader metadataResourceLoader;
 
 
 	public AppResourceCommon(MavenProperties mavenProperties, ResourceLoader resourceLoader) {
+		Assert.notNull(mavenProperties, "Non null Maven Properties are required!");
 		this.mavenProperties = mavenProperties;
-		this.resourceLoader = resourceLoader;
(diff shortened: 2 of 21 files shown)


spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/support/AppResourceCommon.java
```diff
@@ -34,10 +35,66 @@
...
-public class ResourceUtils {
+public class AppResourceCommon {
+
+	/**
+	 * the maven properties to use in case of maven resource
+	 */
+	private MavenProperties mavenProperties;
+
+	/**
+	 *
```
jvalkeal: Maybe just remove empty doc block

spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/support/AppResourceCommon.java
```diff
@@ -34,10 +35,66 @@
...
+	 */
+	private ResourceLoader resourceLoader;
+
+
+	public AppResourceCommon(MavenProperties mavenProperties, ResourceLoader resourceLoader) {
```
jvalkeal: Could you add null asserts for props and loader?

spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/support/AppResourceCommon.java
```diff
@@ -121,71 +170,20 @@ else if (scheme.equals("docker")) {
 				else {
 					resourceLoader = new DownloadingUrlResourceLoader();
 				}
-				return resourceLoader.getResource(uriString);
+				return resourceLoader.getResource(resourceUri);
```
jvalkeal: Not a biggie but could you rename `resourceLoader` used locally here as same symbol `resourceLoader` is already used as a class field. Way too easy to mix when you're just eyeballing this section.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowControllerAutoConfiguration.java
```diff
@@ -450,8 +456,8 @@ public SkipperStreamService skipperStreamDeploymentService(
 
 		@Bean
 		public AppRegistryService appRegistryService(AppRegistrationRepository appRegistrationRepository,
-				DelegatingResourceLoader resourceLoader, MavenProperties mavenProperties) {
-			return new DefaultAppRegistryService(appRegistrationRepository, resourceLoader, mavenProperties);
+				AppResourceCommon appResourceService) {
```
jvalkeal: Should we actually rename `AppResourceCommon` as `AppResourceService` if it's used as a service?

spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/AppRegistryCommon.java
```diff
@@ -95,4 +95,25 @@ default AppRegistration find(String name, ApplicationType type, String version)
...
 	AppRegistration save(AppRegistration app);
+
+	/**
+	 * @param resource
+	 * @return Returns the version for the provided resource
+	 */
+	String getResourceVersion(Resource resource);
+
+	/**
+	 * TODO
```
jvalkeal: TODO what to do? ;)

jvalkeal:
@tzolov When you get to this, please rebase.

spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/support/AppResourceCommon.java
tzolov: Ok will try to infer some description

tzolov: also it seems that `metadataResourceLoader` is better name for this field

spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/support/AppResourceCommon.java
tzolov: I will for the props, but the (metadata)resourceLoader can actually be null for Docker apps

spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/support/AppResourceCommon.java
tzolov: +1

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowControllerAutoConfiguration.java
tzolov: Actually i did just opposite. It use to be called AppResourceService but it seems to have been used mostly as stateful util so @markpollack suggested to common as more descriptive term. I will rename the argument names

tzolov:
@jvalkeal i've pushed a commit that addresses the issues and rebases to the upstream/master

jvalkeal:
Think this looks good now as I really didn't have anything else to nitpick.

jvalkeal:
Merged per e349f827f8b78be26bef5f8e9bc37e8da7bb5830

PR #2470, opened 2018-10-04
Update docs to support Skipper in Docker Compose

### LINKED ISSUE #2270: Create a docker-compose.yml file that includes skipper
May also want to include the metrics server.

chrisjs:
Resolves #2270


chrisjs:
* Skipper added to customization section rather than a separate file / stays consistent with other customizations to the base compose setup

* Removed deprecated `depends_on` tags and other minor cleanups

* Health checks not added to keep the "quick start" content simplified in terms of docs and possibly extra scripts. I have not experienced issues nor have we had any reported yet -- may just be the case services are ready before apps access them. Will revisit a more baked in approach in https://github.com/spring-cloud/spring-cloud-dataflow-acceptance-tests/issues/195

ilayaperumalg:
This is merged as 98b9d51a2e6f659364ffcc98aca211b1ab5b5ca5

Issue #2461, opened 2018-10-02
Audit-trail: Mask sensitive properties for Streams


sabbyanandan:

ghillert:
Depends on https://github.com/spring-cloud/spring-cloud-dataflow/issues/2469

Issue #2469, opened 2018-10-03
Add TaskDefinition support to `ArgumentSanitizer`


ghillert:
On the task side the sanitization support is mostly not there, only `TaskExecutions` are sanitized in the `TaskExecutionController`.

We also need:

- Add support to sanitize Task Arguments to `ArgumentSanitizer`
- Add a TaskDefinition-specific `dslConverter`, similar to `StreamDefinitionToDslConverter`
- Add unit tests

Ensure sanitization works for composed tasks.

See also gh-2461

sabbyanandan:
Addressed in https://github.com/spring-cloud/spring-cloud-dataflow/commit/d679ea60fcc965b6594f8ccb5335d067504ad984 and https://github.com/spring-cloud/spring-cloud-dataflow/commit/120d10b55cb59a3f87e09a3871c61ebc189598ed.

PR #2472, opened 2018-10-05
gh-2461 Audit-trail: Mask sensitive properties for Streams

ghillert:
* Add functionality to the Stream services classes to mask sensitive properties in Streams
* Add tests to StreamController
* Add some stubs in preparation for gh-2469 to ArgumentSanitizer

resolves https://github.com/spring-cloud/spring-cloud-dataflow/issues/2461

### DIFF
--- a/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
+++ b/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
@@ -22,6 +22,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Enumeration;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -44,6 +45,7 @@ import org.springframework.util.StringUtils;
  * @author Mark Fisher
  * @author Janne Valkealahti
  * @author Christian Tzolov
+ * @author Gunnar Hillert
  */
 public final class DeploymentPropertiesUtils {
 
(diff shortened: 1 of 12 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/AbstractStreamService.java
```diff
@@ -118,14 +122,15 @@ public StreamDefinition createStream(String streamName, String dsl, boolean depl
...
+		final StreamDefinition savedStreamDefintion = this.streamDefinitionRepository.save(streamDefinition);
+
 		if (deploy) {
 			this.deployStream(streamName, new HashMap<>());
 		}
 
 		auditRecordService.populateAndSaveAuditRecord(
 				AuditOperationType.STREAM, AuditActionType.CREATE,
-				streamDefinition.getName(), streamDefinition.getDslText());
+				streamDefinition.getName(), this.argumentSanitizer.sanitizeStream(savedStreamDefintion));
```
markpollack: I think that the `ArgumentSanitizer` should be pushed down into the `AuditRecordService` to encapsulate it better.  It would also remove duplication between the Skipper and non-Skipper StreamServices.

markpollack: Will address in #2477.  Merging this for now to avoid extra conflicts with adding argument sanitization for Tasks.

markpollack: merged as 0d207b945190fe8913078c5b06f66b343b9d6514

ghillert:
Was merged with commit 0d207b945190fe8913078c5b06f66b343b9d6514

PR #2473, opened 2018-10-05
Add support for wildcards in stream destination names

### LINKED ISSUE #2425: Add support for additional ASCII characters in named-destination
As a user, I'd like to be able to publish or consume data from named-destinations (i.e., channels) with different ASCII characters as delimiters. 

Currently, there's a strict contract (see [tokenizer.ts](https://github.com/spring-cloud/spring-cloud-dataflow-ui/blob/6d3dc92ca9bfa1383a31bd95484230229a4822e2/ui/src/app/shared/services/tokenizer.ts)) in place and only alpha-numeric, `$`, `.`, `-`,  and `_` are valid delimiters. 

However, I'd like to consume from all the topics w/ the prefix `foooz`, the data pipeline would look like:

> foo = :foooz* | filter | cassandra

Likewise, I'd like to publish data to a topic w/ wildcard regex (depends on the binders that support it) such as:

> bar = :foooz* | filter > :bar*fooz 

There are also cases for when a certain message brokers (e.g., Solace) that can have topics with `/` in it.

> baz = :fooz/baz > hdfs

In summary, it'd be nice to relax the DSL parsers (frontend & backend) to accept `*`, `/`, and `#` as valid delimiters.

### LINKED ISSUE #2430: need some stream tests that have comma at end of options

aclement:
This adds support in the parser/tokenizer to allow stream destination
names to include slash(/), star(*) or hash(#). Extra tests added and
minor refactoring (simplification) to DestinationNode.

Fixes #2425 
Fixes #2430

### DIFF
--- a/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/DSLMessage.java
+++ b/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/DSLMessage.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2017 the original author or authors.
+ * Copyright 2015-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


jschabowsky:
Is there no way to support '>' long term as well? In Solace we support '/' as the topic delimiter, '*' as a level wildcard and '>' as any level wildcard... Example you have a topic structure as: type/order/location... you could subscribe to beer/*/saltlakecity if you wanted all beer orders for saltlakecity or beer/> if you wanted all things under the beer level. I know '>' used just like the '|' but seems like there should be a way support though escaping? Thanks!

PR #2483, opened 2018-10-09
Schedules with an invalid cron expression will return a 400

cppwfs:
Before it would return a 500 error.
Primary change is that all schedule that fail to create should throw a CreateScheduleException as the platform validates the cron expression.


jvalkeal:
LGTM. Usually I'd test this manually but that's slightly diffcult without building everything for cf. Code should do the trick. Merging after my build goes through.

Issue #2370, opened 2018-08-14
Implement audit-trail advice for SchedulerService


sabbyanandan:
As a developer, I'd like to add the advice to automatically track and persist who did, what, and when around the `SchedulerService` APIs.

Child of spring-cloud/spring-cloud-dataflow#2369

PR #2485, opened 2018-10-10
gh-2461 Implement audit-trail advice for SchedulerService

ghillert:
* Add tests
* Fix properties ordering issue in `ArgumentSanitizer`
* Add helper method to `ArgumentSanitizer`

resolves https://github.com/spring-cloud/spring-cloud-dataflow/issues/2370
resolves https://github.com/spring-cloud/spring-cloud-dataflow/issues/2477

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/audit/service/DefaultAuditRecordService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/audit/service/DefaultAuditRecordService.java
@@ -15,6 +15,7 @@
  */
 package org.springframework.cloud.dataflow.server.audit.service;
 
+import java.util.HashMap;
 import java.util.Map;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
@@ -27,6 +28,9 @@ import org.springframework.cloud.dataflow.server.audit.domain.AuditActionType;
 import org.springframework.cloud.dataflow.server.audit.domain.AuditOperationType;
 import org.springframework.cloud.dataflow.server.audit.domain.AuditRecord;
 import org.springframework.cloud.dataflow.server.audit.repository.AuditRecordRepository;
+import org.springframework.cloud.dataflow.server.controller.support.ArgumentSanitizer;
+import org.springframework.cloud.scheduler.spi.core.ScheduleInfo;
+import org.springframework.cloud.scheduler.spi.core.ScheduleRequest;
 import org.springframework.data.domain.Page;
 import org.springframework.data.domain.Pageable;
 import org.springframework.util.Assert;
@@ -39,16 +43,26 @@ import org.springframework.util.Assert;
  */
 public class DefaultAuditRecordService implements AuditRecordService {
 
+	public static final String STREAM_DEFINITION_DSL_TEXT = "streamDefinitionDslText";
+	public static final String TASK_DEFINITION_DSL_TEXT = "taskDefinitionDslText";
+	public static final String TASK_DEFINITION_NAME = "taskDefinitionName";
+	public static final String TASK_DEFINITION_PROPERTIES = "taskDefinitionProperties";
+
+	public static final String DEPLOYMENT_PROPERTIES = "deploymentProperties";
+	public static final String COMMANDLINE_ARGUMENTS = "commandlineArguments";
+
 	private static final Logger logger = LoggerFactory.getLogger(DefaultAuditRecordService.class);
 
 	private final AuditRecordRepository auditRecordRepository;
 
 	private final ObjectMapper objectMapper;
+	private final ArgumentSanitizer argumentSanitizer;
 
 	public DefaultAuditRecordService(AuditRecordRepository auditRecordRepository) {
 		Assert.notNull(auditRecordRepository, "auditRecordRepository must not be null.");
 		this.auditRecordRepository = auditRecordRepository;
 		this.objectMapper = new ObjectMapper();
+		this.argumentSanitizer = new ArgumentSanitizer();
 		this.objectMapper.configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false);
 	}
 
@@ -56,6 +70,7 @@ public class DefaultAuditRecordService implements AuditRecordService {
 		Assert.notNull(auditRecordRepository, "auditRecordRepository must not be null.");
 		Assert.notNull(objectMapper, "objectMapper must not be null.");
 		this.auditRecordRepository = auditRecordRepository;
+		this.argumentSanitizer = new ArgumentSanitizer();
 		this.objectMapper = objectMapper;
 	}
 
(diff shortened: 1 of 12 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/audit/service/DefaultAuditRecordService.java
```diff
@@ -113,4 +128,35 @@ else if (actions != null && operations != null) {
 	public AuditRecord findOne(Long id) {
 		return this.auditRecordRepository.findOne(id);
 	}
+
+	public void recordScheduleCreate(ScheduleRequest scheduleRequest) {
+		Assert.notNull(scheduleRequest, "scheduleRequest must not be null");
+		Assert.hasText(scheduleRequest.getScheduleName(), "The scheduleName of the scheduleRequest must not be null or empty");
+		Assert.notNull(scheduleRequest.getDefinition(), "The appDefinition of the scheduleRequest must not be null");
```
ilayaperumalg: Might be better to say, `taskDefinition` instead of `appDefinition`

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/audit/service/DefaultAuditRecordService.java
```diff
@@ -113,4 +128,35 @@ else if (actions != null && operations != null) {
 	public AuditRecord findOne(Long id) {
 		return this.auditRecordRepository.findOne(id);
 	}
+
+	public void recordScheduleCreate(ScheduleRequest scheduleRequest) {
```
ilayaperumalg: Missing `@Override`

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/support/ArgumentSanitizer.java
```diff
@@ -95,10 +99,12 @@ public String sanitize(String argument) {
 	 * @return the argument with a potentially sanitized value
 	 */
 	public String sanitize(String key, String value) {
-		for (Pattern pattern : this.keysToSanitize) {
-			if (pattern.matcher(key).matches()) {
-				value = REDACTION_STRING;
-				break;
+		if (value != null && StringUtils.hasText(value)) {
```
ilayaperumalg: `value ! = null` can be removed as `StringUtils.hasText(value)` does the non null check as well.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/support/ArgumentSanitizer.java
```diff
@@ -140,10 +146,28 @@ public String sanitizeTask(TaskDefinition taskDefinition) {
 	 */
 	public Map<String, String> sanitizeProperties(Map<String, String> properties) {
 		if (!CollectionUtils.isEmpty(properties)) {
-			return properties.entrySet().stream()
-					.collect(Collectors.toMap(e -> e.getKey(), e -> this.sanitize(e.getKey(), e.getValue())));
+			final Map<String, String> sanitizedProperties = new LinkedHashMap<>(properties.size());
```
ilayaperumalg: Any specific reason for this change to move from `stream()` as I don't seem to find the difference.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/audit/service/DefaultAuditRecordService.java
```diff
@@ -39,23 +43,34 @@
  */
 public class DefaultAuditRecordService implements AuditRecordService {
 
+	public static final String STREAM_DEFINITION_DSL_TEXT = "streamDefinitionDslText";
```
ilayaperumalg: It would be better to move all these constants into `AuditRecordService` as that would avoid having these constants recreated again in some other classes.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultSchedulerService.java
```diff
@@ -137,11 +142,19 @@ public void schedule(String scheduleName, String taskDefinitionName, Map<String,
 		ScheduleRequest scheduleRequest = new ScheduleRequest(revisedDefinition, taskDeploymentProperties,
 				deployerDeploymentProperties, commandLineArgs, scheduleName, getTaskResource(taskDefinitionName));
 		this.scheduler.schedule(scheduleRequest);
+		this.auditRecordService.recordScheduleCreate(scheduleRequest);
 	}
+	public static final String STREAM_DEFINITION_DSL_TEXT = "streamDefinitionDslText";
```
ilayaperumalg: Please see https://github.com/spring-cloud/spring-cloud-dataflow/pull/2485/files#r223937332

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/audit/service/DefaultAuditRecordService.java
markpollack: This is going to happen as part of larger refactoring https://github.com/spring-cloud/spring-cloud-dataflow/issues/2477

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/support/ArgumentSanitizer.java
ghillert: Support for null values :-) See https://stackoverflow.com/questions/24630963/java-8-nullpointerexception-in-collectors-tomap

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/audit/service/DefaultAuditRecordService.java
ghillert: Yes - will be done as part of https://github.com/spring-cloud/spring-cloud-dataflow/issues/2477 I wanted to do it as part of this PR but as it got a big bigger...

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultSchedulerService.java
ghillert: Yes - will be done as part of https://github.com/spring-cloud/spring-cloud-dataflow/issues/2477 I wanted to do it as part of this PR but as it got a big bigger...

ilayaperumalg:
@ghillert Given this PR now addresses #2477, I don't see the changes related to it. Maybe those changes are yet to be updated with this PR?

markpollack:
Maybe there is a commit missing?  Things that should be improved is to encapsulate the `argumentSanitizer` inside the `auditRecordService` and also to remove the `recordScheduleDelete` and `recordScheduleCreate` methods in `AuditRecordService` so that the audit service remains generic. The conversion of the map data can be done by a utility class in the audit service package.

Issue #2457, opened 2018-09-28
Auditing: Polish DB Schema


ghillert:
Add:

- DB indexes: `actionType`, `operationType`, `correlation_id`, `created_time`
- Constraints: `auditOperation `, `auditAction ` (e.g. not-null)
- add a column: `parent_correlation_id` (nullable)

to relevant schema fields

tzolov:
It has been agreed to drop the `parent_correlation_id` addition from this issue.

tzolov:
I have manually tested it with `MySQL`
```sql
mysql> show index from AUDIT_RECORDS;
| Table         | Non_unique | Key_name                          | Seq_in_index | Column_name     | Collation | Cardinality | Sub_part | Packed | Null | Index_type | Comment | Index_comment | Visible |
| AUDIT_RECORDS |          0 | PRIMARY                           |            1 | id              | A         |           0 |     NULL |   NULL |      | BTREE      |         |               | YES     |
| AUDIT_RECORDS |          1 | AUDIT_RECORDS_AUDIT_ACTION_IDX    |            1 | audit_Action    | A         |           0 |     NULL |   NULL |      | BTREE      |         |               | YES     |
| AUDIT_RECORDS |          1 | AUDIT_RECORDS_AUDIT_OPERATION_IDX |            1 | audit_Operation | A         |           0 |     NULL |   NULL |      | BTREE      |         |               | YES     |
| AUDIT_RECORDS |          1 | AUDIT_RECORDS_CORRELATION_ID_IDX  |            1 | correlation_id  | A         |           0 |     NULL |   NULL | YES  | BTREE      |         |               | YES     |
| AUDIT_RECORDS |          1 | AUDIT_RECORDS_CREATED_ON_IDX      |            1 | created_On      | A         |           0 |     NULL |   NULL | YES  | BTREE      |         |               | YES     |

mysql> show create table AUDIT_RECORDS;
| Table         | Create Table                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    | AUDIT_RECORDS | CREATE TABLE `AUDIT_RECORDS` (
  `id` bigint(20) NOT NULL,
  `audit_Action` bigint(20) NOT NULL,
  `audit_data` longtext,
  `audit_Operation` bigint(20) NOT NULL,
  `correlation_id` varchar(255) DEFAULT NULL,
  `created_by` varchar(255) DEFAULT NULL,
[...]

PR #2502, opened 2018-10-17
Revert the schema upper case changes done in #2457

tzolov:
Revert the uppercase jpa schema(s) done in the 960c8599358a691c3ff2786e58df485732015ea5 

Related to #2457


jvalkeal:
Yeah, think this is the most safest bet we can go with for now.

PR #2530, opened 2018-11-01
Format section headers with bold-text

sabbyanandan:
It would be better to have the section headers standout from the body. This change attempts to bring that separation.


cppwfs:
LGTM

cppwfs:
Rebased, Merged.

PR #2546, opened 2018-11-05
Remove explicit dep definition

### LINKED ISSUE #2537: Check reason for adding spring-cloud-common-security-config-web to root pom.xml
Child of #914

jvalkeal:
- Looks like it's not needed anymore to re-define
  spring-cloud-common-security-config-web in a main pom. I believe
  trouble during a boot2 works was that dependencies we horrible
  broken some still being on boot1, etc.
- Maven dep tree is now same when this is removed from pom.
- Fixes #2537


ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Rebased and merged as 58a67ce6

PR #2551, opened 2018-11-06
Revisit use of Optional in repositories

### LINKED ISSUE #2539: Adopt to Spring Data API changes
Support use of Optional in return values in Service Layer classes that delegate to Spring Data Repositories.  e.g. DefaultAuditRecordService.java

Child of #914

jvalkeal:
- Remove temprorary default methods added as
  workaround in origin boot2 PR.
- Change repository methods to align to use Optional
  with method types changed in spring data.
- Try to align custom repo interfaces to also use Optional
  where data is accessed via id/key.
- Fixes #2539

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
@@ -98,10 +98,8 @@ public class StreamDeploymentController {
 	 */
 	@RequestMapping(value = "/{name}", method = RequestMethod.DELETE)
 	public ResponseEntity<Void> undeploy(@PathVariable("name") String name) {
-		StreamDefinition stream = this.repository.findOne(name);
-		if (stream == null) {
-			throw new NoSuchStreamDefinitionException(name);
-		}
+		this.repository.findById(name)
+				.orElseThrow(() -> new NoSuchStreamDefinitionException(name));
 		this.streamService.undeployStream(name);
 		return new ResponseEntity<>(HttpStatus.OK);
 	}
@@ -125,10 +123,8 @@ public class StreamDeploymentController {
 	@RequestMapping(value = "/{name}", method = RequestMethod.GET)
 	@ResponseStatus(HttpStatus.CREATED)
 	public StreamDeploymentResource info(@PathVariable("name") String name) {
-		StreamDefinition streamDefinition = this.repository.findOne(name);
-		if (streamDefinition == null) {
-			throw new NoSuchStreamDefinitionException(name);
-		}
+		StreamDefinition streamDefinition = this.repository.findById(name)
+				.orElseThrow(() -> new NoSuchStreamDefinitionException(name));
 		StreamDeployment streamDeployment = this.streamService.info(name);
 		Map<StreamDefinition, DeploymentState> streamDeploymentStates =
 				this.streamService.state(Arrays.asList(streamDefinition));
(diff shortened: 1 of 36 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
```diff
@@ -98,10 +98,8 @@ public StreamDeploymentController(StreamDefinitionRepository repository, StreamS
 	 */
 	@RequestMapping(value = "/{name}", method = RequestMethod.DELETE)
 	public ResponseEntity<Void> undeploy(@PathVariable("name") String name) {
-		StreamDefinition stream = this.repository.findOne(name);
-		if (stream == null) {
-			throw new NoSuchStreamDefinitionException(name);
-		}
+		this.repository.findById(name)
```
markpollack: We could also add a custom repository method as we did in Skipper, such as `findByIdRequired`

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
```diff
@@ -167,10 +167,8 @@ public void destroyTask(@PathVariable("name") String name) {
 	@RequestMapping(value = "/{name}", method = RequestMethod.GET)
 	@ResponseStatus(HttpStatus.OK)
 	public TaskDefinitionResource display(@PathVariable("name") String name) {
-		TaskDefinition definition = repository.findOne(name);
-		if (definition == null) {
-			throw new NoSuchTaskDefinitionException(name);
-		}
+		TaskDefinition definition = this.repository.findById(name)
+				.orElseThrow(() -> new NoSuchTaskDefinitionException(name));
```
markpollack: Similar here - could add this to the repository interface

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/DeploymentIdRepository.java
```diff
@@ -42,7 +44,7 @@
 	 * @param key the app deployment key
 	 * @return the identifier
 	 */
-	String findOne(String key);
+	Optional<String> findByKey(String key);
```
markpollack: Why not `findById`?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/RdbmsDeploymentIdRepository.java
```diff
@@ -60,7 +61,7 @@ public void delete(String name) {
 	}
 
 	@Override
-	public String findOne(String key) {
-		return findById(key).orElse(null);
+	public Optional<String> findByKey(String key) {
+		return findById(key);
 	}
```
markpollack: `findById` instead?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java
jvalkeal: I don't think that's a good idea as I'm not really fan of adding custom methods which just throws something.

markpollack:
where `orElseThrow` is used in this PR, we could push that functionality down into the repository class.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/RdbmsDeploymentIdRepository.java
jvalkeal: This is a weird interface as it really searches by key(aka value) instead of id.

jvalkeal:
I don't see no way to move `orElseThrow` feature into repo class itself as that's how Optional is supposed to get used. I don't think you're supposed to throw something if you're about to return empty optional. You need to trust how Optional works if we return it from repos.

jvalkeal:
Rebased and force pushed this branch.

markpollack:
merged in 371aeb7e91a81829a65f9ae5688c8d04560db651

PR #2563, opened 2018-11-09
Fix the StreamControllerTests test fixture

### LINKED ISSUE #2562: Refactor StreamControllerTests
See if prepareSkipperClientFixture can be put in an @Before method.

tzolov:
Resolves #2562


spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/StreamControllerTests.java
```diff
@@ -747,7 +784,7 @@ public void testDeployWithSensitiveData() throws Exception {
 
 		final List<AuditRecord> auditRecords = auditRecordRepository.findAll();
 
-		assertEquals(2, auditRecords.size()); // TODO This looks wrong! only one Audit record should exist
+		assertEquals(2, auditRecords.size()); // TODO (Tzolov) This looks wrong! only one Audit record should exist
```
ilayaperumalg: @tzolov Could you create an issue to follow up with your reasoning as it might get missed by just having a comment here.

PR #2573, opened 2018-11-14
Updates the multi-binder example

### LINKED ISSUE #2572: Reword Multi-binder example
In our multi-binder section of dataflow ref docs we refer to the transformer as being multi-binder in the example.  This is a bit confusing, so the example needs to show that it is a specialized transformer the user creates.

cppwfs:
Now example shows multi-binder transformer and specifies that user creates it.  Prevents confusion.
Some bug fixes and version updates

resolves #2572


ilayaperumalg:
LGTM

sobychacko:
Looks good. Merged upstream.

PR #2577, opened 2018-11-15
Schema updates for classic mode removal

### LINKED ISSUE #2568: Review schema requirements for classic mode
We will retrieve this information from Skipper repository instead.

child of #2548

ilayaperumalg:
- Remove DeploymentIdRepository and its usages
  - Update migration doc to add `drop schema` for `STREAM_DEPLOYMENTS` and `DEPLOYMENT_IDS` tables

Resolves #2568


ilayaperumalg:
After some investigation, I noticed that the `DeploymentIdRepository` is not used anymore and hence removed it as part of this PR. I might have missed something related to the use of it. So, please feel free to reject this proposal if there is a valid usage of `DeploymentIdRepository`.

Also, the `UriRegistryPopulator` doesn't seem to be used as well and if we don't find any valid uses of it we could possibly drop the `URI_REGISTRY` schema and related code as well.

cppwfs:
We to also go through the log samples in the reference docs and makes sure they are updated so that they show the version number in the deployment id.  ie. `deploymentId httptest.log-v2`  instead of just `httptest.log`

PR #2588, opened 2018-11-20
Conditionally qualify the update properties

### LINKED ISSUE #2586: Download the app metadata resource only if there are deployment properties
During Stream Deploy/Update, the application metadata resource is downloaded everytime irrespective of the existence of deployment properties. This download happens to `qualify` the whitelisted properties. Hence, we can download only when there are properties to update.

ilayaperumalg:
- The stream update properties goes through the whitelist properties qualification only when there are properties to handle.

Resolves #2586


markpollack:
merged onto 1.7.x branch as eac7302b16c7c4b17e74e42ceb69d75cc0c520d9
merged onto master branch as 6e0894f28ce927155edb33a3b2e947cee5a497bf

Issue #2587, opened 2018-11-20
Parsing error when deploying a stream with SpEL expression


sabbyanandan:
As a user, while deploying the following streams in v1.7 BS, I observe the attached error. 

_(notice that the `''` is not preserved at the time of deployment)_

```
dataflow:>stream create timer --definition "time --date-format=hhmmss | log"
Created new stream 'timer'

dataflow:>stream create minutes --definition ":timer.time > transform --expression='payload.substring(2,4)' | log"
Created new stream 'minutes'

dataflow:>stream deploy --name timer --properties deployer.*.bootMajorVersion=2
Deployment request has been sent for stream 'timer'

dataflow:>stream deploy --name minutes --properties deployer.*.bootMajorVersion=2
Command failed org.springframework.cloud.dataflow.rest.client.DataFlowClientException: 115E:(pos 140): unexpected data in stream definition ')'
:timer.time > transform --spring.cloud.stream.bindings.applicationMetrics.destination=metrics --transformer.expression=payload.substring(2,4) | log --spring.cloud.stream.bindings.applicationMetrics.destination=metrics
```

I also see the same error while creating the stream without `''`.

```
dataflow:>stream create minutes --definition ":timer.time > transform --expression=payload.substring(2,4) | log"
Command failed org.springframework.cloud.dataflow.rest.client.DataFlowClientException: 115E:(pos 58): unexpected data in stream definition ')'
:timer.time > transform --expression=payload.substring(2,4) | log
                                                          ^
```

sabbyanandan:
When this occurs, I see the following in the Dashboard. 

![image](https://user-images.githubusercontent.com/1562654/48755899-e2687900-ec4b-11e8-8e9b-637cf8114707.png)

tzolov:
@sabbyanandan , what apps version are you testing with? Darwin? 
Wonder if `--expression='payload.substring(2,4)'` shouldn't be --expression='new 
 String(payload).substring(2,4)'` instead?

sabbyanandan:
It doesn't matter; same error on both Celsius and Darwin. I believe there are issues with parsing and how we persist.

sabbyanandan:
@aclement: Do you have any ideas?

aclement:
will have a look

aclement:
[...]

PR #2608, opened 2018-11-26
Quotes

markpollack:
Resolves spring-cloud/spring-cloud-dataflow#2587

### DIFF
--- a/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/DefinitionUtils.java
+++ b/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/DefinitionUtils.java
@@ -34,12 +34,14 @@ public class DefinitionUtils {
 		}
 
 		if (!propertyValue.contains("'")) {
-			if (propertyValue.contains(" ") || propertyValue.contains(";") || propertyValue.contains("*")) {
+			if (propertyValue.contains(" ") || propertyValue.contains(";") || propertyValue.contains("*") ||
+					propertyValue.contains(",")) {
 				return "'" + propertyValue + "'";
 			}
 		}
 		else {
-			if (propertyValue.contains(" ") || propertyValue.contains(";") || propertyValue.contains("*")) {
+			if (propertyValue.contains(" ") || propertyValue.contains(";") || propertyValue.contains("*") ||
+					propertyValue.contains(",")) {
 				return "\"" + propertyValue + "\"";
 			}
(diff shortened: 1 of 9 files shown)


markpollack:
```
dataflow:>stream create test3 --definition "time | filter --expression='#jsonPath(payload,'$.lang')=='en'' | log"
Created new stream 'test3'
dataflow:>stream list
╔══════════╤═══════════════════════════════════════════════════════════════════╤═══════════════════════════════════════════════════════════════════╗
║  Stream  │                         Stream Definition                         │                              Status                               ║
║   Name   │                                                                   │                                                                   ║
╠══════════╪═══════════════════════════════════════════════════════════════════╪═══════════════════════════════════════════════════════════════════╣
║test3     │time | filter --expression='#jsonPath(payload,'$.lang')=='en'' |   │The app or group is known to the system, but is not currently      ║
║          │log                                                                │deployed                                                           ║
╚══════════╧═══════════════════════════════════════════════════════════════════╧═══════════════════════════════════════════════════════════════════╝

dataflow:>stream de
stream deploy     stream destroy    
dataflow:>stream deploy --name test3
Deployment request has been sent for stream 'test3'
dataflow:>stream list
╔═══════════╤═══════════════════════════════════════════════════════════════════════════╤═════════════════════════════╗
║Stream Name│                             Stream Definition                             │           Status            ║
╠═══════════╪═══════════════════════════════════════════════════════════════════════════╪═════════════════════════════╣
║test3      │time | filter --filter.expression='#jsonPath(payload,'$.lang')=='en'' | log│The stream is being deployed.║
╚═══════════╧═══════════════════════════════════════════════════════════════════════════╧═════════════════════════════╝
```

markpollack:
I tried to use the `quoteUsed` property to know when to strip off quotes if the argument name contained the word `expresion`, however, it seems for a value of `'#jsonPath(payload,'$.lang')=='en''` the ArgumentNode gives a quoteUsed=null value....

markpollack:
Maybe we can use a more bizarre character other than `,` for the case of the alternate stream application DSL syntax to avoid this conflict?  That would be a breaking change for users who are adopting this new DSL, but that is far fewer than those using the existing DSL.

```
--definition "aaa ; bbb"
--definition "aaa <-> bbb"
```

markpollack:
Another option is to have an additional command option that would be used to indicate which type of DSL is being parsed.

sabbyanandan:
Closing this in favor of #2618.

Issue #2548, opened 2018-11-05
Epic: Review the approach to deprecate classic-mode for streams


sabbyanandan:
As a developer, I'd like to study the approach to deprecate the classic-mode for streams.

**Acceptance:**
The goal of this spike to familiarize with the classic-mode internals and its relationship with streams and tasks. Once when we have a grip on the internals, we would have to come up with the approach to deprecate and refactor the internals to continue with a single mode (i.e., skipper) for SCDF 2.0.

sabbyanandan:
All the stories scoped under this epic is complete including the adjustment of docs in CF and K8s-servers.

PR #2610, opened 2018-11-27
Fix forgotten classic-mode deprecation changes

tzolov:
- Remove redundant class and folder
- fix docker-compose.yum - add the skipper URI

Child of #2548


ilayaperumalg:
LGTM.

PR #2611, opened 2018-11-27
Fix classic mode deprecation left overs

tzolov:
- Remove redundant class and folder
- fix docker-compose.yum - add the skipper URI

Child of #2548


ilayaperumalg:
@tzolov  I merged this already into master. Looks like a duplicate?

tzolov:
Ah, awkward, i will close is as a duplicate

tzolov:
duplicate of 2a2ae91

PR #2640, opened 2018-12-03
Added an index page to serve as a starting point for three guides

### LINKED ISSUE #2614: Cleanup ref. guide to toggle between Local, CF, and K8s platform specific docs
child of spring-cloud/spring-cloud-dataflow#2613

Buzzardo:
Per a discussion with Sabby, I am adding a file called index-multi.adoc. It contains links to three files that will hold the reference documentation for Spring Cloud Data Flow under local, Kubernetes, and Cloud Foundry. At present, those files do not exist, so the links are placeholders for the moment.

Resolves  #2614.


markpollack:
I'm in the middle of a PR that copies over the content from the other two server repos so we can have a single reference guide - though not ideally structured.  This leaves the index page open for usage along the lines of other spring projects.

PR #2651, opened 2018-12-05
Optimize the performance of job step count query

### LINKED ISSUE #2583: Performance issue: job page with lots of executions is slow
Hello,

**Description:**
On the Jobs page I have 657 executions. The request to get the executions from aws takes 38 seconds to load. 

The endpoint is `/jobs/executions?page=0&size=30`, but it doesn't look like it's paginating due to the slowness.

**Release versions:**
This is using `1.7.0.RELEASE`.

**Steps to reproduce:**
Let a stream run with a time source for a while, in order to have a lot of executions.

**Screenshots:**
image

Thank you,

cppwfs:
resolves #2583

The purpose of this PR is to optimize query required to populate the job executions for the UI and shell.

In both cases they require only the number of step executions.   However the current query returns a fully qualified JobExecution object which contains all the step executions associated with the step execution.  The problem is that if a user has a large number of steps with the JobExecution, the result takes a long time to return.   Up to 60 seconds ffor 100 JobExecutions with 500  step executions per.

To resolve this issue a new endpoint was created job/executions/jobexecutioninfoonly that returns a JobResource that contains a JobExecution without its associated stepexecutions, however the stepExecutionCount is populated in the resource.

The new endpoint was added so that if others were currently using the fully qualified jobs from the /job/executions endpoint, they would not be impacted.


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/JobExecutionController.java
```diff
@@ -93,6 +93,25 @@ public JobExecutionController(TaskJobService taskJobService) {
...
+	 * Return a page-able list of {@link JobExecutionResource} defined jobs that
+	 * do not contain step execution detail.
+	 *
+	 * @param pageable page-able collection of {@code TaskJobExecution}s.
+	 * @param assembler for the {@link TaskJobExecution}s
+	 * @return a list of Task/Job executions(job executions do not contain step executions.
+	 * @throws NoSuchJobExecutionException in the event that a job execution id specified
+	 * is not present when looking up stepExecutions for the result.
+	 */
+	@RequestMapping(value = "/jobexecutioninfoonly", method = RequestMethod.GET, produces = "application/json")
```
ilayaperumalg: Instead of a separate REST endpoint, can we have a new request parameter with the name similar to `excludeStepExecutions`  to the existing `jobs/executions` endpoint?

Moreover the name `/jobexecutioninfoonly` highlights the word `foo` than `info only` ;)

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskJobService.java
```diff
@@ -237,6 +195,25 @@ public void stopJobExecution(long jobExecutionId)
...
 	private TaskJobExecution getTaskJobExecution(JobExecution jobExecution) {
+		return new TaskJobExecution(getTaskExecutionId(jobExecution), jobExecution,
+				isTaskDefined(jobExecution), jobExecution.getStepExecutions().size());
+	}
+
+	private List<TaskJobExecution> getTaskJobExecutionsWithStepCountForList(Collection<JobExecution> jobExecutions) {
+		Assert.notNull(jobExecutions, "jobExecutions must not be null");
+		List<TaskJobExecution> taskJobExecutions = new ArrayList<>();
+		for (JobExecution jobExecution : jobExecutions) {
+			taskJobExecutions.add(getTaskJobExecutionWithStepCount(jobExecution));
```
ilayaperumalg: This could potentially become performance bottleneck as we issue a JDBC query for **each** job execution to get the step execution count. Maybe we could do some joins at the `list` query for job executions to get the step execution count?

cppwfs: Updated.   Retrieves count directly from the JobExecutionStepCount instance

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/JobExecutionStepCount.java
```diff
@@ -0,0 +1,43 @@
...
+
+import org.springframework.batch.core.JobExecution;
+
+/**
+ * Supports the base {@link JobExecution} attributes along with a step count.
+ * @author Glenn Renfro
+ *
+ * @since 2.0
+ */
+public class JobExecutionStepCount extends JobExecution {
```
ilayaperumalg: We can probably rename this to `JobExecutionWithStepCount`. I can do it when merging.

ilayaperumalg:
@cppwfs We still need to update the shell command(s) to use the `/jobs/thinexecutions` endpoint. I have created the same for Data Flow UI: https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/1006

ilayaperumalg:
@cppwfs I don't see the new REST endpoint exposed in the root resource though I see the update in the `RootController`. Could you confirm if you could see that?

cppwfs:
@ilayaperumalg hmmm...  I see this in the `http://localhost:9393`   
jobs/thinexecutions: {
href: "http://localhost:9393/jobs/thinexecutions"
},
Is that what you are asking about?

ilayaperumalg:
@cppwfs ok, I will check why I don't see it in the root REST endpoint. 
Not sure if we want to add all the REST endpoints as the performance bottleneck only arises for the `list` query and other endpoints should be OK.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/JobExecutionStepCount.java
cppwfs: done

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/JobExecutionController.java
cppwfs: done

cppwfs:
@oodamien Thanks for catching that.  Updated.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Rebased and merged into `master` as e3b8896e139433566086d4d9a666c93268cb67ce

ilayaperumalg:
@cppwfs Would you mind issuing a separate PR for 1.7.x as there are many conflicts to handle?

sabbyanandan:
This is addressed in `master` vis https://github.com/spring-cloud/spring-cloud-dataflow/commit/e3b8896e139433566086d4d9a666c93268cb67ce. THe backport will be handled in a different PR, closing it.

PR #2660, opened 2018-12-10
Upgrade to Boot 2.1.x

### LINKED ISSUE #2606: Upgrade to Boot 2.1.x
As a developer, I'd like to upgrade 2.x/master to Boot 2.1.x. This is a prerequisite to security rewrite [spring-cloud/spring-cloud-dataflow#2579].

Child of #914.

jvalkeal:
- Pump up all deps to Boot 2.1.x stack
  - Spring Data deps now managed by boot
  - Hibernate deps now managed by boot
  - Server core now have runtime dep to jayway jsonpath as it
    was removed from skipper not being compile nor runtime dep there
    as it's only runtime dep on dataflow side because of hal.
- For tasks and deployers, we now rely on their auto-config packages,
  so all recreated beans are now removed.
- PlatformTransactionManager is now coming from boot. This was originally
  added to dataflow before boot 1.5.x which added it, so became redundant.
- There is now CustomRedisHealthIndicatorAutoConfiguration to not mess up
  boot auto-config.
- All security tests are now disabled as spring-cloud-common-security-config has
  a PR lined up and that project needs to implement auto-config to overcome
  boot21 issues.
- Some testing changes related to boot21 removing deprecated methods.
- @EnableMapRepositories can now only exist once, boot21 override thingy, thus
  removed from TaskConfiguration
- Some tests now rely on spring.main.allow-bean-definition-overriding=true
  as it was way too difficult not to override some beans in tests.
- Fixes #2606

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -16,7 +16,7 @@
 		<!-- Note: Also make sure the parent pom version in spring-cloud-dataflow-dependencies/pom.xml is in sync -->
 		<groupId>org.springframework.cloud</groupId>
 		<artifactId>spring-cloud-build</artifactId>
-		<version>2.0.3.RELEASE</version>
+		<version>2.1.0.RC2</version>
 		<relativePath />
 	</parent>
 	<scm>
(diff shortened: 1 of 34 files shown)


ilayaperumalg:
LGTM

tzolov:
@jvalkeal, 
* The `spring-cloud-dataflow-dependencies` parent version must match the main's pom `spring-cloud-build` parent version.  E.g.  `2.1.0.RC2` instead `2.0.0.BUILD-SNAPSHOT`
```
	<parent>
		<artifactId>spring-cloud-dependencies-parent</artifactId>
		<version>2.1.0.RC2</version>
		<relativePath />
	</parent>
```
* Both spring cloud `2.1.0.RC2` and `Greenwich.M1` are still at boot `2.1.0.RELEASE`. The `2.1.0.BUILD-SNAPSHOT` and `Greenwich.BUILD-SNAPSHOT` are at Boot 2.1.1 RELEASE and miles stones for them should be released today.  I've tested SCDF with those and all tests pass.

tzolov:
Fix a missing dependency and merge at de2abebcd88569d4d240be713865762a831e00ed

PR #2664, opened 2018-12-10
Fix tests that were failing on my machine

markpollack:


jvalkeal:
This maybe be because tests are run with different order depending on a system. We've seen this before. I'll merge this.

PR #2705, opened 2018-12-20
Enable LocalSchedulerAutoConfiguration on local platform only

### LINKED ISSUE #2704: Activate Local Scheduler Configuration only for Local platform deployment
Existing conditional expressions in LocalSchedulerAutoConfiguration (https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-server-local-autoconfig/src/main/java/org/springframework/cloud/dataflow/autoconfigure/local/LocalSchedulerAutoConfiguration.java#L37) may allow activating LocalSchedulerAutoConfiguration in K8s platforms

tzolov:
- Implement negation condition for not-K8s platform.

 Resolves #2704


tzolov:
Will be resolved as part of #2708

PR #2715, opened 2018-12-21
Document usage of SAJ for env properties

### LINKED ISSUE #2712: Update from SPRING_CLOUD_DEPLOYER_CLOUDFOUNDRY_STREAM/TASK to new format
SPRING_CLOUD_DEPLOYER_CLOUDFOUNDRY_STREAM_USE_SPRING_APPLICATION_JSON: false` and `SPRING_CLOUD_DEPLOYER_CLOUDFOUNDRY_TASK_USE_SPRING_APPLICATION_JSON

is not relevant anymore, need to provide different config samples for different servers.

### LINKED ISSUE #2714: Document configuration properties that no longer work via environment properties
Currently, there are few properties that no longer work when using environment properties. We need to use `SPRING_APPLICATION_JSON` for these. This is mainly applicable for CF platform.

ilayaperumalg:
- For maven properties
 - For enable Local Platform in Skipper

Resolves #2714

Document the usage of CF Deployment properties for stream/task

 - For stream (update Skipper)
 - For task (update SCDF)

Resolves #2712

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/getting-started-cloudfoundry.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/getting-started-cloudfoundry.adoc
@@ -97,7 +97,7 @@ applications:
   path: <PATH TO THE DOWNLOADED SKIPPER SERVER UBER-JAR>
   env:
     SPRING_APPLICATION_NAME: skipper-server
-    SPRING_CLOUD_SKIPPER_SERVER_ENABLE_LOCAL_PLATFORM: false
+    SPRING_APPLICATION_JSON: '{ "spring": { "cloud": { "skipper": { "server": { "enableLocalPlatform": "false"} } } } }'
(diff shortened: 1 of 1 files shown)


spring-cloud-dataflow-docs/src/main/asciidoc/getting-started-cloudfoundry.adoc
```diff
@@ -97,7 +97,7 @@ applications:
   path: <PATH TO THE DOWNLOADED SKIPPER SERVER UBER-JAR>
   env:
     SPRING_APPLICATION_NAME: skipper-server
-    SPRING_CLOUD_SKIPPER_SERVER_ENABLE_LOCAL_PLATFORM: false
+    SPRING_APPLICATION_JSON: '{ "spring": { "cloud": { "skipper": { "server": { "enableLocalPlatform": "false"} } } } }'
```
sabbyanandan: I don't think this prpoperty works whether in this format or as flat env-var. That's a bug outside of this change, in skipper perhaps.

Issue #2675, opened 2018-12-12
Remove all deprecation warnings


markpollack:
Many new deprecation warning have appeared after the upgrade to various dependent libraries, try to prune these.

PR #2721, opened 2018-12-27
Remove all deprecation warnings #2675

sylank:
Many new deprecation warning have appeared after the upgrade to various dependent libraries, try to prune these.

Check and collect the deprecated functions:
mvn clean install -Dmaven.compiler.showDeprecation=true -Dmaven.compiler.showWarnings=true > warnings.txt
cat warnings.txt | grep -i deprecated > deprecated.txt

UserCredentials is appears only in tests
UserCredentials: @deprecated since 2.1 without replacement as it was meant for internal use only anyway.

ISO8601DateFormatWithMilliSeconds
StdDateFormat was extended with two constructor because the ISO8601DateFormat and utils was deprecated.

Old format:
2018-12-27T16:24:34.514Z
New format:
2018-12-27T16:24:34.514+0000

/**
* Defines a commonly used date format that conforms
* to ISO-8601 date formatting standard, when it includes basic undecorated
* timezone definition.
*/
public final static String DATE_FORMAT_STR_ISO8601 = "yyyy-MM-dd'T'HH:mm:ss.SSSZ";


markpollack:
Apologies for letting this PR sit so long.  It is very much appreciated.  The difference between `ISO8601DateFormat` and `StdDateFormat` is OK for our usage.  We will also make a copy of the UserCredentials class from spring data commons and add it into the SCDF repo as part of some other work that @ghillert is doing on security improvements. 

Thanks again!

rebased and merged as 45114f730e1e42f2da7d275352013d94b2ffaa79

PR #2745, opened 2019-01-10
Resolves issue where task launches before setup is complete [1.7.x]

### LINKED ISSUE #2744: Need to commit TaskExecution table entry prior to task launch (This is the backport for 1.7)
writing this update so that it will work for 1.7.x

cppwfs:
In this case a transaction is created and completes before the task is launched.
Then a transaction is created to capture audit information after the task is launched.

resolves #2744 for 1.7.x

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
@@ -16,11 +16,17 @@
 
 package org.springframework.cloud.dataflow.server.service.impl;
 
+import java.sql.Types;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+
+import javax.sql.DataSource;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import org.springframework.boot.autoconfigure.jdbc.DataSourceProperties;
 import org.springframework.cloud.dataflow.configuration.metadata.ApplicationConfigurationMetadataResolver;
@@ -54,6 +60,7 @@ import org.springframework.cloud.task.repository.TaskExecution;
 import org.springframework.cloud.task.repository.TaskExplorer;
 import org.springframework.cloud.task.repository.TaskRepository;
 import org.springframework.core.io.Resource;
+import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.transaction.PlatformTransactionManager;
 import org.springframework.transaction.TransactionStatus;
 import org.springframework.transaction.annotation.Transactional;
@@ -78,6 +85,10 @@ import org.springframework.util.StringUtils;
  * @author David Turanski
  */
 public class DefaultTaskService implements TaskService {
+
+	private final Logger logger = LoggerFactory.getLogger(DefaultTaskService.class);
+
+	private DataSource dataSource;
 
 	private final DataSourceProperties dataSourceProperties;
 
@@ -121,9 +132,9 @@ public class DefaultTaskService implements TaskService {
 
 	private final ArgumentSanitizer argumentSanitizer = new ArgumentSanitizer();
 
-	public static final String TASK_DEFINITION_DSL_TEXT = "taskDefinitionDslText";
-	public static final String TASK_DEPLOYMENT_PROPERTIES = "taskDeploymentProperties";
-	public static final String COMMAND_LINE_ARGS = "commandLineArgs";
+	private static final String TASK_DEFINITION_DSL_TEXT = "taskDefinitionDslText";
+	private static final String TASK_DEPLOYMENT_PROPERTIES = "taskDeploymentProperties";
+	private static final String COMMAND_LINE_ARGS = "commandLineArgs";
 
 	/**
 	 * Initializes the {@link DefaultTaskService}.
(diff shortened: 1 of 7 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
```diff
@@ -171,75 +179,110 @@ public DefaultTaskService(DataSourceProperties dataSourceProperties,
...
+	private RequestCriteria createLaunchRequest (final String taskName, final Map<String, String> taskDeploymentProperties,
+			final List<String> commandLineArgs) {
+		TransactionTemplate template = new TransactionTemplate(transactionManager);
+
+		return template.execute(status -> {
+			Assert.hasText(taskName, "The provided taskName must not be null or empty.");
+			Assert.notNull(taskDeploymentProperties, "The provided runtimeProperties must not be null.");
+
+			Map<String, String> launchTaskDeploymentProperties = new HashMap<>(taskDeploymentProperties);
+			if (maxConcurrentExecutionsReached()) {
```
markpollack: This check should move out of this method and be the first line of code in the executeTask method as it isn't related to creating the request.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
```diff
@@ -171,75 +179,110 @@ public DefaultTaskService(DataSourceProperties dataSourceProperties,
...
 
-		this.auditRecordService.populateAndSaveAuditRecordUsingMapData(
-				AuditOperationType.TASK, AuditActionType.DEPLOY,
-				taskDefinition.getName(), getAuditata(taskDefinition, taskDeploymentProperties, updatedCmdLineArgs));
+				taskExecutionRepository.updateExternalExecutionId(
+						requestCriteria.getTaskExecution().getExecutionId(),
+						externalExecutionId);
 
-		return taskExecution.getExecutionId();
+				auditRecordService.populateAndSaveAuditRecordUsingMapData(
```
markpollack: was it intentional that `AuditRecordService` implementation is not annotated with `@Transactional`?  @ghillert

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
```diff
@@ -171,75 +179,110 @@ public DefaultTaskService(DataSourceProperties dataSourceProperties,
...
+		String id = this.taskLauncher.launch(requestCriteria.getAppDeploymentRequest());
 		if (!StringUtils.hasText(id)) {
 			throw new IllegalStateException("Deployment ID is null for the task:" + taskName);
 		}
-		this.taskExecutionRepository.updateExternalExecutionId(taskExecution.getExecutionId(), id);
+		postLaunchProcessing(requestCriteria, id);
+		return requestCriteria.getTaskExecution().getExecutionId();
+	}
+
+	private RequestCriteria createLaunchRequest (final String taskName, final Map<String, String> taskDeploymentProperties,
```
markpollack: It seems like from the usage of the class `RequestCriteria` all the information can be obtained from the `AppDeploymentRequest`, so this method can change to `AppDeploymentRequest createLaunchRequest`

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
```diff
@@ -171,75 +179,110 @@ public DefaultTaskService(DataSourceProperties dataSourceProperties,
...
-			TaskServiceUtils.updateDataFlowUriIfNeeded(this.dataflowServerUri, appDeploymentProperties,
-					commandLineArgs);
-		}
-		AppDefinition revisedDefinition = TaskServiceUtils.mergeAndExpandAppProperties(taskDefinition, metadataResource,
-				appDeploymentProperties, this.whitelistProperties);
-		List<String> updatedCmdLineArgs = this.updateCommandLineArgs(commandLineArgs, taskExecution);
-		AppDeploymentRequest request = new AppDeploymentRequest(revisedDefinition, appResource,
-				deployerDeploymentProperties, updatedCmdLineArgs);
-		String id = this.taskLauncher.launch(request);
+		String id = this.taskLauncher.launch(requestCriteria.getAppDeploymentRequest());
```
markpollack: What if the taskLauncher throws an exception, then there has been a commit into the database for something that hasn't launched.

cppwfs: @markpollack The TaskExecution table will have an entry with the Task_Execution_id and nothing else.  It would be similar to the case where a Task failed to launch (for whatever reason).   Except in this case there would be a stack trace in SCDF.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
cppwfs: Wouldn't it be safer to keep grouped with the other db calls within the TransactionTemplate?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskService.java
cppwfs: RequestCriteria is still being used because we need the DSL in TaskDefinition that is not stored in AppDefinition stored in the AppDeploymentRequest.  Also while the executionId is available in the AppDefinition it is stored in the commandLineArgs and would have to be extracted via a search of the list to find it and then extract it from the string.  So storing it in the RequestCriteria appeared to be cleaner.

markpollack:
closing is favor of an approach that will separate out the call to persist a task execution record before launching a task into a new service that uses tx propagation requiresNew.

PR #2747, opened 2019-01-11
Int casts from long in pageable job queries are being checked

### LINKED ISSUE #2549: Check long to int cast in batch
As `org.springframework.data.domain.Pageable.getOffset()` changed type from `int` to `long`, in `DefaultTaskJobService` we need to cast as other parts of a batch still uses `int`.

We need to check that this is ok, or do some further fixes.

Child of #914

cppwfs:
According to the notes we need to return a qualified exception if the user attempts a query with offest greater than MAX_INT

resolves #2549


jvalkeal:
lgtm

jvalkeal:
I'll remove those TODO comments with merge

jvalkeal:
Merged per 71c1a04d2867c7f1961f8433c6206016094b773d

PR #2755, opened 2019-01-15
Remove Spring Cloud Connectors dependency

### LINKED ISSUE #2724: Remove connectors dependency from the classpath
As a developer, I'd like to remove the dependency of spring-cloud-connectors from the classpath; instead, we would switch to Boot's autoconfiguration proper.

Notes:

In 2.0 baseline, we have already had to work around MySQL and Redis connectors (see: https://github.com/spring-cloud/spring-cloud-dataflow/commit/da40c8f8c2c26a034eb24322c74f3f2fa5f39802) for Cloud Foundry. The goal is to remove the connector footprint in entirety in favor of what's natively possible via Boot itself. 

This story assumes the core redesign would be in the [spring-cloud-connectors](https://github.com/spring-cloud/spring-cloud-connectors) project, and it would be available to use in SCDF and wherever else necessary.

markpollack:
* Uses the spring-cfenv library to set spring boot properties for DB/SSO/Config Server
* Support setting DataSource properties, only MySQL DB supported currently
* Support setting SSO properties
* Support setting config server client properties

When pushing to Cloud Foundry, be sure to add
```
    SPRING_PROFILES_ACTIVE: cloud
    JBP_CONFIG_SPRING_AUTO_RECONFIGURATION: '{enabled: false}'    
```
to your manfiest. You should see a log message very early (before the Boot banner) 

```
2019-01-15T12:27:13.077-05:00 [APP/PROC/WEB/0] [OUT] 2019-01-15 17:27:13.076 INFO 21 --- [ main] s.b.CfDataSourceEnvironmentPostProcessor : Setting spring.datasource.url property from bound service [mysql]
2019-01-15T12:27:13.077-05:00 [APP/PROC/WEB/0] [OUT] 2019-01-15 17:27:13.076 INFO 21 --- [ main] loudConfigClientEnvironmentPostProcessor : Setting spring.cloud.config.client properties from bound service [configserver]
``` 

Fixes #2724


spring-cloud-dataflow-platform-cloudfoundry/pom.xml
```diff
@@ -33,25 +33,13 @@
...
-			</exclusions>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-starter-config</artifactId>
 		</dependency>
 		<dependency>
-			<groupId>io.pivotal.spring.cloud</groupId>
-			<artifactId>spring-cloud-sso-connector</artifactId>
-			<version>${spring-cloud-sso.version}</version>
+			<groupId>org.springframework.cfenv</groupId>
+			<artifactId>spring-cfenv-boot</artifactId>
```
jvalkeal: Think we need to rename module as we always get heat from boot guys if anyone else uses `boot` in naming.

spring-cloud-dataflow-platform-cloudfoundry/pom.xml
```diff
@@ -33,25 +33,13 @@
...
+			<version>1.0.0.BUILD-SNAPSHOT</version>
```
jvalkeal: Version to dep management.

spring-cloud-dataflow-platform-cloudfoundry/pom.xml
```diff
@@ -33,25 +33,13 @@
 		  <optional>true</optional>
 		</dependency>
 		<dependency>
-			<groupId>io.pivotal.spring.cloud</groupId>
-			<artifactId>spring-cloud-services-starter-config-client</artifactId>
-			<version>${spring-cloud-services.version}</version>
```
jvalkeal: `spring-cloud-services.version` can be removed from pom properties.

spring-cloud-dataflow-platform-cloudfoundry/pom.xml
jvalkeal: Maybe not necessarily right now but before we actually release something.

spring-cloud-dataflow-platform-cloudfoundry/pom.xml
jvalkeal: You know, I meant putting `spring-cfenv-boot` in a dependency management section in a parent pom so that in a module we can just add dependency without specifying version.

jvalkeal: Right, there is other deps defined like this. Maybe we'll polish those poms with a different commit.

jvalkeal:
Merged per 6ad6c74a803679d041aa0aebf7d6cb30950e3eab

PR #2762, opened 2019-01-16
Produce error if single char stream element arg value is special char

### LINKED ISSUE #2499: Perhaps address DSL parsing inconsistency for short arg values using special chars
Discovered whilst chatting with Mark P.

We have a number of special characters that terminate an arg value (if not in quotes):
```
"|"   ";"   " "   "\t"   "\n"   ">"   ","   "\r"
```

So this is fine:
`foo --aaa=bbb, bar`
the value of `aaa` is `bbb`, and there are two apps here `foo` and `bar`.

However, if you only have an arg value of a single character that is the special character, the special character fails to terminate the argument value:

`foo --aaa=| bbb`

What do you expect there? 

Currently you will get an error because `|` is not treated as terminating the arg value. and so it considers the first argument to be `aaa`=`|` and then it is looking for a special character or another argument but it immediately hits `bbb`.  This goes for all the special characters, we found this because of Marks test:

`foo --aaa=,, bbb`

Right now this parses as valid - for the first app `foo` the argument `aaa` is `,` and then there is a `,` then the second app `bbb`.

Strictly speaking if `,` should terminate an arg value unless quoted then it should fail to parse because there is no app specified between the two commas (and the value of `aaa` should be `""`)

I haven't jumped on fixing this ... because it is more widespread than I first thought. I initially thought it was a problem just for the new `,` but it is actually affecting all the special symbols. That means changes will affect anyone currently relying on this behaviour.  What do we think? Fix for consistency or is this ability to pass single special chars without quoting convenient?

aclement:
Without this change the system was not consistent in how it treated
special characters in argument values (e.g. | or >). They were
not treated as special if they were the first character in the
argument value (e.g. 'app --param=|'). When used like that they
should still be treated as special (i.e. end of stream def) and
so an error should come out about a missing argument value. If
they are truly the argument value they should be quoted.

Resolves #2499

### DIFF
--- a/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/TaskTokenizer.java
+++ b/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/dsl/TaskTokenizer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 the original author or authors.
+ * Copyright 2017-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Rebased and merged as 01056363

PR #2769, opened 2019-01-18
Fix PageImpl usages

### LINKED ISSUE #2742: Investigate the explicit creation of Page resources
Look for all the `new PageImpl()` hits and determine what to do with them. We should not be doing that.

ilayaperumalg:
- Use Pageable and ResourceAssembler to construct the REST resource instead of constructing PageImpl

Resolves #2742

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskSchedulerController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskSchedulerController.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 the original author or authors.
+ * Copyright 2018-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -46,6 +46,7 @@ import org.springframework.web.bind.annotation.RestController;
  *
  * @author Glenn Renfro
  * @author Christian Tzolov
+ * @author Ilayaperumal Gopinathan
  */
 @RestController
 @RequestMapping("/tasks/schedules")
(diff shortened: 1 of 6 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskSchedulerController.java
```diff
@@ -77,16 +78,10 @@ public TaskSchedulerController(SchedulerService schedulerService) {
...
 			PagedResourcesAssembler<ScheduleInfo> assembler) {
 		List<ScheduleInfo> result = this.schedulerService.list();
-		int resultSize = result.size();
-		Pageable pageable = PageRequest.of(0,
-				(resultSize == 0) ? resultSize = 1 : resultSize); //handle empty result set
-
-		Page<ScheduleInfo> page = new PageImpl<>(result, pageable,
-				result.size());
-		return assembler.toResource(page, taskAssembler);
+		return assembler.toResource(new PageImpl<>(result, pageable, result.size()), taskAssembler);
```
markpollack: why do we still need a `new PageImpl<>` here?

ilayaperumalg: We still need to do this because the `scheduler` service [doesn't have the `list(Pageable)` implemented](https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultSchedulerService.java#L170) yet.

jvalkeal:
I'll go ahead and merge this.

jvalkeal:
Merged per 77e3ff488f61be83fbfc2ea8e717e7ba426c47e6

PR #2771, opened 2019-01-18
Use platform option for task launch

### LINKED ISSUE #2770: Task launch using specific platform name
When launching the task, the `task launch` command's `platform-name` needs to be used. Currently, the TaskTemplate is not updated with this.

Child of #2720

ilayaperumalg:
- Update TaskOperations method to use platformName
    - Update its implementation at TaskTemplate
  - Fix TaskCommand `task launch` to pass `platformName`

Resolves #2770

### DIFF
--- a/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/TaskOperations.java
+++ b/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/TaskOperations.java
@@ -64,9 +64,10 @@ public interface TaskOperations {
 	 * @param name the name of the task
 	 * @param properties the deployment properties
 	 * @param arguments the command line arguments
+	 * @param platformName the name of the platform to use
 	 * @return long containing the TaskExecutionId
 	 */
-	long launch(String name, Map<String, String> properties, List<String> arguments);
+	long launch(String name, Map<String, String> properties, List<String> arguments, String platformName);
 
 	/**
 	 * Destroy an existing task.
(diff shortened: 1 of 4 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskExecutionController.java
```diff
@@ -156,7 +156,7 @@ public TaskExecutionController(TaskExplorer explorer, TaskExecutionService taskE
 	public long launch(@RequestParam("name") String taskName,
 			@RequestParam(required = false) String properties,
 			@RequestParam(required = false) String arguments,
-			@RequestParam(required = false, defaultValue = "default") String platformName) {
+			@RequestParam(required = false, defaultValue = "default", name="platform-name") String platformName) {
```
markpollack: how do we know this works?

spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/TaskOperations.java
```diff
@@ -64,9 +64,10 @@
 	 * @param name the name of the task
 	 * @param properties the deployment properties
 	 * @param arguments the command line arguments
+	 * @param platformName the name of the platform to use
 	 * @return long containing the TaskExecutionId
 	 */
-	long launch(String name, Map<String, String> properties, List<String> arguments);
+	long launch(String name, Map<String, String> properties, List<String> arguments, String platformName);
```
markpollack: Another option, which is how it is done for skipper, is to put a special key in the properties map, just as for other deployer properties.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskExecutionController.java
ilayaperumalg: This is our property that we set from the client side as in [here](https://github.com/spring-cloud/spring-cloud-dataflow/pull/2771/files#diff-67f6359ec02f9f7a13fb5c84968a3596R156)

markpollack:
Closing this as I incorporated the hashmap approach in the upcoming PR for implementing the multiple platform support.

PR #2772, opened 2019-01-18
Overhaul flyway usage

### LINKED ISSUE #2540: Implement new boot approach for creating database indexes
See BatchTaskIndexesDatabaseInitializer

Child of #914

### LINKED ISSUE #2759: Dataflow upgrade with flyway
Implement features to start dataflow 2.x when existing database is created from dataflow 1.7.x.

Child of #914.

jvalkeal:
- Moving away from sql scripts if favour of doing everything from java. This
  doesn't mean that we could not use those in future.
- Remove all old sql scripts except embedded H2 as it will never have any migrations
  as it's safe to assume that db is always started from scratch.
- Change to baseline from version 1 and disable location check as otherwise
  boot fails as we don't have sql scripts anymore.
- In case empty db, we go directly to V1 and create stuff to db.
- In case of not empty db, we either need to baseline from case where we have
  other data in a db, i.e. skipper we baseline to V1, if we have old dataflow
  tables we do real migration. That's reason we do all from java as we
  can be more clever to check what's in a database.
- New DataFlowFlywayConfigurationCustomizer which allows to define java baseline
  classes per vendor as boot currently doesn't allow to define flyway callbacks
  for different vendors.
- No added tests in this commit as things are actually tests with acceptance tests
  what comes for different databases.
- Fixes #2759
- Fixes #2540


markpollack:
Tested by upgrading a previous installation of scdf/skipper on PCF.  Tweaked logging.

Squashed and merged as 8e90f0aee2e918c898da319d1f1f9785c0842e50

PR #2775, opened 2019-01-22
[DO NOT MERGE] Add Runtime Apps based Metrics Controller

### LINKED ISSUE #2736: Revise the metrics endpoint to include application count/health stats through Micrometer
Currently the run-time application health/count information is provided metrics collector (along with the `rate` info)

Use the Runtime apps command or something lighter to provide the UI with this information instead.

Child of #2601

### LINKED ISSUE #2737: Map the `/metrics` endpoint to collect application stats from the underlying Micrometer backend
The goal of this story is to map the existing `/metrics` endpoint to instead query the desired Micrometer backend for application stats including message-rates, error-counts etc.

The UI would eventually interact with this endpoint to redirect to the Grafana dashboard.

Child of #2601

tzolov:
This change allows the SCDF UI to show the green dots corresponding to the running app instances while it drops support for stream rate metrics.
 - Add new Metrics controller that converts the run time apps statuses into metrics streams. The rate metrics sections skipped.
 - Remove all Metrics Collector related components such as: MetricsController, MetricsStore, MetricsProperties, MockTest artifacts.
 - Fix test and rest documentation.

 Resolves #2736
 Resolves #2737
 
Depends on https://github.com/spring-cloud/spring-cloud-skipper/pull/802


markpollack:
I know #2737 specifically called out reusing the existing `/metrics` endpoint for this functionality but it doesn't look right now that I see the PR.  There are no metrics related information in the modified `ApplicationsMetrics` class.  It looks like it belong in the `RuntimeAppController`, which delegates to `streamDeployer` for getting the `AppStatus` all the apps or the apps or just the apps in a given stream.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/RuntimeAppsController.java
```diff
@@ -77,13 +87,65 @@ public RuntimeAppsController(StreamDeployer streamDeployer) {
 		this.streamDeployer = streamDeployer;
 	}
 
-	@RequestMapping
+	@RequestMapping("/stream")
```
markpollack: should be `streams` since it accepts an array of names.

markpollack:
Changed to use `/runtime/streams`.  Squashed and merged as c157740d343e6ad9da5b85c788a23b221d7efbdd

I was not able to run the test `RuntimeStreamStatusForStreamAppsWithoutCollectorDocumentation`  (BTW, win for longest class name) in the IDE but running it via the command line worked.

also there are a few files that still have the name 'metrics' in them that could be renamed.  Will create some follow up issues.

PR #2783, opened 2019-01-24
Manage logging to disk based on profile

### LINKED ISSUE #2672: Manage logging to disk based on profile
With a single server, the logging to disk should only happen when running locally (other platforms capture stdout).  Add 'local' profile to logback-spring.xml, eg, <springProfile name="local"> or maybe figure out if we can avoid using a profile at all.

chrisjs:
Resolves #2672


markpollack:
Made it more extensible here - https://github.com/markpollack/spring-cloud-dataflow/tree/GH-2672

markpollack:
Added some extensibility for other platforms and merged as 7b0cb2ec4f10c0be2fb6b1fade9e1250def0e157

PR #2785, opened 2019-01-25
Unused MavenProperties removed from AppRegistryController's constructor

sylank:
Unused field removed from AppRegistryController's constructor


markpollack:
Thanks!

PR #2787, opened 2019-01-25
Add and expose Grafana connection properties

### LINKED ISSUE #2786: Expose the Grafana Access Info if configured
- Introduce SCDF properties to define the Grafana connection parameter such as URL and security tokens. 
- Add Grafana Feature in the about resource to indicate when the Grafana connection parameters are set. 
- Expose the Grafana access properties through the AboutResource

child of #2601

tzolov:
- Add s.c.d.grafana-info.url and s.c.grafana-info.token properties.
 - Expose the Grafana connection properties via the GrafanaInfo About Resource. (e.g. under http:localhost:9393/about)
 - Add GrafanEnabled feature to the list of About resource features to indicate if the Grafana connection properties are provided or not. If the s.c.d.grafana-info.url is set then the grafanaEnabled=true.
 - Add Ids to the StreamStatus's applications and instances.
 - Fix REST documentation issues with /runtime/apps and /runtime/streams.

Resolves #2786


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AboutController.java
```diff
@@ -93,8 +94,17 @@
 
 	private LauncherRepository launcherRepository;
 
+	@Value("${spring.cloud.dataflow.grafana-info.url:#{null}}")
+	private String grafanaUrl;
+
+	@Value("${spring.cloud.dataflow.grafana-info.token:#{null}}")
+	private String grafanaToken;
+
+	@Value("${spring.cloud.dataflow.grafana-info.refreshInterval:15}")
```
jvalkeal: These should really go via configuration props as using a @Value like will have issues. i.e. `--spring.cloud.dataflow.grafana-info.refresh-interval=20` doesn't work.

tzolov: Those properties and the related `GrafanaInfo` are not used internally by the SCDF server but only as a mean to configure the SCDF-UI. So perhaps using spring configuration props might be an overkill? 

For i've successfully tested with properties like:
```
--spring.cloud.dataflow.grafana-info.url=http://localhost:3000
--spring.cloud.dataflow.grafana-info.refreshInterval=30
```

Not sure if the `xxx.refresh-interval` relaxation would work. Probably not.

tzolov: @jvalkeal, as suggested I've replaced the ad-hoc `@Value` properties with proper GrafanaInfoProperties object.  This provides a cleaner separation of concerns.

jvalkeal:
lgtm, merging

jvalkeal:
Merged per 2d749d3b5b1873f8df38223ca1fdafc566e77b60

Issue #2739, opened 2019-01-08
Adapt CTR to take advantage of improved security features


sabbyanandan:
As a developer, I'd like to refactor and adapt CTR to the latest OAuth by default experience, so the general security configurations are consistent not only in SCDF/Skipper, but also in other microservice components in the ecosystem. 

The support for it in the CTR is a prerequisite.

child of #2574

sabbyanandan:
Addressed in https://github.com/spring-cloud/spring-cloud-dataflow/commit/c1cfb0a40a37e817318eb1f806f3ad39ac801f26.

Issue #2789, opened 2019-01-28
Add Tests to validate scope of the OAuth token


ghillert:
See https://github.com/spring-cloud/spring-cloud-common-security-config/issues/33

Requires #2582

PR #2790, opened 2019-01-28
gh-2789 Add Tests to validate scope of the OAuth token

ghillert:
requires https://github.com/spring-cloud/spring-cloud-common-security-config/issues/33

Resolves https://github.com/spring-cloud/spring-cloud-dataflow/issues/2739


ghillert:
Rebased

spring-cloud-starter-dataflow-server-local/src/test/java/org/springframework/cloud/dataflow/server/single/security/LocalServerSecurityWithOAuth2Tests.java
```diff
@@ -59,6 +63,16 @@ public void testAccessRootUrlWithBasicAuthCredentials() throws Exception {
 				.andExpect(status().isOk());
 	}
 
+	@Test
+	public void testAccessRootUrlWithBasicAuthCredentialsTwice() throws Exception {
+		localDataflowResource.getMockMvc()
+				.perform(get("/").header("Authorization", basicAuthorizationHeader("user", "secret10"))).andDo(print())
+				.andExpect(status().isOk());
+		localDataflowResource.getMockMvc()
```
markpollack: I guess we should eventually have some assert that the token is cached when making the second call.

markpollack:
squashed and merged as c1cfb0a40a37e817318eb1f806f3ad39ac801f26

PR #2792, opened 2019-01-28
Silence hibernate clob error with postgres

### LINKED ISSUE #2750: Hibernate with postgres throws error
With new hibernate version we're getting this error on startup, which doesn't prevent server to start:
```
Caused by: java.sql.SQLFeatureNotSupportedException: Method org.postgresql.jdbc.PgConnection.createClob() is not yet implemented.
	at org.postgresql.Driver.notImplemented(Driver.java:688)
	at org.postgresql.jdbc.PgConnection.createClob(PgConnection.java:1269)
```
There's some info here how to silence it https://github.com/pgjdbc/pgjdbc/issues/1102

Looks like we're just going down a route to change default logging for `org.hibernate.engine.jdbc.env.internal.LobCreatorBuilderImpl` to WARN as that exception is logged on INFO level and is just notification and doesn't cause app to fail to start or function properly.

jvalkeal:
- As postgres driver we use doesn't support PgConnection.createClob(),
  during a startup exception is thrown on INFO level basically just
  for notification, silence this by changing log level for that
  class to WARN.
- github.com/spring-cloud/spring-cloud-dataflow/issues/2750 has links
  to github.com/pgjdbc/pgjdbc/issues/1102 with more discussion about
  this.
- Fixes #2750

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
+++ b/spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
@@ -17,6 +17,8 @@ logging:
     org.springframework.boot.autoconfigure.security: 'INFO'
     # The following INFO is to log the embedded tomcat port
     org.springframework.boot.web.embedded.tomcat: 'INFO'
+    # Silence annoying exception trace logged on info level - gh-2750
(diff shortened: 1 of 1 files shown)


ilayaperumalg:
LGTM, merging.

PR #2796, opened 2019-01-28
Add ability to unregister all applications from shell

### LINKED ISSUE #2507: Add ability to unregister all applications from shell

chrisjs:
Resolves #2507

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppRegistryController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppRegistryController.java
@@ -336,6 +336,7 @@ public class AppRegistryController {
 	@ResponseStatus(HttpStatus.OK)
 	public void unregisterAll() {
 		List<AppRegistration> appRegistrations = appRegistryService.findAll();
+		List<AppRegistration> appRegistrationsToUnregister = new ArrayList<>();
 
 		for (AppRegistration appRegistration : appRegistrations) {
 			String applicationName = appRegistration.getName();
@@ -345,15 +346,17 @@ public class AppRegistryController {
 			if (applicationType != ApplicationType.task) {
 				String streamWithApp = findStreamContainingAppOf(applicationType, applicationName, applicationVersion);
 
-				if (streamWithApp != null) {
-					throw new UnregisterAppException(String.format("The app [%s:%s:%s] you're trying to unregister is " +
-							"currently used in stream '%s'.", applicationName, applicationType, applicationVersion,
-							streamWithApp));
+				if (streamWithApp == null) {
+					appRegistrationsToUnregister.add(appRegistration);
 				}
+			} else {
+				appRegistrationsToUnregister.add(appRegistration);
 			}
 		}
 
-		appRegistryService.deleteAll();
+		if (!appRegistrationsToUnregister.isEmpty()) {
+			appRegistryService.deleteAll(appRegistrationsToUnregister);
+		}
 	}
 
 	/**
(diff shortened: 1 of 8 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppRegistryController.java
```diff
@@ -331,6 +332,30 @@ public void unregister(@PathVariable("type") ApplicationType type, @PathVariable
...
+
+		for (AppRegistration appRegistration : appRegistrations) {
+			String applicationName = appRegistration.getName();
+			String applicationVersion = appRegistration.getVersion();
+			ApplicationType applicationType = appRegistration.getType();
+
+			if (applicationType != ApplicationType.task) {
+				String streamWithApp = findStreamContainingAppOf(applicationType, applicationName, applicationVersion);
+
+				if (streamWithApp != null) {
```
ilayaperumalg: Are we OK exiting with the exception without unregistering other apps that are not being used? If so, we could change the exception message to say something like unregister all cannot be done. 

or, We could possibly use `deleteAll(Iterable)` method from the `AppRegistrationRepository` to delete only the apps that are not being used.

jvalkeal:
I have a feeling that this feature should be handled together with UI, which just sends individual `DELETE` requests. Instead of adding feature to delete all, would it be better to just have feature to delete multiple apps with a same request as then we could potentially work on a transaction and do additional checks, etc?

jvalkeal:
Then UI could use same feature to delete multiple apps and both shell/UI would work same way. If deleting all apps, then shell could first query all apps and do a second request to delete that set. At least then you might have an option to first show that what you're actually about to delete.

ilayaperumalg:
> I have a feeling that this feature should be handled together with UI, which just sends individual DELETE requests. Instead of adding feature to delete all, would it be better to just have feature to delete multiple apps with a same request as then we could potentially work on a transaction and do additional checks, etc?

The operation `deleteAll(Iterable)` from the `AppRegistrationRepository` (- with the iterable list containing the `apps` that are not being used) can fit in this situation.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppRegistryController.java
chrisjs: i copied the logic from `unregister` unless ive overlooked something here

chrisjs: nm i get what your saying - dont stop unregistration if there are streams using apps; just dont unregister ones in use. i guess that may be a question for @sabbyanandan

sabbyanandan: +1 to not removing the apps from the registry, if they are already in use. Maybe also hint that in the return message. Something like:

> The unregister operation is complete with an exception of a few apps that are associated with the stream definition.

chrisjs: @sabbyanandan they don't get removed if they are in use. so seems like the options are:

a) do a blanket delete of apps **only** if none are assigned to streams, which is the current behavior - all or nothing.

b) delete apps that are not bound to streams, report back saying as you mentioned above something like "The unregister operation is complete with an exception of a few apps that are associated with the stream definition."

sounds like you are in favor of option, "b", right?

sabbyanandan: Yeah, #b seems to be better from UX perspective.

chrisjs: ok thx

sabbyanandan:
@jvalkeal: I created spring-cloud/spring-cloud-dataflow-ui#1049 in the UI repo, so we can use the new endpoint as opposed to making 1 delete-call for each application from the Dashboard. We will do it in 2.0 RC1.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed, rebased and merged as 400310b7

PR #2819, opened 2019-01-31
Add qualified prefix for task platform name

### LINKED ISSUE #2818: Task platform name property requires to be qualified
Currently, we avoid validating the task platform name property `spring.cloud.dataflow.task.platformName` without any qualified prefix `app.<registered-app-name>`. But, this will not work at least from the Dashboard UI since we always validate all the properties being passed. 

Hence, we need to always add `spring.cloud.dataflow.task.platformName`  with the qualified prefix.

ilayaperumalg:
- Since validation of task deployment properties is bound to happen (via Dashboard),
we always need the qualified prefix for the task platform name
 - Update test

Resolves #2818


ilayaperumalg:
Closing this for now as we can change the UI code to pass the property as is (without the prefix validation).

PR #2837, opened 2019-02-02
Streams search results are incomplete

### LINKED ISSUE #2793: Streams search results are incomplete
_From @oodamien on January 25, 2019 15:18_

The stream search is not returning the expected result.
Create a stream `foo` (file | log)

**Tests search:**
fo: no result
foo: 1 result
oo: no result

**The search should be return:**
fo: 1 result
foo: 1 result
oo: 1 result

**Note: Other tests**
fo%: 1 result
foo: 1 result
%oo: 1 result

_Copied from original issue: spring-cloud/spring-cloud-dataflow-ui#1040_

sylank:
In StreamDefinitionRepository function called findByNameLike changed to findByNameContains. Due to this modification, we can eliminate '%' character using in UI.

@ilayaperumalg I have just created a short PR to show what should we change.

resolves #2793


ilayaperumalg:
@sylank Thank you for the PR. It looks good and merging.

ilayaperumalg:
Rebased and merged as e4c69c70

PR #2841, opened 2019-02-02
[1.7.x] Update to mariadb client 2.4.0

### LINKED ISSUE #2835: Upgrade MariaDB driver to v2.4.0
Apparently, in this release, there's a fix for an issue re: TLS certs on PCF.

This also needs to be backport'd to 1.7.x.

markpollack:
Fixes #2835

### DIFF
--- a/spring-cloud-dataflow-dependencies/pom.xml
+++ b/spring-cloud-dataflow-dependencies/pom.xml
@@ -24,7 +24,7 @@
 			<dependency>
 				<groupId>org.mariadb.jdbc</groupId>
 				<artifactId>mariadb-java-client</artifactId>
-				<version>${mariadb-java-client.version}</version>
+				<version>2.4.0</version>
 			</dependency>
 			<dependency>
 				<groupId>org.hibernate</groupId>
--- a/spring-cloud-dataflow-dependencies/pom.xml
+++ b/spring-cloud-dataflow-dependencies/pom.xml
@@ -24,7 +24,7 @@
 			<dependency>
 				<groupId>org.mariadb.jdbc</groupId>
 				<artifactId>mariadb-java-client</artifactId>
-				<version>2.2.5</version>
+				<version>${mariadb-java-client.version}</version>
 			</dependency>
 			<dependency>
 				<groupId>org.hibernate</groupId>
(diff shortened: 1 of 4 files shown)


spring-cloud-dataflow-dependencies/pom.xml
```diff
@@ -24,7 +24,7 @@
 			<dependency>
 				<groupId>org.mariadb.jdbc</groupId>
 				<artifactId>mariadb-java-client</artifactId>
-				<version>2.2.5</version>
+				<version>${mariadb-java-client.version}</version>
```
jvalkeal: We can't use `mariadb-java-client.version` here as it's not defined in parent pom.

spring-cloud-dataflow-server-core/pom.xml
```diff
@@ -147,6 +147,7 @@
 		<dependency>
 			<groupId>org.mariadb.jdbc</groupId>
 			<artifactId>mariadb-java-client</artifactId>
+            <version>${mariadb-java-client.version}</version>
```
jvalkeal: Version would already come from parent pom's dep management?

jvalkeal:
I was too quick to merge that other PR as it had same issues. This PR should also follow fixes I did in PR #2846

markpollack:
Thanks, I always seem not to be successful `maven pom` surgeon. :(

spring-cloud-dataflow-server-core/pom.xml
markpollack: I just try to be explicit, can test.

spring-cloud-dataflow-dependencies/pom.xml
markpollack: That sort of sucks, yet it is building fine.  While I don't think we should change the parent pom here in the 1.7.x branch, maybe we should not have `spring-cloud-dependencies-parent` as the parent for `spring-cloud-dataflow-dependencies` in the master branch for 2.0?

spring-cloud-dataflow-server-core/pom.xml
markpollack: seems to work w/o it, so removing it.

markpollack:
closing in favor of https://github.com/spring-cloud/spring-cloud-dataflow/pull/2865

PR #2865, opened 2019-02-07
Update MariaDB client to 2.4.0

markpollack:


jvalkeal:
think this should be ok. I'll test and merge.

PR #2878, opened 2019-02-11
Database migration from classic mode

### LINKED ISSUE #2826: Database migration from classic mode
With 1.7.x classic mode only `URI_REGISTRY` table is used and `APP_REGISTRATION` stays empty as it's only used with skipper mode. As 2.x requires skipper, `URI_REGISTRY` table is not currently migrated.

Should try to tweak flyway migrations so that registered apps are in place with this combination as well.

### LINKED ISSUE #2863: Possible double insert into hibernate_sequence with mysql
Both dataflow and skipper does a simple `insert into hibernate_sequence values ( 1 )` which is causing two rows in a table as mysql don't have a proper sequence. Should make sure that double insert doesn't happen.

### LINKED ISSUE #2868: Migration clob fails with postgresql
For tables in 1.7.x like `STREAM_DEFINITIONS`, some field types are `text` as they are also in 2.x with JPA. However in 2.x entity class type for field is `@Lob` which causes wrapped CLOB and data between migrated tables cannot be copied as is.

Reason for this is because with postgres real data in a field is a long which is an oid to a large object.

jvalkeal:
- As mysql don't have a real sequence, initial value
  needs to be inserted. This change makes sure that
  we only insert value once and that value will be 1.
  Similar fix will also be done on a skipper side.
- Fix clob handling with postgres.
- New base impl AbstractMigrateUriRegistrySqlCommand which
  all db's implement to support migration from URI_REGISTRY
  to app_registration table.
- Fixes #2863
- Fixes #2826
- Fixes #2868

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/db/migration/AbstractMigrateUriRegistrySqlCommand.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/db/migration/AbstractMigrateUriRegistrySqlCommand.java
@@ -1,0 +1,211 @@
+/*
+ * Copyright 2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.cloud.dataflow.server.db.migration;
+
+import java.sql.Connection;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.springframework.cloud.dataflow.core.ApplicationType;
+import org.springframework.cloud.dataflow.registry.support.AppResourceCommon;
+import org.springframework.cloud.deployer.resource.maven.MavenProperties;
+import org.springframework.cloud.deployer.resource.support.DelegatingResourceLoader;
+import org.springframework.jdbc.core.JdbcTemplate;
+
+/**
+ * Base implementation for a {@link SqlCommand} copying data from
+ * {@code URI_REGISTRY} table into {@code app_registration} table.
+ *
+ * @author Janne Valkealahti
+ *
+ */
+public abstract class AbstractMigrateUriRegistrySqlCommand extends SqlCommand {
+
+	private static final Logger logger = LoggerFactory.getLogger(AbstractMigrateUriRegistrySqlCommand.class);
+
+	@Override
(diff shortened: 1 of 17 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/db/migration/AbstractMigrateUriRegistrySqlCommand.java
```diff
@@ -0,0 +1,211 @@
...
+				if (!defaultVersions.contains(key)) {
+					armd.setDefaultVersion(true);
+					defaultVersions.add(key);
+				}
+				if (split.length == 2) {
+					// we got *.* first
+					armd.setName(split[1]);
+					armd.setType(ApplicationType.valueOf(split[0]).ordinal());
+					armd.setUri(uri);
+					armd.setVersion(arc.getResourceVersion(arc.getResource(uri)));
```
markpollack: I guess we need to discuss how to inform users who are in classic mode and don't have a version field that their stream def isn't going to be migrated.  Also, I believe the old table gets dropped, so we prob shouldn't do that if we could not migrate a registered app....

ilayaperumalg:
I went through the upgrade scenarios with MySQL, DB2, Postgres and MSSQL so far.

With MSSQL, I see some issues after the migration.

Originally I had SCDF classic mode server `1.7.4.RELEASE` with a stream definition:

```
dataflow:>stream list
╔═══════════╤═════════════════╤═════════════════════════════════════════╗
║Stream Name│Stream Definition│                 Status                  ║
╠═══════════╪═════════════════╪═════════════════════════════════════════╣
║a1         │time | log       │The stream has been successfully deployed║
╚═══════════╧═════════════════╧═════════════════════════════════════════╝

```

After the migration on SCDF 2.x, I see this:

```
dataflow:>stream list
Command failed org.springframework.cloud.dataflow.rest.client.DataFlowClientException: 115E:(pos 0): unexpected data in stream definition '楴'
楴敭簠氠杯
^
```

The `app list` shows the correct information after the migration though.

jvalkeal:
Haven't seen that and just tried it manually. Maybe something to do with DB language settings running on your system.

jvalkeal:
You'd need poke around database tables to see if there's something weird.

ilayaperumalg:
The database table `stream_definitions` seems to have the correct entry:

```
definition_name                                                                                                                                                                                                                                                 definition
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
a1                                                                                                                                                                                                                                                              time | log

(1 rows affected)
```

ilayaperumalg:
If the SCDF server is started before Skipper server (with SQLServer) then we see this:

```
Caused by: org.flywaydb.core.api.FlywayException: Error while executing beforeBaseline callback: StatementCallback; uncategorized SQLException for SQL [select next value for hibernate_sequence]; SQL state [S0001]; error code [11726]; Object 'hibernate_sequence' is not a sequence object.; nested exception is com.microsoft.sqlserver.jdbc.SQLServerException: Object 'hibernate_sequence' is not a sequence object.

```

ilayaperumalg:
Created https://github.com/spring-cloud/spring-cloud-dataflow/issues/2903 for the hibernate_sequence table issue.

ilayaperumalg:
MSSQL server issue mentioned above disappeared after using the correct SQL server docker image that has the necessary temp tables. 

LGTM, merging this PR.

ilayaperumalg:
Rebased and merged as 9cdef0f7

PR #2879, opened 2019-02-12
Remove local from pom and package names

### LINKED ISSUE #2810: Remove use of world 'local' in poms and classes
`	<artifactId>spring-cloud-dataflow-server-local-autoconfig</artifactId>` should change to `single-autoconfig`

and classes prefixed with `local` should be changed to `single`.

jvalkeal:
- Rename spring-cloud-dataflow-server-local-autoconfig to spring-cloud-dataflow-autoconfigure
  which would be the one where all boot related autoconfiguration classess should eventually go.
- Rename spring-cloud-starter-dataflow-server-local to spring-cloud-starter-dataflow-server as
  we're really only building one server.
- Fixes #2810


pom.xml
```diff
@@ -72,7 +72,7 @@
 		<module>spring-cloud-dataflow-registry</module>
 		<module>spring-cloud-dataflow-rest-resource</module>
 		<module>spring-cloud-dataflow-server-core</module>
-		<module>spring-cloud-dataflow-server-local-autoconfig</module>
+		<module>spring-cloud-dataflow-autoconfigure</module>
```
ilayaperumalg: Can we make this as `spring-cloud-dataflow-server-autoconfigure`?

jvalkeal: Point is that my hope is that eventually, maybe somewhere in 2.1.x line we have one autoconfigure jar instead of 7 we use now.

jvalkeal: We need to stop doing this:
```
$ grep EnableAutoConfiguration */src/main/resources/META-INF/spring.factories
spring-cloud-dataflow-configuration-metadata/src/main/resources/META-INF/spring.factories:org.springframework.boot.autoconfigure.EnableAutoConfiguration:\
spring-cloud-dataflow-platform-cloudfoundry/src/main/resources/META-INF/spring.factories:org.springframework.boot.autoconfigure.EnableAutoConfiguration=\
spring-cloud-dataflow-platform-kubernetes/src/main/resources/META-INF/spring.factories:org.springframework.boot.autoconfigure.EnableAutoConfiguration=\
spring-cloud-dataflow-rest-client/src/main/resources/META-INF/spring.factories:org.springframework.boot.autoconfigure.EnableAutoConfiguration=\
spring-cloud-dataflow-server-core/src/main/resources/META-INF/spring.factories:org.springframework.boot.autoconfigure.EnableAutoConfiguration=\
spring-cloud-dataflow-server-local-autoconfig/src/main/resources/META-INF/spring.factories:org.springframework.boot.autoconfigure.EnableAutoConfiguration=\
spring-cloud-dataflow-shell-core/src/main/resources/META-INF/spring.factories:org.springframework.boot.autoconfigure.EnableAutoConfiguration=\
```
Every separate jar makes app startup slower.

ilayaperumalg:
Rebased and merged as 9c873735

PR #2880, opened 2019-02-12
Refactor DefaultTaskExecutionService's constructor

### LINKED ISSUE #2768: Refactor Constructor in DefaultTaskExecutionService
Refactor the constructor so that `ApplicationConfigurationMetadataResolver`, `String dataflowServerUri` and `CommonApplicationProperties commonApplicationProperties` can be moved into a class like the current `AppDeploymentRequestCreator` but for Tasks.

Since curing AppDeploymentRequest creation, there needs to be a write to the database, we can create a `TaskAppDeploymentRequestService` and use @Transactional annotation support.

This will also make that part of the code more easily testable.

cppwfs:
resolves #2768

When creating the TaskAppDeploymentRequestCreator I couldn't see the need for creating a service with @Transactional.
It could be referring to the creation of the TaskExecution that does require a new transaction, however this transaction information is required by the Auditing and for post task launch request processing.  Similarly TaskExecutionInformation is required for both request creation and auditing.

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/TaskConfiguration.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/TaskConfiguration.java
@@ -60,6 +60,7 @@ import org.springframework.cloud.dataflow.server.service.impl.DefaultTaskExecuti
 import org.springframework.cloud.dataflow.server.service.impl.DefaultTaskExecutionService;
 import org.springframework.cloud.dataflow.server.service.impl.DefaultTaskJobService;
 import org.springframework.cloud.dataflow.server.service.impl.DefaultTaskSaveService;
+import org.springframework.cloud.dataflow.server.service.impl.TaskAppDeploymentRequestCreator;
 import org.springframework.cloud.dataflow.server.service.impl.TaskConfigurationProperties;
 import org.springframework.cloud.deployer.spi.local.LocalDeployerProperties;
 import org.springframework.cloud.deployer.spi.local.LocalTaskLauncher;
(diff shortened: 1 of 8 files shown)


spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/configuration/TaskServiceDependencies.java
```diff
@@ -216,20 +217,25 @@ public TaskExecutionCreationService taskExecutionRepositoryService(TaskRepositor
 		return new DefaultTaskExecutionRepositoryService(taskRepository);
 	}
 
+	@Bean
+	TaskAppDeploymentRequestCreator taskAppDeploymentRequestCreator(
```
ilayaperumalg: Is there any reason why this is not `public`?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/TaskConfiguration.java
```diff
@@ -164,19 +165,26 @@ public TaskExecutionCreationService taskExecutionRepositoryService(TaskRepositor
 		return new DefaultTaskExecutionRepositoryService(taskRepository);
 	}
 
+	@Bean
+	TaskAppDeploymentRequestCreator taskAppDeploymentRequestCreator(
```
ilayaperumalg: Is there any reason why this is not `public`?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/TaskAppDeploymentRequestCreator.java
```diff
@@ -0,0 +1,119 @@
...
+import org.springframework.cloud.dataflow.server.config.apps.CommonApplicationProperties;
+import org.springframework.cloud.dataflow.server.controller.WhitelistProperties;
+import org.springframework.cloud.deployer.spi.core.AppDefinition;
+import org.springframework.cloud.deployer.spi.core.AppDeploymentRequest;
+import org.springframework.cloud.task.repository.TaskExecution;
+import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
+
+/**
+ * Create the list of {@link AppDeploymentRequest}s
```
ilayaperumalg: For the task, Isn't this always a single `AppDeploymentRequest` instead of `list`?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/TaskAppDeploymentRequestCreator.java
```diff
@@ -0,0 +1,119 @@
...
+	private final CommonApplicationProperties commonApplicationProperties;
+
+	private final WhitelistProperties whitelistProperties;
+
+	private final String dataflowServerUri;
+
+	/**
+	 * Initializes the {@link TaskAppDeploymentRequestCreator}.
+	 *
+	 * @param metaDataResolver the metadata resolver
```
ilayaperumalg: The order is incorrect.

cppwfs: Updated.  Thanks!

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/TaskAppDeploymentRequestCreator.java
cppwfs: OOPS!   Thinking streams, Updated thanks!

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/TaskConfiguration.java
cppwfs: ooops update!

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/configuration/TaskServiceDependencies.java
cppwfs: oops updated!

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed and merged as e81ef345

PR #2911, opened 2019-02-15
Running kubernetes non-root with RabbitMQ.

### LINKED ISSUE #2881: Run Applications as Non Root
**Problem description:**
We are trying to get Spring Cloud Data Flow to run in Kubernetes. Our organization's Kubernetes cluster has a `PodSecurityPolicy` configured such that the pod containers must run as non-root users. However, I have not found a way to configure the stream applications to run as non-root user.

**Solution description:**
It would be great to have a way to set the Kubernetes pod's `spec.securityContext.runAsUser` configuration. Or even the container level securiryContext (`spec.containers[].securityContext.runAsUser`).  The `spec.securityContext.fsGroup` configuration should probably be configurable as well.

It's probably a good idea to have applications deployed to run as non-root as the default behavior for the Spring Cloud Dataflow server/skipper.

**Additional context:**
Running as non-root is a security best practice for Kubernetes/containers. I was able to configure the Kubernetes definition files (for server, skipper, mysql, kafka) to run the different SCDF components as non-root in Kubernetes.  I had to switch Kafka to a different image from what is currently defined in your repo. The Kafka image that is currently there now (wurstmeister/kafka:1.1.0) creates files as root during the image creation but it then needs to modify it after startup. 
I wonder if running as non-root is something that is worth pursuing as a default configuration for your Kubernetes implementation.

allanmoso:
Includes changes to kubernetes manifest files to have the images run as non-root users. It also includes changes that are needed to be made as a consequence of using the non-root users(e.g., non-root users cannot bind on ports below 1024).

I did not create new users in the images but instead used the users that are already available in the images. However, most of those built-in users do not have home directories defined. The SCDF server uses the home directory of the user for some of its function (e.g. maven). I had to override it to use `/tmp` instead.

Resolves #2881

### DIFF
--- a/src/kubernetes/mysql/mysql-deployment.yaml
+++ b/src/kubernetes/mysql/mysql-deployment.yaml
@@ -11,6 +11,9 @@ spec:
       labels:
         app: mysql
     spec:
+      securityContext:
+        runAsUser: 999
+        fsGroup: 999
       containers:
       - image: mysql:5.6
         name: mysql
--- a/src/kubernetes/rabbitmq/rabbitmq-deployment.yaml
+++ b/src/kubernetes/rabbitmq/rabbitmq-deployment.yaml
@@ -11,6 +11,9 @@ spec:
       labels:
         app: rabbitmq
     spec:
+      securityContext:
+        runAsUser: 999
+        fsGroup: 999
       containers:
       - image: rabbitmq:3.6.10
         name: rabbitmq
--- a/src/kubernetes/server/server-deployment.yaml
+++ b/src/kubernetes/server/server-deployment.yaml
@@ -14,12 +14,15 @@ spec:
       labels:
         app: scdf-server
     spec:
+      securityContext:
+        runAsUser: 65534
+        fsGroup: 65534
       containers:
       - name: scdf-server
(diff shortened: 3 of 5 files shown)


pivotal-issuemaster:
@allanmoso Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-dataflow&pullRequestId=2911)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-dataflow&pullRequestId=2911) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@allanmoso Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-dataflow&pullRequestId=2911)!

chrisjs:
@allanmoso thanks for making this. ill take a look soon and work out any additional parts and update when i have more info.

chrisjs:
closing; see https://github.com/spring-cloud/spring-cloud-dataflow/issues/2881#issuecomment-479139155 for more info

PR #2912, opened 2019-02-15
Fix docs to run server locally against a remote CF environment

### LINKED ISSUE #2858: Docs: Instructions for CF Local Installation not working
The instructions provided at:

http://docs.spring.io/spring-cloud-dataflow/docs/2.0.0.BUILD-SNAPSHOT/reference/htmlsingle/#getting-started-cloudfoundry-on-local

*Do not work* - You cannot specify square brackets on the command line and will experience `not a valid identifier` in _bash_.

We should possibly provide a `yml` example instead in the docs.

See also:

- https://github.com/spring-projects/spring-boot/issues/5315
- https://github.com/spring-projects/spring-boot/wiki/Relaxed-Binding-2.0

Also, replace 1 occurrence of `spring.config.location` in the docs at:

http://docs.spring.io/spring-cloud-dataflow/docs/2.0.0.BUILD-SNAPSHOT/reference/htmlsingle/#dataflow-multiple-brokers

That should be replaced with: `spring.config.additional-location`

sabbyanandan:
I attempted to fix the export-commands that aren't bash friendly by having them as env-vars in a property file. 

I verified the change locally by starting the server pointing to a CF environment. I had confirmed it with a successful start along with the properties loaded up as a map at: http://localhost:9393/management/configprops.

Resolves #2858

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/appendix-howto.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/appendix-howto.adoc
@@ -333,7 +333,7 @@ These properties can be supplied in a `.properties` file that is accessible to t
 
 [source,bash,subs=attributes]
 ----
-java -jar spring-cloud-dataflow-server/target/spring-cloud-dataflow-server-{project-version}.jar --spring.config.location=<PATH-TO-FILE>/foo.properties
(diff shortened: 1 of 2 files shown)


ilayaperumalg:
LGTM, merging.

PR #2916, opened 2019-02-17
Updated version of image used for docker test

cppwfs:


jvalkeal:
There was some discussion in slack for this and it's related to #2917. Looks like this fixes current build, merging..

PR #2920, opened 2019-02-18
Fix sqlserver hibernate sequence

### LINKED ISSUE #2903: Migration to 2.x: SCDF server with sqlserver fails to start when it is started before Skipper
The SCDF server has this exception message and exits:

```
Caused by: org.flywaydb.core.api.FlywayException: Error while executing beforeBaseline callback: StatementCallback; uncategorized SQLException for SQL [select next value for hibernate_sequence]; SQL state [S0001]; error code [11726]; Object 'hibernate_sequence' is not a sequence object.; nested exception is com.microsoft.sqlserver.jdbc.SQLServerException: Object 'hibernate_sequence' is not a sequence object.
```

jvalkeal:
- As hibernate_sequence in old classic mode is already
  wrong, we need to fix it with baseline as well which
  happens before migration where this same fix applies.
- Fixes #2903

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/db/migration/sqlserver/MsSqlMigrateUriRegistrySqlCommand.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/db/migration/sqlserver/MsSqlMigrateUriRegistrySqlCommand.java
@@ -33,6 +33,20 @@ public class MsSqlMigrateUriRegistrySqlCommand extends AbstractMigrateUriRegistr
 
 	@Override
 	protected void updateAppRegistration(JdbcTemplate jdbcTemplate, List<AppRegistrationMigrationData> data) {
+
+		// check and do same fix as in R__Hibernate_Sequence
+		// needed if we're coming from classic mode and skipper is not
+		// started first
+		boolean fixHibernateSequence = false;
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/db/migration/sqlserver/MsSqlMigrateUriRegistrySqlCommand.java
```diff
@@ -33,6 +33,20 @@
...
 	protected void updateAppRegistration(JdbcTemplate jdbcTemplate, List<AppRegistrationMigrationData> data) {
+
+		// check and do same fix as in R__Hibernate_Sequence
+		// needed if we're coming from classic mode and skipper is not
+		// started first
+		boolean fixHibernateSequence = false;
+		try {
+			jdbcTemplate.execute("select 1 from hibernate_sequence");
+			fixHibernateSequence = true;
+		} catch (Exception e) {
```
ilayaperumalg: Should we throw an exception or at least some error logging?

ilayaperumalg:
LGTM, verified the changes with the migration (and also starting the SCDF server before Skipper). Merging.

ilayaperumalg:
Rebased and merged as 6f0dc64e

PR #2921, opened 2019-02-18
Update to Einstein.GA bit.ly links

sabbyanandan:


ilayaperumalg:
It looks like there are few other places require updates for the same. Will submit a separate PR for that. Merging this for now.

PR #2936, opened 2019-02-20
Any property suffixed with 'url' will be redacted

### LINKED ISSUE #2892: Password is not masked for the "spring.datasource.url" field in the audit-history
As a user, I launched Tasks against a common datasource, and while reviewing the results in audit-history, I notice `spring.datasource.url` field with the username/password in clear-text - it should not be.

The other fields such as `spring.datasource.username` and `spring.datasource.password` are masked correctly.

I see this behavior both in the REST response and in the UI.

Note:
We attempted to fix related areas in https://github.com/spring-cloud/spring-cloud-dataflow/issues/2516; a similar fix may need to be extended to Audit flows, too.

cppwfs:
resolves #2892


sabbyanandan:
I randomly tried a bunch of custom properties with the keywords that we mask, and I was able to verify that they are masked as expected.

1) **Task execution:**
image

2) **Audit record:**
image

sabbyanandan:
I see the same behavior in Shell, too.

PR #2942, opened 2019-02-21
Skip importing missing versions

### LINKED ISSUE #2933: UriRegistry migration without version
Currently if `URI_REGISTRY` table contains an address without version, migration fails. Most of a databases like postgres handles ddl error within transaction well, meaning user can fix that table and re-start and migration continues. I.e mysql blows up as ddl operations cannot be done in a transaction.

Having said that. Trying this fails anyway:
```
dataflow:>app register --name time --type source --uri http://localhost/time-source-rabbit-2.0.1.RELEASE.jar
dataflow:>app register --name log --type sink --uri http://localhost/log-sink-rabbit.jar

dataflow:>stream create --name ticktock --definition "time|log" --deploy 
Command failed org.springframework.cloud.dataflow.rest.client.DataFlowClientException: Could not parse version from http://localhost/log-sink-rabbit.jar, expected format is <artifactId>-<version>.jar
```

So I'm kinda questioning if we could simply skip migrating those apps without version as with dataflow 1.7.x those cannot get deployed with classic mode anyway.

jvalkeal:
- Modify URI_REGISTRY migration so that now we just log
  a warning if version cannot be calculated using a same way
  dataflow already expects during a deployment. As in this
  case version in a data item is left empty, only use
  those items where version is defined.
- Fixes #2933

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/db/migration/AbstractMigrateUriRegistrySqlCommand.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/db/migration/AbstractMigrateUriRegistrySqlCommand.java
@@ -16,12 +16,12 @@
 package org.springframework.cloud.dataflow.server.db.migration;
 
 import java.sql.Connection;
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -31,6 +31,7 @@ import org.springframework.cloud.dataflow.registry.support.AppResourceCommon;
 import org.springframework.cloud.deployer.resource.maven.MavenProperties;
 import org.springframework.cloud.deployer.resource.support.DelegatingResourceLoader;
 import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.util.ObjectUtils;
 
 /**
  * Base implementation for a {@link SqlCommand} copying data from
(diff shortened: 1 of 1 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/db/migration/AbstractMigrateUriRegistrySqlCommand.java
```diff
@@ -122,7 +123,12 @@ public boolean canHandleInJdbcTemplate() {
 					armd.setName(split[1]);
 					armd.setType(ApplicationType.valueOf(split[0]).ordinal());
 					armd.setUri(uri);
-					armd.setVersion(arc.getResourceVersion(arc.getResource(uri)));
+					try {
+						armd.setVersion(arc.getResourceVersion(arc.getResource(uri)));
+					}
+					catch (Exception e) {
+						logger.warn("Skipping URI_REGISTRY item {} migration with uri {} due to error", name, uri, e);
```
ilayaperumalg: Should this be ERROR level instead of warning? Also, can we say this requires the app to be re-registered with a version for the deployment in Skipper?

markpollack: we should explain the error better, but a resource that has no version would never be able to be deployed even in classic mode.  The error occurs at deployment time.

markpollack: `Skipping URI_REGISTRY item {} migration with uri {} due to lack of version number in uri.`

markpollack: I would leave it as a warning level.  That resource would never have been able to have been deployed in anycase.

ilayaperumalg: ok, will update the log message and merge.

ilayaperumalg:
Rebased and merged as ff59fad9

Issue #2851, opened 2019-02-05
Epic: Address TODOs in scope for v2.0


sabbyanandan:
See [here](https://github.com/spring-cloud/spring-cloud-dataflow/search?q=todo&unscoped_q=todo).

sabbyanandan:
The immediate scope for v2.0 is complete. Marking this as done.

PR #2948, opened 2019-02-22
Remove unused file/single security tests

jvalkeal:
- These are already ingored and not coming back
  in any form.
- Relates to #2851

child of #2851


ilayaperumalg:
LGTM

PR #2952, opened 2019-02-25
Added execution id to shell response of task launch

### LINKED ISSUE #2877: Display execution id for launched task in the shell
Currently if I launch a task from the shell it does not return the execution id (it is returned in the rest API) 

It would be useful if the id was returned meaning that the user can then directly query the execution status eg 

```
dataflow:> task launch test-task
Launched task 'test-task' with id 3035
dataflow:> task execution status --id 3035
```

cppwfs:
resolves #2877


ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Rebased and merged as 99b6615b

PR #2957, opened 2019-02-26
Fixes applied to task and task dev guide sections

cppwfs:


spring-cloud-dataflow-docs/src/main/asciidoc/tasks.adoc
```diff
@@ -20,7 +20,7 @@ The entry point for code execution in a Spring Cloud Task application is most of
 The Spring Batch project is probably what comes to mind for Spring developers writing short lived applications.
 Spring Batch provides a much richer set of functionality than Spring Cloud Task and is recommended when processing large volumes of data.
 A use case might be to read many CSV files, transform each row of data, and write each transformed row to a database.
-Spring Batch provides its own database schema with a much more rich https://docs.spring.io/spring-batch/{spring-batch-version}/reference/html/schema-appendix.html#metaDataSchema[set of information] about the execution of a Spring Batch job.
+Spring Batch provides its own database schema with a much more rich https://docs.spring.io/spring-batch/{spring-batch-doc-version}/reference/html/schema-appendix.html#metaDataSchema[set of information] about the execution of a Spring Batch job.
```
sabbyanandan: Any reason why we cannot pull it this doc from the same version as `spring-batch-version` instead of `4.1.x`? Shouldn't they be one and the same?

cppwfs: Spring Batch uses a different format for designating the version in the URL for the documentation. It uses `4.1.x`.  While in our documentation we usually version for each release, for example `4.1.1.RELEASE`.   So to overcome this difference, we have to use the version designation that Spring-Batch utilizes.

cppwfs:
Closing this PR to aggregate its changes into #2962

PR #2958, opened 2019-02-26
Add support to disable ProfileApplicationListener via env-var

### LINKED ISSUE #2955: Add support to disable ProfileApplicationListener via env-vars
The current approach uses system properties, but those can not be set on k8s/cf.

Use

`SPRING_CLOUD_DATAFLOW_SERVER_PROFILEAPPLICATIONLISTENER_IGNORE`

markpollack:
Added implementation.
Adding tests was a rabbit hole due to us using mockito's mock-maker-inline, which does not get along well with jmockit to mock env-vars.
The same code is tested in skiper FWIW

Fixes #2955

### DIFF
--- a/spring-cloud-dataflow-autoconfigure/src/main/java/org/springframework/cloud/dataflow/autoconfigure/local/ProfileApplicationListener.java
+++ b/spring-cloud-dataflow-autoconfigure/src/main/java/org/springframework/cloud/dataflow/autoconfigure/local/ProfileApplicationListener.java
@@ -51,6 +51,9 @@ public class ProfileApplicationListener
 	 * CloudProfileProvider implementations discovered from the ServiceLoader.
 	 */
 	public static final String IGNORE_PROFILEAPPLICATIONLISTENER_PROPERTY_NAME = "spring.cloud.dataflow.server.profileapplicationlistener.ignore";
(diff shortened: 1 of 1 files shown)


jvalkeal:
As same code, review open on a skipper side.

spring-cloud-dataflow-autoconfigure/src/main/java/org/springframework/cloud/dataflow/autoconfigure/local/ProfileApplicationListener.java
```diff
@@ -104,6 +108,20 @@ public void onApplicationEvent(ApplicationEnvironmentPreparedEvent event) {
 
 	}
 
+	private boolean ignoreFromSystemProperty() {
+		return Boolean.getBoolean(IGNORE_PROFILEAPPLICATIONLISTENER_PROPERTY_NAME);
+	}
+
+	private boolean ignoreFromEnvironmentVariable() {
```
jvalkeal: I'll change this as it was changed in a skipper PR

jvalkeal:
Merged per 36c72c4ea4d11eab835c95c426cb8cc6e2d504d9

PR #2960, opened 2019-02-27
Resolves problem with testTaskLaunch

cppwfs:
While running locally the task execution id was 1. But in CI we have pre-existing task launches.
To resolve this, the test was updated to retrieve a task execution list for a unique test name.  And then the execution id is extracted.
This extracted execution id is then used to determine success


jvalkeal:
lgtm

Issue #2963, opened 2019-02-28
Security: URI `/tasks/platforms` not covered 


ghillert:
We should add mapping: 

`- GET    /tasks/platforms/**             => hasRole('ROLE_VIEW')`

to https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml

Without this setting, the launching of tasks in the UI does not work.

See also: 
https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskPlatformController.java

PR #2969, opened 2019-02-28
gh-2963 Fix Security: URI `/tasks/platforms` not covered

ghillert:
* Add rule to `dataflow-server-defaults.yml`
* Add tests to `LocalServerSecurityWithUsersFileTests`

resolves https://github.com/spring-cloud/spring-cloud-dataflow/issues/2963


ilayaperumalg:
LGTM, merging.

PR #2974, opened 2019-03-03
Fix incorrect application references

sabbyanandan:
Following incorrect references were corrected.

1) Fix duplicate rows for "bridge" and "aggregator"
2) Add missing apps ("pmml", "counter", "tasklauncher-dataflow")
3) Add deprecations where applicable

There were a few references to Clark train and that's fixed now.

Finally, also, I flipped Elston from RC1 to GA albeit the release is not done yet. Elston is waiting on SCDF 2.0 (CTR 2.1 GA depends on SCDF 2.0 GA) -> 🐔 or 🥚? When we release SCDF 2.0, it'd be better not to have RC1 references.


ilayaperumalg:
LGTM, merging.

PR #2980, opened 2019-03-04
Switch from MySQL to PostgreSQL in PWS sample

### LINKED ISSUE #2977: update PCF getting started guide to reference postgresql instead of mysql
mysql (paid or free) does not work for data flow on PCF.

free - too few connections
paid - 5 year old version of mysql that doesn't like our schema.

sabbyanandan:
I switched to a paid plan of PostgreSQL that would work with SCDF v2.0.

> cf create-service elephantsql panda my_postgres

I added a note re: concurrent connections, too.

Also, I switched the references from mysql to postgresql where applicable.  

Resolves #2977.


markpollack:
merged as 5bd87b97d0a470af6fa7adbd5436a86092f1b2cd

PR #2994, opened 2019-03-07
Change from http to https in pom.xml for repos

### LINKED ISSUE #2993: Change from http to https in pom.xml for repos

markpollack:
Fixes #2993


ilayaperumalg:
Noticed there are few other places we can add a similar fix. Will update the commit when merging.

ilayaperumalg:
Updated and merged as baef04c4

Issue #2836, opened 2019-02-02
Use json path matcher in controller test classes


sylank:
**Problem description:**
It would be nice if we could use json path matcher in `AppRegistryControllerTests`, `RuntimeAppsControllerTests`, ` RuntimeStreamsControllerTests`, `StreamControllerTests`, `TaskPlatformControllerTests`. Check `AuditRecordControllerTests` for more information.

**Solution description:**
Check the controller test classes and the others where we use `mockMvc.perform(...)` and change the assertion to `jsonPath` assertion if it is possible. I think the test code could be cleaner. For more information and example implementation check `AuditRecordControllerTests` class.

- [x] AppRegistryControllerTests
- [x] RuntimeAppsControllerTests
- [x] RuntimeStreamsControllerTests
- [x] StreamControllerTests
- [x] TaskPlatformControllerTests
- [x] StreamDefinitionControllerTests

sabbyanandan:
Hi, @sylank. Please feel free to take a stab at it; we can collaborate on the PR.

PR #3017, opened 2019-03-14
Use json path matcher in controller test classes

sylank:
This PR contains controller test class refactor and I removed an audit record insertion because it was called twice.

resolves: #2836


sylank:
Something went wrong during the build process in TravisCI, I don't really understand the cause of the error.

jvalkeal:
Thx finding it out, lgtm, merging...

jvalkeal:
Merged per 5e858def1bed88334c597cf5b4bb7751f99c68f7

PR #3020, opened 2019-03-15
Remove H2 related task configs

### LINKED ISSUE #3019: Remove H2 configs for tasks
As we don't support running H2 as non-embedded mode, `H2ServerConfiguration` and `NoH2ServerConfiguration` should go away.

jvalkeal:
- Move JobRepositoryFactoryBean out from custom H2 configs
  as we don't no longer support H2 as non-embedded mode. This
  removes all tweaks and batch/task db initializers should not
  exist at all as db is managed by flyway.
- Remove spring.dataflow.embedded.database.enabled from dataflow-server.yml
  and WebConfiguration as setting itself is now pointless.
- We keep actual datasource config in dataflow-server.yml as that
  together with WebConfiguration starts H2 with tcp mode.
- Fixes #3019

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/configuration-local.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/configuration-local.adoc
@@ -83,12 +83,7 @@ java -jar spring-cloud-dataflow-server/target/spring-cloud-dataflow-server-{proj
     --spring.datasource.driver-class-name=org.hsqldb.jdbc.JDBCDriver &
 ----
 
-NOTE: If you wish to use an external H2 database instance instead of the one
-embedded with Spring Cloud Data Flow, set the
-`spring.dataflow.embedded.database.enabled` property to false.  If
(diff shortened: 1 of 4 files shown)


cppwfs:
The PR looks LGTM (haven't kicked the tire s on it). The only issue I see is that our documentation still states that we support external H2 (Non-Production). Found here: https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-docs/src/main/asciidoc/configuration-local.adoc . Do we need to update the docs to explicitly state we don't support external H2?

jvalkeal:
I think we should as it would not work for H2. Not sure if we should open a channel to boot guys as they previously denied tcp config in https://github.com/spring-projects/spring-boot/issues/7190. Fact is that as of now we tried H2 as fully embedded and if one tries to use it as external, flyway will blow up.

cppwfs:
LGTM 
Tested, Rebased, Squashed, Merged

PR #3035, opened 2019-03-18
Add SCDF docker-compose with Prometheus

### LINKED ISSUE #3034: Configure the getting started docker-compose to use Prometheus instead of (or in addition to)  InfluxDB
**Problem description:**
SCDF 2.0 provides OOTB support for the InfuxDB and Prometheus time-series databases (TSDB). 
Currently the [getting started docker-compose](https://docs.spring.io/spring-cloud-dataflow/docs/current/reference/htmlsingle/#getting-started-local-deploying-spring-cloud-dataflow-docker) is pre-configured with` InfluxDB`, which has some significant limitations when it comes to arrange and order time series. Furthermore it seems that the majority of the dev community has converged to Prometheus as TSDB.   
Therefore it would beneficial to have by default the getting started experience be based on Prometheus rather than InfluxDB.  Or better we can have docker compose scripts for both. 

**Solution description:**
Unlike InfluxDB (which is push-based TSDB), Prometheus is pull-based datastore that continuously pulls the metrics from the SCDF app starters. To enable this an additional Service-Discovery is used to retrieve the urls of the deployed app starters and share them with Prometheus. Later regularly scrapes the `/actuator/prometheus` entry points of the provided urls.

tzolov:
- add new docker-compose-prometheus.yml that uses Prometheus instead of InfluxDB
  	- internally uses the file service-discovery application that continuously retrieves the apps URL entry points (form SCDF REST) and share them with prometheus as targets.json file.
	- configure Skipper's local deployer to use a constrained port range (20000-20105) for the apps deployed.
  - Build and publish on dockerhub a new scdf-prometheus-local docker image, that extends the original prometheus image with custom configuration based on file sd.

  Resolves #3034


spring-cloud-dataflow-server/docker-compose-prometheus.yml
```diff
@@ -0,0 +1,99 @@
...
+    image: springcloud/openjdk:latest
+    depends_on:
+      - dataflow-server
+    command: >
+      /bin/sh -c "
+        while ! nc -z dataflow-server 9393;
+        do
+          sleep 1;
+        done;
+        wget -qO- 'http://dataflow-server:9393/apps' --post-data='uri=http://bit.ly/Einstein-GA-stream-applications-kafka-maven&force=true';
```
ilayaperumalg: I think we can use Einstein-SR2 here. I can make this change when merging.

spring-cloud-dataflow-server/docker-compose-prometheus.yml
```diff
@@ -0,0 +1,99 @@
...
+    container_name: 'grafana'
+    ports:
+      - '3000:3000'
+    depends_on:
+      - prometheus
+
+  # The service-discovery container.
+  # Use `docker exec -it service-discovery /bin/sh` to logging the container
+  service-discovery:
+    image: springcloud/spring-cloud-dataflow-prometheus-service-discovery:0.0.3
```
ilayaperumalg: What's the reasoning to go with the `0.0.3` version here? Shouldn't we use the SCDF versioning for the same?

tzolov: At the moment the service-discovery project is not part of the SCDF build ecosystem and therefore is not possible to automatically generate service-discovery  aligned with the SCDF versions.

ilayaperumalg: ok, let's discuss on what would be the appropriate place for this project and have it versioned before merging this PR.

spring-cloud-dataflow-server/docker-compose-influxdb.yml
```diff
@@ -0,0 +1,66 @@
...
+      - "2181"
+  dataflow-server:
+    image: springcloud/spring-cloud-dataflow-server:${DATAFLOW_VERSION:?DATAFLOW_VERSION is not set!}
+    container_name: dataflow-server
+    ports:
+      - "9393:9393"
+    environment:
+      - spring.cloud.dataflow.applicationProperties.stream.spring.cloud.stream.kafka.binder.brokers=kafka:9092
+      - spring.cloud.dataflow.applicationProperties.stream.spring.cloud.stream.kafka.binder.zkNodes=zookeeper:2181
+      - spring.cloud.skipper.client.serverUri=http://skipper-server:7577/api
```
ilayaperumalg: We need to include the fix made from https://github.com/spring-cloud/spring-cloud-dataflow/commit/f96aa1a307cf415c9b63fe298ff61ab0938745eb

ilayaperumalg:
LGTM, merging. Created https://github.com/spring-cloud/spring-cloud-dataflow/issues/3111 to track the issue exploring stream applications at the Grafana dashboard.

sabbyanandan:
Addressed in https://github.com/spring-cloud/spring-cloud-dataflow/commit/5335a6e1373bde3f4853aa091331bd10b341353f.

PR #3056, opened 2019-03-25
Fix docker compose yml

### LINKED ISSUE #3055: Docker compose require Kafka Streams binder configuration properties
The `docker-compose` yaml doesn't have Kafka streams binder configuration properties for brokers and zookeepr. This makes the SCDF server not allowing any of the Kafka Streams apps to be deployed when using `docker compose` with the current `docker-compose.yml`.

ilayaperumalg:
- to include Kafka Streams config

Resolves #3055

### DIFF
--- a/spring-cloud-dataflow-server/docker-compose.yml
+++ b/spring-cloud-dataflow-server/docker-compose.yml
@@ -22,7 +22,9 @@ services:
       - "9393:9393"
     environment:
       - spring.cloud.dataflow.applicationProperties.stream.spring.cloud.stream.kafka.binder.brokers=kafka:9092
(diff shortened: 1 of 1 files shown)


sabbyanandan:
I remember bringing this up with @sobychacko in the past. It is highly unlikely that Kafka and KStreams connect to different cluster configurations - the whole point of KStreams is to correlate data from topics and its partitions through KStreams, KTable or GlobalTable semantics. I don't see how this could be accomplished with core Kafka and KStreams pointing to 2 different clusters. 

 If one is configured, the other should be automatically inferred, IMHO.

sobychacko:
LGTM - merged upstream.

PR #3058, opened 2019-03-25
Enable configuring different platform types without setting profiles

### LINKED ISSUE #3040: Multiple TaskPlatform configuration requires setting multiple active profiles to be enabled
This was probably a mistake in order to make a test case pass.  Need to enable a local server to have all three platforms enabled at the same time.

dturanski:
Resolves #3040 [2.0.x backport]

### DIFF
--- a/spring-cloud-dataflow-classic-docs/src/test/resources/rest-docs-config.yml
+++ b/spring-cloud-dataflow-classic-docs/src/test/resources/rest-docs-config.yml
@@ -9,7 +9,14 @@ spring:
         collector:
           uri: http://localhost:${fakeMetricsCollector.port}
   autoconfigure:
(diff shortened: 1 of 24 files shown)


markpollack:
This doesn't seem to address the referenced issue , instead it is cleaning up how the default tasklauncher is created when on the local platform.

dturanski:
Removing the conditions on `CloudFoundryTaskPlatformAutoConfiguration` and `KubernetesTaskPlatformAutoConfiguration` created a situation where the default local platform would not be created if any remote platform configured. I assert it does resolve the referenced issue, but it's currently not possible to test. Hey, it works on my machine :)

markpollack:
Removed use of KubernetesPlatformClientProvider

squashed, rebased and merged as 8c981694178772edfa8c3d18b8ab3e6cc2ff216e

PR #3059, opened 2019-03-25
Enable configuring different platform types without setting profiles

### LINKED ISSUE #3040: Multiple TaskPlatform configuration requires setting multiple active profiles to be enabled

dturanski:
Fixes #3040


markpollack:
removed `KubernetesPlatformClientProvider`
squashed and merged as 5df5bdbd7b6ec903a56d46a517689509fa4077ba

PR #3112, opened 2019-03-28
Use sl4fj for logging in all the classes

### LINKED ISSUE #3108: Review logging classes in use
Seems we have some commons logging usage in the code base, should be consistent and stick with sl4fj.

ilayaperumalg:
- Remove commons logging and replace them with sl4fj

Resolves #3108


markpollack:
rebased and merged as 9ff48ce228614a53e1c0b71f3462d0e693726151  backported to 2.0.x

PR #3122, opened 2019-04-02
Make TaskPlatform where SCDF is deployed primary

dturanski:
Fix for error running on cloud platforms with `schedules-enabled=true`

### DIFF
--- a/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/TaskPlatform.java
+++ b/spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/TaskPlatform.java
@@ -19,10 +19,13 @@ import java.util.List;
 
 /**
  * @author Donovan Muller
+ * @author David Turanski
  */
 public class TaskPlatform {
 
 	private String name;
+
+	private boolean primary;
 
 	private List<Launcher> launchers;
 
@@ -45,6 +48,13 @@ public class TaskPlatform {
 
 	public void setLaunchers(List<Launcher> launchers) {
 		this.launchers = launchers;
+	}
+
+	public boolean isPrimary() {
+		return primary;
+	}
+	public void setPrimary(boolean primary) {
+		this.primary = primary;
 	}
 }
 
--- a/spring-cloud-dataflow-platform-cloudfoundry/src/main/java/org/springframework/cloud/dataflow/server/config/cloudfoundry/CloudFoundryTaskPlatformAutoConfiguration.java
+++ b/spring-cloud-dataflow-platform-cloudfoundry/src/main/java/org/springframework/cloud/dataflow/server/config/cloudfoundry/CloudFoundryTaskPlatformAutoConfiguration.java
@@ -17,12 +17,18 @@ package org.springframework.cloud.dataflow.server.config.cloudfoundry;
 
 import java.util.Optional;
 
-import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnCloudPlatform;
+import org.springframework.boot.cloud.CloudPlatform;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.cloud.dataflow.core.TaskPlatform;
+import org.springframework.cloud.dataflow.server.config.CloudProfileProvider;
+import org.springframework.cloud.dataflow.server.config.features.ConditionalOnTasksEnabled;
+import org.springframework.cloud.dataflow.server.config.features.SchedulerConfiguration.SchedulerConfigurationPropertyChecker;
 import org.springframework.cloud.scheduler.spi.cloudfoundry.CloudFoundrySchedulerProperties;
 import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Conditional;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.core.env.Environment;
 
 /**
  * Creates TaskPlatform implementations to launch/schedule tasks on Cloud Foundry.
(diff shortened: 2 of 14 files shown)


chrisjs:
i tested this change in k8s by doing a basic `time | log` and `timestamp` test - things seem fine on K8s side

spring-cloud-dataflow-server/src/test/java/org/springframework/cloud/dataflow/server/single/KubernetesSchedulerTests.java
```diff
@@ -0,0 +1,52 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * @author David Turanski
+ **/
+@ActiveProfiles("local")
+@SpringBootTest(classes = {
+		DataFlowServerApplication.class }, webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+@RunWith(SpringRunner.class)
+public class KubernetesSchedulerTests {
```
markpollack: Should be `LocalSchedulerTests` right?

markpollack: We have `SchedulerPerPlatformTest` but I think it isn't actually testing anything other than the conditions on creating KubernetesSchedulerProperties and CloudFoundrySchedulerProperties

spring-cloud-dataflow-platform-cloudfoundry/src/main/java/org/springframework/cloud/dataflow/server/config/cloudfoundry/CloudFoundryTaskPlatformAutoConfiguration.java
```diff
@@ -38,6 +41,13 @@ public TaskPlatform cloudFoundryPlatform(CloudFoundryTaskPlatformFactory cloudFo
 		return cloudFoundryTaskPlatformFactory.createTaskPlatform();
 	}
 
+	@ConditionalOnCloudPlatform(CloudPlatform.CLOUD_FOUNDRY)
+	@Primary
+	@Bean
+	public TaskPlatform primaryTaskPlatform(CloudFoundryTaskPlatformFactory cloudFoundryTaskPlatformFactory) {
+		return cloudFoundryPlatform(cloudFoundryTaskPlatformFactory);
```
markpollack: Seems like we will have these beans created even if tasks are disabled.  I think we need to add `ConditionalOnTasksEnabled` to the class definition.

markpollack: Problem is that class is living in server-core... I wonder why we have this as an autoconfiguration then....

markpollack: Problem is that `ConditionalOnTasksEnabled` is in the core module.

jvalkeal: Plan was that all auto-config classes would eventually go to `spring-cloud-dataflow-autoconfigure` which can then access everything.

spring-cloud-dataflow-platform-cloudfoundry/src/main/java/org/springframework/cloud/dataflow/server/config/cloudfoundry/CloudFoundryTaskPlatformAutoConfiguration.java
```diff
@@ -38,6 +41,13 @@ public TaskPlatform cloudFoundryPlatform(CloudFoundryTaskPlatformFactory cloudFo
 		return cloudFoundryTaskPlatformFactory.createTaskPlatform();
 	}
 
+	@ConditionalOnCloudPlatform(CloudPlatform.CLOUD_FOUNDRY)
+	@Primary
```
jvalkeal: Use of `@Primary` like this is actually pointless as it exists on other beans as well. It would then fail if two primary beans exists.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/TaskConfiguration.java
```diff
@@ -102,10 +95,13 @@ public LauncherInitializationService launcherInitializationService(
 		return new LauncherInitializationService(launcherRepository, platforms);
 	}
 
+	@Profile({"local", "default"})
```
markpollack: what is the `default` profile?

markpollack: I removed "default" and all the tests still passed, prob was some cruft left over.

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/TaskPlatform.java
```diff
@@ -46,5 +49,12 @@ public void setName(String name) {
 	public void setLaunchers(List<Launcher> launchers) {
 		this.launchers = launchers;
 	}
+
+	public boolean isPrimary() {
```
markpollack: will add some javadoc to clarify what this means.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/TaskConfiguration.java
dturanski: TLDR; The default profile is active when no other profile is set.  https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-external-config.html#boot-features-external-config-profile-specific-properties
SCDF sets the profile to `local` via the `ProfileApplicationListener`, but since this is registered in `spring.factories` some of the Local*Tests were failing since the `local` profile was not set. Rather than explicitly setting the profile for the tests, which may cause future tests to fail if they forget to do that, and mislead some to think it must be explicitly set,  I decided it was safe to default to the local TaskPlatform if no other profile is set.

markpollack: yea, I just bumped into the other failing tests....

markpollack:
added javadoc, rebased and merged as 22e93697b03ca725c18f18c229fbab4d16e8f767

PR #3157, opened 2019-04-15
Use Confluent Kafka/Zookeeper images in docker compose files

### LINKED ISSUE #3156: Docker compose to use Confluent Kafka docker image

ilayaperumalg:
Resolves #3156

### DIFF
--- a/spring-cloud-dataflow-server/docker-compose-influxdb.yml
+++ b/spring-cloud-dataflow-server/docker-compose-influxdb.yml
@@ -2,28 +2,31 @@ version: '3'
 
 services:
   kafka:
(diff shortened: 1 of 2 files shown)


ilayaperumalg:
@chrisjs Without the security protocol, the Kafka broker name doesn't get resolved with the Confluent Kafka docker image. It has to be one of http://kafka.apache.org/11/javadoc/org/apache/kafka/common/security/auth/SecurityProtocol.html

chrisjs:
ah ok thanks, makes sense. looks like that was the default on the prior image

chrisjs:
merged in 2e214e1

PR #3159, opened 2019-04-15
Enhance local scheduler mock implementation

jvalkeal:
- Issue with current no-op local impl is that it
  throws error with UI just by going into a tasks and
  schedules page making UI side development more difficult.
  Error is java.lang.IllegalStateException: Could not find a default scheduler
  which is caused because existing `localScheduler` is not
  wired into a `Launcher`.
- Locally use `localScheduler` from `LocalSchedulerAutoConfiguration`
  so that we can delay UI errors to a point where actual
  scheduling would then throw an error. This allows UI
  to so some meaninfull errors instead of a cryptic messages.


dturanski:
LGTM; Merging

dturanski:
Merged via 996a660

PR #3162, opened 2019-04-15
Fix default local with no scheduler

dturanski:
Fix for local server not starting when `schedules-enabled=false`


chrisjs:
LGTM - property didn't have to be explicitly defined, when done so and set to `false`, scheduling wasn't available and when set to `true`, upon trying to schedule a message popped up indicating scheduling not available on local

chrisjs:
merged in 1bc64ac1d67

PR #3203, opened 2019-05-03
Switch to MySQL by default in Docker Compose

### LINKED ISSUE #3198: Switch to MySQL by default in Docker Compose
As a developer, I'd like to switch to MySQL from default in the SCDF `docker-compose.yml` experience. 

This is a prerequisite for all the task/batch sample developer guide in the Microsite. 

When switching, we would want to remove MySQL from the customization section of the reference guide.

chrisjs:
Resolves #3198

### DIFF
--- a/spring-cloud-dataflow-server/docker-compose.yml
+++ b/spring-cloud-dataflow-server/docker-compose.yml
@@ -1,6 +1,15 @@
 version: '3'
 
 services:
+  mysql:
+    image: mysql:5.7.25
+    environment:
+      MYSQL_DATABASE: dataflow
+      MYSQL_USER: root
+      MYSQL_ROOT_PASSWORD: rootpw
+    expose:
+      - 3306
+
   kafka:
     image: confluentinc/cp-kafka:5.2.1
     expose:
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-server/docker-compose.yml
```diff
@@ -1,6 +1,15 @@
 version: '3'
 
 services:
+  mysql:
+    image: mysql:5.7.25
+    environment:
+      MYSQL_DATABASE: dataflow
+      MYSQL_USER: root
+      MYSQL_ROOT_PASSWORD: rootpw
```
sabbyanandan: Not sure what the best practice is when using user/pass in Docker Compose experience. Folks question the reasoning behind having this as clear-text. We encode it in K8s world; I wonder if there's anything similar in plain Docker world.

chrisjs: agreed plaintext is not a production grade thing, but for a "quick start" it requires no work on the end user side.

docker secrets could be an option (https://docs.docker.com/compose/compose-file/#secrets) - (https://hub.docker.com/_/mysql) but would add extra steps to explain and have the end user setup - not sure it would make for a good "quick start" experience

if its commonly asked how to avoid plain text, this might be a candidate for an entry in the customization section

chrisjs:
merged in 433c9f98963a9b7cd4311ac4f252d09b54e18747

cytgjf:
请问，你们将H2数据库换成mysql数据库有出现缺失表的情况吗，Table 'dataflow2.uri_registry' doesn't exist

PR #3239, opened 2019-05-15
Removed AS clause from sql query for step count.

### LINKED ISSUE #3227: Issue while getting step counts with Oracle
**Description:**
While using SCDF with Oracle database for batch processing, the query used to get steps execution count fails. On checking found that the class JdbcSearchableJobExecutionDao uses "AS" for alias on table (in FIELDS_WITH_STEP_COUNT) which is syntactically an error in Oracle. 

**Release versions:**
Observed while migrating from SCDF 1.7.4 to SCDF 2.0.x

Is there any solution for this issue?

cppwfs:
resolves #3227


cppwfs:
Tested against
* H2
* DB2
* Postgres
* MySql
* Oracle

jvalkeal:
lgtm

PR #3266, opened 2019-06-05
Fix app registration behaviour with version

### LINKED ISSUE #3241: Stream Application version is deployed with a URI that does not match with what was registered
**Description:**
When deploying a stream, 
  and a _version_ is specified for a stream application that is different from the _default version_, 
  and the  _specified version_ and the _default version_ is registered with different docker image names on their URIs,
then the application is deployed in K8s with a _docker image name_ of the _default version_ but with the _docker image tag_ of the _specified version_,
  and this will result in an image pull failure.

**For example,**
If an application has the following versions and their corresponding URIs:
`0.1.0` -  `docker:prefix1/my-source:0.1.0` 
`0.2.0` -  `docker:prefix2/my-source:0.2.0` 
_Notice the different prefixes on the docker image names_

And I deploy a stream with version `0.1.0` of `my-source`, the result is that it will deploy `prefix2/my-source:0.1.0`.
It should have deployed `prefix1/my-source:0.1.0` because that is the URI registered for that version.

It looks like Skipper is assuming that all versions of a registered application have similar URI to the default version and it's just switching the version tag.

**Release versions:**
```
{
  "versionInfo": {
    "implementation": {
      "name": "spring-cloud-dataflow-server",
      "version": "2.0.0.RELEASE"
    },
    "core": {
      "name": "Spring Cloud Data Flow Core",
      "version": "2.0.0.RELEASE"
    },
    "dashboard": {
      "name": "Spring Cloud Dataflow UI",
      "version": "2.0.0.RELEASE"
    },
    "shell": {
      "name": "Spring Cloud Data Flow Shell",
      "version": "2.0.0.RELEASE",
      "url": "https://repo.spring.io/libs-release/org/springframework/cloud/spring-cloud-dataflow-shell/2.0.0.RELEASE/spring-cloud-dataflow-shell-2.0.0.RELEASE.jar"
    }
  },
  "featureInfo": {
    "streamsEnabled": true,
    "tasksEnabled": true,
    "schedulerEnabled": true,
    "grafanaEnabled": true
  },
  "securityInfo": {
    "isAuthenticationEnabled": false,
    "isAuthenticated": false,
    "username": null,
    "roles": []
  },
  "runtimeEnvironment": {
    "appDeployer": {
      "platformSpecificInfo": {},
      "deployerImplementationVersion": "2.0.0.RELEASE",
      "deployerName": "Spring Cloud Skipper Server",
      "deployerSpiVersion": "2.0.0.RELEASE",
      "javaVersion": "1.8.0_192",
      "platformApiVersion": "",
      "platformClientVersion": "",
      "platformHostVersion": "",
      "platformType": "Skipper Managed",
      "springBootVersion": "2.1.2.RELEASE",
      "springVersion": "5.1.4.RELEASE"
    },
    "taskLaunchers": [
      {
        "platformSpecificInfo": {},
        "deployerImplementationVersion": "2.0.0.RELEASE",
        "deployerName": "KubernetesTaskLauncher",
        "deployerSpiVersion": "2.0.0.RELEASE",
        "javaVersion": "1.8.0_192",
        "platformApiVersion": "v1",
        "platformClientVersion": "unknown",
        "platformHostVersion": "unknown",
        "platformType": "Kubernetes",
        "springBootVersion": "2.1.2.RELEASE",
        "springVersion": "5.1.4.RELEASE"
      }
    ]
  },
  "grafanaInfo": {
    "url": "http://grafana:3000",
    "token": "",
    "refreshInterval": 15
  }
}
```

jvalkeal:
- Previously you were able to register an app with different base uri
  while we're always assumed that versions works so that only version
  part can be different. i.e. docker:docker:prefix1/my-source:0.1.0
  vs docker:docker:prefix3/my-source:0.3.0. This would always fail with
  an actual upgrade and depending on which one is a default app also
  initial installation might fail.
- Fixing this on a registration time by denying user to use
  anything else than a changed version when registering an app
  with existing default app.
- Fixes #3241

```
dataflow:>app register --name test --uri docker:docker:prefix1/my-source:0.1.0 --type sink
Successfully registered application 'sink:test'
dataflow:>app register --name test --uri docker:docker:prefix1/my-source:0.2.0 --type sink
Successfully registered application 'sink:test'
dataflow:>app register --name test --uri docker:docker:prefix3/my-source:0.3.0 --type sink
Command failed org.springframework.cloud.dataflow.rest.client.DataFlowClientException: Existing default application [docker:docker:prefix1/my-source:0.1.0] can only differ by a version but is [docker:docker:prefix3/my-source:0.3.0]
```

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppRegistryController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppRegistryController.java
@@ -60,7 +60,6 @@ import org.springframework.hateoas.PagedResources;
 import org.springframework.hateoas.mvc.ResourceAssemblerSupport;
 import org.springframework.http.HttpStatus;
 import org.springframework.util.CollectionUtils;
-import org.springframework.util.ObjectUtils;
 import org.springframework.util.StringUtils;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.RequestMapping;
@@ -204,19 +203,7 @@ public class AppRegistryController {
 			@RequestParam("uri") String uri, @RequestParam(name = "metadata-uri", required = false) String metadataUri,
 			@RequestParam(value = "force", defaultValue = "false") boolean force) {
 
-		// compare existing default app and deny registration if base path of uri's
-		// don't match as registering app with different version can only differ by version
-		AppRegistration defaultApp = appRegistryService.getDefaultApp(name, type);
-		if  (defaultApp != null && StringUtils.hasText(version)) {
-			String defaultAppUri = defaultApp.getUri().toString();
-			String defaultAppUriNoVersion = removeLastMatch(defaultAppUri, defaultApp.getVersion());
-			String newAppUriNoVersion = removeLastMatch(uri, version);
-			if (!ObjectUtils.nullSafeEquals(defaultAppUriNoVersion, newAppUriNoVersion)) {
-				throw new IllegalArgumentException("Existing default application [" + defaultAppUri
-						+ "] can only differ by a version but is [" + uri + "]");
-			}
-		}
-
+		appRegistryService.validate(appRegistryService.getDefaultApp(name, type), uri, version);
 		AppRegistration previous = appRegistryService.find(name, type, version);
 		if (!force && previous != null) {
 			throw new AppAlreadyRegisteredException(previous);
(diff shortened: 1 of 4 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppRegistryController.java
```diff
@@ -202,6 +203,20 @@ public void register(@PathVariable("type") ApplicationType type, @PathVariable("
 			@PathVariable("version") String version,
 			@RequestParam("uri") String uri, @RequestParam(name = "metadata-uri", required = false) String metadataUri,
 			@RequestParam(value = "force", defaultValue = "false") boolean force) {
+
+		// compare existing default app and deny registration if base path of uri's
+		// don't match as registering app with different version can only differ by version
+		AppRegistration defaultApp = appRegistryService.getDefaultApp(name, type);
```
markpollack: I think we should move this logic out of the controller and into a `validate` method on `appRegistryService` ?

markpollack:
squashed and merged as ac5fa963ac2e0eb48173c75768fdc5fa3a0eec52

PR #3270, opened 2019-06-06
Polish build files

jvalkeal:
- Move everything we can into main dependency management
- Some tabs vs spaces dancing
- Remove redundant sections from poms which already comes from parent
- Add missing modules to bom
- Don't define own module versions in deps as those comes from parent
- Use correct spring-cloud-kubernetes.version version in parent as it
  had wrong one and mess in poms masked that issue
- All these changes creates fatjar with same deps


markpollack:
Added scm/license section in root pom - rebase, squashed and merged as d6a02b98e0fff87b02cc15962424b5b696471e9d

PR #3283, opened 2019-06-10
Use a value from secret for MYSQL_ROOT_PASSWORD

making:
`MYSQL_ROOT_PASSWORD ` should be retrieved from Secret instead of a hardcoded value to make sure the same value is shared between mysql, server and skipper.


chrisjs:
merged in 1c4843c372dd71c23efd5ce549d554e0706d81b6

PR #3284, opened 2019-06-11
Polish database support

### LINKED ISSUE #3255: Support mariadb 10.3+
Starting from mariadb 10.3 a proper sequence has been added and things we do with a sequence table originating from remains of mysql doesn't work anymore.

Need to fully move to this and think about all migration nonsense this brings.

### LINKED ISSUE #3256: Support mysql 8
Doesn't work with current mariadb driver so fix would be to bundle mysql driver. User can then choose a driver without re-building a server.

jvalkeal:
- Drop hsqldb from poms and docs as it's not supported anymore.
- Drop manual migration section from docs as things are handled by flyway.
- Polish db sections for local server and add notes about
  MySQL 8+ and MariaDB 10.3+.
- Testing work has been done in db acceptance tests.
- Fixes #3255
- Fixes #3256


ilayaperumalg:
LGTM, merging.

PR #3285, opened 2019-06-13
Separate dsl parser in its own module

jvalkeal:
- This is a simple module split by extracting
  org.springframework.cloud.dataflow.core.dsl
  from spring-cloud-dataflow-core to
  spring-cloud-dataflow-core-dsl.
- Makes it easier to consume just a raw dsl bits
  without a need to pollute classpath with all
  dependencies what spring-cloud-dataflow-core
  itself brings in.


markpollack:
rebased and merged as fbee12e8775b26f1e37c468239ee4f9e8f2bd750

PR #3291, opened 2019-06-24
Update app import links

### LINKED ISSUE #3288: Switch all the bulk import links to "latest" maven/docker links server by dataflow.spring.io
Including all the docker-compose YAMLs, it'd be better to switch all the demo/sample links in the ref. docs to point to "latest" URLs from dataflow.spring.io.

Also, add a table to list all the stream/task "latest" links in the [ref. guide](https://docs.spring.io/spring-cloud-dataflow/docs/2.1.0.RELEASE/reference/htmlsingle/#supported-apps-and-tasks). There's no central place for them right now.

See if we are using it in the unit-tests; if so, let's clean that up also! 

This is in some ways related to https://github.com/spring-cloud/spring-cloud-dataflow-acceptance-tests/pull/261.

ilayaperumalg:
- Update docker compose and docs to use latest app import links

Resolves #3288

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
@@ -376,22 +376,26 @@ and Spring Boot 2.1.x:
 
 [width="100%",frame="topbot",options="header"]
 |======================
-|Artifact Type |Stable Release |SNAPSHOT Release
+|Artifact Type |Latest Release|Stable Release |SNAPSHOT Release
 
 |RabbitMQ + Maven
-|https://dataflow.spring.io/Einstein-SR2-stream-applications-rabbit-maven
+|https://dataflow.spring.io/rabbitmq-maven-latest
+|https://dataflow.spring.io/Einstein-SR3-stream-applications-rabbit-maven
 |https://dataflow.spring.io/Einstein-BUILD-SNAPSHOT-stream-applications-rabbit-maven
 
 |RabbitMQ + Docker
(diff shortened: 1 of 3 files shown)


spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -376,21 +376,25 @@ and Spring Boot 2.1.x:
 
 [width="100%",frame="topbot",options="header"]
 |======================
-|Artifact Type |Stable Release |SNAPSHOT Release
+|Artifact Type |Latest Release|Stable Release |SNAPSHOT Release
 
 |RabbitMQ + Maven
+|https://dataflow.spring.io/rabbitmq-maven-latest
 |https://dataflow.spring.io/Einstein-SR2-stream-applications-rabbit-maven
```
sabbyanandan: I believe Einstein is at SR3 right now.

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -428,13 +432,15 @@ The following table includes the available Task Application Starters based on Sp
 
 [width="100%",frame="topbot",options="header"]
 |======================
-|Artifact Type |Milestone Release |SNAPSHOT Release
+|Artifact Type |Latest Release|Milestone Release |SNAPSHOT Release
```
sabbyanandan: It'd be good to change the column headers to be consistent for both stream and tasks.

> |Latest Release|Stable Release |SNAPSHOT Release

... probably is better

PR #3298, opened 2019-06-28
Support importing multiple app versions

### LINKED ISSUE #3287: Cannot import apps with multiple versions
Trying to import app uri's from a same file containing multiple versions silently fails. For example:
```
source.time=maven://org.springframework.cloud.stream.app:time-source-rabbit:2.0.1.RELEASE
source.time.metadata=maven://org.springframework.cloud.stream.app:time-source-rabbit:jar:metadata:2.0.1.RELEASE
source.time=maven://org.springframework.cloud.stream.app:time-source-rabbit:2.0.0.RELEASE
source.time.metadata=maven://org.springframework.cloud.stream.app:time-source-rabbit:jar:metadata:2.0.0.RELEASE
sink.log=maven://org.springframework.cloud.stream.app:log-sink-rabbit:2.0.2.RELEASE
sink.log.metadata=maven://org.springframework.cloud.stream.app:log-sink-rabbit:jar:metadata:2.0.2.RELEASE
sink.log=maven://org.springframework.cloud.stream.app:log-sink-rabbit:2.0.1.RELEASE
sink.log.metadata=maven://org.springframework.cloud.stream.app:log-sink-rabbit:jar:metadata:2.0.1.RELEASE
```
Only latest versions per type/name are actually imported. This is most likely because import functionality were created before we really had versioned apps, thus import code probably tracks imports per type/name and latter always overrides previous definitions.

With above example you end up having just `sink.log 2.0.1.RELEASE` and `source.time 2.0.0.RELEASE`

jvalkeal:
- In DefaultAppRegistryService and AppRegistryController, move away
  from using jdk's Properties. Move to a custom app import parser
  which is able to understand that `source.time`, `sink.log` can have
  different versions defined in a same file.
- In AppRegistryController change incoming app string not to get
  converted into Properties as that causes all sort of issues
  how jdk handles escaping, etc. i.e. it adds header lines and
  escapes things from `maven:` into `maven\:`.
- Fix AuditRecordControllerTests as it used to assume an exact
  order of how things were inserted into a db where this order
  just magically happened to be a one what hashtable gave out from
  a Properties. Fixed 3 tests to just check that response has needed
  fields on order and not checking actual order in a json response.
- Few new tests to see that usual cases like, spaces, empty lines,
  comment lines, metadata before app, missing metadata and missing
  main app url, etc.
- Fixes #3287

Essentially with this we can do:
```
source.time=maven://org.springframework.cloud.stream.app:time-source-rabbit:2.0.1.RELEASE
source.time.metadata=maven://org.springframework.cloud.stream.app:time-source-rabbit:jar:metadata:2.0.1.RELEASE
source.time=maven://org.springframework.cloud.stream.app:time-source-rabbit:2.0.0.RELEASE
source.time.metadata=maven://org.springframework.cloud.stream.app:time-source-rabbit:jar:metadata:2.0.0.RELEASE
```

### DIFF
--- a/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/service/DefaultAppRegistryService.java
+++ b/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/service/DefaultAppRegistryService.java
@@ -16,14 +16,15 @@
 
 package org.springframework.cloud.dataflow.registry.service;
 
+import java.io.BufferedReader;
 import java.io.IOException;
+import java.io.InputStreamReader;
 import java.net.URI;
-import java.net.URISyntaxException;
-import java.util.AbstractMap;
+import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
 import java.util.Properties;
-import java.util.function.Function;
+import java.util.function.BiFunction;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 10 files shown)


ilayaperumalg:
LGTM, merging.

PR #3308, opened 2019-07-04
Retrieve stream/task application logs

### LINKED ISSUE #3306: Add REST endpoint to retrieve logs of stream applications via Runtime apps
The runtime apps REST endpoint `/runtime/apps` can serve the application logs of a stream. We can also have a REST endpoint that retrieves a single application inside the streaming pipeline.

### LINKED ISSUE #3307: Add REST endpoint to retrieve task application logs
- Use TaskExecutionController to serve the REST endpoint to access application logs of a given task
- Update CF platform configuration to update the DopplerClient for CF operations to fetch logs
- Update K8s config to provide access to logs API

ilayaperumalg:
- Add REST endpoints for stream apps log retrieval at Runtime apps controller
 - Add REST endpoint for task app log retrieval at TaskExecution controller
 - Add getLog() method SPI for TaskExecutionService and implemented it in DefaultTaskExecutionService
 - Add getLog() method in StreamDeployer and have it implemented in SkipperStreamDeployer
 - Add and update tests
 - Update CF task platform configuration to update doppler client to CF operations

Resolves #3306
Resolves #3307


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/RuntimeAppsController.java
```diff
@@ -90,6 +90,16 @@ public AppStatusResource display(@PathVariable String appId) {
 		return statusAssembler.toResource(status);
 	}
 
+	@RequestMapping("/logs/{streamName}")
```
jvalkeal: We need to figure out a different path as `http://localhost:9393/runtime/apps/logs/ticktock` would not work as we already use `http://localhost:9393/runtime/apps/ticktock.time-v1`. Either something in `/runtime/logs/` and/or `/runtime/apps/ticktock.time-v1/logs`.

ilayaperumalg: yeah, it makes sense. I will update the PR

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/RuntimeAppsController.java
```diff
@@ -90,6 +90,16 @@ public AppStatusResource display(@PathVariable String appId) {
 		return statusAssembler.toResource(status);
 	}
 
+	@RequestMapping("{streamName}/logs")
```
jvalkeal: You can't have this as `logs` is expected to be an `appName` for mapping just below.

jvalkeal: And we already have `/{appId}` at root, so you can't just switch to `streamName` as it gets confusing. `RuntimeAppController` is really only meant for apps, not streams.

ilayaperumalg: Yeah, missed that. Now the PR is updated with the changes which move the new REST endpoints into separate controllers.

jvalkeal: These should then go away now that you moved it to a new controller?

tzolov:
squashed and merged at 7179de442907482a6695778d6feadecf3abdd566

PR #3309, opened 2019-07-04
Document adding mysql driver to a server

### LINKED ISSUE #2489: Add Custom JDBC Driver to Spring Cloud Data Flow Server
Hi,

In the documentation it is specified that to add a new JDBC driver it is necessary to change the pom.xml and rebuild the project.  I was not confortable with this approach, so I implemented an extension of the Spring Cloud Data Flow Server image:

```
FROM springcloud/spring-cloud-dataflow-server-kubernetes:1.5.2.RELEASE

#Get the oracle driver jar
COPY ojdbc8-12.2.0.1.jar /maven/lib/ojdbc8-12.2.0.1.jar

ENTRYPOINT ["java", "-cp", "/maven/spring-cloud-dataflow-server-kubernetes.jar", "-Dloader.path=/maven/lib", "org.springframework.boot.loader.PropertiesLauncher"]
```
Is this ok or could it have any side effect? In my opinion to implement an image extension is better than to change the project's dependencies.

Thanks!

jvalkeal:
- As an temporary fix for this documentation issue, provide
  gradle and maven samples how to build a custom dataflow server
  adding mysql driver. These custom build files examples will
  produce same fatjar as project itself with an addition of
  mysql-connector-java-8.0.16.jar.
- This lays groundwork to make dependency management more clear
  as there is a way too much manual configuration for dependencies
  and versions for custom gradle and maven builds.
- Fixes #2489

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/configuration-local.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/configuration-local.adoc
@@ -196,6 +196,173 @@ spring:
     driver-class-name:org.postgresql.Driver
 ----
 
+[start=4]
+. Or build a custom server with your own build files. Here are examples of bundling MySQL driver using gradle and maven, which
+will result same jar as `2.1.2.RELEASE` with addition of a `mysql-connector-java-8.0.16.jar`:
+
+[source, groovy]
+.gradle
+----
+plugins {
+        id 'org.springframework.boot' version '2.1.3.RELEASE'
+        id 'java'
+}
+
+ext {
+        springCloudVersion = 'Greenwich.SR1'
+        springCloudDataflowVersion = '2.1.2.RELEASE'
+        springSecurityOauth2AutoconfigureVersion = '2.1.3.RELEASE'
+        springSecurityOauth2Version = '2.3.4.RELEASE'
+        kubernetesClientVersion = '4.0.4'
+        mysqlJdbcDriverVersion = '8.0.16'
+}
+
+apply plugin: 'io.spring.dependency-management'
+
+group = 'com.example'
+version = '0.0.1-SNAPSHOT'
+sourceCompatibility = '1.8'
+
+repositories {
+        mavenCentral()
+        maven { url "https://repo.springsource.org/libs-snapshot" }
+}
+
+dependencyManagement {
+        imports {
+                mavenBom "org.springframework.cloud:spring-cloud-dependencies:${springCloudVersion}"
+                mavenBom "org.springframework.cloud:spring-cloud-dataflow-dependencies:${springCloudDataflowVersion}"
+        }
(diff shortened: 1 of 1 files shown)


spring-cloud-dataflow-docs/src/main/asciidoc/configuration-local.adoc
```diff
@@ -196,6 +196,173 @@ spring:
     driver-class-name:org.postgresql.Driver
 ----
 
+[start=4]
```
sabbyanandan: We need to remove any references to statements like this:

> Since there is a Spring Cloud Data Flow Server for each target platform, you need to modify the appropriate maven `pom.xml` for each platform. 

This is right above the changes where you're attempting to make.

sabbyanandan: Perhaps also change:

> There are tags in each GitHub repository for each server version.

to

> There are GA release tags in GitHub repository, so you can switch to desired GA tags to add the drivers on the production-ready codebase.

spring-cloud-dataflow-docs/src/main/asciidoc/configuration-local.adoc
```diff
@@ -196,6 +196,173 @@ spring:
...
+. Or build a custom server with your own build files. Here are examples of bundling MySQL driver using gradle and maven, which
```
sabbyanandan: It could be:

> Alternatively, you can build a custom server with your build files. Here are the examples of bundling MySQL driver using Gradle and Maven, which will result in the same jar as `2.1.2.RELEASE` with the addition of a `mysql-connector-java-8.0.16.jar`:

spring-cloud-dataflow-docs/src/main/asciidoc/configuration-local.adoc
```diff
@@ -196,6 +196,173 @@ spring:
...
+will result same jar as `2.1.2.RELEASE` with addition of a `mysql-connector-java-8.0.16.jar`:
+
+[source, groovy]
+.gradle
```
sabbyanandan: Should show an empty project with the gradle and maven files in it? Maybe easy that way for folks to copy the file to repeat it.

If you think it is a good idea, we could host that empty project in [samples](https://github.com/spring-cloud/spring-cloud-dataflow-samples), and cross-link it here.

ilayaperumalg: Created https://github.com/spring-cloud/spring-cloud-dataflow-samples/issues/107 to track this.

ilayaperumalg:
Merging after fixing the review comments.

ilayaperumalg:
Fixed review comments, rebased and merged as 1d25dee7

PR #3312, opened 2019-07-04
Allow splits with a single entry for some composed task scenarios

### LINKED ISSUE #3263: Improve Composed Task DSL to support multiple transitions to converge at a common step
As a user, I have a requirement to branch out to multiple transitions from a step, and also have all the transitions to run a common downstream step.

For example, 
I'd like to branch out from the `Import` step to form transitions like this:

image
----
I'd then want all these transitions to _still_ run a common step, which in this case is `Backwards`.

image
----
Though it is possible to build that representation in the UI today, it results in bogus generated DSL. 

The workaround today is to add a no-op step in the DSL to make the transitions all run in as a composite unit, while the no-op runs in parallel.

image
----

Associated DSL:
```
<Import: timestamp 'ERROR'->t1: timestamp 'ERROR2'->t2: timestamp || t3: timestamp> && Backwards: timestamp
```

aclement:
With this change the DSL parser allows a single entry in a split - this
is so that you can have a task that uses transitions and rather than
those transitions jumping straight to the END after they are done, they
will jump to the task following the split. Effectively the split is
being used to create a nested flow. This is not a new usage of it, but
previously it only allowed nested flows involving multiple tasks to
run in parallel. Now you can use it for just a single task in a
nested flow.

The parser has been changed and the graph-to-text and text-to-graph
algorithms have been updated to support it.

Resolves #3263

### DIFF
--- a/spring-cloud-dataflow-core-dsl/src/main/java/org/springframework/cloud/dataflow/core/dsl/SplitNode.java
+++ b/spring-cloud-dataflow-core-dsl/src/main/java/org/springframework/cloud/dataflow/core/dsl/SplitNode.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 the original author or authors.
+ * Copyright 2017-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


aclement:
If you want to test this, you need to create DSLs like this:
```
<Import: timestamp 'Error2'->T2: timestamp 'Error'->T1: timestamp> && Backwards: timestamp 
```
With the `<...>` both T1 and T2 will link to Backwards.  Without the `<..>` they will link to END. I am having trouble building the composed-task-runner to see if it is affected. The only way it might be is if it is testing that <..> with a single task is an error, it no longer will be.

aclement:
If someone has composed-task-runner compiling properly, here are two tests you could add to `ComposedRunnerVisitorTests` which verify the behaviour when <...> is used (or not):
```

	@Test
	public void testVerifyingFlowFromTransitionAppsWiithoutSingleSplit() {
		// With no <...> BBB will go straight to end and skip DDD
		setupContextForGraph("failedStep 'FAILED' -> BBB 'E2' -> CCC && DDD");
		Collection<StepExecution> stepExecutions = getStepExecutions();
		Set<String> stepNames = getStepNames(stepExecutions);
		assertEquals(2, stepExecutions.size());
		assertTrue(stepNames.contains("failedStep_0"));
		assertTrue(stepNames.contains("BBB_0"));
		assertFalse(stepNames.contains("DDD_0"));
		List<StepExecution> sortedStepExecution =
				getSortedStepExecutions(stepExecutions);
		assertEquals("failedStep_0", sortedStepExecution.get(0).getStepName());
		assertEquals("BBB_0", sortedStepExecution.get(1).getStepName());
	}

	@Test
	public void testVerifyingFlowFromTransitionAppsWithSplit() {
		// With <...> BBB will go to DDD 
		setupContextForGraph("<failedStep 'FAILED' -> BBB 'E2' -> CCC> && DDD");
		Collection<StepExecution> stepExecutions = getStepExecutions();
		Set<String> stepNames = getStepNames(stepExecutions);
		assertEquals(3, stepExecutions.size());
		assertTrue(stepNames.contains("failedStep_0"));
		assertTrue(stepNames.contains("BBB_0"));
		assertTrue(stepNames.contains("DDD_0"));
		List<StepExecution> sortedStepExecution =
				getSortedStepExecutions(stepExecutions);
		assertEquals("failedStep_0", sortedStepExecution.get(0).getStepName());
		assertEquals("BBB_0", sortedStepExecution.get(1).getStepName());
		assertEquals("DDD_0", sortedStepExecution.get(2).getStepName());
	}
```
No code changes needed so it is doing the right thing and needs no real changes because of this parser change.

cppwfs:
Aside some format issues.   The code looks ok.   I haven't tested it.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Rebased and merged as 6f53b19f

PR #3319, opened 2019-07-08
Fix deletion of backing task containers

### LINKED ISSUE #3259: task execution cleanup does not delete kubernetes pods
**Description:**
We are running the Spring Cloud Dataflow version 2.0.1.RELEASE in kubernetes cluster.  I set up a simple hello-world app in a docker image.  The task execution creates a pod that correctly prints the hello-world message and the pod moves to the completed state.  

According to the documentation, “task execution cleanup” should cleanup that completed kubernetes pod. When I tried that, the pod is NOT cleaned up. The server prints out a message from DefaultTaskDeleteService “Did not find TaskDeployment for taskname….”  “Nothing to clean up”.  

Looking at the DefaultTaskDeleteService, it appears there may be a bug in the cleanupExecution method.  It takes the execution id to get the TaskExecution. From there it calls getExternalExecutionId and names it launchId. That launchId is the name of the pod. It looks like that launchId should be used to look up the TaskDeployment as the task_deployment_id in the database is the name of the pod.  However, the task execution id is used instead.  The lookup therefore finds nothing.

If line 101 is changed from this:
TaskDeployment taskDeployment = this.taskDeploymentRepository.findByTaskDeploymentId(String.valueOf(id));

to this:
  TaskDeployment taskDeployment = this.taskDeploymentRepository.findByTaskDeploymentId(launchId);

then the TaskDeployment is found and the pod is deleted.

chrisjs:
Resolves #3259

tested task destroy on k8s/cf/local and all deleted tasks as expected

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskDeleteService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskDeleteService.java
@@ -98,7 +98,7 @@ public class DefaultTaskDeleteService implements TaskDeleteService {
 		Assert.notNull(taskExecution, "There was no task execution with id " + id);
 		String launchId = taskExecution.getExternalExecutionId();
 		Assert.hasLength(launchId, "The TaskExecution for id " + id + " did not have an externalExecutionId");
-		TaskDeployment taskDeployment = this.taskDeploymentRepository.findByTaskDeploymentId(String.valueOf(id));
+		TaskDeployment taskDeployment = this.taskDeploymentRepository.findByTaskDeploymentId(launchId);
 		if (taskDeployment == null) {
 			logger.warn(String.format("Did not find TaskDeployment for taskName = [%s], taskId = [%s].  Nothing to clean up.",
 					taskExecution.getTaskName(), id));
--- a/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/TaskExecutionControllerTests.java
+++ b/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/TaskExecutionControllerTests.java
@@ -179,7 +179,7 @@ public class TaskExecutionControllerTests {
 			taskBatchDao.saveRelationship(taskExecution, jobExecution);
 			TaskDeployment taskDeployment = new TaskDeployment();
 			taskDeployment.setTaskDefinitionName(TASK_NAME_ORIG);
-			taskDeployment.setTaskDeploymentId("1");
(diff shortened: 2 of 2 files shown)


cppwfs:
I tried on cloud foundry but the artifact was not removed.    :-(

chrisjs:
ugh.. i tested this against PCF 2.4 yesterday and worked fine.. what version are you testing against?

cppwfs:
violet 2.5

Using command:
```
 curl -X DELETE http://myserverloc/tasks/executions/<task-exec-id>
```

chrisjs:
im deploying to 2.4 - using the shell:

```
server-unknown:>dataflow config server --uri http://HOST
Successfully targeted http://HOST
dataflow:>app import --uri https://dataflow.spring.io/Elston-GA-task-applications-maven
Successfully registered 3 applications from [task.composed-task-runner, task.timestamp.metadata, task.composed-task-runner.metadata, task.timestamp-batch.metadata, task.timestamp-batch, task.timestamp]
dataflow:>task create timestamp --definition "timestamp --timestamp.format=\"yyyy\""
Created new task 'timestamp'
dataflow:>task launch timestamp
Launched task 'timestamp' with execution id 1
dataflow:>
```

that resulted in the following in `cf apps`:

```
name               requested state   instances   memory   disk   urls
data-flow-server   started           1/1         2G       2G     HOST
skipper-server     started           1/1         1G       1G     HOST
timestamp          stopped           0/1         1G       1G
```

then in shell:

```
dataflow:>task destroy timestamp
Destroyed task 'timestamp'
dataflow:>
```

then running `cf apps` again, `timestamp` is gone as expected:

```
name               requested state   instances   memory   disk   urls
data-flow-server   started           1/1         2G       2G     HOST
skipper-server     started           1/1         1G       1G     HOST
```

chrisjs:
just tried the above on 2.5 via shell and all seems fine as well

cppwfs:
This is the steps you would want to reproduce the issue:
1) Execute the run of the task
2) Verify that the app is deployed on CF
3) task execution cleanup <task exec id>
4) Verify that the app is removed from CF

chrisjs:
ah your using `cleanup` -- correct me if im seeing things wrong here, but i don't think `cleanup` is implemented in CF deployer (providing "cleanup" means the same thing here):

`CloudFoundry2630AndLaterTaskLauncher` extends `AbstractCloudFoundryTaskLauncher` - `AbstractCloudFoundryTaskLauncher` defines an empty implementation of `public void cleanup(String id)` and `CloudFoundry2630AndLaterTaskLauncher` does not override it.

in kubernetes deployer we treat `cleanup` and `destroy` the same way, so guessing thats why it works

not sure why `cleanup` isn't implemented in CF deployer?

cppwfs:
ahh...    makes sense.  
Then LGTM!

cppwfs:
Rebased Merged

PR #3323, opened 2019-07-09
Add ability to stop running task

### LINKED ISSUE #858: Add a REST endpoint to stop parent and child Task executions
As a developer, I'd like to add support to stop all the parent and child the task executions. 

**Acceptance:**
- New API to stop an already running Task is added and exposed in the Root Controller
- New API is included in the default security/role mapping 
- Include new RESTdocs tests
- Integrate with RESTdocs auto-docs generation process

cppwfs:
resolves #858


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskExecutionController.java
```diff
@@ -213,6 +214,17 @@ public void cleanup(@PathVariable("id") long id) {
 		this.taskDeleteService.cleanupExecution(id);
 	}
 
+	/**
+	 * Stop a set of task executions.
+	 *
+	 * @param ids the ids of the {@link TaskExecution}s to stop
+	 */
+	@RequestMapping(value = "/stop/{id}", method = RequestMethod.POST)
```
cppwfs: At second look... we can drop the stop from the request mapping.  This will simplify things further.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskExecutionController.java
```diff
@@ -213,6 +214,17 @@ public void cleanup(@PathVariable("id") long id) {
...
+	@ResponseStatus(HttpStatus.OK)
+	public void stop(@PathVariable("id") Set<Long> ids) {
```
ilayaperumalg: I see that this would users to pass comma-separated list of task execution IDs and stop them (along with their child task executions), do we intend to expose this as an option to stop a set of executions (and their child executions)?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
```diff
@@ -192,6 +213,29 @@ public String getLog(String platformName, String taskId) {
...
 	}
 
+	@Override
+	public void stopTaskExecution(Set<Long> ids) {
+		final Map<String, Object> auditData = new LinkedHashMap<>();
+		logger.info("Stopping {} task executions.", ids.size());
+		Set<TaskExecution> taskExecutions = getTaskExecutions(ids);
+		Set<Long> childTaskExecutionIds = this.dataflowTaskExecutionDao.findChildTaskExecutionIds(ids);
+		Set<TaskExecution> childTaskExecutions = getTaskExecutions(childTaskExecutionIds);
+		for (TaskExecution taskExecution : taskExecutions) {
```
ilayaperumalg: Should we check the current status of task execution before we try to cancel?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskExecutionController.java
cppwfs: Good question.
In this case users would submit a list of task executions they wish to stop.  We can think of it as a user clicking a set of `TaskExecution`s from the UI and requesting a stop.    

While a user could request a stop for both parent and child `TaskExecution`s, they would not need to specify the child IDs if the child `TaskExecution` contained a parent id,  in that the code would stop the parent then search for the child ids to stop.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
cppwfs: Another good question.   I thought about this for a little bit.
While that would be useful, for a task that is currently running, but there are reported conditions where a pod is hung because of memory problems even though the task is complete.   In this case they want the pod deleted (which the stop for k8's effectively does).    So we would still allow for a stop request to be issued for those tasks that are marked `COMPLETE`.

cppwfs:
PR has been rebased

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/TaskExecutionService.java
```diff
@@ -51,4 +52,10 @@
 	 */
 	String getLog(String platformName, String taskId);
 
+	/**
+	 * Request the platform to stop the task executions for the id's provided.
```
ilayaperumalg: id's -> ids

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Rebased and merged as 8474daf01453b6316edb2b3c9a7999a16553b4b3

PR #3327, opened 2019-07-10
Supports task execution stop command for shell

### LINKED ISSUE #3303: Add "task execution stop <id>" command in the Shell
As a developer, I'd like to amend `task execution stop <id>` command in the Shell, so I can stop either a CTR or standalone tasks that are currently running.

**Acceptance:**
- Integrate with the endpoint delivered via https://github.com/spring-cloud/spring-cloud-dataflow/issues/858 
- Unit tests added
- Docs updated

cppwfs:
resolves #3303

Depends on PR #3323

### DIFF
--- a/spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/TaskCommands.java
+++ b/spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/TaskCommands.java
@@ -74,6 +74,7 @@ public class TaskCommands implements CommandMarker {
 	// Deploy Role
 
 	private static final String LAUNCH = "task launch";
+	private static final String STOP = "task execution stop";
 
 	// Destroy Role
 
@@ -106,6 +107,11 @@ public class TaskCommands implements CommandMarker {
 
 	@CliAvailabilityIndicator({ LAUNCH })
 	public boolean availableWithDeployRole() {
+		return dataFlowShell.hasAccess(RoleType.DEPLOY, OpsType.TASK);
+	}
+
+	@CliAvailabilityIndicator({ STOP })
+	public boolean availableWithUnDeployRole() {
 		return dataFlowShell.hasAccess(RoleType.DEPLOY, OpsType.TASK);
 	}
 
@@ -211,6 +217,12 @@ public class TaskCommands implements CommandMarker {
 		}
 		long taskExecutionId = taskOperations().launch(name, propertiesToUse, argumentsToUse);
 		return String.format("Launched task '%s' with execution id %d", name, taskExecutionId);
+	}
+
+	@CliCommand(value = STOP, help = "Stop executing tasks")
+	public String stop(@CliOption(key = { "", "ids" }, help = "the task execution id", mandatory = true) String ids) {
+		taskOperations().stop(ids);
+		return String.format("Request to stop task execution %s has been submitted", ids);
 	}
 
 	@CliCommand(value = DESTROY, help = "Destroy an existing task")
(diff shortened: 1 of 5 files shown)


application.ymlold
```diff
@@ -0,0 +1,43 @@
+info:
```
ilayaperumalg: Looks like this got added after some testing ;)

ilayaperumalg:
@cppwfs Now this PR is ready as #3323 is merged. Could you rebase this PR against master (that has the squashed version of #3323 commits)?

cppwfs:
Now ready for review.   Rebased and Squashed

spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/TaskCommands.java
```diff
@@ -213,6 +219,12 @@ public String launch(
 		return String.format("Launched task '%s' with execution id %d", name, taskExecutionId);
 	}
 
+	@CliCommand(value = STOP, help = "Stop executing tasks")
+	public String stop(@CliOption(key = { "", "ids" }, help = "the task execution id", mandatory = true) String ids) {
+		taskOperations().stop(ids);
+		return String.format("Request to stop task execution %s has been submitted", ids);
```
ilayaperumalg: Since there are multiple ids, it might be better to use:

Request to stop task execution(s) [%s] has been submitted

I can fix this on merge if you are ok with this.

ilayaperumalg:
LGTM

spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/TaskCommands.java
sabbyanandan: It seems the way how it is logged would be correct at runtime, though. For instance, if the IDs passed were, let's say: `22, 45, 33`, then the return would be:

 > Request to stop task execution 22, 45, 33 has been submitted"

Perhaps we could change it to:

 > Request to stop the task execution with id(s): 22, 45, 33 has been submitted"

spring-cloud-dataflow-shell-core/src/test/java/org/springframework/cloud/dataflow/shell/command/TaskCommandTemplate.java
```diff
@@ -81,8 +81,20 @@ public void  launch(String taskName) {
 
 		long value = (long) result.getModel().getValue(1, 1);
 		assertTrue(cr.toString().contains("with execution id " + value));
+		return value;
 	}
 
+	/**
+	 * Stop a task execution.
+	 *
+	 * @param id the id of a {@link org.springframework.cloud.task.repository.TaskExecution}
```
sabbyanandan: twice:

> id the id of a  ....

sabbyanandan: never mind; "id" is the param and what follows is just description - having no delimiter in between caused the confusion - please ignore.

spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/TaskCommands.java
ilayaperumalg: yeah, this is better!

cppwfs: 👍

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Fixed the response message and merged as 1804b735

PR #3342, opened 2019-07-15
Allows user to specify different CTR app when launching composed tasks

### LINKED ISSUE #2290: Feature request: customizable composed-task-runner application  name
Hello
Would you please consider `CTR` name to be optionally provided to  `TaskExecutionController` that will default to today's hard-coded `composed-task-runner` if not specified ?
This will allow users to  provide their own implementation (by registering the CTR task app with custom name )  for specific composed tasks  cases and at the same time  to be able to run OOTB provided  `composed- task-runner` .
Thanks

### LINKED ISSUE #3303: Add "task execution stop <id>" command in the Shell

cppwfs:
resolves #2290

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/TaskExecutionInfoService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/TaskExecutionInfoService.java
@@ -28,11 +28,19 @@ import org.springframework.cloud.dataflow.server.service.impl.TaskExecutionInfor
  * @author Daniel Serleg
  * @author Mark Pollack
  * @author David Turanski
+ * @author Glenn Renfro
  */
 public interface TaskExecutionInfoService {
 
+	/**
+	 * Create a the {@link TaskExecutionInformation} instance for the information provided.
+	 * @param taskName the name of the task definition
+	 * @param taskDeploymentProperties the deployment properties to use for the {@link TaskExecutionInformation}
+	 * @param userComposedTaskRunnerName user provided CTR app name to use.
+	 * @return instance of {@link TaskExecutionInformation}
+	 */
 	TaskExecutionInformation findTaskExecutionInformation(String taskName,
-			Map<String, String> taskDeploymentProperties);
+			Map<String, String> taskDeploymentProperties, String userComposedTaskRunnerName);
 
 	AllPlatformsTaskExecutionInformation findAllPlatformTaskExecutionInformation();
 }
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/TaskExecutionInfoService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/TaskExecutionInfoService.java
@@ -36,11 +36,11 @@ public interface TaskExecutionInfoService {
 	 * Create a the {@link TaskExecutionInformation} instance for the information provided.
 	 * @param taskName the name of the task definition
 	 * @param taskDeploymentProperties the deployment properties to use for the {@link TaskExecutionInformation}
-	 * @param userComposedTaskRunnerName user provided CTR app name to use.
+	 * @param composedTaskRunnerName user provided CTR app name to use.
 	 * @return instance of {@link TaskExecutionInformation}
 	 */
 	TaskExecutionInformation findTaskExecutionInformation(String taskName,
-			Map<String, String> taskDeploymentProperties, String userComposedTaskRunnerName);
+			Map<String, String> taskDeploymentProperties, String composedTaskRunnerName);
 
 	AllPlatformsTaskExecutionInformation findAllPlatformTaskExecutionInformation();
 }
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/TaskServiceUtils.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/TaskServiceUtils.java
@@ -65,12 +65,13 @@ public class TaskServiceUtils {
 
 	/**
 	 * Creates a properly formatted CTR definition based on the graph provided.
+	 * @param alternateComposedTaskRunnerName a ctr name to be used instead of the default.
 	 * @param graph the graph for the CTR to execute.
 	 * @param taskConfigurationProperties the properties that contain the name
 	 * of the CTR app to be launched.
 	 * @return String containing the CTR task definition.
 	 */
-	public static String createComposedTaskDefinition(String userComposedTaskRunnerName, String graph,
+	public static String createComposedTaskDefinition(String alternateComposedTaskRunnerName, String graph,
 			TaskConfigurationProperties taskConfigurationProperties) {
 		Assert.hasText(graph, "graph must not be empty or null");
 		Assert.notNull(taskConfigurationProperties,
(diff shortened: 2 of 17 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/TaskExecutionInfoService.java
```diff
@@ -28,11 +28,19 @@
...
+	/**
+	 * Create a the {@link TaskExecutionInformation} instance for the information provided.
+	 * @param taskName the name of the task definition
+	 * @param taskDeploymentProperties the deployment properties to use for the {@link TaskExecutionInformation}
+	 * @param userComposedTaskRunnerName user provided CTR app name to use.
+	 * @return instance of {@link TaskExecutionInformation}
+	 */
 	TaskExecutionInformation findTaskExecutionInformation(String taskName,
-			Map<String, String> taskDeploymentProperties);
+			Map<String, String> taskDeploymentProperties, String userComposedTaskRunnerName);
```
markpollack: 'user' feels a bit odd as a prefix for the argumetn, maybe 'composedTaskRunnerName' or 'customComposedTaskRunnerName'

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionServiceTests.java
```diff
@@ -193,6 +194,21 @@ public void executeSingleTaskTest() {
 			assertNotNull("TaskDeployment createdOn field should not be null", taskDeployment.getCreatedOn());
 		}
 
+		@Test(expected = InvalidCTRLaunchRequestException.class)
+		@DirtiesContext
+		public void executeSingleTaskTestWithCtrNameSpecified() {
+			initializeSuccessfulRegistry(appRegistry);
+			when(taskLauncher.launch(any())).thenReturn("0");
+			this.taskExecutionService.executeTask(TASK_NAME_ORIG, new HashMap<>(), new LinkedList<>(), "anotherctr");
```
markpollack: It isn't clear to me how the assertions below are going to check the ctr name as I don't see "anotherctr" in those assertions.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/service/impl/TaskServiceUtilsTests.java
```diff
@@ -53,10 +53,15 @@
 	public void testCreateComposedTaskDefinition() {
 		TaskConfigurationProperties props = new TaskConfigurationProperties();
 		props.setComposedTaskRunnerName("FOO");
-		TaskServiceUtils.createComposedTaskDefinition(BASE_GRAPH, props);
 		assertThat(TaskServiceUtils.createComposedTaskDefinition(BASE_GRAPH, props)).isEqualTo("FOO --graph=\"AAA && BBB\"");
 	}
 
+	@Test
+	public void testCreateComposedTaskDefinitionWithUserCTRName() {
```
markpollack: how does this test make assertions on a custom ctr name?

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionServiceTests.java
cppwfs: In this case we are expecting a InvalidCTRLaunchRequestException because the user attempted to apply a CTR app name against a non Composed Task.  So checking for the anotherctr is not necessary.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/TaskExecutionInfoService.java
cppwfs: makes sense.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/service/impl/TaskServiceUtilsTests.java
cppwfs: Actually I can delete this.   Replacing a CTR name only occurs during the task launch.   Good catch.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/TaskServiceUtils.java
```diff
@@ -59,12 +60,28 @@ public static boolean isComposedTaskDefinition(String dsl) {
...
+			TaskConfigurationProperties taskConfigurationProperties) {
 		Assert.hasText(graph, "graph must not be empty or null");
 		Assert.notNull(taskConfigurationProperties,
 				"taskConfigurationProperties must not be null");
 		Assert.hasText(taskConfigurationProperties.getComposedTaskRunnerName(),
 				"taskConfigurationProperties.composedTaskRunnerName must not be null");
-		return String.format("%s --graph=\"%s\"", taskConfigurationProperties.getComposedTaskRunnerName(), graph);
+		String composedTaskRunnerName = taskConfigurationProperties.getComposedTaskRunnerName();
+		if(StringUtils.hasText(userComposedTaskRunnerName)) {
+			composedTaskRunnerName = userComposedTaskRunnerName;
```
markpollack: Since the custom composed task runner name should be registered, we can check early to see if it has in fact been registered to avoid discovering at runtime that it doesn't exist, for example due to a type or other user error.  Not sure this is the best spot to do that, I guess we can d such a check much earlier in the request.

markpollack:
Shouldn't there be a change to DataFlowTemplate and the shell that takes into account passing the name of the ctr to use?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/TaskServiceUtils.java
cppwfs: Updated based on your request.  It now checks to see if the specified ctr app is valid before this method is called.

cppwfs:
With the latest commit you have the ability to use the shell.  
`task launch --name delme --alternateComposedTaskRunnerApp timestamp`

spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/TaskCommands.java
```diff
@@ -201,7 +202,10 @@ public String launch(
 					ARGUMENTS_OPTION }, help = "the commandline arguments for this launch") String arguments,
 			@CliOption(key = {
 					PLATFORM_NAME}, help = "the platform name to use for this launch",
-					unspecifiedDefaultValue = "default") String platformName)
+					unspecifiedDefaultValue = "default") String platformName,
+			@CliOption(key = {
+					ALTERNATE_CTR_APP}, help = "alternate Composed Task Runner app to use for this launch") String alternateComposedTaskRunnerApp)
```
markpollack: Can we give this a shorter name, if we are specifying it, it is implicitly an 'alternative' name, perhaps just `--composedTaskRunnerName` ?

markpollack:
rebased and merged as 80562b90a11c73e9a4686904f9c4d2b0ae5906ad

PR #3350, opened 2019-07-16
Fix CF task application log retrieval

### LINKED ISSUE #3337: Task execution logs retrieval for CF platform should use the task name instead of external execution ID
Currently, we use the external task execution ID (the task deployment ID) to fetch the logs from the platform. Since CF task execution instances are destroyed after launch is complete and the task container is available by the task name in CF, we need to use task name to get the logs

ilayaperumalg:
- Use task definition name instead of task external execution ID as the task instance is destroyed upon completion
  - Add tests

Resolves #3337

### DIFF
--- a/spring-cloud-dataflow-platform-cloudfoundry/src/main/java/org/springframework/cloud/dataflow/server/config/cloudfoundry/CloudFoundryTaskPlatformFactory.java
+++ b/spring-cloud-dataflow-platform-cloudfoundry/src/main/java/org/springframework/cloud/dataflow/server/config/cloudfoundry/CloudFoundryTaskPlatformFactory.java
@@ -45,13 +45,13 @@ import org.springframework.cloud.scheduler.spi.cloudfoundry.CloudFoundryAppSched
 import org.springframework.cloud.scheduler.spi.cloudfoundry.CloudFoundrySchedulerProperties;
 import org.springframework.cloud.scheduler.spi.core.Scheduler;
 
+import static org.springframework.cloud.dataflow.server.service.impl.DefaultTaskExecutionService.CLOUDFOUNDRY_PLATFORM_TYPE;
+
 /**
  * @author David Turanski
  * @author Ilayaperumal Gopinathan
  **/
 public class CloudFoundryTaskPlatformFactory extends AbstractTaskPlatformFactory<CloudFoundryPlatformProperties> {
-
-	private final static String PLATFORM_TYPE = "Cloud Foundry";
 
 	private final static Logger logger = LoggerFactory.getLogger(CloudFoundryTaskPlatformFactory.class);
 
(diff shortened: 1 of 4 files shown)


spring-cloud-dataflow-platform-cloudfoundry/src/main/java/org/springframework/cloud/dataflow/server/config/cloudfoundry/CloudFoundryTaskPlatformFactory.java
```diff
@@ -45,14 +45,14 @@
 import org.springframework.cloud.scheduler.spi.cloudfoundry.CloudFoundrySchedulerProperties;
 import org.springframework.cloud.scheduler.spi.core.Scheduler;
 
+import static org.springframework.cloud.dataflow.server.service.impl.DefaultTaskExecutionService.CLOUDFOUNDRY_PLATFORM_TYPE;
```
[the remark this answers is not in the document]
ilayaperumalg: Sorry about that. Fixed now along with some refactoring.

sabbyanandan:
@cppwfs: When you get a chance, it'd be great if you also can weigh-in on the last commit. If it is good to go, feel free to merge it.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
```diff
@@ -210,6 +208,13 @@ public long executeTask(String taskName, Map<String, String> taskDeploymentPrope
 
 	@Override
 	public String getLog(String platformName, String taskId) {
+		Launcher launcher = this.launcherRepository.findByName(platformName);
+		// In case of Cloud Foundry, fetching logs by external execution Id isn't valid as the execution instance is destroyed.
+		// We need to use the task name instead.
+		if (launcher != null && launcher.getType().equals(TaskPlatformFactory.CLOUDFOUNDRY_PLATFORM_TYPE)) {
+			TaskDeployment taskDeployment = this.taskDeploymentRepository.findByTaskDeploymentId(taskId);
+			taskId = taskDeployment.getTaskDefinitionName();
```
cppwfs: We get a NullPointerException if the user enters an invalid external-execution-id.   We should see an exception or something like what we get from the local deployer, "Log could not be retrieved as the task instance is not running."

tzolov: I'm afraid that this will have to be addressed in an additional PR as i've just merged this one

ilayaperumalg: I will have a PR to address this issue. 
@cppwfs Thank you for pointing this out.

ilayaperumalg: Here is the PR for this: https://github.com/spring-cloud/spring-cloud-dataflow/pull/3356. Closing this PR as it is already merged.

Issue #3351, opened 2019-07-17
Postgres: Task Search is not working


ghillert:
![image](https://user-images.githubusercontent.com/685363/61390182-04a3ac80-a856-11e9-8760-fcc58dcb4381.png)

Logs:

```
2019-07-17 05:44:13.636 ERROR 4219 --- [nio-9393-exec-7] o.s.c.d.s.c.RestControllerAdvice         : Caught exception while handling a request

org.springframework.orm.jpa.JpaSystemException: Unable to access lob stream; nested exception is org.hibernate.HibernateException: Unable to access lob stream
	at org.springframework.orm.jpa.vendor.HibernateJpaDialect.convertHibernateAccessException(HibernateJpaDialect.java:351)
	at org.springframework.orm.jpa.vendor.HibernateJpaDialect.translateExceptionIfPossible(HibernateJpaDialect.java:253)
	at org.springframework.orm.jpa.AbstractEntityManagerFactoryBean.translateExceptionIfPossible(AbstractEntityManagerFactoryBean.java:527)
	at org.springframework.dao.support.ChainedPersistenceExceptionTranslator.translateExceptionIfPossible(ChainedPersistenceExceptionTranslator.java:61)
	at org.springframework.dao.support.DataAccessUtils.translateIfNecessary(DataAccessUtils.java:242)
	at org.springframework.dao.support.PersistenceExceptionTranslationInterceptor.invoke(PersistenceExceptionTranslationInterceptor.java:153)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:186)
	at org.springframework.data.jpa.repository.support.CrudMethodMetadataPostProcessor$CrudMethodMetadataPopulatingMethodInterceptor.invoke(CrudMethodMetadataPostProcessor.java:138)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:186)
	at org.springframework.aop.interceptor.ExposeInvocationInterceptor.invoke(ExposeInvocationInterceptor.java:93)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:186)
	at org.springframework.data.repository.core.support.SurroundingTransactionDetectorMethodInterceptor.invoke(SurroundingTransactionDetectorMethodInterceptor.java:61)
[...]

PR #3362, opened 2019-07-20
gh-3351 Fix - Postgres: Task Search is not working

ghillert:
resolves https://github.com/spring-cloud/spring-cloud-dataflow/issues/3351


ilayaperumalg:
To add some context:

From @ghillert: Postgres Clob/Lob (Large Objects) even for read operation need to be wrapped in transactions https://www.postgresql.org/docs/current/largeobjects.html?origin_team=T024LQKAS

ilayaperumalg:
We need to create the follow up stories based on the following observation from @ghillert :

- Our Controllers are still too busy, to many operations that should rather be pushed down into a Service
- Controllers should never access Dao/Repositories directly (We do that all over the place)
- Repositories should not be transactional but rather the services should be (Right now some services are, some repos are etc.)

ilayaperumalg:
LGTM, merging.

PR #3364, opened 2019-07-22
TaskValidation now handles invalid task definition name

### LINKED ISSUE #3363: Task Validate with invalid task name gives NPE
Should return a Invalid Task Definition exception.

cppwfs:
resolves #3363


ilayaperumalg:
LGTM, merging.

PR #3375, opened 2019-07-24
Default to a single task platform

### LINKED ISSUE #3367: Task launch fails when "default" platform configuration isn't set
As a user, I have a custom platform configuration for Task launches (see below); however, when launching the Task without the `platformName`, I get the following error.

Config:
```
    MAVEN_REMOTEREPOSITORIES[REPO1]_URL: http://repo.spring.io/test-libs-snapshot-local
    SPRING_CLOUD_DATAFLOW_TASK_PLATFORM_CLOUDFOUNDRY_ACCOUNTS[sabbycfdev]_CONNECTION_URL: 
    SPRING_CLOUD_DATAFLOW_TASK_PLATFORM_CLOUDFOUNDRY_ACCOUNTS[sabbycfdev]_CONNECTION_ORG: 
    SPRING_CLOUD_DATAFLOW_TASK_PLATFORM_CLOUDFOUNDRY_ACCOUNTS[sabbycfdev]_CONNECTION_SPACE: 
    SPRING_CLOUD_DATAFLOW_TASK_PLATFORM_CLOUDFOUNDRY_ACCOUNTS[sabbycfdev]_CONNECTION_DOMAIN: 
    SPRING_CLOUD_DATAFLOW_TASK_PLATFORM_CLOUDFOUNDRY_ACCOUNTS[sabbycfdev]_CONNECTION_USERNAME: 
    SPRING_CLOUD_DATAFLOW_TASK_PLATFORM_CLOUDFOUNDRY_ACCOUNTS[sabbycfdev]_CONNECTION_PASSWORD: 
    SPRING_CLOUD_DATAFLOW_TASK_PLATFORM_CLOUDFOUNDRY_ACCOUNTS[sabbycfdev]_DEPLOYMENT_SERVICES: 
    SPRING_PROFILES_ACTIVE: cloud
    JBP_CONFIG_SPRING_AUTO_RECONFIGURATION: '{enabled: false}'
```

Error:
```
dataflow:>task launch --name task1
Command failed org.springframework.cloud.dataflow.rest.client.DataFlowClientException: No Launcher found for the platform named 'default'.  Available platform names are [sabbycfdev]
No Launcher found for the platform named 'default'.  Available platform names are [sabbycfdev]
org.springframework.cloud.dataflow.rest.client.DataFlowClientException: No Launcher found for the platform named 'default'.  Available platform names are [sabbycfdev]
    at org.springframework.cloud.dataflow.rest.client.VndErrorResponseErrorHandler.handleError(VndErrorResponseErrorHandler.java:65)
    at org.springframework.web.client.ResponseErrorHandler.handleError(ResponseErrorHandler.java:63)
    at org.springframework.web.client.RestTemplate.handleResponse(RestTemplate.java:778)
    at org.springframework.web.client.RestTemplate.doExecute(RestTemplate.java:736)
    at org.springframework.web.client.RestTemplate.execute(RestTemplate.java:670)
    at org.springframework.web.client.RestTemplate.postForObject(RestTemplate.java:414)
    at org.springframework.cloud.dataflow.rest.client.TaskTemplate.launch(TaskTemplate.java:162)
    at org.springframework.cloud.dataflow.shell.command.TaskCommands.launch(TaskCommands.java:222)
    at java.lang.reflect.Method.invoke(Method.java:498)
    at org.springframework.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:282)
    at org.springframework.shell.core.SimpleExecutionStrategy.invoke(SimpleExecutionStrategy.java:68)
    at org.springframework.shell.core.SimpleExecutionStrategy.execute(SimpleExecutionStrategy.java:59)
    at org.springframework.shell.core.AbstractShell.executeCommand(AbstractShell.java:134)
    at org.springframework.shell.core.JLineShell.promptLoop(JLineShell.java:533)
    at org.springframework.shell.core.JLineShell.run(JLineShell.java:179)
    at java.lang.Thread.run(Thread.java:748)
```
Since I only have 1 platform defined, it'd be nice to use that instead of complaining about no "default" platform. 

At the same time, it'd be better to throw this error, when  "default" is not available, but if we have >1 platform configurations. In other words, it is the user's responsibility to make a decision as to where to launch the Task.

Also: What do we do for Streams in Skipper?

jvalkeal:
- If user have only one task platform and that named platform is not
  given with properties, switch to that in a DefaultTaskExecutionService.
- We also log that a single platform was replaced.
- Other option would have been to forcefully register single platform as
  'default' as it's done in a skipper but looks like with tasks that
  might be more intrusive.
- Fixes #3367

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
@@ -153,15 +153,23 @@ public class DefaultTaskExecutionService implements TaskExecutionService {
 		this.dataflowTaskExecutionDao = dataflowTaskExecutionDao;
 	}
 
-
-
 	@Override
-	public long executeTask(String taskName, Map<String, String> taskDeploymentProperties, List<String> commandLineArgs, String composedTaskRunnerName) {
+	public long executeTask(String taskName, Map<String, String> taskDeploymentProperties, List<String> commandLineArgs,
+			String composedTaskRunnerName) {
 
 		String platformName = taskDeploymentProperties.get(TASK_PLATFORM_NAME);
+
+		// If not given, use 'default'
 		if (!StringUtils.hasText(platformName)) {
 			platformName = "default";
 		}
+		// In case we have exactly one launcher, we override to that
+		List<String> launcherNames = StreamSupport.stream(launcherRepository.findAll().spliterator(), false)
+				.map(Launcher::getName).collect(Collectors.toList());
+		if (launcherNames.size() == 1) {
+			platformName = launcherNames.get(0);
+		}
+
 		// Remove since the key for task platform name will not pass validation for app, deployer, or scheduler prefix
 		if (taskDeploymentProperties.containsKey(TASK_PLATFORM_NAME)) {
 			taskDeploymentProperties.remove(TASK_PLATFORM_NAME);
--- a/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionServiceTests.java
+++ b/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionServiceTests.java
@@ -172,6 +172,34 @@ public abstract class DefaultTaskExecutionServiceTests {
 	@Autowired
 	TaskConfigurationProperties taskConfigurationProperties;
 
+	@AutoConfigureTestDatabase(replace = Replace.ANY)
+	public static class SimpleDefaultPlatformTests extends DefaultTaskExecutionServiceTests {
+
+		@Before
+		public void setupMocks() {
+			// not adding platform name as default as we want to check that this only one
+			// gets replaced
+			this.launcherRepository.save(new Launcher("fakeplatformname", "local", taskLauncher));
+			taskDefinitionRepository.save(new TaskDefinition(TASK_NAME_ORIG, "demo"));
+			taskDefinitionRepository.findAll();
+		}
+
+		@Test
+		@DirtiesContext
+		public void executeSingleTaskDefaultsToExistingSinglePlatformTest() {
+			initializeSuccessfulRegistry(appRegistry);
+			when(taskLauncher.launch(any())).thenReturn("0");
+			assertEquals(1L, this.taskExecutionService.executeTask(TASK_NAME_ORIG, new HashMap<>(), new LinkedList<>()));
+
+			TaskDeployment taskDeployment = taskDeploymentRepository.findByTaskDeploymentId("0");
+			assertNotNull("TaskDeployment should not be null", taskDeployment);
+			assertEquals("0", taskDeployment.getTaskDeploymentId());
+			assertEquals(TASK_NAME_ORIG, taskDeployment.getTaskDefinitionName());
+			assertEquals("fakeplatformname", taskDeployment.getPlatformName());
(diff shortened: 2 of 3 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
```diff
@@ -264,9 +264,18 @@ private TaskLauncher findTaskLauncher(String platformName) {
 					StreamSupport.stream(launcherRepository.findAll().spliterator(), false)
 							.map(Launcher::getName)
 							.collect(Collectors.toList());
-			throw new IllegalStateException(String.format("No Launcher found for the platform named '%s'.  " +
-							"Available platform names are %s",
-					platformName, launcherNames));
+			if (launcherNames.size() == 1) {
```
ilayaperumalg: I think we should fail fast when the platform name is explicitly given and it doesn't exist. Instead of always using the single platform even if the platform doesn't exist with the provided platform name, I think we need to use the single platform only if the platform name is not given explicitly. We probably need to fix our code [here](https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java#L161)

sabbyanandan: Whatever we end up deciding, let's please also make sure Skipper (for streams) follow the same pattern. We absolutely don't want to have a different platform-resolution behavior between streams vs. tasks.

ilayaperumalg: In a single platform with the name other than `default`, Skipper inserts a new deployer named `default` which means we would have the same platform named `default` and the actual name. We can either change this logic in Skipper or change the behavior here in this PR.

sabbyanandan: I'd vote for making the change here in this PR for consistency. @markpollack: What are your thoughts?

markpollack: I don't remember why we picked that strategy in skipper, but consistency is a good argument.

ilayaperumalg:
LGTM, merging.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
ilayaperumalg: Was about to merge, but found a reason why we have the `default` platform being created.

There are some invocations to fetch the details about task executions use `default` as the platform name if none is provided. For instance, getLog() is one such case. I think we need to apply the same logic in those places as well.

ilayaperumalg:
I think we also need to change the client-side to handle this case as well. For instance, the `task launch` command sets the platform name to be `default` if it is not explicitly set.

jvalkeal:
Client side should not matter even if it uses `default`. Even if we remove that, user can always set it to `default` so it needs to work either way.

jvalkeal:
Though maybe we should just remove `default` from shell. For streams shell doesn't use any default platform name. It would be same then.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed, rebased and merged as a4267ca3

PR #3376, opened 2019-07-24
Verifies that all task executions that are being stopped have a external-execution-id

### LINKED ISSUE #3369: Stopping an app that has been started outside SCDF results in a NPE for SCDF
SCDF should throw exception if any parent or child does not have an external-execution-id populated.

cppwfs:
resolves #3369

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/TaskExecutionMissingExternalIdException.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/TaskExecutionMissingExternalIdException.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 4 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/TaskExecutionMissingExternalIdException.java
```diff
@@ -0,0 +1,51 @@
...
+public class TaskExecutionMissingExternalIdException extends RuntimeException {
+
+	private static final long serialVersionUID = -303914974762305117L;
+
+	/**
+	 * Create a new exception that can handle a single {@link TaskExecution} id.
+	 *
+	 * @param id the id of the {@link TaskExecution} that is missing external execution id.
+	 */
+	public TaskExecutionMissingExternalIdException(long id) {
```
ilayaperumalg: Should we add this exception into RESTControllerAdvice by marking them with the response status of HttpStatus.CONFLICT (409 Error)?

ghillert: I think `400` /  `Bad Request` is fine, no?

ghillert:
rebased and merged with commit eb965aa7ce4845576d379f23ae26c8b0b417aee4

Issue #1654, opened 2017-09-14
As stream creator I should be able to provide description for a stream & task definition.


mbhuie:
As stream creator I should be able to provide description for a stream & task definition.

This can make it easy for a new joiner to know what is the purpose of the stream and task.

PR #3388, opened 2019-07-27
Add description field to StreamDefinition and TaskDefinition

sylank:
As stream creator, I should be able to provide a description for a stream & task definition

Resolves: #1654


spring-cloud-dataflow-server-core/src/main/resources/migration/1.7.x/migration-db2.sql
```diff
@@ -8,3 +8,6 @@
         created_On timestamp,
         primary key (id)
     );
+
+    alter table stream_definitions add description varchar(255)
```
sylank: I' am not sure these lines are needed.

spring-cloud-dataflow-server-core/src/main/resources/org/springframework/cloud/dataflow/server/db/migration/h2/V1__INITIAL_SETUP.sql
```diff
@@ -37,12 +37,14 @@ create table audit_records (
 create table stream_definitions (
   definition_name varchar(255) not null,
   definition clob,
+  description varchar(255),
```
jvalkeal: Actual database migrations are in https://github.com/spring-cloud/spring-cloud-dataflow/tree/master/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/db/migration and this would result a new schema version. With embedded we can really just keep existing V1 schema as DB is always created from scratch when server starts.

Also those old `1.7.x` migration scripts are not used anymore as things are handled via flyway.

sabbyanandan:
Hi, @sylank. Glad to see your contribution once again! Let us know when this is ready, and we would love to collaborate and merge it.

jvalkeal:
Sorry if I were a little unclear about schema changes. We can't modify existing migrations like `V1__Initial_Setup.java`. We need to have a version 2 like `V2__Add_Descriptions.java` either extending our `AbstractMigration` or flyways `BaseJavaMigration` and in that modify table. For example postgres something like `alter table task_definitions add description varchar(255);`.

Schema for embedded database is a little different thing as its always getting re-created so it really doesn't matter if you always use `V1` or do same dance by adding `V2` as well.

For example if you first run `spring-cloud-dataflow-server-2.2.0.RELEASE.jar` with same db settings and then launch snapshot with your changes, you can see that schema is not upgraded as flyway already has `V1` schema applied.
```
$ docker run --name test-postgres -d -p 5432:5432 -e POSTGRES_PASSWORD=spring -e POSTGRES_USER=spring -e POSTGRES_DB=dataflow postgres:10

$ java -jar spring-cloud-dataflow-server/target/spring-cloud-dataflow-server-2.3.0.BUILD-SNAPSHOT.jar --spring.datasource.url='jdbc:postgresql://localhost:5432/dataflow' --spring.datasource.username=spring --spring.datasource.password=spring --spring.datasource.driverClassName=org.postgresql.Driver

$ docker exec -it test-postgres psql -U spring -d dataflow
dataflow=# \d+ task_definitions
                                          Table "public.task_definitions"
     Column      |          Type          | Collation | Nullable | Default | Storage  | Stats target | Description 
-----------------+------------------------+-----------+----------+---------+----------+--------------+-------------
 definition_name | character varying(255) |           | not null |         | extended |              | 
 definition      | text                   |           |          |         | extended |              | 
Indexes:
    "task_definitions_pkey" PRIMARY KEY, btree (definition_name)
```
Adding a column should be straightforward but we usually get surprises with oracle and db2 and sometimes with sqlserver. We have a separate db tests in our acceptance test suite where we test against real databases.

sylank:
@jvalkeal, your comment was really helpful. I will check it, thank you. 😄
I had some issues with git e-mail addresses. I would like to use my private email address in the commits. Could you select my `@live.com` address when you merge it? 😃

jvalkeal:
Thanks, I think this is looking fairly good. I'll play it with a bit more to see how all this might look like from shell/ui/dsl perspective.

jvalkeal:
thanks mate, squashed and merged per 7280e270735fad6a76541a68560299a4e904b02b

PR #3389, opened 2019-07-29
Added docs for launching a CT with a custom CTR

### LINKED ISSUE #3377: Document how to use a custom CTR when launching Composed Tasks
As a user, to use the newly added ability to switch to a custom CTR (see: https://github.com/spring-cloud/spring-cloud-dataflow/commit/80562b90a11c73e9a4686904f9c4d2b0ae5906ad), I'd like to know how to use it with an example. Preferably with a Shell sample.

A rough take on it:

```
dataflow:>app register --name best-ctr --type task --uri maven://the.best.ctr.composed-task-runner:1.0.0.RELEASE

dataflow:>task create foo --definition "te:timestamp &&  tr:timestamp"
Created new task 'foo'

dataflow:>task launch --name foo
foo      foo-te   foo-tr

dataflow:>task launch --name foo --
task launch --name foo --arguments                 task launch --name foo --composedTaskRunnerName
task launch --name foo --platformName              task launch --name foo --properties
task launch --name foo --propertiesFile

dataflow:>task launch --name foo --composedTaskRunnerName best-ctr
```

cppwfs:
resolves #3377

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/tasks.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/tasks.adoc
@@ -625,9 +625,20 @@ In some cases a user will need to launch a composed task using a Composed Task R
 To do this a user will need to specify the `--composedTaskRunnerName`  at task launch time as shown below:
 [source,bash,options="nowrap"]
 ----
-dataflow:>task launch myComposedTask --composedTaskRunnerName myComposedTaskRunnerApp
-Launched task 'myComposedTask' with execution id 110
-dataflow:>
+dataflow:>app register --name best-ctr --type task --uri maven://the.best.ctr.composed-task-runner:1.0.0.RELEASE
+
+dataflow:>task create mycomposedtask --definition "te:timestamp &&  tr:timestamp"
+Created new task 'mycomposedtask'
+
+dataflow:>task launch --name mycomposedtask
+mycomposedtask      mycomposedtask-te   mycomposedtask-tr
+
+dataflow:>task launch --name mycomposedtask --
+task launch --name mycomposedtask --arguments                 task launch --name mycomposedtask --composedTaskRunnerName
+task launch --name mycomposedtask --platformName              task launch --name mycomposedtask --properties
+task launch --name mycomposedtask --propertiesFile
+
+dataflow:>task launch --name mycomposedtask --composedTaskRunnerName best-ctr
 ----
 
 NOTE: The app specified by the  `composedTaskRunnerName` needs to be a task registered in the Application Registry.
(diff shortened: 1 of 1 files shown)


spring-cloud-dataflow-docs/src/main/asciidoc/tasks.adoc
```diff
@@ -619,6 +619,19 @@ dataflow:>task launch my-composed-task --arguments "--increment-instance-enabled
 Launched task 'my-composed-task'
 ----
 
+===== Launching a Composed Task using Custom Composed Task Runner
+
+In some cases a user will need to launch a composed task using a Composed Task Runner other than the one specified with `spring.cloud.dataflow.task.composedTaskRunnerName` property.
```
sabbyanandan: To make it more descriptive, what if we show the registration of a custom CTR and then highlight its use of it at the time of launch?

I had dropped an example in the issue description: https://github.com/spring-cloud/spring-cloud-dataflow/issues/3377

sabbyanandan: Sorry, one other suggestion. Can we rephrase this to make the customized version of CTR stand out? Something like:

> In some cases a user will need to launch a composed task using a custom version of a Composed Task Runner other than default application that is shipped out-of-the-box.

spring-cloud-dataflow-docs/src/main/asciidoc/tasks.adoc
```diff
@@ -619,6 +619,30 @@ dataflow:>task launch my-composed-task --arguments "--increment-instance-enabled
...
+To do this a user will need to specify the `--composedTaskRunnerName`  at task launch time as shown below:
```
sabbyanandan: > To do this, a user will need to register the custom version of the Composed Task Runner and then specify the `--composedTaskRunnerName` property pointing to the custom application at task launch as shown below:

spring-cloud-dataflow-docs/src/main/asciidoc/tasks.adoc
```diff
@@ -619,6 +619,30 @@ dataflow:>task launch my-composed-task --arguments "--increment-instance-enabled
...
+[source,bash,options="nowrap"]
+----
+dataflow:>app register --name best-ctr --type task --uri maven://the.best.ctr.composed-task-runner:1.0.0.RELEASE
+
+dataflow:>task create mycomposedtask --definition "te:timestamp &&  tr:timestamp"
+Created new task 'mycomposedtask'
+
+dataflow:>task launch --name mycomposedtask
```
sabbyanandan: Maybe we can remove this:

```
dataflow:>task launch --name mycomposedtask
mycomposedtask      mycomposedtask-te   mycomposedtask-tr
```

Just noise from my test trials. Sorry.

cppwfs:
Updated based on review

cppwfs:
rebased squashed merged.   @sabbyanandan thanks for the review

PR #3404, opened 2019-08-05
Remove checks for missing steps in step execution list code.

### LINKED ISSUE #3381: Step execution list throws exception when a step execution status is UNKNOWN
**Description:**
When I use the method stepExecutionList() of the org.springframework.cloud.dataflow.rest.client.JobOperations class I sporadically receive the following error: 

```
org.springframework.cloud.dataflow.rest.client.DataFlowClientException: Id must not be null!
        at org.springframework.cloud.dataflow.rest.client.VndErrorResponseErrorHandler.handleError(VndErrorResponseErrorHandler.java:65) ~[spring-cloud-dataflow-rest-client-2.1.2.RELEASE.jar!/:2.1.2.RELEASE]
        at org.springframework.web.client.ResponseErrorHandler.handleError(ResponseErrorHandler.java:63) ~[spring-web-5.1.8.RELEASE.jar!/:5.1.8.RELEASE]
        at org.springframework.web.client.RestTemplate.handleResponse(RestTemplate.java:778) ~[spring-web-5.1.8.RELEASE.jar!/:5.1.8.RELEASE]
        at org.springframework.web.client.RestTemplate.doExecute(RestTemplate.java:736) ~[spring-web-5.1.8.RELEASE.jar!/:5.1.8.RELEASE]
        at org.springframework.web.client.RestTemplate.execute(RestTemplate.java:670) ~[spring-web-5.1.8.RELEASE.jar!/:5.1.8.RELEASE]
        at org.springframework.web.client.RestTemplate.getForObject(RestTemplate.java:311) ~[spring-web-5.1.8.RELEASE.jar!/:5.1.8.RELEASE]
        at org.springframework.cloud.dataflow.rest.client.JobTemplate.stepExecutionList(JobTemplate.java:140) ~[spring-cloud-dataflow-rest-client-2.1.2.RELEASE.jar!/:2.1.2.RELEASE]
        at de.ubs.xdm3.core.dataflow.access.RestDataFlowAccess.getSteps(RestDataFlowAccess.java:262) ~[xdm3-core-dataflow-access-1.0-SNAPSHOT.jar!/:na]
```

**Release versions:**
Spring Boot: 2.1.1.RELEASE
Spring Cloud Data Flow: 2.1.2.RELEASE

**Steps to reproduce:**

1. Execute a SCDF Task that forks a spring batch job with a couple of steps. 

1. Actively polling the status of the running job sometimes causes the "Id must not be null!" error. 

**Screenshots:**
screen

cppwfs:
resolves #3381

In short the code that was removed made sure that stepExecutions that are associated with a job returned from the jobLocator are represented in the step execution list.
However, the map within the jobLocator is never populated by dataflow.  In fact jobLocator's usage in this class is really a no-op.

The cause of the problem the user reported came from the fact that roughly the same query was executed 2 times within the same method and from time to time the batch app would update the batch_step_execution table with a new entry, and thus the SCDF code would interpret this as a stepExecution that was in the jobLocator job and thus enter a stubbed entry into the ResultSet.
The fix was to remove the jobLocator code from within step.

We need to create a story to review the SimpleJobService and remove the usage of the jobLocator and get rid of unused methods.  https://github.com/spring-cloud/spring-cloud-dataflow/issues/3405

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
@@ -105,12 +105,6 @@ public class SimpleJobService implements JobService, DisposableBean {
 
 	public SimpleJobService(SearchableJobInstanceDao jobInstanceDao, SearchableJobExecutionDao jobExecutionDao,
 			SearchableStepExecutionDao stepExecutionDao, JobRepository jobRepository, JobLauncher jobLauncher,
-			ListableJobLocator jobLocator, ExecutionContextDao executionContextDao) {
-		this(jobInstanceDao, jobExecutionDao, stepExecutionDao, jobRepository, jobLauncher, jobLocator, executionContextDao, null);
-	}
-
-	public SimpleJobService(SearchableJobInstanceDao jobInstanceDao, SearchableJobExecutionDao jobExecutionDao,
-			SearchableStepExecutionDao stepExecutionDao, JobRepository jobRepository, JobLauncher jobLauncher,
 			ListableJobLocator jobLocator, ExecutionContextDao executionContextDao, JobOperator jsrJobOperator) {
 		super();
 		this.jobInstanceDao = jobInstanceDao;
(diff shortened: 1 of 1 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
```diff
@@ -139,40 +133,6 @@ public SimpleJobService(SearchableJobInstanceDao jobInstanceDao, SearchableJobEx
 
 		stepExecutionDao.addStepExecutions(jobExecution);
 
-		String jobName = jobExecution.getJobInstance() == null ? jobInstanceDao.getJobInstance(jobExecution).getJobName() : jobExecution.getJobInstance().getJobName();
-		Collection<String> missingStepNames = new LinkedHashSet<String>();
-
-		if (jobName != null) {
```
ilayaperumalg: It would be good if we know why we added this logic of finding missing step names at first place.

cppwfs: @ilayaperumalg Great Question! This looks to be a copy paste from Spring Batch Admin https://github.com/spring-attic/spring-batch-admin/blob/master/spring-batch-admin-manager/src/main/java/org/springframework/batch/admin/service/SimpleJobService.java.  This was pulled in as a part of SCDF 2.2.0 when we removed the need to have the Spring Batch Admin. 
SCDF never used these features and thus only passed in the `MapJobRegistry` just to fulfill the requirement of the `SimpleJobServiceFactoryBean` to have a `JobLocator`.   This can be seen in the SCDF 1.7 release.

ilayaperumalg:
LGTM, merging.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
ilayaperumalg: @cppwfs Thanks for the description!

ilayaperumalg:
Rebased and merged as 6b62d01073dc712bd1d05b111ee7f4f3532443ad. Cherry-picked and pushed to 2.2.x as 139f7430f71debb162b4dae660b3fc92df2042d4

PR #3394, opened 2019-07-31
Remove k8s "jobs" permissions from incorrect apigroup

jcrowthe:
In Kubernetes, the `jobs` API resource is under the `batch` API Group, not `extensions`. This PR is a successor to #2956 which properly added jobs under the batch apigroup, but did not remove it from the extensions apigroup.

Kubernetes doesn't error out when attempting to apply `server-roles.yaml` as-is and when the operator is using the cluster-admin role. However when the operator does not have cluster-admin (for example in a multi-tenant kubernetes cluster where tenants have admin only in their namespace), this will error out. As such, this issue has likely gone unnoticed.

Evidence:
```
$ kubectl api-resources
NAME                               SHORTNAMES        APIGROUP                       NAMESPACED   KIND
cronjobs                           cj                batch                          true         CronJob
jobs                                                 batch                          true         Job
...
```

and notably the lack of `jobs` under `extensions`:
```
$ kubectl api-resources | grep extensions
customresourcedefinitions          crd,crds          apiextensions.k8s.io           false        CustomResourceDefinition
daemonsets                         ds                extensions                     true         DaemonSet
deployments                        deploy            extensions                     true         Deployment
ingresses                          ing               extensions                     true         Ingress
networkpolicies                    netpol            extensions                     true         NetworkPolicy
podsecuritypolicies                psp               extensions                     false        PodSecurityPolicy
replicasets                        rs                extensions                     true         ReplicaSet
```

PR #3417, opened 2019-08-06
"pods/log" does not have create/delete/update

jcrowthe:
In Kubernetes, the "pods/log" resource type do not have verbs for `"create", "delete", or "update"`. This PR moves "pods/logs" down to have the proper list of verbs.

Similar to what is noted in #3394:
> Kubernetes doesn't error out when attempting to apply server-roles.yaml as-is and when the operator is using the cluster-admin role. However when the operator does not have cluster-admin (for example in a multi-tenant kubernetes cluster where tenants have admin only in their namespace), this will error out. As such, this issue has likely gone unnoticed.

Attempt to apply this role before this PR:
```
Error from server (Forbidden): error when creating "STDIN": roles.rbac.authorization.k8s.io "scdf-role" is forbidden: user ... is attempting to grant RBAC permissions not currently held:
{APIGroups:[""], Resources:["pods/log"], Verbs:["create" "delete" "update"]}
```

Successful attempt after this PR:
```
role.rbac.authorization.k8s.io/scdf-role created
```

### DIFF
--- a/src/kubernetes/server/server-roles.yaml
+++ b/src/kubernetes/server/server-roles.yaml
@@ -4,10 +4,10 @@ metadata:
   name: scdf-role
 rules:
   - apiGroups: [""]
-    resources: ["services", "pods", "pods/log", "replicationcontrollers", "persistentvolumeclaims"]
+    resources: ["services", "pods", "replicationcontrollers", "persistentvolumeclaims"]
     verbs: ["get", "list", "watch", "create", "delete", "update"]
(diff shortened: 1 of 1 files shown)


ilayaperumalg:
@jcrowthe Thanks for fixing this!

ilayaperumalg:
LGTM

chrisjs:
merged in a5b0f5bdbff3d9c40b875918ae9ec8fee41eed5b

PR #3422, opened 2019-08-09
Fix restdocs generation

### LINKED ISSUE #3421: RESTFul API Docs not being generated for Reference guide
Starting at build https://build.spring.io/browse/SCD-BMASTER-3165 the RESTFUL-API docs stopped producing results.

You can add 
```
		<dependency>
			<groupId>org.junit.vintage</groupId>
			<artifactId>junit-vintage-engine</artifactId>
			<version>5.2.0</version>
			<scope>test</scope>
		</dependency>
```
to the spring-cloud-dataflow-classic-docs as a workaround.

jvalkeal:
- Cloud build is now using latest junit5 deps which caused
  mismatch with surefire plugin. At least internally it got
  NoSuchMethodError as surefire also depends on junit5.
- We use explicit surefire configuration and this commit
  changes it to same version as cloud build which seems
  to fix this junit5 issue and thus putting restdocs
  back to business.
- Fixes #3421

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -275,7 +275,7 @@
 			<plugin>
 				<groupId>org.apache.maven.plugins</groupId>
 				<artifactId>maven-surefire-plugin</artifactId>
-				<version>2.22.0</version>
+				<version>2.22.1</version>
 				<configuration>
 					<includes>
 						<include>**/*Tests.java</include>


cppwfs:
LGTM, Rebased

PR #3427, opened 2019-08-12
Add ability for user to set platform to use when stopping tasks

### LINKED ISSUE #3382: Cannot stop a remote partitioned job on the platform where it was launched
In this scenario a user launches a Spring Cloud Task outside of SCDF and they populate the external-execution-id with the correct platform execution id  ( Spring Batch Partition app).  When the user attempts to stop a task execution that was launched outside of dataflow and has an external-execution-id it will fail with a null pointer exception when attempting to get the platform name from the taskDeploymentRepository.   This because the external deployer or external execution does not insert the entry into the taskDeploymentRepository.

cppwfs:
This is for those cases where user executes a task with a Spring Cloud Deployer outside SCDF and populates the external-execution-id
resolves #3382

### DIFF
--- a/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/TaskTemplate.java
+++ b/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/TaskTemplate.java
@@ -176,6 +176,13 @@ public class TaskTemplate implements TaskOperations {
 	}
 
 	@Override
+	public void stop(String ids, String platform) {
+		MultiValueMap<String, Object> values = new LinkedMultiValueMap<>();
+		values.add("platform", platform);
+		restTemplate.postForLocation(executionLink.expand(ids).getHref(),values);
+	}
+
+	@Override
 	public void destroy(String name) {
 		restTemplate.delete(definitionLink.expand(name).getHref(), Collections.singletonMap("name", name));
 	}
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskExecutionController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskExecutionController.java
@@ -49,6 +49,7 @@ import org.springframework.hateoas.server.ExposesResourceFor;
 import org.springframework.hateoas.server.mvc.RepresentationModelAssemblerSupport;
 import org.springframework.http.HttpStatus;
 import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestMethod;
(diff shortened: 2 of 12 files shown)


cppwfs:
Rebased to latest and resolved conflicts

spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/TaskTemplate.java
```diff
@@ -175,6 +175,13 @@ public void stop(String ids) {
 		restTemplate.postForLocation(executionLink.expand(ids).getHref(),values);
 	}
 
+	@Override
+	public void stop(String ids, String platform) {
+		MultiValueMap<String, Object> values = new LinkedMultiValueMap<>();
```
markpollack: MultiValueMap is for storing multiple values under the same key, that isn't needed here, a standard Map/HashMap will be sufficient.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskExecutionController.java
```diff
@@ -231,8 +232,14 @@ public void cleanup(
 	 */
 	@RequestMapping(value = "/{id}", method = RequestMethod.POST)
 	@ResponseStatus(HttpStatus.OK)
-	public void stop(@PathVariable("id") Set<Long> ids) {
-		this.taskExecutionService.stopTaskExecution(ids);
+	public void stop(@PathVariable("id") Set<Long> ids,
+	@RequestParam(defaultValue = "", name="platform") String platform) {
+		if(StringUtils.hasText(platform)) {
```
markpollack: This logic can be pushed down into the service.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
```diff
@@ -265,24 +265,51 @@ public String getLog(String platformName, String taskId) {
...
 		childTaskExecutions.forEach(childTaskExecution -> {
 			cancelTaskExecution(childTaskExecution);
 		});
 
-		long numberOfExecutionsStopped = ids.size() + childTaskExecutionIds.size();
+		updateAuditInfoForTaskStops(taskExecutions.size() + childTaskExecutions.size());
+	}
+
+	@Override
+	public void stopTaskExecution(Set<Long> ids, String platform) {
```
markpollack: There is code duplication between this method and stopTaskExecution(ids), can be removed.

spring-cloud-dataflow-shell-core/src/main/java/org/springframework/cloud/dataflow/shell/command/TaskCommands.java
```diff
@@ -65,7 +65,7 @@
 	private static final String PROPERTIES_OPTION = "properties";
 	private static final String PROPERTIES_FILE_OPTION = "propertiesFile";
 	private static final String ARGUMENTS_OPTION = "arguments";
-	private static final String PLATFORM_OPTION = "platformName";
+	private static final String PLATFORM_OPTION = "platform";
```
markpollack: In StreamCommands the option `platformName` is used, should probably switch to that to be consistent.  `platformName` is also used in `TasksLogsController`

markpollack:
I addressed my own comments and merged it in 422683622d8309d65dcfc616990ea96d7b93925e

PR #3441, opened 2019-08-16
Fix RootController test

### LINKED ISSUE #3412: Revisit RootControllerTests
`RootControllerTests` broke during boot/hateoas upgrades. Looking original test class it doesn't even test anything what makes sense. Output from a root what only test is asserting is not even something what a real server outputs.

ilayaperumalg:
- Fix the test to check for the SCDF REST response for root path `/`
 - Updated the current root path response

Resolves #3412


cppwfs:
LGTM

cppwfs:
Rebased Merged

PR #3459, opened 2019-08-22
Provide a hint as to why the Exception was thrown at SCDF startup

### LINKED ISSUE #3458: Provide hint when BadSqlGrammarException is thrown at startup
Currently users will get a `BadSqlGrammarException` when attempting to start a SCDF server in a schema where Spring Batch or Spring Cloud Tasks have already been run.  We should give them a hint on why SCDF failed to start.

cppwfs:
* Tried to use the callback framework however, there was no callback for beforeBaseline errors.
* Had to add some cleanup code to SimpleDefaultPlatformTests.  The test was failing even in master.  This had nothing to do with the code introduced

resolves #3458

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/db/migration/AbstractCallback.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/db/migration/AbstractCallback.java
@@ -72,11 +72,13 @@ public abstract class AbstractCallback implements Callback {
 			runner.execute(context.getConnection(), getCommands(event, context));
 		}
 		catch(Exception sqe) {
-			if (sqe instanceof BadSqlGrammarException && sqe.getMessage().contains("already exists")) {
+			if (sqe instanceof BadSqlGrammarException) {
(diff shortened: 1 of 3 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/db/migration/AbstractCallback.java
```diff
@@ -67,7 +68,18 @@ public boolean canHandleInTransaction(Event event, Context context) {
 
 	@Override
 	public void handle(Event event, Context context) {
-		runner.execute(context.getConnection(), getCommands(event, context));
+		try {
+			runner.execute(context.getConnection(), getCommands(event, context));
+		}
+		catch(Exception sqe) {
+			if (sqe instanceof BadSqlGrammarException && sqe.getMessage().contains("already exists")) {
```
jvalkeal: I'm not sure if `already exists` is returned with every database as only reliable way is to check sql error code.

jvalkeal:
Merged per bf830a23053fa19c63a574904d0d4f23eff7a72f

Issue #1885, opened 2018-01-04
Add support for wildcard searching/listing of job executions by name


annasarte:
Hello. I am wondering if wildcard searching/listing is supported when listing job executions by name.  This is the API in question: https://docs.spring.io/spring-cloud-dataflow/docs/current-SNAPSHOT/reference/htmlsingle/#api-guide-resources-job-executions-list-by-name.

It seems this cannot have wildcards in the name. Is this true? I have not seen answers to my question on stack overflow as well.

My scenario:  I have many job executions starting with the name "matching."  For example, one job name is matching-simu-12345 and hence this API works just fine: GET on http://localhost:9393/jobs/executions?name=matching-simu-12345

But I also have jobs that all have the same prefix, as in: matching-simu-**_timestamp_** where **_timestamp_** is the suffix for the actual timestamp. Basically, I want to list all job executions that start with this specific prefix of "matching-simu." Is it possible?

I have tried this GET: http://localhost:9393/jobs/executions?name=matching*

and the response is:
[
    {
        "logref": "NoSuchJobException",
        "message": "No Job with that name either current or historic: [matching*]",
        "links": []
    }
]

I also tried using "%" for wildcard and received a similar response.

Thanks.

sabbyanandan:
Hi, @annasarte. Questions like this are better discussed in [StackOverflow](stackoverflow.com/tags/spring-cloud-dataflow/). Please re-post it there and someone from the team will review and respond soon.

sabbyanandan:
Maybe we missed your question on that forum; let us dig-in and get back to you.

annasarte:
Hello.

Here is a link to the question there.

https://stackoverflow.com/questions/48087540/spring-cloud-dataflow-list-job-executions-by-name-with-wildcard-supported

annasarte:
I see the answer from you in StackOverflow is that it is currently not supported, and I was told to open an issue here.  Since this issue I have here is closed, can't you just reopen this instead of me opening a new one ? Thanks.

annasarte:
Something similar to what Elasticsearch provides for wildcard searches might be comparable perhaps?  For ex:  in Elasticsearch, if I wanted to list all the documents using a wildcard for the index name (anything that has **_tenant_** in the index name for example), I would do a GET on http://localhost:9200/*tenant*?pretty

sabbyanandan:
[...]

PR #3463, opened 2019-08-24
Add support for wildcard searching/listing of job executions by name

sylank:
Add wildcard filtering to API when a user query the job executions by name.

Resolves: #1885 

Todo:
- [x] Update documentation
- [x] Create controller tests

Usage:
Support the SQL server specific wildcards. It could be different from each SQL server type like MySQL, PostgreSQL, ...
https://www.w3schools.com/sql/sql_wildcards.asp
https://www.postgresql.org/docs/9.0/functions-matching.html

You have to URL Encode the query filtering parameters like this:
`http://localhost:9393/jobs/executions?name=alm%` which will `http://localhost:9393/jobs/executions?name=alm%25`

### DIFF
--- a/spring-cloud-dataflow-shell-core/src/test/java/org/springframework/cloud/dataflow/shell/command/TaskCommandTests.java
+++ b/spring-cloud-dataflow-shell-core/src/test/java/org/springframework/cloud/dataflow/shell/command/TaskCommandTests.java
@@ -25,6 +25,7 @@ import javax.sql.DataSource;
 import org.junit.AfterClass;
 import org.junit.Before;
 import org.junit.BeforeClass;
+import org.junit.Ignore;
 import org.junit.Test;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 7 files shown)


spring-cloud-dataflow-shell-core/src/test/java/org/springframework/cloud/dataflow/shell/command/TaskCommandTests.java
```diff
@@ -129,6 +130,7 @@ public void testTaskLaunchCTRUsingAltCtrName() {
 	}
 
 	@Test
+	@Ignore
 	public void testGetLog() throws Exception{
```
sylank: Investigate why failing this test case (add more waiting time or check why is it needed).

ilayaperumalg:
@sylank Thanks for your contribution!  This PR looks good to me. 

Related to the ignored test, we have our CI plan runs the same test fine and same with my local setup. I will remove the ignore and explicit sleep. It may be that the execution of that task inside the test is stuck with some environment-specific issues in your setup. But, let's keep #3464 open to see if someone else has a similar experience.

Merging this PR now.

ilayaperumalg:
Rebased and merged as 7ff63bc7afa6689fe65e3631252d55ea95a42348

PR #3465, opened 2019-08-26
Persist original Stream Definition

### LINKED ISSUE #3429: Filter server influenced deployment properties from the stream/task definitions
As a user, after deploying the following stream, I see several deployment properties listed as part of the definition. Though they are meant to be used by Skipper/Monitoring features behind the scenes, I have no reason to see a change to my stream-definition from what I had originally submitted.

Before:
```
minutes = :timer.time > transform --expression='payload.substring(2,4)' | log
```

After:
```
:timer.time > transform --management.endpoints.web.exposure.include=prometheus,info,health --spring.cloud.streamapp.security.enabled=false --management.metrics.export.prometheus.enabled=true --transformer.expression=payload.substring(2,4) | log --management.endpoints.web.exposure.include=prometheus,info,health --spring.cloud.streamapp.security.enabled=false --management.metrics.export.prometheus.enabled=true:timer.time > transform --management.endpoints.web.exposure.include=prometheus,info,health --spring.cloud.streamapp.security.enabled=false --management.metrics.export.prometheus.enabled=true --transformer.expression=payload.substring(2,4) | log --management.endpoints.web.exposure.include=prometheus,info,health --spring.cloud.streamapp.security.enabled=false --management.metrics.export.prometheus.enabled=true
```

ilayaperumalg:
- Currently,the stream DSL is updated by Skipper after successful deploy/update and this
causes the original DSL text updated with the DSL that has all the deployment properties etc.,
To avoid this, added a new column `ORIGINAL_DEFINITION`  in `STREAM_DEFINITIONS` table
 - This new column value is retrieved for the stream definition list
 - Added database related changes including migration code
 - Updated documentation and tests
 - The stream deployment properties can still be retrieved using `stream manifest`

Resolves #3429

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
@@ -207,7 +207,7 @@ public class StreamDefinitionController {
 		@Override
 		public StreamDefinitionResource instantiateModel(StreamDefinition stream) {
 			final StreamDefinitionResource resource = new StreamDefinitionResource(stream.getName(),
-					new ArgumentSanitizer().sanitizeStream(stream), stream.getDescription());
+					new ArgumentSanitizer().sanitizeStream(stream), stream.getOriginalDslText(), stream.getDescription());
 			DeploymentState deploymentState = streamDeploymentStates.get(stream);
 			if (deploymentState != null) {
 				final DeploymentStateResource deploymentStateResource = ControllerUtils
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
@@ -207,7 +207,7 @@ public class StreamDefinitionController {
 		@Override
 		public StreamDefinitionResource instantiateModel(StreamDefinition stream) {
 			final StreamDefinitionResource resource = new StreamDefinitionResource(stream.getName(),
-					new ArgumentSanitizer().sanitizeStream(stream), stream.getOriginalDslText(), stream.getDescription());
+					new ArgumentSanitizer().sanitizeStream(stream), new ArgumentSanitizer().sanitizeOriginalStreamDsl(stream), stream.getDescription());
 			DeploymentState deploymentState = streamDeploymentStates.get(stream);
 			if (deploymentState != null) {
 				final DeploymentStateResource deploymentStateResource = ControllerUtils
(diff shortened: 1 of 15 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/db/migration/db2/V2_Add_Descriptions_And_OriginalDefinition.java
```diff
@@ -25,23 +25,26 @@
...
 			"alter table task_definitions add description varchar(255)";
 
 	private final SqlCommandsRunner runner = new SqlCommandsRunner();
 
 	@Override
 	public void migrate(Context context) throws Exception {
 		runner.execute(context.getConnection(), Arrays.asList(
-				SqlCommand.from(ALTER_STREAM_DEFINITION_TABLE),
+				SqlCommand.from(ALTER_STREAM_DEFINITION_TABLE_DESC),
+				SqlCommand.from(ALTER_STREAM_DEFINITION_TABLE_ORIG_DEF),
```
jvalkeal: When we're in a migration for an existing db, should we copy current definition to original? I know it's not original anymore but that's what we have. Otherwise that field is empty.

ilayaperumalg: Yeah, good point. I think we should.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDefinitionController.java
```diff
@@ -207,7 +207,7 @@ public StreamDefinitionResource toModel(StreamDefinition stream) {
 		@Override
 		public StreamDefinitionResource instantiateModel(StreamDefinition stream) {
 			final StreamDefinitionResource resource = new StreamDefinitionResource(stream.getName(),
-					new ArgumentSanitizer().sanitizeStream(stream), stream.getDescription());
+					new ArgumentSanitizer().sanitizeStream(stream), stream.getOriginalDslText(), stream.getDescription());
```
jvalkeal: should we sanitize original as well?

jvalkeal:
I tried with postgres and it seem to work ok even if clob points to same reference after migration. Those then get changed after deploy. Think I'm going to merge as is and then see what a real database tests look like.

```
dataflow=# select * from stream_definitions;
 definition_name | definition | description | original_definition 
-----------------+------------+-------------+---------------------
 ticktock        | 16913      |             | 16913
(1 row)

dataflow=# select * from stream_definitions;
 definition_name | definition | description | original_definition 
-----------------+------------+-------------+---------------------
 ticktock        | 16927      |             | 16928
(1 row)

```

jvalkeal:
Merged per 1e03c3d777054dc02ee5a5b364e5c4ee827ba3c8

PR #3487, opened 2019-09-09
Preserve stream description upon update

### LINKED ISSUE #3486: Stream Definition's description is not preserved after deploy
When the stream is deployed, it's description is not persisted into Stream Definitions table. This is because, the stream definition which gets updated from the Skipper's release manifest (after successful deployment) doesn't include the original description.

ilayaperumalg:
- When the stream definition is updated after the successful update, the stream description needs to be updated as well.
 - Add test to check

Resolves #3486


jvalkeal:
lgtm

Issue #3505, opened 2019-09-17
SCDF Sends OAuth Token for each CTR Launch, this needs to be configurable


cppwfs:
**Problem description:**
Currently if Oauth is enabled for SCDF it sends an OAuth token for each CTR launch.
This needs to be configurable, because the user may want to set their own values for each launch.
This will need to be fixed prior to releasing to Tile.

PR #3511, opened 2019-09-20
gh-3505 Make sending of OAuth Token for each CTR Launch more configurable

ghillert:
resolves gh-3505

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
@@ -196,23 +196,28 @@ public class DefaultTaskExecutionService implements TaskExecutionService {
 
 		if (taskExecutionInformation.isComposed()) {
 			boolean containsAccessToken = false;
+			boolean useUserAccessToken = false;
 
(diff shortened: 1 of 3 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
```diff
@@ -196,23 +196,28 @@ public long executeTask(String taskName, Map<String, String> taskDeploymentPrope
 
 		if (taskExecutionInformation.isComposed()) {
 			boolean containsAccessToken = false;
+			boolean useUserAccessToken = false;
 
-			final String dataflowAccessTokenKey = "dataflow-server-access-token";
+			final String dataflowServerAccessTokenKey = "dataflow-server-access-token";
+			final String dataflowServerUseUserAccessToken = "dataflow-server-use-user-access-token";
```
jvalkeal: This new key should somewhat get documented somewhere in `spring-cloud-dataflow-docs/src/main/asciidoc/tasks.adoc`.

ghillert: Added docs.

ghillert:
Added docs + rebased.

jvalkeal:
lgtm, may need to wait for master to become green before merging.

PR #3516, opened 2019-09-24
Retrieve log for latest task execution on CF

### LINKED ISSUE #3515: Show logs only for the latest execution for a given Task
As a user, while testing multiple task launches, I notice that the logs for **all** the executions in the history always show the logs from the latest execution, and that is incorrect.

The same task can be launched multiple times either manually, via streams, or through schedules. Each execution is unique; of course, the logs would be different as well. However, we are fetching the "recent" logs always (see: [example](https://github.com/spring-cloud/spring-cloud-deployer-cloudfoundry/blob/master/src/main/java/org/springframework/cloud/deployer/spi/cloudfoundry/CloudFoundryTaskLauncher.java#L137)), and that won't match the executions that had occurred in the past.

A quick fix would be to show the logs only for the current/latest execution for a given task. Do not show it for older executions.

ilayaperumalg:
- Since Cloud Foundry manages task executions by the given task name, the
log retrieval needs to be limited only to the latest acceptable task execution only.
 - Make sure to check the task execution is the latest when retrieving the logs in case of CF
 - Update tests

Resolves #3515

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
@@ -257,6 +257,12 @@ public class DefaultTaskExecutionService implements TaskExecutionService {
 	}
 
 	@Override
+	/**
+	 * Return the log content of the task execution identified by the given task deployment ID (external execution ID).
+	 * In case of concurrent task executions on Cloud Foundry, the logs of all the concurrent executions are displayed.
+	 * Also, on Cloud Foundry, the task execution log is retrieved only for the latest execution that matches the
+	 * given deployment ID (external execution ID).
+	 */
 	public String getLog(String platformName, String taskId) {
 		Launcher launcher = this.launcherRepository.findByName(platformName);
 		// In case of Cloud Foundry, fetching logs by external execution Id isn't valid as the execution instance is destroyed.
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
```diff
@@ -267,7 +273,13 @@ public String getLog(String platformName, String taskId) {
 				if (taskDeployment == null) {
 					throw new IllegalArgumentException();
 				}
-				taskId = taskDeployment.getTaskDefinitionName();
+				String taskName = taskDeployment.getTaskDefinitionName();
+				TaskExecution taskExecution = this.taskExplorer.getLatestTaskExecutionForTaskName(taskName);
+				if (taskExecution != null && !taskExecution.getExternalExecutionId().equals(taskId)) {
+					return "";
```
jvalkeal: Should we have same type of logic of throwing an error instead of returning empty log?

ilayaperumalg: This is mostly driven by the Data Flow UI. If the returned log value is empty, then we don't display anything, if there is anything else (including the exception message), that will get displayed at the logs page.

jvalkeal: I was just thinking to show a reason why we don't show anything. It's a difference whether log is empty or if log is missing.

ilayaperumalg: We wanted to display the log only for the latest task execution in case of Cloud Foundry and having this check on the platform type and the latest task execution on the data flow UI side turns out to be cumbersome. Hence, this PR returns empty log for anything other than latest task execution on CF and thereby handling at the backend. Happy to choose any other approach that can help achieve this better.

PR #3518, opened 2019-09-24
Added Task Scheduler Launcher to SCDF

### LINKED ISSUE #3494: Add TaskLauncher-Proxy application
As a continuation from spring-cloud/spring-cloud-dataflow#3406, we would require a TaskLauncher-Proxy application, an ephemeral Task application that can be the proxy to invoke the desired Tasks when triggered by the Scheduler implementations.

**Acceptance:**

- A new `spring-cloud-dataflow-tasklauncher` module would host the new application
- This module would be included in the parent build lifecycle
- A new uber-jar would be generated as part of the project build
- Include security components
- Tests included
- Docs/README included

Depends on https://github.com/spring-cloud/spring-cloud-dataflow/issues/3407.

cppwfs:
Provides a boot application that will be launched by the platform's scheduling framework that will in turn launch the appropriate task using Spring Cloud Data Flows Restful API

Resolves #3494


spring-cloud-dataflow-scheduler-launcher/pom.xml
```diff
@@ -0,0 +1,107 @@
...
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+	<parent>
+		<groupId>org.springframework.cloud</groupId>
+		<artifactId>spring-cloud-dataflow-parent</artifactId>
+		<version>2.3.0.BUILD-SNAPSHOT</version>
+	</parent>
+	<artifactId>spring-cloud-dataflow-scheduler-task-launcher</artifactId>
```
ilayaperumalg: The artifact id is different from the actual module name. Is that intentional?

spring-cloud-dataflow-scheduler-launcher/src/main/docker/assembly.xml
```diff
@@ -0,0 +1,38 @@
...
+  ~  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~  See the License for the specific language governing permissions and
+  ~  limitations under the License.
+  -->
+
+<assembly
+		xmlns="http://maven.apache.org/plugins/maven-assembly-plugin/assembly/1.1.2"
+		xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+		xsi:schemaLocation="http://maven.apache.org/plugins/maven-assembly-plugin/assembly/1.1.2 https://maven.apache.org/xsd/assembly-1.1.2.xsd">
+	<id>pring-cloud-dataflow-scheduler-task-launcher</id>
```
ilayaperumalg: Typo: spring

ilayaperumalg: If we happen to change the artifact id to align with the module name, we need to change the id as well as other references in this file.

spring-cloud-dataflow-scheduler-launcher/src/main/docker/assembly.xml
```diff
@@ -0,0 +1,38 @@
...
+			</includes>
+			<outputDirectory>maven</outputDirectory>
+			<outputFileNameMapping>scheduler-task-launcher.jar</outputFileNameMapping>
+		</dependencySet>
+	</dependencySets>
+	<files>
+		<file>
+			<source>./target/spring-cloud-dataflow-scheduler-task-launcher-${project.version}.jar</source>
+			<outputDirectory>.</outputDirectory>
+			<destName>shell.jar</destName>
```
ilayaperumalg: Not sure why is this `shell.jar`.

spring-cloud-dataflow-scheduler-launcher/src/main/java/org/springframework/cloud/dataflow/scheduler/launcher/SchedulerTaskLauncherApplication.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+package org.springframework.cloud.dataflow.scheduler.launcher;
+
+import org.springframework.boot.Banner;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.builder.SpringApplicationBuilder;
+
+/**
+ * @author Glenn Renfro
```
ilayaperumalg: Please add some description of the functionality of this main application in the Javadoc.

spring-cloud-dataflow-scheduler-launcher/src/main/java/org/springframework/cloud/dataflow/scheduler/launcher/configuration/LauncherConfiguration.java
```diff
@@ -0,0 +1,74 @@
...
+				schedulerTaskLauncherProperties, environment);
+	}
+
+	@Bean
+	public TaskOperations getTaskOperations(SchedulerTaskLauncherProperties schedulerTaskLauncherProperties) {
+		try {
+			final URI dataFlowUri = new URI(schedulerTaskLauncherProperties.getDataflowServerUri());
+			final DataFlowOperations dataFlowOperations = new DataFlowTemplate(dataFlowUri);
+			if (dataFlowOperations.taskOperations() == null) {
+				throw new SchedulerTaskLauncherException("The SCDF server does not support task operations");
```
ilayaperumalg: Instead of saying `The SCDF server does not support task operations`, we can say something like `The task operations are not enabled in the Spring Cloud Data Flow server`.

spring-cloud-dataflow-scheduler-launcher/src/main/java/org/springframework/cloud/dataflow/scheduler/launcher/configuration/SchedulerTaskLauncher.java
```diff
@@ -0,0 +1,138 @@
...
+import org.springframework.util.StringUtils;
+
+/**
+ * Sends a request to the Spring Cloud Data Flow server to launch a task.
+ *
+ * @author Glenn Renfro
+ **/
+public class SchedulerTaskLauncher {
+
+	static final String TASK_PLATFORM_NAME = "spring.cloud.dataflow.task.platformName";
```
ilayaperumalg: We already have this property in a few other places and hence, can we move this constant to a common accessible helper class?

spring-cloud-dataflow-scheduler-launcher/src/main/java/org/springframework/cloud/dataflow/scheduler/launcher/configuration/LauncherConfiguration.java
```diff
@@ -0,0 +1,74 @@
...
+@Configuration
+@EnableConfigurationProperties({SchedulerTaskLauncherProperties.class})
+@EnableAutoConfiguration(exclude = {DataFlowClientAutoConfiguration.class})
+public class LauncherConfiguration {
+
+	@Bean
+	public SchedulerTaskLauncher launchRequestConsumer(
+			SchedulerTaskLauncherProperties schedulerTaskLauncherProperties,
+			TaskOperations taskOperations, Environment environment) {
+		return new SchedulerTaskLauncher(schedulerTaskLauncherProperties.getTaskName(),
```
ilayaperumalg: Given we already pass the `SchedulerTaskLauncherProperties` as one of the arguments, can we simplify this constructor to just use `new SchedulerTaskLauncher(SchedulerTaskLauncherProperties schedulerTaskLauncherProperties, TaskOperations taskOperations, Environment environment)` ?

spring-cloud-dataflow-scheduler-launcher/src/main/java/org/springframework/cloud/dataflow/scheduler/launcher/configuration/SchedulerTaskLauncher.java
```diff
@@ -0,0 +1,138 @@
...
+
+	private Environment environment;
+
+	public SchedulerTaskLauncher(String taskName, String platformName,
+			TaskOperations taskOperations, SchedulerTaskLauncherProperties schedulerTaskLauncherProperties,
+			Environment environment) {
+		Assert.hasText(taskName, "`taskName` cannot be null");
+		Assert.notNull(taskOperations, "`taskOperations` must not be null");
+		Assert.hasText(platformName, "`platformName` cannot be null");
+		Assert.notNull(schedulerTaskLauncherProperties, "`taskLauncherProperties` must not be null");
```
ilayaperumalg: taskLauncherProperties -> schedulerTaskLauncherProperties

spring-cloud-dataflow-scheduler-launcher/src/main/java/org/springframework/cloud/dataflow/scheduler/launcher/configuration/SchedulerTaskLauncher.java
```diff
@@ -0,0 +1,138 @@
...
+		if (!deploymentProperties.containsKey(TASK_PLATFORM_NAME)) {
+			enrichedProperties.put(TASK_PLATFORM_NAME, this.platformName);
+		}
+		return enrichedProperties;
+	}
+
+
+	private Map<String, String> getDeploymentProperties() {
+		Map<String, String> props = new HashMap<>();
+		final String taskLauncherPrefix = this.schedulerTaskLauncherProperties.getTaskLauncherPropertyPrefix() + ".";
```
ilayaperumalg: I am not sure how this is supposed to work when we have the environment using the task deployment properties with `_` instead of `.`.

spring-cloud-dataflow-scheduler-launcher/src/test/java/org/springframework/cloud/dataflow/scheduler/launcher/TestConfig.java
```diff
@@ -0,0 +1,91 @@
...
+			@Override
+			public List<ScheduleInfo> list() {
+				return Collections.emptyList();
+			}
+		};
+	}
+
+	@Bean
+	@Primary
+	public SkipperClient skipperClientMock() {
```
ilayaperumalg: Do we need `SkipperClient` bean in here?

spring-cloud-dataflow-scheduler-launcher/src/main/java/org/springframework/cloud/dataflow/scheduler/launcher/configuration/SchedulerTaskLauncher.java
cppwfs: In this case when multiple properties are sent to SchedulerTaskLauncher app it needs to know what properties need to be sent to SCDF for the task it is to launch. This prefix is used to mark those properties that are required for the app to be launched.   
For example if a user wants to schedule task definition `tstamp` (a timestamp app) with a property  of `app.timestamp.timestamp.format=YYYY` along with the other properties that are added by SCDF to be used by the SchedulerTaskLauncher.  We needed a method to mark this property as to be used for the `tstamp` definition launch.   So we append `tasklauncher.` to the property when it is scheduled with the platform's scheduler. i.e. it will look like `tasklauncher.app.timestamp.timestamp.format=YYYY`.   Thus when SchedulerTaskLauncher has been launched by the platform scheduler and it is associating the properties to be used for the SCDF API task launch call it will find these properties and marked with the `tasklauncher.` and associate them with the task launch.

cppwfs:
@ilayaperumalg 
Thanks for the review!

1) Since it is possible to use this new module as a REST client application to launch the task, what are our plans to make this as a generic task launcher client along with how CI/CD SPI uses?
This is  one part of how SCDF will handle the launching of scheduled apps.  SCDF will deploy this application to the platform and schedule its launch with the platform's scheduler.    There is a branch of code that will be submitted as soon as https://github.com/spring-cloud/spring-cloud-dataflow/pull/3520 that will support these  features. 

To read more about the ScheduledTaskLauncher you can find the design document here:
https://docs.google.com/document/d/1tu4a5TyagZBSxWoQhlVwcnht_J8PIdeXXsqVMk6pOE4/edit#heading=h.v8lgr3kx0x2

Are we planning to use this a generic place so that CTR can be built on top of this?
REALLY GOOD QUESTION!!!   Had to think about this question.   At this time no, while they do share the ability to launch tasks via SCDF, that is about it as far as I can tell.

spring-cloud-dataflow-scheduler-launcher/src/main/java/org/springframework/cloud/dataflow/scheduler/launcher/configuration/SchedulerTaskLauncher.java
cppwfs: Hmm...   🤔 
I like the idea but this is used in 3 different modules 
a) Task Command 
b) SchedulerTask Launcher
c) And DefaultTaskExecutionService.

While we could technically add it to a helper class  in the  spring-cloud-dataflow-server-core and that would work for SchedulerTaskLauncher and DefaultTaskExecutionService.  However I don't feel comfortable adding that dependency to the module containing TaskCommand.  
Do you know of a good place to put the helper class?

spring-cloud-dataflow-scheduler-launcher/pom.xml
cppwfs: Oops :-)  fixed

spring-cloud-dataflow-scheduler-launcher/src/main/docker/assembly.xml
cppwfs: Good catch!  Fixed

spring-cloud-dataflow-scheduler-task-launcher/src/main/java/org/springframework/cloud/dataflow/scheduler/launcher/SchedulerTaskLauncherApplication.java
```diff
@@ -0,0 +1,37 @@
...
+ * Spring Cloud Task apps is to utilize the Restful API provided by Spring Cloud
+ * Data Flow to launch the applications.
+ *
+ * @author Glenn Renfro
+ */
+@SpringBootApplication
+public class SchedulerTaskLauncherApplication {
+
+	public static void main(String[] args) {
+		new SpringApplicationBuilder().sources(SchedulerTaskLauncherApplication.class).bannerMode(Banner.Mode.OFF).run(args);
```
jvalkeal: I think with these new apps we could put `spring.main.banner-mode` in `application.yml` as we may want to put other stuff there like logging settings, etc.

spring-cloud-dataflow-scheduler-task-launcher/src/main/java/org/springframework/cloud/dataflow/scheduler/launcher/configuration/LauncherConfiguration.java
```diff
@@ -0,0 +1,73 @@
...
+			TaskOperations taskOperations, Environment environment) {
+		return new SchedulerTaskLauncher(taskOperations, schedulerTaskLauncherProperties,
+				environment);
+	}
+
+	@Bean
+	public TaskOperations getTaskOperations(SchedulerTaskLauncherProperties schedulerTaskLauncherProperties) {
+		try {
+			final URI dataFlowUri = new URI(schedulerTaskLauncherProperties.getDataflowServerUri());
+			final DataFlowOperations dataFlowOperations = new DataFlowTemplate(dataFlowUri);
```
jvalkeal: Now, be carefull with this as `DataFlowTemplate` sucks in a way that it fails fast if dataflow server is not running or if root request fails. This is thanks to all hateoas nonsense we have.

spring-cloud-dataflow-scheduler-task-launcher/src/main/java/org/springframework/cloud/dataflow/scheduler/launcher/configuration/SchedulerTaskLauncherProperties.java
```diff
@@ -0,0 +1,71 @@
...
+package org.springframework.cloud.dataflow.scheduler.launcher.configuration;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * Properties used to configure the task launcher.
+ *
+ * @author Glenn Renfro
+ **/
+@ConfigurationProperties
```
jvalkeal: I think we should have a prefix for these properties.

spring-cloud-dataflow-scheduler-task-launcher/src/main/java/org/springframework/cloud/dataflow/scheduler/launcher/configuration/LauncherConfiguration.java
```diff
@@ -0,0 +1,73 @@
...
+import org.springframework.core.env.Environment;
+
+/**
+ * Configure the SchedulerTaskLauncher application.
+ *
+ * @author Glenn Renfro
+ */
+@Configuration
+@EnableConfigurationProperties({SchedulerTaskLauncherProperties.class})
+@EnableAutoConfiguration(exclude = {DataFlowClientAutoConfiguration.class})
```
jvalkeal: Add this exclude to `application.yml`. Main `@SpringBootApplication` is meta annotation for this and it's a bad practise to have `@EnableAutoConfiguration` used more than once.

spring-cloud-dataflow-scheduler-task-launcher/src/main/java/org/springframework/cloud/dataflow/scheduler/launcher/configuration/LauncherConfiguration.java
jvalkeal: Ah yeah, so app is not kept running as commandLineRunner is only thing it does. But then I need to ask why dont' just use `DataFlowClientAutoConfiguration` to get it instead of rebuilding it manually?

spring-cloud-dataflow-scheduler-launcher/src/main/docker/assembly.xml
jvalkeal: I think skeleton was copied from shell module :)

jvalkeal:
Merged per 11cbb16de589ceec2fbef418345603628f646a27

PR #3528, opened 2019-09-30
Add Scheduling to CI/CD Task Effort

cppwfs:
Note:
1) I will open a issus efor the UI and for core:
Currently if a user try to delete all schedules associated with a task definition it will fail to work.   This is because we no longer use just the app name.  It is now scdf-appname-schedulename
However if you delete the task definition it will delete all associated schedules (as it should).

2) You will see in the the application.yml and in TaskConfigurationProperties the maven URI for the scheduler-task-launcher is pointing to a snapshot.  Is there a way we can automate this value vs. having to do it via the release process?

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/TaskConfigurationProperties.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/TaskConfigurationProperties.java
@@ -50,7 +50,7 @@ public class TaskConfigurationProperties {
 	/**
 	 * The URI of the resource to be used for launching apps via a schedule.
 	 */
-	private String schedulerTaskLauncherUrl = "maven://org.springframework.cloud:spring-cloud-dataflow-scheduler-task-launcher:2.3.0.BUILD-SNAPSHOT";
+	private String schedulerTaskLauncherUrl;
 	/**
 	 * The prefix to be applied to schedule names.
 	 */
@@ -60,6 +60,11 @@ public class TaskConfigurationProperties {
 	 * The prefix to attach to the application properties to be sent to the schedule task launcher.
 	 */
 	private String taskLauncherPrefix = "tasklauncher.";
+
+	/**
+	 * The Spring Cloud Data Flow version.
+	 */
+	private String version;
 
 	public String getComposedTaskRunnerName() {
 		return composedTaskRunnerName;
@@ -99,5 +104,13 @@ public class TaskConfigurationProperties {
 
 	public void setSchedulerTaskLauncherUrl(String schedulerTaskLauncherUrl) {
 		this.schedulerTaskLauncherUrl = schedulerTaskLauncherUrl;
+	}
+
+	public String getVersion() {
+		return version;
+	}
+
+	public void setVersion(String version) {
+		this.version = version;
 	}
 }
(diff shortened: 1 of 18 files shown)


src/kubernetes/server/server-deployment.yaml
```diff
@@ -40,6 +40,8 @@ spec:
           value: 'true'
         - name: SPRING_CLOUD_DATAFLOW_FEATURES_SCHEDULES_ENABLED
           value: 'true'
+        - name: SPRING_CLOUD_DATAFLOW_TASK_SCHEDULER_TASK_LAUNCHER_URL
+          value: 'docker://springcloud/spring-cloud-dataflow-scheduler-task-launcher:latest'
```
jvalkeal: Should be `2.3.0.BUILD-SNAPSHOT` for now as we currently need to manually replace all other image versions during a release.

jvalkeal: Ah yeah, we need to update CI to push out this image.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/TaskConfigurationProperties.java
```diff
@@ -41,9 +41,26 @@
...
+	/**
+	 * The schedule application name to be used for the scheduler task launcher.
+	 */
 	@NotBlank
 	private String schedulerTaskLauncherName = "scheduler-task-launcher";
 
+	/**
+	 * The URI of the resource to be used for launching apps via a schedule.
+	 */
+	private String schedulerTaskLauncherUrl = "maven://org.springframework.cloud:spring-cloud-dataflow-scheduler-task-launcher:2.3.0.BUILD-SNAPSHOT";
```
jvalkeal: We could leave this empty and set value in `dataflow-server-defaults.yml` where we can replace value via maven build with `@project.version@`. Though this has an annoying issue that when running via IDE that value would be wrong.

src/kubernetes/server/server-deployment.yaml
cppwfs: Would this be `@project.version@` as well?

spring-cloud-dataflow-server/src/main/resources/application.yml
```diff
@@ -37,6 +37,9 @@ info:
 #     config:
 #        discovery:
 #           enabled: true
+#     dataflow:
```
jvalkeal: This file is totally defunc even as an example. No need to update it. We should probably zap all commented lines.

spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
```diff
@@ -66,6 +66,9 @@ spring:
             url: "{repository}/org/springframework/cloud/spring-cloud-dataflow-shell/{version}/spring-cloud-dataflow-shell-{version}.jar"
             checksum-sha1-url: "{repository}/org/springframework/cloud/spring-cloud-dataflow-shell/{version}/spring-cloud-dataflow-shell-{version}.jar.sha1"
             checksum-sha256-url: "{repository}/org/springframework/cloud/spring-cloud-dataflow-shell/{version}/spring-cloud-dataflow-shell-{version}.jar.sha256"
+      task:
+        version: "@project.version@"
```
jvalkeal: I didn't mean to add a new property, instead just put @project.version@ in a url where you now have {version}.

jvalkeal: If you're wondering above `{version}` use in `version-info`, that's just a hack(in AboutController) to get predictable addresses for tile.

jvalkeal:
Merged per 43c616fb3f04acca1cbd10ec87fd413c2b8cf613

PR #3530, opened 2019-10-01
Fixes issue where a user can not search for schedules using task definition name.

cppwfs:
Also Allows users to delete schedules using task definition name


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskSchedulerController.java
```diff
@@ -120,6 +120,18 @@ public ScheduleInfoResource getSchedule(@PathVariable("name") String scheduleNam
 				result.size());
 		return assembler.toModel(page, taskAssembler);
 	}
+	/**
+	 * Return a page-able list of {@link ScheduleInfo}s for a specific
```
chrisjs: the method is returning `void` and its unscheduling vs searching

chrisjs:
merged in bf56523037f1b2535e142676b9384bd7c3857f22

PR #3542, opened 2019-10-04
Restart Task from Shell

sylank:
resolves: #548


cppwfs:
@sylank Thanks for the PR!   Can you add a test for this?

sylank:
Sure, I will write some 😃

sabbyanandan:
Hello, @sylank! We are aiming for 2.3 RC1 on 11/6. It'd be nice to merge this enhancement for the RC release, so if you could add the tests, that'd be greatly appreciated.

sylank:
Sorry about the late reply. I checked the possibilities but there are way too many efforts to mock out tasks and job execution classes to test the restart function.
I couldn't find successful restart tests on the controller side, there are only bad case test scenarios. Could you scope/specify what kind of test you meant?
I checked the `JobExecutionControllerTests` there are no TaskDefinitionRepository defined/mocked.
I tried to test the bad case scenarios in `JobCommandTests` but the `DataFlowShell#executeCommand` was blocked the result check in `AbstractShellIntegrationTest` class.

cppwfs:
LGTM.  I'll add some tests next week.  I want to update those tests a bit.

cppwfs:
@sylank Thank you for another great contribution!

Rebased, Squashed, Merged.

PR #3548, opened 2019-10-14
Update docker Kafka image to support partitions

### LINKED ISSUE #3547: Kafka version used for Kubernetes setup doesn't support CREATE_PARTITIONS
Current Kafka version (`wurstmeister/kafka:2.11-0.11.0.3`) doesn't support partition creation. For example deploy a pipeline with minPartitionCount of 2: 
```
stream create --name old-kafka --definition "time | log"
stream deploy --name old-kafka --properties "app.log.spring.cloud.stream.kafka.binder.autoAddPartitions=true,app.log.spring.cloud.stream.kafka.binder.minPartitionCount=2"
```
Above will trigger an exception: `org.apache.kafka.common.errors.UnsupportedVersionException: The broker does not support CREATE_PARTITIONS`

In addition the `wurstmeister/kafka` is not a confluent supported image.  Instead the `confluentinc/cp-kafka` images should be used.

tzolov:
- update kafka image used for k8s installation from wurstmeister/kafka:2.11-0.11.0.3 to confluentinc/cp-kafka:5.3.1
 - update kafka zookeeper image used for k8s installation from digitalwonderland/zookeeper to confluentinc/cp-zookeeper:5.3.1
 - adjust related Kafka configuration parameters (remove deprecated host and port)
 - rename the k8s kafka service to kafka-broker due to the https://github.com/confluentinc/cp-docker-images/issues/388
 - update the skipper config map environmentVariables to reflect the new kafka broker name
 - align the docker-compose kafka/zookeer image versions to confluentinc/cp-kafka:5.3.1 and confluentinc/cp-zookeeper:5.3.1

 Resolves #3547

### DIFF
--- a/spring-cloud-dataflow-server/docker-compose.yml
+++ b/spring-cloud-dataflow-server/docker-compose.yml
@@ -11,21 +11,21 @@ services:
     expose:
       - 3306
 
-  kafka:
-    image: confluentinc/cp-kafka:5.2.1
+  kafka-broker:
+    image: confluentinc/cp-kafka:5.3.1
     container_name: dataflow-kafka
     expose:
       - "9092"
     environment:
-      - KAFKA_ADVERTISED_LISTENERS=PLAINTEXT://kafka:9092
+      - KAFKA_ADVERTISED_LISTENERS=PLAINTEXT://kafka-broker:9092
       - KAFKA_ZOOKEEPER_CONNECT=zookeeper:2181
-      - KAFKA_ADVERTISED_HOST_NAME=kafka
+      - KAFKA_ADVERTISED_HOST_NAME=kafka-broker
       - KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR=1
     depends_on:
       - zookeeper
 
   zookeeper:
-    image: confluentinc/cp-zookeeper:5.2.1
+    image: confluentinc/cp-zookeeper:5.3.1
     container_name: dataflow-kafka-zookeeper
     expose:
       - "2181"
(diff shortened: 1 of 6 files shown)


spring-cloud-dataflow-server/docker-compose.yml
```diff
@@ -11,21 +11,21 @@ services:
     expose:
       - 3306
 
-  kafka:
-    image: confluentinc/cp-kafka:5.2.1
+  kafka-broker:
```
ilayaperumalg: Is there any specific reason to change this name to `kafka-broker` ?

ilayaperumalg: I guess it is just convenient.

tzolov: No, unfortunately this is an enforced requirement due to  https://github.com/helm/charts/issues/5089 and https://github.com/helm/charts/pull/5104/files#diff-96171ed0f9aefa1650847d0c34f3ed60R175

If you try to use `kafka` as a name, the official kafka docker images you fail with an error like this: `port is deprecated. Please use KAFKA_ADVERTISED_LISTENERS instead.`

While the error message is misleading the core of the issue is that an implicit env. variable is created with name `<Your-Kafka-Service-Name>_PORT`.  E.g. for `kafka` name a `KAFKA_PORT` variable is created. But the `KAFKA_PORT` is listed as deprecated and the image script aborts.

There is an open issue to relax the failure on deprecated variables: https://github.com/confluentinc/cp-docker-images/issues/388

But until it is resolved we can not use `kafka` as a name.  Btw i've been using `dataflow-kafka` in our the docker-compose files.

src/kubernetes/kafka/kafka-deployment.yaml
```diff
@@ -3,35 +3,33 @@ kind: Deployment
...
-            value: "9092"
-          - name: KAFKA_ADVERTISED_PORT
-            value: "9092"
           - name: KAFKA_ADVERTISED_HOST_NAME
             valueFrom:
               fieldRef:
                 fieldPath: status.podIP
           - name: KAFKA_ZOOKEEPER_CONNECT
             value: kafka-zk:2181
+          - name: KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR
```
ilayaperumalg: I guess we set this as it is required for the single node cluster?

src/kubernetes/kafka/kafka-svc.yaml
```diff
@@ -1,16 +1,16 @@
...
-    name: kafka-port
+    name: kafka-broker-port
     targetPort: 9092
     protocol: TCP
   selector:
-    app: kafka
-    component: kafka-broker
\ No newline at end of file
+    app: kafka-broker
+    component: kafka-broker-c
```
ilayaperumalg: why `c`? anything special about it except the name `Christian` ;)

tzolov: because the component name was already `kafka-broker` when the app name was `kafka`. After we renamed the app to `kafka-broker` i had to change the component name.  In a lack of imaginations i picked `c` (from component) but feel free to rename it.

ilayaperumalg:
LGTM, merging.

src/kubernetes/kafka/kafka-deployment.yaml
tzolov: Removed this property.

PR #3549, opened 2019-10-15
Conditionally import SchedulerService

### LINKED ISSUE #3546: SCDF fails to start with Scheduling disabled
As a user, when deploying SCDF to CF without Scheduling (i.e., `SPRING_CLOUD_DATAFLOW_FEATURES_SCHEDULES_ENABLED=false`), I get the following error on `master` and on `2.3 M2`.

```
2019-10-07T21:36:17.35-0700 [APP/PROC/WEB/0] OUT ***************************
   2019-10-07T21:36:17.35-0700 [APP/PROC/WEB/0] OUT APPLICATION FAILED TO START
   2019-10-07T21:36:17.35-0700 [APP/PROC/WEB/0] OUT ***************************
   2019-10-07T21:36:17.35-0700 [APP/PROC/WEB/0] OUT Description:
   2019-10-07T21:36:17.35-0700 [APP/PROC/WEB/0] OUT Parameter 8 of method deleteTaskService in org.springframework.cloud.dataflow.server.config.features.TaskConfiguration required a bean of type 'org.springframework.cloud.dataflow.server.service.SchedulerService' that could not be found.
```

The same error also appeared when deploying `master` on K8s, but it could have been due to a corrupted/unstable image.

ilayaperumalg:
- The DefaultTaskDeleteService should import the scheduler service when the task and schedules are enabled.
   - Conditionally import the `SchedulerService` bean

 - Remove explicit `local` scheduler no-op which was used only for tests
 - Update documentation and scheduler service tests

Resolves #3546

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/TaskConfiguration.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/TaskConfiguration.java
@@ -15,9 +15,7 @@
  */
 package org.springframework.cloud.dataflow.server.config.features;
 
-import java.util.Collections;
 import java.util.List;
-import java.util.Map;
 
 import javax.sql.DataSource;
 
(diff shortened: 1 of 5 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/TaskConfiguration.java
```diff
@@ -130,55 +128,6 @@ public TaskPlatform localTaskPlatform(LocalPlatformProperties localPlatformPrope
...
 	}
 
-	/**
-	 * The default profile is active when no other profiles are active. This is configured so
-	 * that several tests will pass without having to explicitly enable the local profile.
-	 * @return a no-op {@link SchedulerService}
-	 */
-	@Profile({ "local", "default" })
-	@Bean
-	public SchedulerService schedulerService() {
```
jvalkeal: It wasn't just for tests. We did this dump thing for UI not to throw errors too early as essentially we wanted to fail only when schedule we tried to use. Put it back please.

ilayaperumalg: ok, Though this bean might be needed to fail fast, the availability of `SchedulerService` by default or in the `local` setup looks incorrect. Could you tell me which part of the UI requires this bean? I will try to see if we can work around it by not having this bean explicitly?

jvalkeal: Hmm hold on, I was expecting to get an error before even appempting to create a schedule. Think I was wrong with my initial comment.

jvalkeal: We added some of these dump empty impls for local not to fail until attempting to actually schedule a task. It was a convenience for developer and not to throw 505's in UI. All these seem to work as is with this pr.

cppwfs:
LGTM But we do need to create an issue on adding the doc tests back.

PR #3562, opened 2019-10-16
Update SCDF docker-compose suite

### LINKED ISSUE #3561: Update SCDF docker-compose suite
- [x] update Palantir version.
- [x] support docker-compose file composition.
- [x] add Task and compose tasks tests.
- [x] add more stream tests to get in pair with the ATs stream tests suite. 
- [x] add multi-platform support.
- [x] self-document the DockerComposeIT to make easy to create different docker-compose tests.

tzolov:
- Update Palantir from 0.32.0 to 1.3.0.
 - Allow support for multiple docker-compose files.
 - Factor out the inner utility classes in util sub-package.
 - Add Task and Composed Tasks Tests.
 - Make Stream AutoClosalbe, calling destroy() on exit.
 - Add deocker-compose-cf.yml and docker-compose-k8s.yml templates to register CF and K8s runtime platforms.
 - Make RuntimeApplicationHelper platform aware, by resolving the platformType and handle platform specific app-instance-url resolution.
 - Muve utility classes under the util sub-package.
 - Add partition and transform stream tests DockerComposeIT
 - Add detail JavaDoc docs to DockerComposeIT.
 - Add multi-platform support to run Stream tests on CF and K8s.

 Resolves #3561


spring-cloud-dataflow-server/pom.xml
```diff
@@ -41,7 +41,7 @@
 		<dependency>
 			<groupId>com.palantir.docker.compose</groupId>
 			<artifactId>docker-compose-rule-junit4</artifactId>
-			<version>0.35.0</version>
+			<version>1.3.0</version>
```
sabbyanandan: That's a big jump. 😄

spring-cloud-dataflow-server/docker-compose-k8s.yml
```diff
@@ -0,0 +1,12 @@
+version: '3'
+
+services:
+
+  skipper-server:
+    environment:
+      - SPRING_CLOUD_SKIPPER_SERVER_PLATFORM_KUBERNETES_ACCOUNTS_K8S_FABRIC8_MASTER_URL=<MASTER URL>
+      - SPRING_CLOUD_SKIPPER_SERVER_PLATFORM_KUBERNETES_ACCOUNTS_K8S_FABRIC8_USERNAME=<USER>
```
sabbyanandan: This assumes Basic-auth is enabled at the cluster level, right?

spring-cloud-dataflow-server/docker-compose-cf.yml
```diff
@@ -0,0 +1,17 @@
+version: '3'
+
+services:
+
+  skipper-server:
```
sabbyanandan: I'm not there's much value in keeping these standalone files. IMO, no one would be in a position to start SCDF and everything in locally running Docker daemon, but instead deploy streams/tasks to a cloud platform. I know we wanted to showcase the possibility, but I wonder if it would end up confusing the users.

spring-cloud-dataflow-server/docker-compose-k8s.yml
sabbyanandan: Also, by default, Basic auth is disabled in GKE at least.

spring-cloud-dataflow-server/pom.xml
tzolov: it is, but it works fine.

spring-cloud-dataflow-server/docker-compose-k8s.yml
tzolov: I have successfully tested it with K8s/GKE cluster. Did not experienced any issues.
Furthermore this is a basic setup/template to show how you can do it.

spring-cloud-dataflow-server/docker-compose-cf.yml
tzolov: I believe that it is useful especially in combination with `DockerComposeIT` like testing. It will be awkward to explain how to do this on the DockerComposeIT javadoc. 
Instead i've added it as a  bullet to the updated customized docker-compose guide (still not pushed)

tzolov: I mean after i push the changes to the microsite will update the java doc to reference multiplatform support.

tzolov: I've pushed a commit with simplified miltiplatform testing explanation, referring to https://dataflow.spring.io/docs/2.3.0.SNAPSHOT/installation/local/docker-customize/

ilayaperumalg:
LGTM, this is really a nice addition to our Integration Tests across platforms. Merging.

ilayaperumalg:
Squashed, rebased and merged as b1e69ce85

Issue #3572, opened 2019-10-24
Upgrade to Spring Security 5.2


ghillert:
Depends on https://github.com/spring-cloud/spring-cloud-common-security-config/issues/57

PR #3573, opened 2019-10-24
gh-3572 Upgrade to Spring Security 5.2

ghillert:
* Upgrade Spring Cloud Common Security dependency
* Fix tests

resolves https://github.com/spring-cloud/spring-cloud-dataflow/issues/3572


spring-cloud-dataflow-dependencies/pom.xml
```diff
@@ -110,6 +110,20 @@
 				<artifactId>spring-cloud-dataflow-platform-cloudfoundry</artifactId>
 				<version>2.3.0.BUILD-SNAPSHOT</version>
 			</dependency>
+			<dependency>
+				<groupId>org.springframework.security</groupId>
+				<artifactId>spring-security-bom</artifactId>
+				<version>5.2.0.RELEASE</version>
```
ilayaperumalg: We need to find an alternative way to fix this as this way to explicit declaration might become stale after a few upcoming releases and will be painful to manually manage.

spring-cloud-dataflow-platform-cloudfoundry/src/main/java/org/springframework/cloud/dataflow/server/config/cloudfoundry/security/support/CloudFoundryDataflowAuthoritiesExtractor.java
```diff
@@ -34,14 +35,14 @@
...
- * This Spring Cloud Data Flow {@link AuthoritiesExtractor} will assign all
+ * This Spring Cloud Data Flow {@link AuthorityMapper} will assign all
  * {@link CoreSecurityRoles} to the authenticated OAuth2 user IF the user is a "Space
  * Developer" in Cloud Foundry.
  *
  * @author Gunnar Hillert
  *
  */
-public class CloudFoundryDataflowAuthoritiesExtractor implements AuthoritiesExtractor {
+public class CloudFoundryDataflowAuthoritiesExtractor implements AuthorityMapper {
```
ilayaperumalg: Please rename this class to be a `AuthorityMapper`.

spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/config/DataFlowClientProperties.java
```diff
@@ -37,7 +37,7 @@
 	/**
 	 * Skip Ssl validation.
 	 */
-	private boolean skipSslValidation = true;
+	private boolean skipSslValidation;
```
ilayaperumalg: I thought we wanted to enable SSL validation by default.

spring-cloud-dataflow-server-core/pom.xml
```diff
@@ -170,6 +170,11 @@
 			<groupId>org.flywaydb</groupId>
 			<artifactId>flyway-core</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>io.projectreactor</groupId>
+			<artifactId>reactor-core</artifactId>
+			<version>3.3.0.RELEASE</version>
```
ilayaperumalg: We definitely don't want to be explicit about versioning as it should come from the underlying dependency management. If there is any specific reason, could you elaborate why to do that and if so, we can move the version to a variable for better management.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/TaskConfiguration.java
```diff
@@ -182,12 +183,14 @@ public TaskExecutionService taskService(LauncherRepository launcherRepository,
 			TaskAppDeploymentRequestCreator taskAppDeploymentRequestCreator,
 			TaskExplorer taskExplorer,
 			DataflowTaskExecutionDao dataflowTaskExecutionDao,
-			DataflowTaskExecutionMetadataDao dataflowTaskExecutionMetadataDao) {
+			DataflowTaskExecutionMetadataDao dataflowTaskExecutionMetadataDao,
+			OAuth2TokenUtilsService oauth2TokenUtilsService) {
```
ilayaperumalg: Instead of having this as a required dependency and bringing in `DummyOAuth2TokenUtilsService` when OAuth2 is disabled, we can have this as a non-required dependency in `DefaultTaskExecutionService` only when OAuth2 is enabled.

spring-cloud-dataflow-server/pom.xml
```diff
@@ -19,6 +19,11 @@
 	</properties>
 
 	<dependencies>
+		<dependency>
+			<groupId>io.projectreactor</groupId>
+			<artifactId>reactor-core</artifactId>
+			<version>3.3.0.RELEASE</version>
```
ilayaperumalg: Similar to the previous comment on version management.

spring-cloud-starter-dataflow-server/pom.xml
```diff
@@ -46,6 +46,12 @@
 			<artifactId>spring-cloud-dataflow-rest-client</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework.security.oauth</groupId>
```
ilayaperumalg: Is there a way to avoid using this in tests?

spring-cloud-starter-dataflow-server/src/test/resources/org/springframework/cloud/dataflow/server/single/security/oauthConfig.yml
```diff
@@ -11,24 +11,40 @@ spring:
...
-      client-id: myclient
-      client-secret: mysecret
-      access-token-uri: http://127.0.0.1:${oauth2.port}/oauth/token
-      user-authorization-uri: http://127.0.0.1:${oauth2.port}/oauth/authorize
-    resource:
-      user-info-uri: http://127.0.0.1:${oauth2.port}/me
-      token-info-uri: http://127.0.0.1:${oauth2.port}/oauth/check_token
-    authorization:
-      check-token-access: isAuthenticated()
+          provider-role-mappings:
```
ilayaperumalg: We need to add the documentation specifically on the changes in the properties configuration

spring-cloud-starter-dataflow-server/src/test/resources/org/springframework/cloud/dataflow/server/single/security/oauthConfig.yml
```diff
@@ -11,24 +11,40 @@ spring:
...
+              map-oauth-scopes: true
+              role-mappings:
+                ROLE_VIEW: dataflow.view
+                ROLE_CREATE: dataflow.create
+                ROLE_MANAGE: dataflow.manage
+                ROLE_DEPLOY: dataflow.create
+                ROLE_DESTROY: dataflow.create
+                ROLE_MODIFY: dataflow.create
+                ROLE_SCHEDULE: dataflow.create
+  security:
```
ilayaperumalg: This is another important property hierarchical change as well. We need to be explicit about this in the documentation.

spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/config/DataFlowClientProperties.java
ghillert: We should never "skip ssl validation" by default. I am not sure how that got set to `true` in the first place.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/TaskConfiguration.java
ghillert: Made the change

spring-cloud-starter-dataflow-server/pom.xml
jvalkeal: And version dep needs to go to main pom.

pom.xml
```diff
@@ -66,23 +66,21 @@
 
 		<!-- note - check version of reactor at deployer-cf/cf-java-client uses -->
 		<spring-cloud-deployer-cloudfoundry.version>2.1.0.BUILD-SNAPSHOT</spring-cloud-deployer-cloudfoundry.version>
-		<reactor.version>3.2.0.RELEASE</reactor.version>
+		<reactor.version>3.3.0.RELEASE</reactor.version>
```
jvalkeal: I know that we're going to reactor 3.3.x when cf java client gets to it but why this change with this PR?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/TaskConfiguration.java
```diff
@@ -181,12 +182,14 @@ public TaskExecutionService taskService(LauncherRepository launcherRepository,
 			TaskAppDeploymentRequestCreator taskAppDeploymentRequestCreator,
 			TaskExplorer taskExplorer,
 			DataflowTaskExecutionDao dataflowTaskExecutionDao,
-			DataflowTaskExecutionMetadataDao dataflowTaskExecutionMetadataDao) {
+			DataflowTaskExecutionMetadataDao dataflowTaskExecutionMetadataDao,
+			@Autowired(required = false) OAuth2TokenUtilsService oauth2TokenUtilsService) {
```
jvalkeal: In spring5 we can now use `Optional` or even `@Nullable`. I think Optional is preferred way these days as this `required` thing was a spring4 thing.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowControllerAutoConfiguration.java
```diff
@@ -363,7 +365,8 @@ public StreamDeploymentController updatableStreamDeploymentController(
 
 		@Bean
 		public SkipperClient skipperClient(SkipperClientProperties properties,
-				RestTemplateBuilder restTemplateBuilder, ObjectMapper objectMapper) {
+				RestTemplateBuilder restTemplateBuilder, ObjectMapper objectMapper,
+				@Autowired(required = false) OAuth2TokenUtilsService oauth2TokenUtilsService) {
```
jvalkeal: Same `Optional` comment here vs using `@Autowired`.

spring-cloud-dataflow-server/pom.xml
jvalkeal: Actually this dep doesn't belong here. It needs to come from a transitive deps, which currently is only a cf java client.

pom.xml
ghillert: There is a call in Spring Security `WebClient` that uses an API call of 3.3.x.

spring-cloud-starter-dataflow-server/src/test/resources/org/springframework/cloud/dataflow/server/single/security/oauthConfig.yml
```diff
@@ -11,24 +11,40 @@ spring:
...
+    oauth2:
+      client:
+        registration:
+          uaa:
+            redirect-uri: '{baseUrl}/login/oauth2/code/{registrationId}'
```
jvalkeal: I can't find where baseUrl and registrationId are coming from?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/TaskConfiguration.java
ghillert: Changed to `@Nullable`

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowControllerAutoConfiguration.java
ghillert: Changed to `@Nullable`

spring-cloud-starter-dataflow-server/src/test/resources/org/springframework/cloud/dataflow/server/single/security/oauthConfig.yml
ghillert: Spring Security feature, see under https://docs.spring.io/spring-security/site/docs/5.2.0.RELEASE/reference/htmlsingle/#initiating-the-authorization-request Also, see the associated commit under: https://github.com/spring-projects/spring-security/pull/6900

spring-cloud-dataflow-docs/src/main/asciidoc/configuration-local.adoc
```diff
@@ -798,44 +835,105 @@ While the UAA is used by https://www.cloudfoundry.org/[Cloud Foundry],
...
+uaac target http://uaa:8080/uaa
+uaac token client get admin -s adminsecret
+uaac client add dataflow \
+  --name dataflow \
+  --secret dataflow \
+  --scope cloud_controller.read,cloud_controller.write,openid,password.write,scim.userids,foo.create,foo.view \
+  --authorized_grant_types password,authorization_code,client_credentials,refresh_token \
+  --authorities uaa.resource,dataflow.create,dataflow.deploy,dataflow.destroy,dataflow.manage,dataflow.modify,dataflow.schedule,dataflow.view,foo.view,foo.create\
+  --redirect_uri http://localhost:9393/login \
+  --autoapprove openid \
```
ilayaperumalg: We can remove the new line character here.

spring-cloud-dataflow-docs/src/main/asciidoc/configuration-local.adoc
```diff
@@ -798,44 +835,105 @@ While the UAA is used by https://www.cloudfoundry.org/[Cloud Foundry],
...
+./mvnw clean install
+java -jar target/uaa-bundled-1.0.0.BUILD-SNAPSHOT.jar
+----
+
+The configuration of the UAA is driven by either a Yaml file `uaa.yml` or you can script the configuration
+using the UAA Command Line Client:
+
+[source,bash]
+----
+uaac target http://uaa:8080/uaa
```
ilayaperumalg: Though it is implicit, we can say, `uaa` is the local host name in `uaa:8080`

ilayaperumalg:
This PR needs to be updated based on the recent package name changes on `spring-cloud-common-security-config`. Will update when merging.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed, rebased, fixed the package name conflicts, updated docs on the above comments and merged as 06adfa64af584f76fd3691e89d43c00ef324154a and d61938db4faa0249c67587d20c37c447fcd836b0

PR #3575, opened 2019-10-24
Fix task manifest sanitization

### LINKED ISSUE #3574: TaskManifest data are not sanitized
When the task execution is retrieved, we also fetch the task manifest data associated with it. But, the sensitive values in the task manifest are not sanitized/masked.

ilayaperumalg:
- Fix the return value of task manifest sanitization
 - Add tests for task execution and manifest sanitization
 - Move the sanitization code into Util

Resolves #3574

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskExecutionController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskExecutionController.java
@@ -23,15 +23,14 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.stream.Collectors;
 
 import org.springframework.cloud.dataflow.core.PlatformTaskExecutionInformation;
 import org.springframework.cloud.dataflow.core.TaskManifest;
 import org.springframework.cloud.dataflow.rest.job.TaskJobExecutionRel;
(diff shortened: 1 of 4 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskExecutionController.java
```diff
@@ -246,55 +243,19 @@ public void stop(@PathVariable("id") Set<Long> ids,
...
-				sanitizedAppProperties);
-		List<String> sanitizedCommandLineArgs = existingAppDeploymentRequest.getCommandlineArguments().stream()
-				.map(argument -> (this.argumentSanitizer.sanitize(argument))).collect(Collectors.toList());
-		AppDeploymentRequest sanitizedAppDeploymentRequest = new AppDeploymentRequest(
-				sanitizedAppDefinition,
-				existingAppDeploymentRequest.getResource(),
-				sanitizedDeploymentProperties,
-				sanitizedCommandLineArgs);
-		sanitizedTaskManifest.setTaskDeploymentRequest(sanitizedAppDeploymentRequest);
-		return taskManifest;
```
ilayaperumalg: Returning the `taskManifest` is the root cause of the issue.

cppwfs:
LGTM I'll polish the test above.

cppwfs:
Rebased, polished, merged.

PR #3577, opened 2019-10-24
Updated so that the we will rely on the deployer to give us the correct task def name

### LINKED ISSUE #3550: On the Schedule Tab Show Detail and Add New Schedule fail
This is a server side problem brought about by the changes we made to scheduling due to the CI/CD work for tasks.
In the past we had an expectation that a schedule had a direct relationship to the task for which it is scheduled. 
i.e. I could have:
Schedule1->timestamp
Schedule2->batch-timestamp
And I could query the scheduler and get the associated task.

With the changes we made to SCDF all schedules are mapped to the scheduler-launcher. 
i.e. 
Schedule1->scheduler-launcher
Schedule2->scheduler-launcher

Now that this relationship is severed we may have to store metadata on the SCDF side for each schedule to state what task it is associated (in order to support these features.

cppwfs:
When testing on CF be sure to use the CF deployer update found in this PR. https://github.com/spring-cloud/spring-cloud-deployer-cloudfoundry/pull/321

resolves #3550


cppwfs:
Closing for now.  We are implementing a new method of passing task-definition names to deployer-schedulers that is less invasive.

PR #3582, opened 2019-10-27
HAETOS compliant /runtime/streams endpoint

### LINKED ISSUE #2799: Make the /runtime/streams endpoint HAETOS compliant
The `/runtime/streams/` REST endpoint should be unified with the rest of the SCDF controller by providing a self-descriptive HAETOS responses. 

- The `StreamStatus`, `StreamStatus.Application` and `StreamStatus.Instance` classes have to be converted into Resources. 

- Consider replacing the `StreamStatus.Application` and `StreamStatus.Instance` with the existing `AppStatusResource` and `AppInstanceStatusResource` instead

tzolov:
- Add new StreamStatusResource that aggregates (name, CollectionModel<AppStatusResource>).
 - Add RuntimeStreamsControllerV2 with /runtime/streams/{streamNames} interface and returns StreamStatusResource.
 - Mark RuntimeStreamsController and exposed /runtime/streams?names={streamNames} interface depreciated.
 - Add AppInstanceStatusResource#getGuid() method/field that dynamically resolves the guid for the skipper property or falls back to the instance-id.
 - Add AppStatusResource#getName() to provide dynamically resolvable (from skipper.application.name) app name.
 - Add RuntimeOperations#streamStatus(String... streamNames) that returns PagedModel<StreamStatusResource>.
 - Extends Stream with runtimeApps() method to query the stream's runtime app instance statuses.
 - Add tests.

 Resolves #2799


ilayaperumalg:
Though the changes are related to the new REST endpoint and some changes around that, I think we can postpone merging this PR for the next minor release (2.3.1) as it could be risky at this point in time for 2.3.x release.

spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/resource/AppStatusResource.java
```diff
@@ -44,6 +44,12 @@ public AppStatusResource(String deploymentId, String state) {
 		this.state = state;
 	}
 
+	public String getName() {
+		AppInstanceStatusResource instance = this.instances.iterator().next();
+		return (instance != null && instance.getAttributes() != null) ?
+				instance.getAttributes().get("skipper.application.name") : "no-instances";
```
ilayaperumalg: I think we need to find a better place to fit these attribute names as constants (the skipper.application.name and skipper.release.version). Currently, they are declared in the controller package but not in scope of the REST resources.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/RuntimeStreamsController.java
```diff
@@ -37,12 +37,17 @@
...
+ * Deprecated and replaced by the {@link RuntimeStreamsControllerV2} instead.
+ * The new controller uses the /runtime/streams/{streamNames} query syntax.
+ *
  * Exposes runtime status of deployed streams.
  *
  * @author Christian Tzolov
  */
 @RestController
 @RequestMapping("/runtime/streams")
+@Deprecated
```
ilayaperumalg: Instead of deprecating this class, can we move the new changes into this controller itself and have the new REST endpoint updated?

tzolov: Currently the old (non compliant) REST endpoints are used by the UI for the metrics visualization (for example the dots under the deployed apps).  This as temp hack back then to transit from the pre-micrometer metrics collection.  
In short, imo we should keep the old endpoint until we migrate the UI to use the new one. And then we can remove it. 
The UI work could involving as we may have to change/adjust the JS domain model away from the proprietary classes used in the old endpoint.

spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/resource/StreamStatusResource.java
```diff
@@ -0,0 +1,61 @@
...
+		return name;
+	}
+
+	public void setName(String name) {
+		this.name = name;
+	}
+
+	public String getVersion() {
+		try {
+			return applications.iterator().next()
```
ilayaperumalg: I hope we won't get into NPE here. Is there a way to simplify this logic? Also, please try to move these keys (skipper.release.version etc.,) into constants and refer them

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/RuntimeStreamsControllerV2.java
```diff
@@ -0,0 +1,122 @@
...
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.RestController;
+
+/**
+ * @author Christian Tzolov
+ */
+@RestController
+@RequestMapping("/runtime/streams")
```
ilayaperumalg: I just learned from this change that we can have multiple controllers using the same `RequestMapping`.

tzolov: Yep, as long as the defined paths don't overlap (and they don't in this case)

spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/resource/StreamStatusResource.java
tzolov: sure i will add some checks to prevent NPE response and will factor out the constant.

spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/resource/AppStatusResource.java
tzolov: I will factor those at least as constants. 
But i agree they should be pulled in some parent (`spring-cloud-dataflow-core` perhaps) that is shared with the controllers as well.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed, rebased and merged as cbb216a8d

PR #3600, opened 2019-10-30
Support reusing deployer props and command-line args in tasks

### LINKED ISSUE #3555: Reuse app/deployer properties on subsequent launches
As part of the Task relaunches, we want to make sure the previously submitted app/deployer properties are reused and it is not expected from the user once again.

Notes: This behavior is already supported in Skipper for streams.

jvalkeal:
- Bring over analyzer and diff utils from a skipper order
  to see what has been changed in a task launches.
- PropertiesDiff impl copied over from skipper has been enhanced
  to support understanding of a deletion of a property. This now
  means that we understand a common property(not changed), changed
  property, added property, removed(doesn't exist in a new set) and
  deleted property(value is effectively missing).
- Deleted property vs removed property has a distiction in a sense
  that we then know that if user explicitely deleted it should not
  get back while removed property should get back to a new launch.
- Change in TaskAppDeploymentRequestCreator to stash all deploy props
  so that we know what to resurrect when app specific props were passed
  in via deploy props.
- DefaultTaskExecutionService has been modified(mostly executeTask has
  been partially rewritten to reuse props and args).
- Fixes #3555

NOTE: This pr probably needs heavy scrutiny as things get complex as I've
      most likely missed something.

Generic way to test these manually is to play with parameters and arguments
respectively:

```
app.timestamp.format=yyyy
app.timestamp.format=yyyy-MM
app.timestamp.format=

--timestamp.format=yyyy
--timestamp.format=yyyy-MM
--timestamp.format=
```

These will show how `format` is still in place if you use launch with not arguments
or properties. You can see how these can be modified and if value is letf empty it
will essentially get removed and then default `format` value takes place. Then
see a latest log what gets printed there and also what gets into a `TaskManifest`.

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
@@ -53,6 +53,8 @@ import org.springframework.cloud.dataflow.server.repository.TaskExecutionMissing
 import org.springframework.cloud.dataflow.server.service.TaskExecutionCreationService;
 import org.springframework.cloud.dataflow.server.service.TaskExecutionInfoService;
 import org.springframework.cloud.dataflow.server.service.TaskExecutionService;
+import org.springframework.cloud.dataflow.server.service.impl.diff.TaskAnalysisReport;
+import org.springframework.cloud.dataflow.server.service.impl.diff.TaskAnalyzer;
 import org.springframework.cloud.deployer.spi.core.AppDeploymentRequest;
 import org.springframework.cloud.deployer.spi.task.TaskLauncher;
 import org.springframework.cloud.task.repository.TaskExecution;
@@ -124,6 +126,8 @@ public class DefaultTaskExecutionService implements TaskExecutionService {
 
 	private final Map<String, List<String>> tasksBeingUpgraded = new ConcurrentHashMap<>();
 
+	private final TaskAnalyzer taskAnalyzer = new TaskAnalyzer();
+
 	/**
 	 * Initializes the {@link DefaultTaskExecutionService}.
 	 *
@@ -172,25 +176,27 @@ public class DefaultTaskExecutionService implements TaskExecutionService {
 	@Override
 	public long executeTask(String taskName, Map<String, String> taskDeploymentProperties, List<String> commandLineArgs,
 			String composedTaskRunnerName) {
-
+		// Get platform name and fallback to 'default'
 		String platformName = getPlatform(taskDeploymentProperties);
 
+		// Naive local state to prevent parallel launches to break things up
 		if(this.tasksBeingUpgraded.containsKey(taskName)) {
 			List<String> platforms = this.tasksBeingUpgraded.get(taskName);
 			if(platforms.contains(platformName)) {
-				throw new IllegalStateException(String.format("Unable to launch %s on platform %s because it is being upgraded", taskName, platformName));
+				throw new IllegalStateException(String.format(
+						"Unable to launch %s on platform %s because it is being upgraded", taskName, platformName));
 			}
 		}
 
-		// Remove since the key for task platform name will not pass validation for app, deployer, or scheduler prefix
+		// Remove since the key for task platform name will not pass validation for app,
+		// deployer, or scheduler prefix.
+		// Then validate
 		if (taskDeploymentProperties.containsKey(TASK_PLATFORM_NAME)) {
 			taskDeploymentProperties.remove(TASK_PLATFORM_NAME);
 		}
-
 		DeploymentPropertiesUtils.validateDeploymentProperties(taskDeploymentProperties);
 
-		TaskLauncher taskLauncher = findTaskLauncher(platformName);
-
+		// TODO: Same task name can only exist in one platform, um why?
 		TaskDeployment existingTaskDeployment = taskDeploymentRepository
 				.findTopByTaskDefinitionNameOrderByCreatedOnAsc(taskName);
 		if (existingTaskDeployment != null) {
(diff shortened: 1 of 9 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
```diff
@@ -339,15 +347,19 @@ private void saveExternalExecutionId(TaskExecution taskExecution, String taskDep
 	 * @param previousManifest manifest from the last execution of the same task definition
 	 * @return an updated {@code AppDeploymentRequest}
 	 */
-	private AppDeploymentRequest updateDeploymentProperties(List<String> commandLineArgs, String platformName, TaskExecutionInformation taskExecutionInformation, TaskExecution taskExecution, TaskManifest previousManifest) {
+	// private AppDeploymentRequest updateDeploymentProperties(List<String> commandLineArgs, String platformName, TaskExecutionInformation taskExecutionInformation, TaskExecution taskExecution, TaskManifest previousManifest) {
```
jvalkeal: Looks like forget to remove comment lines.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
```diff
@@ -172,25 +176,27 @@ public DefaultTaskExecutionService(LauncherRepository launcherRepository,
...
+		// Then validate
 		if (taskDeploymentProperties.containsKey(TASK_PLATFORM_NAME)) {
 			taskDeploymentProperties.remove(TASK_PLATFORM_NAME);
 		}
-
 		DeploymentPropertiesUtils.validateDeploymentProperties(taskDeploymentProperties);
 
-		TaskLauncher taskLauncher = findTaskLauncher(platformName);
-
+		// TODO: Same task name can only exist in one platform, um why?
```
ilayaperumalg: I think this is because our task execution model (and the corresponding batch model) is not aware of the platform.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
```diff
@@ -202,53 +208,54 @@ public long executeTask(String taskName, Map<String, String> taskDeploymentPrope
...
 				.findTaskExecutionInformation(taskName, taskDeploymentProperties, composedTaskRunnerName);
-
 		if (taskExecutionInformation.isComposed()) {
 			handleAccessToken(commandLineArgs, taskExecutionInformation);
 		}
 
+		// Build app deploy request and stash deploy props there
 		TaskExecution taskExecution = taskExecutionRepositoryService.createTaskExecution(taskName);
-
 		AppDeploymentRequest appDeploymentRequest = this.taskAppDeploymentRequestCreator.
```
ilayaperumalg: Can we create the `AppDeploymentRequest` later in the method instead of updating it again?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
```diff
@@ -202,53 +208,54 @@ public long executeTask(String taskName, Map<String, String> taskDeploymentPrope
...
-			}
-		}
+		// We now have a new props and args what should really get used.
+		Map<String, String> mergedTaskDeploymentProperties = report.getMergedDeploymentProperties();
+		List<String> mergedCommandLineArguments = report.getMergedCommandLineArguments();
+
+		taskExecutionInformation.setTaskDeploymentProperties(mergedTaskDeploymentProperties);
+		appDeploymentRequest = updateDeploymentProperties(mergedCommandLineArguments, platformName, taskExecutionInformation,
+				taskExecution, mergedTaskDeploymentProperties);
+		taskDeploymentProperties = mergedTaskDeploymentProperties;
```
ilayaperumalg: This `taskDeploymentProperties` is not used now.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
```diff
@@ -202,53 +208,54 @@ public long executeTask(String taskName, Map<String, String> taskDeploymentPrope
...
 
 		AppDeploymentRequest request = new AppDeploymentRequest(appDeploymentRequest.getDefinition(),
```
ilayaperumalg: We can probably create the AppDeploymentRequest at this line after we updated the necessary task deployment and command line arguments.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
jvalkeal: This flow is so complex with so many side effects even in a original code that for now you kinda need to. Eventually these methods need a complete rewrite.

ilayaperumalg:
LGTM

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
ilayaperumalg: will remove this while merging.

ilayaperumalg:
Merging.

ilayaperumalg:
Rebased and merged as 1ce362a23679a54ef24241c503de0078f4f92f1c

PR #3603, opened 2019-10-30
Restarting Jobs with incrementer now will work

### LINKED ISSUE #3483: Restarting failed job which uses incrementer creates new JobInstance
**Description:**
I created a job which uses RunIdIncrementer. When launching task it increments id as expected. However if job fails and I try to restart it using UI, id is also incremented, which results in new JobInstance being created. Is this the correct behavior?
I can restart failed JobInstance by manually entering job parameters but is there a way to do it using restart button?

**Release versions:**
Tested on Spring Cloud Data Flow 2.2.1.RELEASE local and kubernetes

cppwfs:
resolves #3483

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskJobService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskJobService.java
@@ -27,6 +27,8 @@ import org.slf4j.LoggerFactory;
 
 import org.springframework.batch.core.JobExecution;
 import org.springframework.batch.core.JobInstance;
+import org.springframework.batch.core.JobParameter;
+import org.springframework.batch.core.JobParameters;
 import org.springframework.batch.core.StepExecution;
 import org.springframework.batch.core.launch.JobExecutionNotRunningException;
 import org.springframework.batch.core.launch.NoSuchJobException;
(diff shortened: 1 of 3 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskJobService.java
```diff
@@ -190,14 +192,45 @@ public void restartJobExecution(long jobExecutionId) throws NoSuchJobExecutionEx
 			deploymentProperties.put(DefaultTaskExecutionService.TASK_PLATFORM_NAME, platformName);
 			String taskAppName = taskJobExecution.getJobExecution().getJobParameters().getString("-spring.cloud.data.flow.taskappname");
 			taskExecutionService.executeTask(taskDefinition.getName(), deploymentProperties,
-					taskExecution.getArguments(), taskAppName);
+					restartExecutionArgs(taskExecution.getArguments(),
```
ilayaperumalg: The only caveat with this approach is that we persist the Job parameters as part of the task execution arguments (as soon as the failed task gets re-started for the first time) which I think is ok.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Verified the changes with a batch job that fails and gets restarted (with the previous job parameters applied) and most importantly, the job instance for the restarted batch job is the same as that of the original batch job that failed.

ilayaperumalg:
Rebased and merged as 26155f6e6

PR #3607, opened 2019-10-31
Add SCDF App Tool to the docs

sabbyanandan:
The no-maven situation continues to come up at customers. Unfortunately, we don't yet list the SCDF App Tool in the docs, so this commit adds just that.

It'd be nice to backport it to 2.2.x.


dturanski:
LGTM; merging

PR #3611, opened 2019-11-02
commandline args must be prefixed

### LINKED ISSUE #3608: CommandLine args of task is passed to Scheduler task launcher
With the new Scheduler task launcher in effect, the command line arguments passed for the underlying task definition are also passed to the scheduler task launcher application as well.

cppwfs:
This is so that the args will not have an affect on the scheduler launcher.

resolves #3608

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultSchedulerService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultSchedulerService.java
@@ -68,6 +68,7 @@ public class DefaultSchedulerService implements SchedulerService {
 
 	private final static String APP_PREFIX = "app.";
 	private final static String DEPLOYER_PREFIX = "deployer.";
+	private final static String COMMAND_ARGUMENT_PREFIX = "cmdarg.";
 	private final static String DATA_FLOW_URI_KEY = "spring.cloud.dataflow.client.serverUri";
 
 	private CommonApplicationProperties commonApplicationProperties;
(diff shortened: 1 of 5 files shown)


ilayaperumalg:
Can we fix this by not passing the commandline args from task launch arguments to the scheduler launcher [here](https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultSchedulerService.java#L202) ?

cppwfs:
Hmmm, I don't think that will work.   We need for the command line arguments to be passed to to the application that is being launched.

ilayaperumalg:
I believe that the command-line arguments that are passed into the `schedule` method are intended to be used only for the task being launched by the `scheduler task launcher` and hence, we don't have to construct the cmd line args for the `scheduler task launcher` with the passed command-line args [here](https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultSchedulerService.java#L202). Please correct me the part I miss here.

cppwfs:
Agreed.  The goal was to utilize the existing scheduler infrastructure to pass command line args to the scheduler task launcher by tagging those args meant for the task to be executed (following the same pattern as properties).   If we don't pass them through this method we will have to pass them through the properties and that introduces parsing complexity.

cppwfs:
I should say I agree with the first sentence of your response.  I disagree with the second ;-)

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultSchedulerService.java
```diff
@@ -214,6 +215,23 @@ public void schedule(String scheduleName, String taskDefinitionName, Map<String,
...
+		List<String> taggedArgs = new ArrayList<>();
+
+		for(String arg : args) {
+			if(arg.contains("spring.cloud.task.name")) {
+				continue;
+			}
+			String updatedArg = arg;
+			if (!arg.startsWith(DATA_FLOW_URI_KEY) && !"--".concat(arg).startsWith(DATA_FLOW_URI_KEY)) {
+					updatedArg = COMMAND_ARGUMENT_PREFIX +
+							taskConfigurationProperties.getTaskLauncherPrefix() + arg;
```
ilayaperumalg: We need to add the explicit dot here (Given we now have the task launcher prefix without the dot). Will fix when merging.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Verified the changes on K8s by running the scheduled tasks with cmd-line args and they are invoked with the correct command-line arguments.

ilayaperumalg:
Fixed the task launcher prefix usage inside the command line arguments for the scheduler task launcher. Rebased and merged as 65b8f540d

PR #3613, opened 2019-11-04
Removed the "." from TaskConfigurationProperties from prefix

### LINKED ISSUE #3609: Consider removing dot on Scheduler task launcher prefix
Currently, the scheduler task launcher prefix is set to `tasklauncher.`. Given, we use this prefix to filter the app/deployer properties associated with the underlying task definition when launching, if the user sets a new prefix without the **dot**, then the prefixed properties wouldn't get applied into the launching task.

cppwfs:
resolves #3609


ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Rebased and merged as 9b963a627

PR #3616, opened 2019-11-05
Support task arguments with single and no dash

jvalkeal:
- This commit adds support for processing task arguments
  which are passed as `-foo=bar` and `foo=bar` format.
- Relates to an issue PR'd with #3555 which broke
  tests as new tests for these pros were added after
  PR were issued.
- Polish DefaultTaskExecutionService


ilayaperumalg:
LGTM, merging.

PR #3619, opened 2019-11-05
Add feature to auto create task definitions if flag is set.

dturanski:
Resolves https://github.com/spring-cloud-stream-app-starters/tasklauncher-dataflow/issues/24

Modify DefaultTaskExecutionService to auto create task definition from registered task app if property is set.  Currently requires a work-around to manually invoke `TaskDefinition.initialize()` in `DefaultTaskExecutionInfoService` still investigating root cause, but should work as is.

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/TaskConfiguration.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/TaskConfiguration.java
@@ -97,6 +97,9 @@ public class TaskConfiguration {
 	@Value("${spring.cloud.dataflow.server.uri:}")
 	private String dataflowServerUri;
 
+	@Autowired
+	private TaskConfigurationProperties taskConfigurationProperties;
+
 	@Bean
 	public DeployerConfigurationMetadataResolver deployerConfigurationMetadataResolver(
 			TaskConfigurationProperties taskConfigurationProperties) {
@@ -181,12 +184,16 @@ public class TaskConfiguration {
 			TaskAppDeploymentRequestCreator taskAppDeploymentRequestCreator,
 			TaskExplorer taskExplorer,
 			DataflowTaskExecutionDao dataflowTaskExecutionDao,
-			DataflowTaskExecutionMetadataDao dataflowTaskExecutionMetadataDao) {
-		return new DefaultTaskExecutionService(
+			DataflowTaskExecutionMetadataDao dataflowTaskExecutionMetadataDao,
+			TaskSaveService taskSaveService) {
+
+		DefaultTaskExecutionService defaultTaskExecutionService = new DefaultTaskExecutionService(
 				launcherRepository, auditRecordService, taskRepository,
 				taskExecutionInfoService, taskDeploymentRepository, taskExecutionRepositoryService,
 				taskAppDeploymentRequestCreator, taskExplorer, dataflowTaskExecutionDao,
-				dataflowTaskExecutionMetadataDao);
+				dataflowTaskExecutionMetadataDao, taskSaveService);
+		defaultTaskExecutionService.setAutoCreateTaskDefinitions(taskConfigurationProperties.isAutoCreateTaskDefinitions());
+		return defaultTaskExecutionService;
 	}
 
 	@Bean
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionInfoService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionInfoService.java
@@ -127,6 +127,8 @@ public class DefaultTaskExecutionInfoService implements TaskExecutionInfoService
 
 		TaskDefinition originalTaskDefinition = taskDefinitionRepository.findById(taskName)
 				.orElseThrow(() -> new NoSuchTaskDefinitionException(taskName));
+		//TODO: This normally called by JPA automatically but `AutoCreateTaskDefinitionTests` fails without this.
+		originalTaskDefinition.initialize();
 		TaskParser taskParser = new TaskParser(originalTaskDefinition.getName(), originalTaskDefinition.getDslText(),
 				true, true);
 		TaskNode taskNode = taskParser.parse();
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
@@ -145,7 +145,6 @@ public class DefaultTaskExecutionService implements TaskExecutionService {
 	 * @param taskExecutionRepositoryService the service used to create the task execution
 	 * @param dataflowTaskExecutionMetadataDao repository used to manipulate task manifests
 	 */
-	//@formatter:off
 	public DefaultTaskExecutionService(LauncherRepository launcherRepository,
 			AuditRecordService auditRecordService,
 			TaskRepository taskRepository,
@@ -157,7 +156,7 @@ public class DefaultTaskExecutionService implements TaskExecutionService {
 			DataflowTaskExecutionDao dataflowTaskExecutionDao,
 			DataflowTaskExecutionMetadataDao dataflowTaskExecutionMetadataDao,
 			TaskSaveService taskSaveService) {
-		//@formatter:on
+
 		Assert.notNull(launcherRepository, "launcherRepository must not be null");
 		Assert.notNull(auditRecordService, "auditRecordService must not be null");
 		Assert.notNull(taskExecutionInfoService, "taskExecutionInfoService must not be null");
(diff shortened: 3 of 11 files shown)


dturanski:
After further investigation, I think its easiest to leave the work-around in place.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
```diff
@@ -285,6 +305,28 @@ public long executeTask(String taskName, Map<String, String> taskDeploymentPrope
...
+
+		TaskExecutionInformation taskExecutionInformation;
+		try {
+			 taskExecutionInformation = taskExecutionInfoService
+					.findTaskExecutionInformation(taskName, taskDeploymentProperties, composedTaskRunnerName);
+
+		} catch (NoSuchTaskDefinitionException e) {
+			if (autoCreateTaskDefinitions) {
+				logger.info("Creating a Task Definition {} for registered app name {}", taskName, taskName);
+				TaskDefinition taskDefinition = new TaskDefinition(taskName, taskName);
```
ilayaperumalg: Not sure if this is a valid assumption to have both the task name and the definition to be of the same.

Perhaps, we can introduce a new task deployment property (something like `spring.cloud.dataflow.task.definition`) and infer the task definition from the task deployment property.

dturanski: Hi @ilayaperumalg. There are a couple of issues with your suggestion. The main one is that `taskName` in the launch request is the task name. I don't think we should overload the meaning of that field. The other is that it allows the task definition to be overridden for each launch request. For example a user could create a new task definition for every launch of the same app. I'm not sure we want to go there.  My preference is to stay with this opinionated approach for now. If the community needs more flexibility, we can revisit it then.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionServiceTests.java
```diff
@@ -417,7 +417,7 @@ public void setup() {
 			this.launcherRepository.save(new Launcher("MyPlatform", "local", taskLauncher));
 
 			taskDefinitionRepository.save(new TaskDefinition(TASK_NAME_ORIG, "demo"));
-			taskDefinitionRepository.findAll();
+			//taskDefinitionRepository.findAll();
```
cppwfs: We can remove these comments.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
cppwfs: My primary concern with this PR.  Is that now we are now saying that task launches can be directly associated with an application registration, which blurs the difference between app registrations and task definitions..   I know this is being used by one user, but is this something we want to support for all users?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionInfoService.java
```diff
@@ -127,6 +127,8 @@ public TaskExecutionInformation findTaskExecutionInformation(String taskName,
 
 		TaskDefinition originalTaskDefinition = taskDefinitionRepository.findById(taskName)
 				.orElseThrow(() -> new NoSuchTaskDefinitionException(taskName));
+		//TODO: This normally called by JPA automatically but `AutoCreateTaskDefinitionTests` fails without this.
```
cppwfs: Do we want to go forward with the initialize here or is it safe to use `@Transactional(propagation = Propagation.SUPPORTS)`?   I know we discussed not using Propagation.NOT_SUPPORTED.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionServiceTests.java
```diff
@@ -152,8 +152,8 @@
 	@Autowired
 	TaskDeleteService taskDeleteService;
 
-	@Autowired
-	TaskExecutionInfoService taskExecutionInfoService;
+	//@Autowired
+	//TaskExecutionInfoService taskExecutionInfoService;
```
cppwfs: We can remove these  lines

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionServiceTests.java
```diff
@@ -708,6 +709,31 @@ public void validateNullResourceTaskTest() {
...
+		public void setup() {
+			this.launcherRepository.save(new Launcher("default", "local", taskLauncher));
+		}
+
+		@Test
+		@DirtiesContext
+		public void executeTaskWithNullDefinitionCreatesDefinitionIfConfigured() {
+			initializeSuccessfulRegistry(appRegistry);
+			when(this.taskLauncher.launch(any())).thenReturn("0");
+			taskExecutionService.executeTask("demo", new HashMap<>(), new LinkedList<>());
```
cppwfs: Can we test the task definitions repo to see if the task definition was created?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/TaskConfiguration.java
```diff
@@ -181,12 +184,16 @@ public TaskExecutionService taskService(LauncherRepository launcherRepository,
...
+			DataflowTaskExecutionMetadataDao dataflowTaskExecutionMetadataDao,
+			TaskSaveService taskSaveService) {
+
+		DefaultTaskExecutionService defaultTaskExecutionService = new DefaultTaskExecutionService(
 				launcherRepository, auditRecordService, taskRepository,
 				taskExecutionInfoService, taskDeploymentRepository, taskExecutionRepositoryService,
 				taskAppDeploymentRequestCreator, taskExplorer, dataflowTaskExecutionDao,
-				dataflowTaskExecutionMetadataDao);
+				dataflowTaskExecutionMetadataDao, taskSaveService);
+		defaultTaskExecutionService.setAutoCreateTaskDefinitions(taskConfigurationProperties.isAutoCreateTaskDefinitions());
```
cppwfs: Just curious why are we using a setter instead of adding it to constructor?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionInfoService.java
dturanski: AFAIK `propogation.SUPPORTS` does not create a transaction but will participate if already running in one. I don't think that will work here.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionServiceTests.java
dturanski: It couldn't hurt but I don't think we would get to the deployer launch if not.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/TaskConfiguration.java
dturanski: Mostly because it's optional. I follow the convention of constructor args for required dependencies.
But also because any change to this constructor has to be replicated in 4 places for the test fixtures :)

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
dturanski: The way I think of it is that the task definition is in most cases is only needed to associate a task definition with a task app. For example, the file ingest sample doesn't need anything else. In fact the convention I use is to use the app name as the task name (In the instructions, I specifically don't do this to make it clear that creating the task definition is not the same as registering the app - but it's natural to think of them as the same IMO). So for those cases, it's just an extra bit of boilerplate to remember, and will save me a step going forward.  I wanted to make this work be default, but agreed to make it opt-in.  My guess is that many will find this useful.  The risk is if someone needs a real task definition, e.g., for CTR, and forget to create the definition, they may inadvertently kick off a misconfigured task. Not much of a down side if you ask me.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/TaskConfiguration.java
cppwfs: 👍

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
cppwfs: 👍

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionInfoService.java
cppwfs: 👍

cppwfs:
LGTM .  @sabbyanandan do we want to merge this for the RC1?

sabbyanandan:
@cppwfs: Correct. Milestone marker should reflect that, as well.

cppwfs:
Rebased, Conflicts resolved, cleanup, merged

PR #3638, opened 2019-11-18
Fix task deployment properties

### LINKED ISSUE #3636: Task deployment properties broken
Commit https://github.com/spring-cloud/spring-cloud-dataflow/commit/1ce362a23679a54ef24241c503de0078f4f92f1c removed call to `DeploymentPropertiesUtils.extractAndQualifyDeployerProperties` as we want to keep all props but side effect in that function is that it modifies those deploy props prefixes i.e. from `deployer.timestamp.local` to `spring.cloud.deployer.local`. Without this binding in deployer impls don't work properly.

jvalkeal:
- Add new qualifyDeployerProperties which also keep around
  app related deploy props.
- Problem with original PR adding deploy props reuse were that
  we never qualified those props so we're missing a proper
  prefix.
- For exaple if we set `deployer.timestamp.local.javaOpts` we
  now get `spring.cloud.deployer.local.javaOpts` to stored in
  a manifest which aligns how things are used and stored with
  streams.
- Fixes #3636

### DIFF
--- a/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
+++ b/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
@@ -234,6 +234,39 @@ public final class DeploymentPropertiesUtils {
 	}
 
 	/**
+	 * Retain all properties that are meant for the <em>deployer</em> of a given app
+	 * (those that start with {@code deployer.[appname]} or {@code deployer.*} or
+	 * {@code app.*}) and qualify all property values with the
+	 * {@code spring.cloud.deployer.} prefix.
+	 *
+	 * @param input   the deplopyment properties
+	 * @param appName the app name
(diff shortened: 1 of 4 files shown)


spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
```diff
@@ -233,6 +233,39 @@ public static void validateSkipperDeploymentProperties(Map<String, String> prope
 		return result;
 	}
 
+	/**
+	 * Retain all properties that are meant for the <em>deployer</em> of a given app
+	 * (those that start with {@code deployer.[appname]} or {@code deployer.*} or
+	 * {@code app.*}) and qualify all property values with the
```
cppwfs: It only supports deployer.[appname] and deployer.*.  It does not support app.*.  will remove comment on merge.

cppwfs:
LGTM.  Will apply the cleanup for the comment above.

cppwfs:
Applied small cleanup on merge.  Squashed.   Thanks!!!!!

Issue #2649, opened 2018-12-04
Add REST resource to filter Jobs by name and status


oodamien:
**Problem description:**
As a user, there is no option to filter the jobs.

**Solution description:**
As a user, I would like to filter the jobs by TaskName and/or status.

sylank:
Hi @oodamien,

Currently, you can filter jobs by job name:
http://localhost:9393/jobs/executions?name=job%

I had job1 and job2 and it returned with two JobExecutionResource.

Would you like to filter jobs by task name as well?

PR #3653, opened 2019-11-24
Add REST resource to filter Jobs by name and status

sylank:
Job execution filtering by batch status.

@oodamien, do we need a filtering option by task name too?

Resolves: #2649


jvalkeal:
I have some reservations for this PR as it seem to rely on passing name like to sql. I think users kinda might expect to be able to use `job*` or some other normal things.
```
http://localhost:9393/jobs/executions?name=job%25&status=COMPLETED
```
But then we can't filter by status:
```
http://localhost:9393/jobs/executions?status=UNKNOWN
```
as it basically returns everything.

What you think @cppwfs

cppwfs:
I agree.

sylank:
I managed to reproduce the issue. I can fix it when I have enough time. Thank you for the debugging 😸 .

sylank:
Now the user/frontend has to send the percent characters in URL encoded form when they want to use wildcards. I wanted to avoid using the `%` chars and use `contains` expression at the service level. Because of spring batch initialization, I can not use the `JpaRepository` and `@Query` pattern without a big refactor.
Then I wanted to inject the `%` chars into the where clause but it failed because of the SQL compiler that replaces the values between two percent mark :P.

The issue was a rooting problem. I had to merge the two query function to solve it :).

jvalkeal:
Sorry for taking so long to look at this. Looks like this now works, afaik. @cppwfs would you mind check and see if you can bless this?

cppwfs:
Basic tests were successful.
The only issue that I see is that our HATEOS homepage request still shows the old URL i.e.
    "jobs/executions/name": {
      "href": "http://localhost:9393/jobs/executions{?name}",
      "templated": true
    },
Should show `name` and `status`

Tests Ran:
.  No Param <Success>
.  Specific Job Name <Success>
. Specific Job Name With Status (Complete and Failed) <Success>
. Specific Status (completed and failed) <Success>

sylank:
Is this format okay?
image

sylank:
@jvalkeal could you validate it again?

jvalkeal:
Yeah, `/jobs/executions/status` is wrong as were existing `/jobs/executions/name`.

jvalkeal:
Polished/squashed/merged per 07cc47105fa94d3a8d96ddd4cc1e77055b0d9e75

Issue #3695, opened 2019-12-11
Can't launch same task definition with different arguments concurrently in 2.3.0


djgeary:
In SCDF 2.2.1 (Kubernetes) we launch tasks with the same definition and different argument/properties frequently.
Properties that don't commonly change we would put in the definition and those that do we supply at launch time.
Some of these executions may overlap.

After updating to SDCF 2.3.0 when launching tasks (through the rest API) with different arguments to ones already running we get a 500 status code and the message body 
`[ { "logref": "IllegalStateException", "message": "Unable to update application due to currently running applications" } ]`

As discussed on gitter (https://gitter.im/spring-cloud/spring-cloud-dataflow?at=5df0fa540995661eb8c923d3) this is due to the new CD functionality in 2.3.0.

The seems like a serious restriction in functionality, as it seems its now not possible to launch a task definition with different arguments or properties from one already running.

In our case this would result in 'random' failures. Surely changing a launch property is nothing to do with CI or updates/upgrades (which I assume is related to the task application version only?).
Also there may be the odd failed/aborted execution in the history with no end time that could impact launches too.
The exception thrown is a 500 and mentions 'update application' which isn't very meaningful either - as we aren't trying to update an application.

It would be useful if there was a way to turn this off or simulate the previous behaviour - running a task definition concurrently with different arguments doesn't seem like an unusual thing to do?

sabbyanandan:
Hi, @djgeary. As discussed in Gitter (and clarified by @mminella), the CLI-args aren't included in the CD flows, so you should be able to launch/re-launch with dynamic arguments. There's currently support to persist the CLI-args in SCDF 2.3, but we will decouple that in #3694. With this change, you should be able to _concurrently_ launch the _same_ task-definition _with_ different CLI-args.

We will also [update the docs](https://docs.spring.io/spring-cloud-dataflow/docs/2.3.0.RELEASE/reference/htmlsingle/#_task_launch_lifecycle) to describe the App Properties, Deployer properties, and CLI Arguments and its applicability on CICD flows.

Let us know if you see any anomaly.

djgeary:
Hi @sabbyanandan, I'm not sure I follow 

> There's currently support to persist the CLI-args in SCDF 2.3

Is this something that would currently prevent us launching the same task concurrently with different arguments?

[...]

PR #3697, opened 2019-12-11
Args are stored from previous execs but not used for the next launch

### LINKED ISSUE #3694: Next execution of scheduled task adds arguments from previous task execution
**Description:**
If task is scheduled and then launched manually with new argument, next scheduled execution will have the same argument. I discovered this after restarting batch job which creates new task execution with additional incrementable argument. Next time Kubernetes job was created, new task execution had the same argument which resulted in error (job instance already complete with provided parameters).

**Release versions:**
Spring Cloud Data Flow 2.3.0.RELEASE on Kubernetes

**Steps to reproduce:**
1. Create task and schedule it.
2. Launch same task with some arguments.
3. Next time scheduled task execution is created it will have the same arguments.

cppwfs:
resolves #3694


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
```diff
@@ -254,10 +254,9 @@ public long executeTask(String taskName, Map<String, String> taskDeploymentPrope
 
 		// We now have a new props and args what should really get used.
 		Map<String, String> mergedTaskDeploymentProperties = report.getMergedDeploymentProperties();
-		List<String> mergedCommandLineArguments = report.getMergedCommandLineArguments();
 
 		taskExecutionInformation.setTaskDeploymentProperties(mergedTaskDeploymentProperties);
-		appDeploymentRequest = updateDeploymentProperties(mergedCommandLineArguments, platformName, taskExecutionInformation,
+		appDeploymentRequest = updateDeploymentProperties(commandLineArgs, platformName, taskExecutionInformation,
```
ilayaperumalg: I wonder how and why was the `mergedCommandLineArguments` used before. By any chance, can we get this `executeTask` method refactored as this has lots of moving parts and becoming incomprehensible?

cppwfs:
Hello Ilaya,
I can break this method up into some manageable pieces .
As to how `mergedCommandLineArguments` was used in the past.   In this case it was not used in previous releases of Spring Cloud Data Flow.   It was introduced in the following commit: https://github.com/spring-cloud/spring-cloud-dataflow/commit/1ce362a23679a54ef24241c503de0078f4f92f1c#diff-ecf6ab80ac41cacc755033ef4163b11f.  However after discussions with @mminella it was determined that this was done in error and should not have been added.

cppwfs:
PR Updated.  Broke up the method so that it is more readable.

sabbyanandan:
@cppwfs: Does this PR also address https://github.com/spring-cloud/spring-cloud-dataflow/issues/3695?

mminella:
Squashed, rebased, and merged as aea7d207a77ccf8570c1bfed753bc2e107f051f1

canerkoseren:
Hi all,

I was using SCDF 2.5.1 and having a problem like launching a task. I could launch the application with the properties of previous execution. 

Today I updated the version of scdf to 2.7.1 after seeing this fix & merge, but the result was same. 

By using scdf-dashboard, I executed the first application with the properties given below
```
deployer.batchapp.kubernetes.environmentVariables=PARAMS1=input_file1.csv,PARAMS2=output_file2.csv
deployer.batchapp.kubernetes.readinessProbeDelay=120
deployer.batchapp.kubernetes.livenessProbePath=/live
deployer.batchapp.kubernetes.readinessProbePath=/ready
deployer.batchapp.kubernetes.livenessProbeDelay=120
```

After the application ends without error, I wanted to launch the task again. On the scdf-dashboard, I opened the task page and click "Launch Task", and in the next page, I saw all parameters of the previous execution.

Regarding to this fix, shouldn't that parameter section be opened empty?

cppwfs:
Helllo  @canerkoseren ,
This fix pertains to command line arguments of a task launch, not the properties.   The properties will be carried forward for each new launch of a task.

canerkoseren:
Hi @cppwfs, thanks for replies.

Issue #3670, opened 2019-12-04
Add new build structure


jvalkeal:
Essentially we want to be less dependant on a spring-cloud build things as we're not exactly part of that train. Work focuses moving away from `spring-cloud-build` in favour of our own structure.

- [x] Create spring-cloud-dataflow-build
- [ ] Skipper
- [ ] Commons Security
- [ ] Deployers

PR #3702, opened 2019-12-13
Move to spring-cloud-dataflow-build

jvalkeal:
- Simple change to use spring-cloud-dataflow-build instead
  of spring-cloud-build.
- Relates spring-cloud/spring-cloud-dataflow#3670

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -15,8 +15,8 @@
 			 release versions -->
 		<!-- Note: Also make sure the parent pom version in spring-cloud-dataflow-dependencies/pom.xml is in sync -->
 		<groupId>org.springframework.cloud</groupId>
-		<artifactId>spring-cloud-build</artifactId>
-		<version>2.2.0.RELEASE</version>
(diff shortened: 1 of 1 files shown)


ilayaperumalg:
@jvalkeal Are we planning to do the other possible refactoring in the POM files based on the `spring-cloud-dataflow-build` changes in a separate PR?

jvalkeal:
Not really sure what you mean but if you mean polishing poms and removing unneeded stuff if favour of spring-cloud-dataflow-build, then yes those needs to happen but probably not within these first PR's.

ilayaperumalg:
> if you mean polishing poms and removing unneeded stuff if favour of spring-cloud-dataflow-build, then yes those needs to happen but probably not within these first PR's.

That's exactly what I meant :-) and sure, let's take care of the changes in separate PRs.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Merged as 7c7e10824dbf71fa436e84ec163c586e084d867c

PR #3727, opened 2020-01-07
Added check to make sure schedule name fits k8's spec

### LINKED ISSUE #3704: Follow the applicable naming conventions for K8s resources
Given we append the task name as part of the scheduler name, the K8s scheduler name validation fails on certain cases. We either need to validate the task name (using the K8s scheduler name format) or use a different approach for appending the scheduler name.

For instance, if I have a task definition name `A1` then, the creation of the schedule would fail on K8s.

cppwfs:
All other checks will be managed by the deployer.
This size check was required so that users could get a useful message when they exceed the 52 char limit
Since now we construct the schedule name

resolves #3704

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultSchedulerService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultSchedulerService.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2019 the original author or authors.
+ * Copyright 2018-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultSchedulerService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultSchedulerService.java
@@ -36,6 +36,7 @@ import org.springframework.cloud.dataflow.core.AuditOperationType;
 import org.springframework.cloud.dataflow.core.Launcher;
 import org.springframework.cloud.dataflow.core.TaskDefinition;
 import org.springframework.cloud.dataflow.core.TaskPlatform;
+import org.springframework.cloud.dataflow.core.TaskPlatformFactory;
 import org.springframework.cloud.dataflow.core.dsl.TaskNode;
 import org.springframework.cloud.dataflow.core.dsl.TaskParser;
 import org.springframework.cloud.dataflow.registry.service.AppRegistryService;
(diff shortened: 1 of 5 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultSchedulerService.java
```diff
@@ -70,6 +70,8 @@
 	private final static String DEPLOYER_PREFIX = "deployer.";
 	private final static String COMMAND_ARGUMENT_PREFIX = "cmdarg.";
 	private final static String DATA_FLOW_URI_KEY = "spring.cloud.dataflow.client.serverUri";
+	private final static String KUBERNETES = "kubernetes";
```
ilayaperumalg: We could perhaps move the PLATFORM_TYPE constant values (kubernetes, cloudfoundry and local) from TaskPlatformFactory classes into some common or core module and that way we can re-use them in other places.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultSchedulerService.java
```diff
@@ -205,11 +207,25 @@ public void schedule(String scheduleName, String taskDefinitionName, Map<String,
...
 		Launcher launcher = getDefaultLauncher();
+		validateScheduleNameForPlatform(launcher.getType(), revisedDefinition.getName());
 		launcher.getScheduler().schedule(scheduleRequest);
 		this.auditRecordService.populateAndSaveAuditRecordUsingMapData(AuditOperationType.SCHEDULE, AuditActionType.CREATE,
 				scheduleRequest.getScheduleName(), this.auditServiceUtils.convertScheduleRequestToAuditData(scheduleRequest),
 				taskPlatform.getName());
 	}
+
+	private void validateScheduleNameForPlatform(String type, String scheduleName) {
+		if(type.equals(KUBERNETES)) {
```
ilayaperumalg: I thought we were planning to convert the case of the schedule name to be lower case and add the corresponding case insensitive checks on task definition names in other places that involves the task name references.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultSchedulerService.java
cppwfs: I love your idea!   But what I will do is to pull the types that are enumerated in each of the impl's of `TaskPlatformFactory` to the `TaskPlatformFactory` itself.  This way we have a single location from which to pull the types.   I'll submit this change in the update to the PR.

cppwfs: Tell me what you think.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultSchedulerService.java
cppwfs: I will lowercase the task-definition-name portion of the schedule name as well.  We do this when we create the pods.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultSchedulerService.java
ilayaperumalg: This is a nice approach! :-)

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed, rebased and merged as 3c551ad5c

ilayaperumalg:
Cherry-picked the changes and backported into 2.3.x as 6a2c9324a8c4df40fc4238fb5299a93db90c3ac7

PR #3742, opened 2020-01-15
Updated so that validate and lock only occurs for CF Tasks

### LINKED ISSUE #3695: Can't launch same task definition with different arguments concurrently in 2.3.0

cppwfs:
resolves #3695

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
@@ -269,7 +269,8 @@ public class DefaultTaskExecutionService implements TaskExecutionService {
 		String taskDeploymentId = null;
 		try {
 			TaskLauncher taskLauncher = findTaskLauncher(platformName);
-			if(!isAppDeploymentSame(previousManifest, taskManifest)) {
+			Launcher launcher = this.launcherRepository.findByName(platformName);
+			if(launcher.getType().equals(TaskPlatformFactory.CLOUDFOUNDRY_PLATFORM_TYPE) && !isAppDeploymentSame(previousManifest, taskManifest)) {
 				validateAndLockUpgrade(taskName, platformName, taskExecution);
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
```diff
@@ -270,7 +270,8 @@ public long executeTask(String taskName, Map<String, String> taskDeploymentPrope
 		String taskDeploymentId = null;
 		try {
 			TaskLauncher taskLauncher = findTaskLauncher(platformName);
-			if(!isAppDeploymentSame(previousManifest, taskManifest)) {
+			Launcher launcher = this.launcherRepository.findByName(platformName);
```
mminella: Why do we need to get the launcher again (we just got it the line above)?

cppwfs: Different launcher ;-).

PR #3744, opened 2020-01-17
 Port the AT schedule tests to the IT test suite

### LINKED ISSUE #3737: Port the AT schedule tests to the IT test suite
With the ability to run the IT test against CF, K8S (same as ATs) we can port the scheduler tests as well.

tzolov:
- Add Task Schedule DSL (under test utils)
 - Port the Schedule ATs into TaskScheduleIT. Later is run only for K8s or CF ATs and ignored for Docker Compose installation.
- Pot the Stream Confg Server test (run only on CF).
- Rename DockerCompseIT to DataFlowIT.
- Improve the logging info.
    
Resolves #3737

### DIFF
--- a/spring-cloud-dataflow-server/src/test/java/org/springframework/cloud/dataflow/integration/test/DataFlowIT.java
+++ b/spring-cloud-dataflow-server/src/test/java/org/springframework/cloud/dataflow/integration/test/DataFlowIT.java
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package org.springframework.cloud.dataflow.integration.test;
 
 import java.io.IOException;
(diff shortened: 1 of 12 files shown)


jvalkeal:
Think this looks ok. What or how was a plan to run those task tests on k8s/cf?

tzolov:
@jvalkeal, i don't think there is a plan yet. Different approaches can be considered. 
Currently the ATs is responsible to provision the target platform, run the AT tests and then destroy the target platform.  
The new ITs tests by default run locally (with docker-compose installation), but can be run against remote k8s or CF platforms (effectively doing AT).  Therefore the remaining purpose of the ATs scripts is to provision and to destroy the target environments. 
An easy hack can be to deprecate the java test code in the AT repo and on AT test run check out the IT test code instead.  But we need to rethink the entire lifecycle, as i believe there is better way to manage it.

For my ITs testing i installed SCDF on Hydra manually.

ilayaperumalg:
LGTM, merging.

PR #3753, opened 2020-01-23
Add k8s template files for build automation

### LINKED ISSUE #3752: Template k8s deployment files
Current problem is that during a release build `src/kubernetes/server/server-deployment.yaml` needs to have a version which doesn't yet exist. We should template these files so that when execution a maven build(probably via profile or similar) so that all versions changing in k8s files would get automatically replaced and committed back to git with a tag.

i.e. template file would be `src/templates/kubernetes/server/server-deployment.yaml` with version as a placeholder. When modifying these files we would just need to remember to modify template and then run specific part of a build if wanting to modify snapshots on github. Release build in CI would then do these automatically based on properties set in pom files.

jvalkeal:
- Now autogenerating `src/kubernetes` from `src/templates/kubernetes`
  by replacing property placeholders in those templates files.
- Generation activated by `deploymentfiles` profile during a
  `copy-resources` phase.
- This commit changed skipper-deployment.yaml as version there didn't
  match one from pom when I ran `mvnw process-resources -P deploymentfiles`.
- Fixes #3752

### DIFF
--- a/src/templates/kubernetes/kafka/kafka-deployment.yaml
+++ b/src/templates/kubernetes/kafka/kafka-deployment.yaml
@@ -1,0 +1,40 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: kafka-broker
+  labels:
+    app: kafka
+    component: kafka-broker
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: kafka
+  template:
+    metadata:
+      labels:
+        app: kafka
+        component: kafka-broker
+    spec:
+      containers:
+      - name: kafka
+        image: wurstmeister/kafka:2.12-2.3.0
+        ports:
+        - containerPort: 9092
+        env:
+          - name: ENABLE_AUTO_EXTEND
+            value: "true"
+          - name: KAFKA_RESERVED_BROKER_MAX_ID
+            value: "999999999"
+          - name: KAFKA_AUTO_CREATE_TOPICS_ENABLE
+            value: "false"
+          - name: KAFKA_PORT
+            value: "9092"
+          - name: KAFKA_ADVERTISED_PORT
+            value: "9092"
+          - name: KAFKA_ADVERTISED_HOST_NAME
+            valueFrom:
+              fieldRef:
+                fieldPath: status.podIP
+          - name: KAFKA_ZOOKEEPER_CONNECT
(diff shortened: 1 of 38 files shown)


src/kubernetes/skipper/skipper-deployment.yaml
```diff
@@ -16,7 +16,7 @@ spec:
     spec:
       containers:
       - name: skipper
-        image: springcloud/spring-cloud-skipper-server:2.2.1.BUILD-SNAPSHOT
+        image: springcloud/spring-cloud-skipper-server:2.3.0.BUILD-SNAPSHOT
```
ilayaperumalg: Is it possible to inject the ${skipper.version} defined from the POM file into the template files?

src/templates/kubernetes/kafka/kafka-deployment.yaml
```diff
@@ -0,0 +1,40 @@
...
+      app: kafka
+  template:
+    metadata:
+      labels:
+        app: kafka
+        component: kafka-broker
+    spec:
+      containers:
+      - name: kafka
+        image: wurstmeister/kafka:2.12-2.3.0
```
ilayaperumalg: At some point of time, we might want to move all these explicit versioning into configuration parameters.

src/kubernetes/skipper/skipper-deployment.yaml
jvalkeal: That's exactly what happens when this file is generated from a template. This change happened as pom is using skipper 2.3.0.BUILD-SNAPSHOT

src/templates/kubernetes/kafka/kafka-deployment.yaml
jvalkeal: Yeah true, I just touched those parts which always need a change during a build.

src/kubernetes/skipper/skipper-deployment.yaml
ilayaperumalg: oh, I see. I get it now.

src/templates/kubernetes/kafka/kafka-deployment.yaml
ilayaperumalg: ok, we can revisit in an iteration.

ilayaperumalg:
LGTM, merging.

PR #3759, opened 2020-01-24
Exclude jobExecution on client side

### LINKED ISSUE #3755: StepExecutionJacksonMixIn mixed with rest client
When upgrading boot which uses hateoas 1.0.2 causes configuration/beans getting created in a slightly different order and thus revealing this issue with `JobCommandTests` where mixins are replaced with a ones from client side.

https://github.com/spring-cloud/spring-cloud-dataflow/blob/878bdbb9944b965bb5fea21fc8afc1a48b17fbcb/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/DataFlowTemplate.java#L275-L286

Need to think how to fix this. Either use same mixins on both sides or rely on different object mappers(this is pita).

jvalkeal:
- To overcome future errors in `JobCommandTests` exclude `jobExecution` if
  client side `StepExecutionJacksonMixIn` is used as what happens with those
  tests when hateoas 1.0.2 is brought in.
- Issue will happen with real server/shell but we can't do more clever
  things until shell is fully upgraded to boot 2.x.
- Fixes #3755


ilayaperumalg:
LGTM

PR #3763, opened 2020-01-29
Resolves https://github.com/spring-io/dataflow.spring.io/issues/181

tzolov:
Resolves https://github.com/spring-io/dataflow.spring.io/issues/181

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
@@ -474,6 +474,10 @@ Make sure to add 'spring-boot-configuration-processor' as an optional dependency
     <optional>true</optional>
 </dependency>
 ----
+
+The whitelisting support works only for uber-jar application artifacts! At the moment, the metadata properties are not retrievable from the dockerized application images directly — a dedicated companion metadata JAR is required.
+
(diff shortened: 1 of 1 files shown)


spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -474,6 +474,10 @@ Make sure to add 'spring-boot-configuration-processor' as an optional dependency
     <optional>true</optional>
 </dependency>
 ----
+
+The whitelisting works only with jar app artifacts! At the moment metadata is not retrievable from the App Docker Images. For later a dedicated metadata artifact is required.
```
sabbyanandan:
```suggestion
The whitelisting support works only for uber-jar application artifacts! At the moment, the metadata properties are not retrievable from the dockerized application images directly — a dedicated companion metadata JAR is required.
```

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -474,6 +474,10 @@ Make sure to add 'spring-boot-configuration-processor' as an optional dependency
...
+
+The `configuration-properties.names` can white list only properties defined in java property class annotated with `@ConfigurationProperties`. The `@Value` defined properties can not be withe listed!
```
sabbyanandan:
```suggestion
The `configuration-properties.names` can whitelist properties that are defined in the property class annotated with `@ConfigurationProperties`. The properties defined with `@Value` annotations cannot be whitelisted.
```

PR #3772, opened 2020-02-05
Clean build

### LINKED ISSUE #3739: Fix build warnings and errors
Current build outputs a lot of warnings and errors so need to polish this a little.

jvalkeal:
- Refactor parts of a build files to remove
  errors and some warnings.
- For @NotNull warning we just import findbugs as
  provided dependency as that's currently what
  framework recommends.
- Polish javadocs so that we can at some point
  start failing build for incomplete javadocs, etc.
- Fixes #3739

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -89,6 +89,7 @@
 		<codearte-props2yml.version>0.5</codearte-props2yml.version>
 		<json-unit.version>2.11.1</json-unit.version>
 		<docker-compose-rule.version>1.4.2</docker-compose-rule.version>
+		<findbugs.version>3.0.2</findbugs.version>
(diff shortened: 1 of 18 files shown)


pom.xml
```diff
@@ -334,22 +339,6 @@
 					<artifactId>maven-source-plugin</artifactId>
 					<version>3.0.1</version>
 				</plugin>
-				<plugin>
```
ilayaperumalg: Does this mean we don't use this plugin at all?

jvalkeal: It's already in a build pom and these re-definitions just cause downgrades of versions.

PR #3775, opened 2020-02-05
Cf deployer issue 335

dturanski:
Depends on https://github.com/spring-cloud/spring-cloud-deployer-cloudfoundry/pull/336 and Fixes https://github.com/spring-cloud/spring-cloud-deployer-cloudfoundry/issues/335

Since SCDF configures the TaskPlatform, this fix requires changes to `CloudFoundryTaskPlatformFactory` and corresponding changes to tests for the CF task platform. 

It would be difficult to automate integration tests for this.  I verified the fix manually using one of our AT environments and an account that has access to multiple orgs and spaces.  I deployed this branch of dataflow server to two different orgs and launched tasks in one while monitoring `tasks/executions/current`  using `watch curl ...` on both dataflow servers.  The instance that launched the task shows 1 task running, while the instance in the other space/org, correctly,  continues to show 0 tasks running. See attached (the missing url error due to an extra space, but the command still worked) 
Screen Shot 2020-02-05 at 10 01 24 AM

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -60,8 +60,10 @@
 		<spring-cloud-deployer.version>2.3.0.BUILD-SNAPSHOT</spring-cloud-deployer.version>
 		<spring-cloud-deployer-local.version>2.3.0.BUILD-SNAPSHOT</spring-cloud-deployer-local.version>
 
+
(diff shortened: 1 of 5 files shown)


jvalkeal:
Squashed and merged

PR #3777, opened 2020-02-05
Checks to make sure any CTR split does not exceed maxConcurrentCount

### LINKED ISSUE #3630: Enhance the ability to control concurrent CTR launches
From @TimoSchaz at https://github.com/spring-cloud-task-app-starters/composed-task-runner/issues/100:

```
When setting the limit of tasks to execute. The composedtaskrunner, stops launching tasks in his task execution workflow when the task limit is reached. This causes problems, because its unpredictable at which time the limit is reached when multiple composed tasks are executed.

Spring Cloud Dataflow should support a limit of concurrent ComposedTasks to solve this problem.
```

cppwfs:
resolves #3630

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
@@ -42,6 +42,11 @@ import org.springframework.cloud.dataflow.core.TaskDefinition;
 import org.springframework.cloud.dataflow.core.TaskDeployment;
 import org.springframework.cloud.dataflow.core.TaskManifest;
 import org.springframework.cloud.dataflow.core.TaskPlatformFactory;
+import org.springframework.cloud.dataflow.core.dsl.SplitNode;
+import org.springframework.cloud.dataflow.core.dsl.TaskAppNode;
+import org.springframework.cloud.dataflow.core.dsl.TaskNode;
+import org.springframework.cloud.dataflow.core.dsl.TaskParser;
+import org.springframework.cloud.dataflow.core.dsl.TaskVisitor;
 import org.springframework.cloud.dataflow.rest.util.ArgumentSanitizer;
 import org.springframework.cloud.dataflow.rest.util.DeploymentPropertiesUtils;
 import org.springframework.cloud.dataflow.server.job.LauncherRepository;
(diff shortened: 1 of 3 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
```diff
@@ -696,4 +704,50 @@ private void cancelTaskExecution(TaskExecution taskExecution, String platformNam
...
+		if(composedRunnerVisitor.highCount > taskLauncher.getMaximumConcurrentTasks()) {
+			throw new IllegalArgumentException(String.format("One or more of the " +
+					"splits in the composed task contains a task count that exceeds " +
+					"the maximumConcurrentTasks count of %s",
+					taskLauncher.getMaximumConcurrentTasks()));
+		}
+	}
+
+
+	private class ComposedRunnerVisitor extends TaskVisitor {
```
ilayaperumalg: Is it possible to move this logic into `spring-cloud-dataflow-core-dsl`?

cppwfs: That is a good question.  I don't think `spring-cloud-dataflow-core-dsl` is a good home for this logic.    This module contains a parser and as thus as a parser it provides syntax checking and the ability for users to utilize its visitor callbacks.    The parser doesn't have knowledge of the rest of SCDF or any restrictions that SCDF applies to itself, i.e.  maximum number of concurrrent tasks.   Also this check  is not a syntax check, its more of a runtime check.  
Thoughts?

ilayaperumalg: I agree this is more of a runtime check but the way I see is that the `ComposedRunnerVisitor` acts just as an extension of the actual `TaskVisitor` with the `high` and `current` counts which only carries the abstract counts on the split and after all this moves into the data flow DSL anyway ;) 

Also, I meant to move just this `ComposedRunnerVisitor` not the other logic associated with it.
Perhaps we could rename the ComposedRunnerVisitor to ComposedTaskRunnerVisitor as well.

cppwfs: but but but... I thought I won that argument :-D
Sounds good !

ilayaperumalg: The game is still on ;)  do you think any other impact on moving this into `spring-cloud-dataflow-core-dsl` ?

cppwfs: Updated the PR based on your recommendation.  It was a good one!

ilayaperumalg: Thank you @cppwfs. Looks good to me. Merging.

PR #3791, opened 2020-02-10
Add Docker Compose Host File System Access file

### LINKED ISSUE #3790: Automate the mounting of Host File Systems with docker-compose
With Docker Compose SCDF installation, if you need to  register apps form the local file system, currently you have to modify the original `docker-compose.yml` to mount the desired host volumes to the `dataflow-server` and `skipper` containers. And then once you're done you have to revert those changes!  While the procedure is [well documented](https://dataflow.spring.io/docs/installation/local/docker-customize/#accessing-the-host-file-system) it is rather inefficient as it repetitively requires you to apply (usually the same) changes  and then revert them.  

A better approach is to automate the process with OOTB docker-compose file like this: 
```
HOST_MOUNT_PATH=~/.m2 \
DOCKER_MOUNT_PATH=/root/.m2/ \
docker-compose -f ./docker-compose.yml -f ./docker-compose-mount-host-folder up
```
The optional `HOST_MOUNT_PATH` and `DOCKER_MOUNT_PATH` variable are used to set the host and docker mount folders.  Above would mount the HOST `~/.m2` folder to the `dataflow` and `skipper` folder called `/root/.m2`. 

* `HOST_MOUNT_PATH` defaults to the local host folder (e.gl. `.`)  where the docker compose is being started. This is a safe default that works on Linux, OSX and Windows!
* `DOCKER_MOUNT_PATH` defaults to `/root/scdf` on `dataflow-server` and `skipper` containers.

tzolov:
Resolves #3790


spring-cloud-dataflow-server/docker-compose.yml
```diff
@@ -1,5 +1,18 @@
 version: '3'
 
+# Configuration environment variables:
+# - DATAFLOW_VERSION and SKIPPER_VERSION specify what DataFlow and Skipper image versions use.
```
ilayaperumalg: use -> used

ilayaperumalg: perhaps, use -> to use

spring-cloud-dataflow-server/docker-compose.yml
```diff
@@ -1,5 +1,18 @@
...
+# - STREAM_APPS_URI and TASK_APPS_URI are used to specify what Stream and Task applications to pre-registered.
```
ilayaperumalg: pre-registered -> pre-register

ilayaperumalg:
LGTM, will merge with the typos fixed.

spring-cloud-dataflow-server/docker-compose.yml
ilayaperumalg:
```suggestion
# - DATAFLOW_VERSION and SKIPPER_VERSION specify what DataFlow and Skipper image versions to use.
```

spring-cloud-dataflow-server/docker-compose.yml
ilayaperumalg:
```suggestion
# - STREAM_APPS_URI and TASK_APPS_URI are used to specify what Stream and Task applications to pre-register.
```

PR #3802, opened 2020-02-23
Use reactive methods to get stream statuses

### LINKED ISSUE #3799: Support reactive api's for non-blocking communication
This is a long running epic to overhaul various external and internal api's to support non-blocking communication to provide better performance throughout a stack. While we've not really needed to use reactive api's, work started for spring-cloud/spring-cloud-skipper#939 showed that easiest way to provide performance enhancements is to use reactor through calls as cf-java-client itself is based on reactor.

Thus, we need to start adding various new api's for dataflow/skipper/deployers so that we can do same work with less traffic and roundabouts, and not trying to be sequential.

jvalkeal:
- StreamDeployer has a new method to ask multiple statuses at a same time
  which allows use to make only one call to skipper.
- /runtime/streams has been changed to use new method in a StreamDeployer
  to use only one request.
- SkipperStreamDeployer uses new method from SkipperClient to request
  all streams at a same time.
- Fixes #3799


jvalkeal:
closing to create against correct issue.

PR #3820, opened 2020-03-05
Make runtime app instance faster

jvalkeal:
- When using `runtime apps --appId xxx` it effectively
  on a server side iterates all apps to find a matching
  deploymentId and then returns AppStatus for that. This
  is painfully slow.
- To make this command even slightly usefull, replace backend
  to use release information from a database and return
  AppStatus from there for matching deploymentId.


ilayaperumalg:
LGTM, merging.

PR #3835, opened 2020-03-17
Verify tasks running with TaskLauncher

### LINKED ISSUE #3733: Task launch fails due to currently running applications even when app is not running
**Description:**
When a task fails internally, the TASK_EXECUTION table is not updated to show the task completed. SCDF thinks the task is running even if the app/pod is deleted.  It would be more accurate to use the TaskLauncher to query the platform to see what is actually running. 

Related to #3695

dturanski:
Fixes #3733


markpollack:
JobCommandTests is failing

dturanski:
I saw that but cannot reproduce locally.

markpollack:
I had my build fail but then pass, not sure what is going on - probably have a lurking issue with dependency test ordering in the package where JobCommandTests is located.  Merged as 91690e47ae979cc5f3f98d0bfa0cdc7f2faac8ad

PR #3846, opened 2020-03-24
SCDF to use Scheduler behavior form version 2.2.x

### LINKED ISSUE #3483: Restarting failed job which uses incrementer creates new JobInstance

cppwfs:
Removes the use of the SchedulerTaskLauncher and sets the scheduling policy to that from 2.2.x.

resolves #3483


ilayaperumalg:
With the changes in place, I believe it now makes sense to bring back https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/issues/331 to allow deployment properties options for scheduled jobs.

Also, I'd like to know what happens to the schedules which were part of post 2.2.x (with the custom scheduler name with the task definition name attached) and how do we port them into the current setup?

cppwfs:
1) I completely agree on reopening #331
2) Migration tool story
3) Update the Acceptance test story
4) Update dataflow.spring.io docs as well.

chrisjs:
LGTM - testing shows pods being created in the normal way

@ilayaperumalg adding you here for any feedback and @cppwfs mentioned there may need to be something done around the TL module

ilayaperumalg:
The changes LGTM. Also, verified on minikube by launching/scheduling tasks with the arguments and properties passed. All look good.

ilayaperumalg:
Merging.

PR #3859, opened 2020-03-31
Updated doc-resources-version and highlight

Buzzardo:
Changed the doc-resources version to get Damien's latest look and feel.
Also changed the highlight to Github, as Damien suggests.


oodamien:
Look good! Thanks Jay

Issue #3880, opened 2020-04-12
Handle authentication to retrieve metadata/label info from the Docker images


sabbyanandan:

chrisjs:
the two related issues are closed so doing so here

PR #3885, opened 2020-04-15
Improve k8s metadata retrieval

tzolov:
Resolves https://github.com/spring-cloud/spring-cloud-dataflow/issues/3880

### DIFF
--- a/spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/ApplicationConfigurationMetadataResolverAutoConfiguration.java
+++ b/spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/ApplicationConfigurationMetadataResolverAutoConfiguration.java
@@ -20,7 +20,11 @@ import java.security.KeyManagementException;
 import java.security.NoSuchAlgorithmException;
 import java.security.cert.X509Certificate;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.TrustManager;
@@ -29,23 +33,29 @@ import javax.net.ssl.X509TrustManager;
 import org.apache.http.conn.ssl.NoopHostnameVerifier;
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.HttpClients;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.web.client.RestTemplateBuilder;
 import org.springframework.cloud.dataflow.configuration.metadata.container.ContainerImageMetadataProperties;
 import org.springframework.cloud.dataflow.configuration.metadata.container.ContainerImageMetadataResolver;
 import org.springframework.cloud.dataflow.configuration.metadata.container.ContainerImageParser;
 import org.springframework.cloud.dataflow.configuration.metadata.container.DefaultContainerImageMetadataResolver;
+import org.springframework.cloud.dataflow.configuration.metadata.container.RegistryConfiguration;
 import org.springframework.cloud.dataflow.configuration.metadata.container.authorization.AwsEcrAuthorizer;
 import org.springframework.cloud.dataflow.configuration.metadata.container.authorization.BasicAuthRegistryAuthorizer;
 import org.springframework.cloud.dataflow.configuration.metadata.container.authorization.DockerHubRegistryAuthorizer;
 import org.springframework.cloud.dataflow.configuration.metadata.container.authorization.RegistryAuthorizer;
+import org.springframework.cloud.dataflow.configuration.metadata.container.authorization.SecretToRegistryConfigurationConverter;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.http.MediaType;
 import org.springframework.http.client.HttpComponentsClientHttpRequestFactory;
 import org.springframework.http.converter.StringHttpMessageConverter;
+import org.springframework.util.StringUtils;
 import org.springframework.validation.annotation.Validated;
 import org.springframework.web.client.RestTemplate;
 
@@ -57,6 +67,8 @@ import org.springframework.web.client.RestTemplate;
  */
 @Configuration
 public class ApplicationConfigurationMetadataResolverAutoConfiguration {
+
+	private static final Logger logger = LoggerFactory.getLogger(ApplicationConfigurationMetadataResolverAutoConfiguration.class);
 
 	@Bean
 	public RegistryAuthorizer dockerHubRegistryAuthorizer() {
@@ -89,9 +101,11 @@ public class ApplicationConfigurationMetadataResolverAutoConfiguration {
 	@ConditionalOnMissingBean(ContainerImageMetadataResolver.class)
 	public DefaultContainerImageMetadataResolver containerImageMetadataResolver(
 			@Qualifier("metadataRestTemplate") RestTemplate metadataRestTemplate,
-			ContainerImageParser imageNameParser, List<RegistryAuthorizer> registryAuthorizers,
-			ContainerImageMetadataProperties properties) {
-		return new DefaultContainerImageMetadataResolver(metadataRestTemplate, imageNameParser, registryAuthorizers, properties);
+			ContainerImageParser imageNameParser,
+			Map<String, RegistryConfiguration> registryConfigurationMap,
+			List<RegistryAuthorizer> registryAuthorizers) {
+		return new DefaultContainerImageMetadataResolver(metadataRestTemplate, imageNameParser,
+				registryConfigurationMap, registryAuthorizers);
 	}
 
 	@Bean
@@ -143,5 +157,31 @@ public class ApplicationConfigurationMetadataResolverAutoConfiguration {
 		return builder.requestFactory(() -> customRequestFactory)
 				.additionalMessageConverters(octetToStringMessageConverter).build();
 
+	}
+
+	@Value("${.dockerconfigjson}")
+	public String dockerconfigjson;
+
+	@Bean
+	public Map<String, RegistryConfiguration> registryConfigurationMap(ContainerImageMetadataProperties properties) {
+
+		logger.info("dockerconfigjson  >>>> :" + dockerconfigjson);
+
+		// Retrieve declared registries.
+		Map<String, RegistryConfiguration> propertiesRegistryConfigurationMap = properties.getRegistryConfigurations().stream()
+				.collect(Collectors.toMap(RegistryConfiguration::getRegistryHost, Function.identity()));
+
+		HashMap<String, RegistryConfiguration> registryConfigurationMap = new HashMap<>(propertiesRegistryConfigurationMap);
+
+		if (!StringUtils.isEmpty(this.dockerconfigjson)) {
+			Map<String, RegistryConfiguration> secretsRegistryConfigurationMap
+					= new SecretToRegistryConfigurationConverter().convert(this.dockerconfigjson);
+			logger.info("secretsRegistryConfigurationMap: " + secretsRegistryConfigurationMap);
+			registryConfigurationMap.putAll(secretsRegistryConfigurationMap);
+		}
+
+		System.out.println(registryConfigurationMap);
+
+		return registryConfigurationMap;
 	}
 }
--- a/spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/ApplicationConfigurationMetadataResolverAutoConfiguration.java
+++ b/spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/ApplicationConfigurationMetadataResolverAutoConfiguration.java
@@ -20,11 +20,10 @@ import java.security.KeyManagementException;
 import java.security.NoSuchAlgorithmException;
 import java.security.cert.X509Certificate;
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.function.Function;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.TrustManager;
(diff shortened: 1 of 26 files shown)


spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/ApplicationConfigurationMetadataResolverAutoConfiguration.java
```diff
@@ -80,4 +114,74 @@ public ApplicationConfigurationMetadataResolver metadataResolver(
...
+
+	}
+
+	@Value("${.dockerconfigjson}")
+	public String dockerconfigjson;
+
+	@Bean
+	public Map<String, RegistryConfiguration> registryConfigurationMap(ContainerImageMetadataProperties properties) {
+
+		logger.info("dockerconfigjson  >>>> :" + dockerconfigjson);
```
chrisjs, line 168: this may contain base64 encoded things one may not want to have in logs

spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/container/authorization/SecretToRegistryConfigurationConverter.java
```diff
@@ -0,0 +1,136 @@
...
+				String tokenAccessUrl = determineAuthorizationType(rc.getRegistryHost(), rc.getUser(), rc.getSecret());
+				if (StringUtils.isEmpty(tokenAccessUrl)) {
+					rc.setAuthorizationType(RegistryConfiguration.AuthorizationType.basicauth);
+				}
+				else {
+					rc.setAuthorizationType(RegistryConfiguration.AuthorizationType.dockerhub);
+					rc.getExtra().put("registryAuthUri", tokenAccessUrl);
+				}
+
+				logger.info("Registry Secret: " + rc.toString());
```
chrisjs, line 74: might not be good to log this data

src/kubernetes/server/server-config.yaml
```diff
@@ -37,8 +37,19 @@ data:
               kubernetes:
                 accounts:
                   default:
+                    imagePullSecret: harborsec
```
chrisjs, line 40: understood "do not merge", noting for updating -- probably want to remove this. we may consider "imagePullSecrets" if theres a usecase for multiple

src/kubernetes/server/server-config.yaml
```diff
@@ -37,8 +37,19 @@ data:
...
                     limits:
                       memory: 1024Mi
+          container:
+            metadata:
+              disableSslVerification: true
+              registry-configurations:
+                - registry-host: 'myregistrydomain.com:5000'
+                  authorization-type: basicauth
+                  user: testuser
```
chrisjs, line 49: can user just be rolled into the secret given its part of creds?

src/kubernetes/server/server-deployment.yaml
```diff
@@ -17,7 +17,14 @@ spec:
       containers:
       - name: scdf-server
         image: springcloud/spring-cloud-dataflow-server:2.5.0.BUILD-SNAPSHOT
-        imagePullPolicy: Always
+        imagePullPolicy: IfNotPresent
+        volumeMounts:
+          - name: database
+            mountPath: /etc/secrets/database
+            readOnly: true
+          - name: harborsec
```
chrisjs, line 25: given do not merge, assuming this and the below refs to "harborsec" wouldn't be included, yet a doc thing on how to add mounts

spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/ApplicationConfigurationMetadataResolverAutoConfiguration.java
tzolov: Right it does and they are even base64 decoded :) will remove all log entires  that may revel registry configurations and secrets.

spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/container/authorization/SecretToRegistryConfigurationConverter.java
tzolov: I've changed the RegistryConfiguration's `toString()` implementation to prevent printing the secrets (e.g. password). 
IMO listing the provided and retrieved registry configurations is useful.  If necessary we can strip back the listed information even further, showing only the registry host uri and the authorization type?

src/kubernetes/server/server-config.yaml
```diff
@@ -37,8 +37,19 @@ data:
               kubernetes:
                 accounts:
                   default:
+                    imagePullSecret: harborsec
                     limits:
                       memory: 1024Mi
+          container:
```
markpollack, line 43: Hi,
Not related to the scope of this PR, but I think that this configuration shouldn't live under 

```
    spring:
      cloud:
        dataflow:
          container:
            metadata:
              disableSslVerification: true
              registry-configurations:
                - registry-host: 'myregistrydomain.com:5000'
```

because we can use the connectivity to a container registry for purposes other than metadata retrieval, for example we could automatically 'import' the images that are under a specific namespace in the registry.

What about   `spring.cloud.dataflow.registries` or `spring.cloud.dataflow.containerRegistries` and under that is the collection that has `registry-host`.  Also, on a per-registry level we should be able to disableSslVerification, so move that down into the list of registries.

src/kubernetes/server/server-config.yaml
```diff
@@ -37,8 +37,19 @@ data:
...
+            metadata:
+              disableSslVerification: true
+              registry-configurations:
+                - registry-host: 'myregistrydomain.com:5000'
+                  authorization-type: basicauth
+                  user: testuser
+                  secret: testpassword
+                - registry-host: registry-1.docker.io
+                  authorization-type: dockerhub
```
markpollack, line 52: Another naming question, why call this `dockerhub` - as there are so many registries out there giving it a product specific name doesn't seem correct.

spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/ApplicationConfigurationMetadataResolverAutoConfiguration.java
```diff
@@ -80,4 +114,74 @@ public ApplicationConfigurationMetadataResolver metadataResolver(
...
+		mediaTypeList.add(MediaType.APPLICATION_OCTET_STREAM);
+		octetToStringMessageConverter.setSupportedMediaTypes(mediaTypeList);
+
+		if (!properties.isDisableSslVerification()) {
+			return builder.additionalMessageConverters(octetToStringMessageConverter).build();
+		}
+
+		// From https://www.javacodemonk.com/disable-ssl-certificate-check-resttemplate-e2c53583
+
+		// Trust manager that blindly trusts all SSL certificates.
```
markpollack, line 134: This doesn't feel right...

spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/ApplicationConfigurationMetadataResolverAutoConfiguration.java
```diff
@@ -80,4 +114,74 @@ public ApplicationConfigurationMetadataResolver metadataResolver(
...
+				.setSSLHostnameVerifier(NoopHostnameVerifier.INSTANCE).build();
+		HttpComponentsClientHttpRequestFactory customRequestFactory = new HttpComponentsClientHttpRequestFactory();
+		customRequestFactory.setHttpClient(httpClient);
+		// Create a RestTemplate that uses custom request factory
+		return builder.requestFactory(() -> customRequestFactory)
+				.additionalMessageConverters(octetToStringMessageConverter).build();
+
+	}
+
+	@Value("${.dockerconfigjson}")
```
markpollack: where does this come from?

spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/ApplicationConfigurationMetadataResolverAutoConfiguration.java
```diff
@@ -80,4 +114,74 @@ public ApplicationConfigurationMetadataResolver metadataResolver(
...
+		HashMap<String, RegistryConfiguration> registryConfigurationMap = new HashMap<>(propertiesRegistryConfigurationMap);
+
+		if (!StringUtils.isEmpty(this.dockerconfigjson)) {
+			Map<String, RegistryConfiguration> secretsRegistryConfigurationMap
+					= new SecretToRegistryConfigurationConverter().convert(this.dockerconfigjson);
+			logger.info("secretsRegistryConfigurationMap: " + secretsRegistryConfigurationMap);
+			registryConfigurationMap.putAll(secretsRegistryConfigurationMap);
+		}
+
+		System.out.println(registryConfigurationMap);
```
markpollack: no system.out.println

src/kubernetes/server/server-config.yaml
tzolov: I agree. I prefer `spring.cloud.dataflow.container.registry` as property prefix. The Registry-Configuration is an inner [List<RegistryConfiguration>](https://github.com/tzolov/spring-cloud-dataflow/blob/improve-k8s-metadata-retrieval/spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/container/ContainerImageMetadataProperties.java#L58) property. So i don't think we can remove it. 

Also currently this functionality is implemented within the `spring-cloud-dataflow-configuration-metadata` project.  Perhaps it should be moved elsewhere. 

For the record, the following code cries to be replaced with the new registry access: https://github.com/spring-cloud/spring-cloud-dataflow/tree/master/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/validation
Also we should replace the existing `DockerImage` class by the more elaborate `ContainerImage`.

src/kubernetes/server/server-config.yaml
tzolov: I've renamed `dockerhub` to `dockerouaht2`

spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/ApplicationConfigurationMetadataResolverAutoConfiguration.java
tzolov: cleaned

spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/ApplicationConfigurationMetadataResolverAutoConfiguration.java
tzolov: this is the default key used by k8s configuration and documentation to add registry credentials.

markpollack:
Unrelated to the issue for the PR - but we should create another issue to create some unit tests for the RegistryAuthorizer implementations.

spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/ApplicationConfigurationMetadataResolverAutoConfiguration.java
```diff
@@ -80,4 +110,115 @@ public ApplicationConfigurationMetadataResolver metadataResolver(
...
+
+	@Bean
+	public DockerConfigJsonSecretToRegistryConfigurationConverter secretToRegistryConfigurationConverter(
+			@Qualifier("noSslVerificationContainerRestTemplate") RestTemplate trustAnySslRestTemplate) {
+		return new DockerConfigJsonSecretToRegistryConfigurationConverter(trustAnySslRestTemplate);
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public RestTemplate containerRestTemplate(RestTemplateBuilder builder) {
```
markpollack, line 181: The conditionalOnMissingBean works against the return type....do you want to test against the name of the bean as well?

spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/container/RegistryConfiguration.java
```diff
@@ -27,31 +27,31 @@
...
+ *    - spring.cloud.dataflow.container.registry-configurations[myamazonaws].secret=[your AWS secretKey]
+ *    - spring.cloud.dataflow.container.registry-configurations[myamazonaws].extra[region]=us-west-1
+ *    - spring.cloud.dataflow.container.registry-configurations[myamazonaws].extra[registryIds]=283191309520
  *
  *  Configure Azure private container registry
- *    - spring.cloud.dataflow.container.metadata.registry-configurations[2].registry-host=tzolovazureregistry.azurecr.io
- *    - spring.cloud.dataflow.container.metadata.registry-configurations[2].authorization-type=basicauth
- *    - spring.cloud.dataflow.container.metadata.registry-configurations[2].user=[your Azure registry username]
- *    - spring.cloud.dataflow.container.metadata.registry-configurations[2].secret=[your Azure registry access password]
+ *    - spring.cloud.dataflow.container.registry-configurations[tzolovazureregistry].registry-host=tzolovazureregistry.azurecr.io
```
chrisjs, line 44: should this be `myazureregistry` like `myamazonaws`?

spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/container/RegistryConfiguration.java
```diff
@@ -27,31 +27,31 @@
...
+ *    - spring.cloud.dataflow.container.registry-configurations[tzolovazureregistry].user=[your Azure registry username]
+ *    - spring.cloud.dataflow.container.registry-configurations[tzolovazureregistry].secret=[your Azure registry access password]
  *
  *  Harbor Registry. Same as DockerHub but with different registryAuthUri
- *    - spring.cloud.dataflow.container.metadata.registry-configurations[3].registry-host=demo.goharbor.io
- *    - spring.cloud.dataflow.container.metadata.registry-configurations[3].authorization-type=dockerhub
- *    - spring.cloud.dataflow.container.metadata.registry-configurations[3].user=admin
- *    - spring.cloud.dataflow.container.metadata.registry-configurations[3].secret=Harbor12345
- *    - spring.cloud.dataflow.container.metadata.registry-configurations[3].extra[registryAuthUri]=https://demo.goharbor.io/service/token?service=harbor-registry&scope=repository:{repository}:pull
+ *    - spring.cloud.dataflow.container.registry-configurations[goharbor].registry-host=demo.goharbor.io
```
chrisjs, line 50: `goharbor` -> `harbor` ?

src/kubernetes/server/server-deployment.yaml
```diff
@@ -18,6 +18,10 @@ spec:
       - name: scdf-server
         image: springcloud/spring-cloud-dataflow-server:2.5.0.BUILD-SNAPSHOT
         imagePullPolicy: Always
+        volumeMounts:
```
chrisjs, line 21: these changes need to go into `src/templates/kubernetes/server/server-deployment.yaml` (note the `templates` in the path) otherwise the build process won't get them /cc @jvalkeal

markpollack:
squashed and merged as fc0dcbf7883bb8fa1afc3de3f4c0ce903a200326

PR #3892, opened 2020-04-18
Symetric encrypt for spring.datasource.password in TaskDefinition

publinchi:
Symetric encrypt for spring.datasource.password value in json record, task_execution_metadata table, task_execution_manifest column, E.g.:

before:

{
    "platformName": "default",
    "taskDeploymentRequest": {
        "commandlineArguments": [
            "--spring.cloud.data.flow.platformname=default",
            "--spring.cloud.task.executionid=2"
        ],
        "definition": {
            "name": "timestamp-task",
            "properties": {
                "management.metrics.export.prometheus.enabled": "true",
                "management.metrics.export.prometheus.rsocket.enabled": "true",
                "management.metrics.export.prometheus.rsocket.host": "localhost",
                "management.metrics.export.prometheus.rsocket.port": "7001",
                "spring.cloud.task.name": "timestamp-task",
                "spring.datasource.driverClassName": "com.microsoft.sqlserver.jdbc.SQLServerDriver",
                "spring.datasource.password": "**P@55w0rd**",
                "spring.datasource.url": "jdbc:sqlserver://localhost:1433;databaseName=master",
                "spring.datasource.username": "sa"
            }
        },
        "deploymentProperties": {},
        "resource": "maven://org.springframework.cloud.task.app:timestamp-task:jar:2.1.1.RELEASE"
    }
}

after:

{
    "platformName": "default",
    "taskDeploymentRequest": {
        "commandlineArguments": [
            "--spring.cloud.data.flow.platformname=default",
            "--spring.cloud.task.executionid=2"
        ],
        "definition": {
            "name": "timestamp-task",
            "properties": {
                "management.metrics.export.prometheus.enabled": "true",
                "management.metrics.export.prometheus.rsocket.enabled": "true",
                "management.metrics.export.prometheus.rsocket.host": "localhost",
                "management.metrics.export.prometheus.rsocket.port": "7001",
                "spring.cloud.task.name": "timestamp-task",
                "spring.datasource.driverClassName": "com.microsoft.sqlserver.jdbc.SQLServerDriver",
                "spring.datasource.password": "**{cipher}bc65ae5a79c36d5cdaa2c5b124b85c7212c0a400641c47cb4e897e2d3c6321c8**",
                "spring.datasource.url": "jdbc:sqlserver://localhost:1433;databaseName=master",
                "spring.datasource.username": "sa"
            }
        },
        "deploymentProperties": {},
        "resource": "maven://org.springframework.cloud.task.app:timestamp-task:jar:2.1.1.RELEASE"
    }
}


mminella:
A couple thoughts here:
1. If we are going to encrypt these values, why limit to that specific value and instead use the same rules that Boot does for masking secrets (as defined here): https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/Sanitizer.java#L46
2. What is the expectation around where the `ENCRYPT_KEY` comes from? I know that Spring Cloud Config Server does something similar, but I want to make sure we have a clearly documented way to manage that key...

publinchi:
Hi Michael, before I start, I'm a fan of your work done for Spring Batch. Regarding point number 1, the problem is that the spring.datasource.password field in the json stored in the task_execution_metadata table, task_execution_manifest column, shows the credentials in clear. If someone with query access to the database can see this field, they will be able to know the credentials used by the transaction and be able to access the database with greater privileges. Which could be risky in environments with sensitive data. Based on your proposed solution in # 1, I see it's related to masking the values ​​in memory and not at the database level.

Regarding point 2, the idea is to encrypt the key from the command line, using spring cli, for which, I generate the encrypted key using:

$ spring encrypt CLEAR_PASS --key SOME_KEY

and then I set in an environment variable the ENCRYPT_KEY

$ export ENCRYPT_KEY=SOME_KEY

in order to decrypt the key at runtime and ensure that the task execution flow is successfully completed.


I'm sorry for my lousy English, but I hope I was clear.


Best Regards.

mminella:
@publinchi Thank you for your kind words.

For point one, I'm not saying we shouldn't encrypt, what I'm saying is that we shouldn't encrypt only that one property value. The manifest is generated as a `String` in places like [this](https://github.com/spring-cloud/spring-cloud-dataflow/blob/f15e50a4f8c2607492eab1d93f50d8087175dc60/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/JdbcDataflowTaskExecutionMetadataDao.java#L105), we should go through and encrypt all values that have `password`, `secret`, etc in the key name.  Basically, I'm saying do the same thing the Spring Boot `Sanatizer` does but instead of masking the value, encrypt it.

For point two, I think my main thing is that we've added the code to support a new configuration property but have not documented it anywhere. 

Related to point two, we also have no tests for this new functionality. Can you please add a test that validates the new functionality?

publinchi:
@mminella Ok, I will add the unit test to validate point 2, regards.

ilayaperumalg:
@publinchi Did you get a chance to work on this?

dturanski:
@publinchi - I'm not comfortable merging this as is. @mminella makes some good points. 
* This applies to other properties
* This feature needs to be documented
* The code should use the Spring environment abstraction instead of explicitly relying on an environment variable
* I don't see anything that the task would use to decrypt these values
* Needs tests

publinchi:
Hi, i need more time for this, sorry i wasted your time.

PR #3896, opened 2020-04-21
Fix for the container metadata SSL handling

tzolov:
- Fix SSL vs NON-SSL RestTemplate interaction.
 - Add Tests for the SSL and  NON-SSL paths

Related to #3880


markpollack:
rebased and merged as d264701771a34203331d41c82da7b1f07889b08a

PR #3907, opened 2020-04-24
Disable k8s api access for secrets and configmaps

### LINKED ISSUE #3869: Investigate and remove spring-cloud-kubernetes library from SCDF & Skipper

trisberg:
Resolves #3869

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/configuration-kubernetes.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/configuration-kubernetes.adoc
@@ -320,21 +320,7 @@ You can configure the Data Flow server that is on Kubernetes to deploy tasks to 
 [[configuration-kubernetes-general]]
 === General Configuration
 
-The Spring Cloud Data Flow server for Kubernetes uses the https://github.com/fabric8io/spring-cloud-kubernetes[`spring-cloud-kubernetes`] module process both the ConfigMap and the secrets settings. To enable the ConfigMap support, pass in an environment variable of `SPRING_CLOUD_KUBERNETES_CONFIG_NAME` and set it to the name of the ConfigMap. The same is true for the secrets, where the environment variable is `SPRING_CLOUD_KUBERNETES_SECRETS_NAME`. To use the secrets, you also need to set `SPRING_CLOUD_KUBERNETES_SECRETS_ENABLE_API` to `true`.
-
-The following example shows a snippet from a deployment script that sets these environment variables:
-
-
-[source,yaml]
-----
-env:
-- name: SPRING_CLOUD_KUBERNETES_SECRETS_ENABLE_API
-  value: 'true'
-- name: SPRING_CLOUD_KUBERNETES_SECRETS_NAME
-  value: mysql
-- name: SPRING_CLOUD_KUBERNETES_CONFIG_NAME
-  value: scdf-server
-----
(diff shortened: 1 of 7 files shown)


sabbyanandan:
@chrisjs: Would any of these changes apply to the helm-chart or whether if it needs to be synchronized also there?

tzolov:
Looks like this PR would affect (and avert) the way the Docker Registry Secrets?
Currently it is assumed that any secrete (including the docker registries such) mounted under the `/etc/secrets` is automatically picked for the purposes of the app metadata retrieving:
- [volume-mounted-secretes](https://dataflow.spring.io/docs/2.5.0.SNAPSHOT/installation/kubernetes/helm/#volume-mounted-secretes).
-  [SPRING_CLOUD_KUBERNETES_SECRETS_PATHS to  load automatically all secrets under the /etc/secrets](https://github.com/spring-cloud/spring-cloud-dataflow/commit/d31ed05052f6e39cc80907e666cf83925c9f568f) .

So, @trisberg , @chrisjs if this change is to go forward then we need to update the App Metadata documentation to explain how to list the Registry Secretes to initialize the app metadata resolver.

spring-cloud-dataflow-docs/src/main/asciidoc/configuration-kubernetes.adoc
```diff
@@ -320,21 +320,7 @@ You can configure the Data Flow server that is on Kubernetes to deploy tasks to
...
-----
-env:
-- name: SPRING_CLOUD_KUBERNETES_SECRETS_ENABLE_API
-  value: 'true'
-- name: SPRING_CLOUD_KUBERNETES_SECRETS_NAME
-  value: mysql
-- name: SPRING_CLOUD_KUBERNETES_CONFIG_NAME
-  value: scdf-server
-----
+The Spring Cloud Data Flow server for Kubernetes uses the https://github.com/fabric8io/spring-cloud-kubernetes[`spring-cloud-kubernetes`] module to process secrets that are mounted under `/etc/secrets`. ConfigMaps must be mounted as `application.yaml` in the `/config` directory that is processed by Spring Boot. To avoid access to the Kubernetes API server the `SPRING_CLOUD_KUBERNETES_CONFIG_ENABLE_API` and `SPRING_CLOUD_KUBERNETES_SECRETS_ENABLE_API` are set to `false`.
```
ilayaperumalg, line 323: Given these changes aim to remove the usage of `spring-cloud-kubernetes`, I wonder why do we still keep these `SPRING_CLOUD_KUBERNETES_*` settings and set them to `false` instead of removing them completely?

ilayaperumalg: or, maybe we still want this to be available until we completely remove the usage (on loading the secrets from mounts to env)?

trisberg: Yes, until Boot supports reading the mounted secrets we need to keep the `spring-cloud-kubernetes`  dependency just to read the secrets from the volume mount.

markpollack:
I suggest waiting to do this in the 2.6 series.

chrisjs:
> @chrisjs: Would any of these changes apply to the helm-chart or whether if it needs to be synchronized also there?

@sabbyanandan yes; for the most part any changes would be candidates

trisberg:
> Looks like this PR would affect (and avert) the way the Docker Registry Secrets?
> Currently it is assumed that any secrete (including the docker registries such) mounted under the `/etc/secrets` is automatically picked for the purposes of the app metadata retrieving:

I think we continue to load any secrets under `/etc/secrets` since we have this setting:

```
        - name: SPRING_CLOUD_KUBERNETES_SECRETS_PATHS
          value: /etc/secrets
```

@tzolov

chrisjs:
@trisberg is this good to be un-drafted?

chrisjs:
rebased, squashed and merged in df91ab22a5a05d9354b48617b011a873f9002721

PR #3910, opened 2020-04-25
Add liveness/readiness probes

trisberg:
Resolves: #3908

### DIFF
--- a/src/kubernetes/server/server-deployment.yaml
+++ b/src/kubernetes/server/server-deployment.yaml
@@ -24,6 +24,16 @@ spec:
             readOnly: true
         ports:
         - containerPort: 80
+        livenessProbe:
+          httpGet:
+            path: /management/health
+            port: 80
+          initialDelaySeconds: 45
+        readinessProbe:
+          httpGet:
+            path: /management/info
+            port: 80
+          initialDelaySeconds: 45
         resources:
           limits:
             cpu: 1.0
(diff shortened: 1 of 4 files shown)


src/kubernetes/server/server-deployment.yaml
```diff
@@ -24,6 +24,16 @@ spec:
             readOnly: true
         ports:
         - containerPort: 80
+        livenessProbe:
+          httpGet:
+            path: /management/health
+            port: 80
+          initialDelaySeconds: 45
```
sabbyanandan, line 31: Just for consistency, I wonder if the time-values can be the same as what we have as defaults in the [K8s-deployer](https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/blob/0cb8e688633faea9377f5068600b183e14b6e703/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java#L377-L401).

jvalkeal: Btw, these changes need to go under `src/templates/kubernetes`. `src/README.md` has notes about this structure.

jvalkeal: ah, never mind as template was also updated.

trisberg: > Just for consistency, I wonder if the time-values can be the same as what we have as defaults in the [K8s-deployer](https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/blob/0cb8e688633faea9377f5068600b183e14b6e703/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesDeployerProperties.java#L377-L401).

These apps take longer to start up than the stream apps, so we need to adjust for that

PR #3915, opened 2020-04-29
Add database schema instructions

### LINKED ISSUE #3480: Re-add a second `launch` method without alternateComposedTaskRunnerApp parameter
In `TaskOperations` we currently have:

```java
long launch(String name, Map<String, String> properties, List<String> arguments, String alternateComposedTaskRunnerApp);
```

This is a recent change and technically a breaking change of the Rest Client. We may want to consider re-adding a second `launch` method without the `alternateComposedTaskRunnerApp` parameter to remove a breaking change.

jvalkeal:
- Fixes #3480


src/db/db2/V1-dataflow.sql
```diff
@@ -0,0 +1,165 @@
+create sequence hibernate_sequence start with 1 increment by 1;
+
+create table app_registration (
```
ilayaperumalg, line 3: Do you think we should switch to a specific case (probably upper case) in these DDL scripts? I remember we had some case sensitive issues in the past.

jvalkeal: We can't as some db's will break. In fact it was task/batch which never should've use uppercase.

ilayaperumalg:
@jvalkeal Do we want to confirm the schema changes by running database-specific tests with these changes before merging this PR?

jvalkeal:
We don't have facilities(prob gonna look into it) to execute tests with these schema files and even if we had, we'd first need to merge these.

ilayaperumalg:
Merging the changes into 2.6.x development line as we'll explore ways to test multiple databases.

PR #3936, opened 2020-05-07
Bundle db ddl files

### LINKED ISSUE #3934: Bundle db ddl files
In #3915 we added ddl files to get run manually by dba and it'd be nice, and would ease testing if we'd bundle those files so that you could reference within classpath. For example we can already do:
```
java -jar spring-cloud-dataflow-server.jar \
  --spring.datasource.url='jdbc:postgresql://localhost:5432/dataflow' \
  --spring.datasource.username=spring \
  --spring.datasource.password=spring \
  --spring.datasource.driverClassName=org.postgresql.Driver \
  --spring.flyway.enabled=false \
  --spring.datasource.initializationMode=always \
  --spring.datasource.schema=file:src/db/postgresql/V1-dataflow.sql,file:src/db/postgresql/V1-skipper.sql,file:src/db/postgresql/V2-dataflow.sql,file:src/db/postgresql/V2-dataflow-after.sql,file:src/db/postgresql/V3-dataflow.sql
```
If you could reference to files by classpath then user could also create schema by starting dataflow with custom attributes or take files out from git and use those manually.

jvalkeal:
- Move all ddl files from src/db into spring-cloud-dataflow-server-core/src/main/resources/schemas/
  to get those packaged with an application to get used manually.
- Bundling these files serves two things, firstly user can bootstap server with a special configuration
  to execute these files during a startup, secondly this makes testing much easier.
- Add special application profile config files which makes it easier for some of these common settings
  if user is attempting to do these things with server bootstrap.
- All these files and profiles are not used at default, just to easy things up.
- spring-cloud-dataflow-server-core/src/main/resources/schemas/README.md contains more info
  about a use of these profiles.
- Fixes #3934


spring-cloud-dataflow-server-core/src/main/resources/schemas/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+For example if there is an existing PostgreSQL server running locally, starting a server
+with below command would automatically create needed schemas:
+
+    java -jar spring-cloud-dataflow-server.jar \
+      --spring.datasource.url='jdbc:postgresql://localhost:5432/dataflow' \
+      --spring.datasource.username=spring \
+      --spring.datasource.password=spring \
+      --spring.datasource.driverClassName=org.postgresql.Driver \
+      --spring.profiles.active=init-postgresql
```
ilayaperumalg, line 47: This is nice!

ilayaperumalg:
A comment outside the scope of this PR;

I think we may want to consider moving the database scripts outside of SCDF as that would help the DBAs to review and modify the DDLs based on their system requirements.

ilayaperumalg:
LGTM

ilayaperumalg:
Merging.

PR #3982, opened 2020-06-03
Uniquely identify the task elements within composed task

### LINKED ISSUE #3930: Error bulk deleting task Composed Tasks (task definition)
image

Reported by customer running SCDF for Tanzu 1.7  I have seen this in my test 1.7 environment as well.  Not sure what causes it.

### LINKED ISSUE #3933: Composed Task Runner: Consequent task launch fails
Since composed task is now launched as batch job, when launching after the first launch, the batch job fails with:

```
Caused by: org.springframework.batch.core.repository.JobInstanceAlreadyCompleteException: A job instance already exists and is complete for parameters={...}.  If you want to run this job again, change the parameters.

```

We need to handle this by introducing a custom logic inside the composed task runner to make the job instance unique.

ilayaperumalg:
- This is part of the changes required to fix #3930 to classify the task definitions that are part of a composed task definition.
 - Add a new property `spring.cloud.dataflow.task.composed-task-element` to identify the tasks within composed tasks
   - Set this property to `true` when creating the tasks of a CTR task
 - Set `increment-instance-enabled=true` for CTR to resolve #3933
 - Add test to verify the changes

Resolves #3930
Resolves #3933

### DIFF
--- a/spring-cloud-dataflow-composed-task-runner/src/main/java/org/springframework/cloud/dataflow/composedtaskrunner/properties/ComposedTaskProperties.java
+++ b/spring-cloud-dataflow-composed-task-runner/src/main/java/org/springframework/cloud/dataflow/composedtaskrunner/properties/ComposedTaskProperties.java
@@ -164,7 +164,7 @@ public class ComposedTaskProperties {
 	 * ComposedTaskRunner instance can only be executed once with a given set
 	 * of parameters, if true it can be re-executed.
 	 */
-	private boolean incrementInstanceEnabled = false;
+	private boolean incrementInstanceEnabled = true;
 
 	/**
 	 * The platform property that will be used for each task in the workflow when it is launched.
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/TaskConfigurationProperties.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/TaskConfigurationProperties.java
@@ -30,10 +30,12 @@ import org.springframework.validation.annotation.Validated;
  * @author David Turanski
  */
 @Validated
-@ConfigurationProperties(prefix = TaskConfigurationProperties.COMPOSED_TASK_PREFIX)
+@ConfigurationProperties(prefix = TaskConfigurationProperties.TASK_PREFIX)
 public class TaskConfigurationProperties {
 
-	public static final String COMPOSED_TASK_PREFIX = DataFlowPropertyKeys.PREFIX + "task";
+	public static final String TASK_PREFIX = DataFlowPropertyKeys.PREFIX + "task";
+
+	public static final String COMPOSED_TASK_ELEMENT = TASK_PREFIX + ".composed-task-element";
 
 	public static final String COMPOSED_TASK_RUNNER_NAME = "composed-task-runner";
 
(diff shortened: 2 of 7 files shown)


cppwfs:
we will need a migration script for existing composed tasks.

spring-cloud-dataflow-composed-task-runner/src/main/java/org/springframework/cloud/dataflow/composedtaskrunner/properties/ComposedTaskProperties.java
```diff
@@ -164,7 +164,7 @@
 	 * ComposedTaskRunner instance can only be executed once with a given set
 	 * of parameters, if true it can be re-executed.
 	 */
-	private boolean incrementInstanceEnabled = false;
+	private boolean incrementInstanceEnabled = true;
```
cppwfs, line 167: Probaboy need a SCDF variable that enables or disables whether this is true or not.  This could affect folks who already have an expectation that incrementInstanceEnabled is false.

ilayaperumalg: Since SCDF owns the CTR and having this flag still be overridable, I think it can be ok for the existing users who have already set this flag to `true` or `false`.

ilayaperumalg:
> we will need a migration script for existing composed tasks.

Good point. I realize it could be cumbersome to handle the migration in this case (though the implementation is straightforward). I have re-worked this PR to use a different approach to identify the composed task element.

ilayaperumalg:
Force pushed the PR as the new changes are a complete re-work.

cppwfs:
Need to add descriptions for composedTaskElement.   I add them on merge.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionServiceTests.java
```diff
@@ -1363,6 +1363,9 @@ private static void verifyTaskExistsInRepo(String taskName, String dsl,
 			TaskDefinitionRepository taskDefinitionRepository) {
 		TaskDefinition taskDefinition = taskDefinitionRepository.findById(taskName).get();
 		assertThat(taskDefinition.getName(), is(equalTo(taskName)));
+		if (taskDefinition.getProperties().containsKey("spring.cloud.dataflow.task.composed-task-element")) {
```
cppwfs, line 1366: Test is no longer needed due to the updates in the controller.  Will fix on merge.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/TaskConfigurationProperties.java
```diff
@@ -30,10 +30,12 @@
...
  */
 @Validated
-@ConfigurationProperties(prefix = TaskConfigurationProperties.COMPOSED_TASK_PREFIX)
+@ConfigurationProperties(prefix = TaskConfigurationProperties.TASK_PREFIX)
 public class TaskConfigurationProperties {
 
-	public static final String COMPOSED_TASK_PREFIX = DataFlowPropertyKeys.PREFIX + "task";
+	public static final String TASK_PREFIX = DataFlowPropertyKeys.PREFIX + "task";
+
+	public static final String COMPOSED_TASK_ELEMENT = TASK_PREFIX + ".composed-task-element";
```
cppwfs, line 38: Dead code.  Will be removed on merge.

cppwfs:
Need to update Composed Task Tests due to test failures.  Will fix on merge.

cppwfs:
Polished, fixed some tests, rebased, squashed, merged.

PR #3986, opened 2020-06-04
Update task manifest for latest task execution in task definition

ilayaperumalg:
- When retrieving the task definition's latest task execution, update the task manifest associated with it.
This will enable the REST clients to refer to the manifest used in the task execution.
 - Modify TaskExecutionResource to allow constructing with TaskManifest
 - Add/update tests

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
@@ -68,8 +68,6 @@ import org.springframework.web.bind.annotation.RestController;
 @ExposesResourceFor(TaskDefinitionResource.class)
 public class TaskDefinitionController {
 
-	private final Assembler taskAssembler = new Assembler();
-
 	private final TaskDefinitionRepository repository;
 
 	private final TaskSaveService taskSaveService;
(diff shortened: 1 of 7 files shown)


cppwfs:
Let's chat about this.  The original story said we needed a new endpoint to retrieve the task defintion with the last task execution.   It looks like we get the task execution information is being obtained when a user does a listing of task definitions (this could cause performance issues).

ilayaperumalg:
> Let's chat about this. The original story said we needed a new endpoint to retrieve the task defintion with the last task execution. It looks like we get the task execution information is being obtained when a user does a listing of task definitions (this could cause performance issues).

Sure, and it is a good point. The reason for adding this feature into the existing endpoint is to help aggregate the functionality into the API where it returns the task definitions along with the task executions. I think we can add a flag to make the retrieval of the task manifest optional.

cppwfs:
That is a great idea!!!!

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
```diff
@@ -219,10 +223,26 @@ public TaskDefinitionResource display(@PathVariable("name") String name) {
 	 */
 	class Assembler extends RepresentationModelAssemblerSupport<TaskExecutionAwareTaskDefinition, TaskDefinitionResource> {
 
+		private boolean enableManifest = false;
```
cppwfs, line 226: Technically we don't have to initialize a boolean at the class level to false.  I'll fix on merge.

cppwfs:
LGTM
Polished Squashed Merged

PR #4002, opened 2020-06-16
Tasks launched by SCDF will now work if previously launched outside 

### LINKED ISSUE #4001: Can't launch task from dataflow if launched from external resource
When a user attempts to launch a task from SCDF that has been launched from an external source  (cron job) the following exception is thrown:
```
java.lang.IllegalArgumentException: taskManifest must not be null
	at org.springframework.util.Assert.notNull(Assert.java:198)
	at org.springframework.cloud.dataflow.rest.resource.TaskExecutionResource.<init>(TaskExecutionResource.java:176)
	at org.springframework.cloud.dataflow.server.controller.TaskDefinitionController$Assembler.updateTaskExecutionResource(TaskDefinitionController.java:280)
```

cppwfs:
resolves #4001


ilayaperumalg:
LGTM, merging.

Issue #3610, opened 2019-11-01
Refactor DefaultTaskExecutionService's executeTask


ilayaperumalg:
The [executeTask](https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java#L173) method in `DefaultTaskExecutionService` needs to be refactored as it gets growing with the number of underlying operational changes.

We need to consider setting the AppDeploymentRequest at the final place of the method as well.

deepak5127:
Hi @ilayaperumalg  Shall i work on this issue ? If so, can you please share some more details on the task.

ilayaperumalg:
Hi @deepak5127, Sure, thanks for your interest in working on this.

The main idea is to split the `executeTask` into multiple smaller methods so that it is easy to understand the flow (especially how the `AppDeploymentRequest` is updated with the properties).

Feel free to submit a initial PR and we can go from there. Thanks again!

deepak5127:
Sure, Thanks @ilayaperumalg . I will start looking into this and submit a PR when my initial version is ready.

deepak5127:
@ilayaperumalg  I have refactored AppDeploymentRequest creation part alone. Please let me know if the approach looks good, so that I can continue refactoring the whole method.

ilayaperumalg:
This was addressed via c23ac5d

ilayaperumalg:
@deepak5127 Thank you for your contributions.

PR #4004, opened 2020-06-17
GH-3610 Refactor DefaultTaskExecutionService's executeTask

deepak5127:
Refactored AppDeploymentRequest creation. If the approach looks good, I will refactor the entire method.

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
@@ -58,8 +58,10 @@ import org.springframework.cloud.dataflow.server.service.TaskExecutionCreationSe
 import org.springframework.cloud.dataflow.server.service.TaskExecutionInfoService;
 import org.springframework.cloud.dataflow.server.service.TaskExecutionService;
 import org.springframework.cloud.dataflow.server.service.TaskSaveService;
+import org.springframework.cloud.dataflow.server.service.impl.diff.PropertiesDiff;
 import org.springframework.cloud.dataflow.server.service.impl.diff.TaskAnalysisReport;
 import org.springframework.cloud.dataflow.server.service.impl.diff.TaskAnalyzer;
+import org.springframework.cloud.dataflow.server.service.impl.diff.PropertiesDiff.PropertyChange;
 import org.springframework.cloud.deployer.spi.core.AppDeploymentRequest;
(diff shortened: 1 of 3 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
```diff
@@ -323,6 +312,40 @@ public long executeTask(String taskName, Map<String, String> taskDeploymentPrope
...
 	}
+	
+	/**
+	 * Method to merge the task deployment properties from current task execution and
+	 * previous manifest.
+	 * @param taskExecutionInformation Current task execution information
+	 * @param previousManifest previous manifest info. *
+	 * @return mergedTaskDeploymentProperties Final Merged task deployment properties map
+	 */
+	private Map<String, String> getMergedProperties(TaskExecutionInformation taskExecutionInformation,
```
ilayaperumalg, line 323: We need to use `TaskAnalysisReport` here instead of the custom comparison logic. Not sure why you removed that at first place.

deepak5127: I had two thoughts. One is to do changes in TaskAnalyzer to accept the DeploymentProperties instead of TaskManifest, so that we can avoid creating AppDeploymentRequest. Another one to write a custom logic. Are  you suggesting to make changes in TaskAnalyzer in a way to accept only the DeploymentProperties  as input ?

I saw a requirement to create AppDeploymentRequest at last, so was trying to avoid where ever it is created before final AppDeploymentRequest object creation

ilayaperumalg: > One is to do changes in TaskAnalyzer to accept the DeploymentProperties instead of TaskManifest

Sure, you can add an additional method in `TaskAnalyzer` to handle this. We don't want to duplicate the properties difference logic away from `TaskAnalyzer`.

deepak5127: Sure, I will do that. Thanks

ilayaperumalg:
LGTM, merging.

Hi @deepak5127, Thank you for your contribution! Looking forward to seeing some more :-)

deepak5127:
Thanks @ilayaperumalg  . Please let me know if you have any issues for me to work on

PR #4008, opened 2020-06-19
Applications list to include all the versions

### LINKED ISSUE #2039: Skipper integration: list of applications grouped by version
As a consumer, I want to get the applications grouped by name/type with a list of versions as a child (with pagination/filter/order).

For now, the service `GET /apps` return one line by version.
The consumer has to retrieve all applications and performed a group by on the name/type (bad perf).

A new service should return the following response:
```
{
  "_embedded": {
    "appRegistrationResourceList": [
      {
        "name": "aggregator",
        "type": "processor",
        "uri": "maven://org.springframework.cloud.stream.app:aggregator-processor-kafka:2.1.2.RELEASE",
        "defaultVersion": "2.1.2.RELEASE",
        "versions": [
          "2.1.2.RELEASE",
          "2.1.1.RELEASE"
        ],
        "_links": {
          "self": {
            "href": "http://localhost:4200/apps/processor/aggregator/2.1.2.RELEASE"
          }
        }
      },
    ]
  },
  "page": {
    "size": 20,
    "totalElements": 1,
    "totalPages": 1,
    "number": 0
  }
}

```

ilayaperumalg:
- When the apps are retrieved, we need to have an ability to include all the versions associated with the app
  - Add a request parameter `defaultVersion=true` to retrieve only the apps that have default version set to true and along with the versions for that app
  - Update REST API documentation
  - Add/update tests

Resolves #2039

### DIFF
--- a/spring-cloud-dataflow-classic-docs/src/test/java/org/springframework/cloud/dataflow/server/rest/documentation/AppRegistryDocumentation.java
+++ b/spring-cloud-dataflow-classic-docs/src/test/java/org/springframework/cloud/dataflow/server/rest/documentation/AppRegistryDocumentation.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2018 the original author or authors.
+ * Copyright 2016-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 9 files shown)


oodamien:
LGTM

PR #4033, opened 2020-07-07
Update to java-cfenv 2.2.0

### LINKED ISSUE #3687: Adopt java-cfenv 2.1.x
In a previous release we were unable to away from `1.1.1` as oauth things broke with config server and while there was newer releases, those we basically re-releases without any changes we required. Hopefully this would allow us to move away from legacy `spring-security-oauth2` package.

markpollack:
Fixes #3687

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -190,6 +190,16 @@
 				<version>${java-cfenv-boot.version}</version>
 			</dependency>
 			<dependency>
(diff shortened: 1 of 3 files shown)


jvalkeal:
Merged per 388ace055f911c0cfe3ec2da49494ecc057bc015

PR #4034, opened 2020-07-07
Wording changes

Buzzardo:
Replacing some terms

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
@@ -493,7 +493,7 @@ The following example shows the contents of such an artifact, for the canonical 
 ----
 $ jar tvf log-sink-rabbit-1.2.1.BUILD-SNAPSHOT-metadata.jar
 373848 META-INF/spring-configuration-metadata.json
-   174 META-INF/spring-configuration-metadata-allowlist.properties
+   174 META-INF/spring-configuration-metadata-whitelist.properties
 ----
 
 Note that the `spring-configuration-metadata.json` file is quite large. This is because it contains the concatenation of _all_ the properties that
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
@@ -149,7 +149,7 @@ One may also specify the longhand version as well, as shown in the following exa
 
 `dataflow:> stream create --definition "http --server.port=8000 | log" --name myhttpstream`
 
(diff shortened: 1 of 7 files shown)


ilayaperumalg:
@Buzzardo Thank you for bringing in these changes!

Since we still need to support the existing applications which might have the application metadata configuration using the old names, I am afraid we still have to maintain the older naming convention. But, I think we can update our documentation to remove the older name references so that the new users could adopt these nicely chosen names.

dturanski:
I'm working on the corresponding code changes, so we need to sync on the name of the properties file.  I went with 
`dataflow-configuration-metadata.properties`.  The variable names and inline docs use `included` but here we should define it more explicitly.  `allowed` doesn't really reflect the intention here.  I think `included` may be too generic but is better.

Buzzardo:
Thanks for working on it, @dturanski. "visible" may well be the best choice. It is definitely true that there is a lot of code and configuration to change, too.

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -475,25 +475,25 @@ Make sure to add 'spring-boot-configuration-processor' as an optional dependency
...
+configuration properties metadata and the allowlisting file described in the previous section.
 
 The following example shows the contents of such an artifact, for the canonical `log` sink:
 
 [source, bash]
 ----
 $ jar tvf log-sink-rabbit-1.2.1.BUILD-SNAPSHOT-metadata.jar
 373848 META-INF/spring-configuration-metadata.json
-   174 META-INF/spring-configuration-metadata-whitelist.properties
+   174 META-INF/spring-configuration-metadata-allowlist.properties
```
cppwfs, line 496: We can't replace this property file name, until we move to the next full release of SCDF.

cppwfs:
I agree with you @Buzzardo that @dturanski 's suggestion of visible is better.

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
Buzzardo: No worries. Clearly, I was moving too quickly and got one I shouldn't have (I think I thought it was a sample rather than a file that really exists, but it doesn't matter - I shouldn't have changed that one).

Buzzardo: I fixed that, an extraneous comma in content that was new since my first commit, and the conflict.

Buzzardo:
I caught enough other problems that I am now doing a full editing pass on Spring Cloud Data Flow's docs. I'll fix the wording changes in that process, so we can let this one go.

PR #4042, opened 2020-07-13
Fix inbound/outbound port mapping metadata for container images

ilayaperumalg:
- Retrieve inbound/outbound port mapping metadata from container image labels


jvalkeal:
lgtm

PR #4049, opened 2020-07-15
Updated schedulerProperties for cf props to schedule

### LINKED ISSUE #4048: Rename scheduler properties
In `CloudFoundryPlatformProperties` we should follow correct naming and have `schedulerProperties` as `scheduler`, just like others `connection` and `deployment`.

In `CloudFoundrySchedulerProperties` as we're already within a scheduler, it's unnecessary to user same prefix in `schedulerUrl`.

cppwfs:
resolves #4048


jvalkeal:
lgtm

Issue #4065, opened 2020-07-28
Go Harbor repos backed with S3 fail fetching configuration metadata


aweigold:
**Description:**
When configured against a Go Harbor repo backed by S3, configuration metadata fails.

Because Go Harbor under this configuration returns an Amazon Signed S3 URL as a 307 redirect in response to docker registry blob requests, http clients should not add an authorization header when following the redirect, as Amazon's API will return a 400 response when given more than one auth mechanism.

Current implementation throws an exception due to the RegistryAuthorizer adding a bearer token to all requests.

```
2020-07-27 16:50:44.707  WARN 1 --- [nio-9393-exec-4] ApplicationConfigurationMetadataResolver : Failed to retrieve properties for resource:Docker Resource [docker:myprivaterepo.com/foobar/taskapp:1.0.0]
org.springframework.web.client.HttpClientErrorException$BadRequest: 400 Bad Request: [<?xml version="1.0" encoding="UTF-8"?>
<Error>InvalidArgumentOnly one auth mechanism allowed; only the X-Amz-Algorithm query parameter, Signature query string parameter or the Au... (1607 bytes)]
    at org.springframework.web.client.HttpClientErrorException.create(HttpClientErrorException.java:101)
    at org.springframework.web.client.DefaultResponseErrorHandler.handleError(DefaultResponseErrorHandler.java:170)
    at org.springframework.web.client.DefaultResponseErrorHandler.handleError(DefaultResponseErrorHandler.java:112)
    at org.springframework.web.client.ResponseErrorHandler.handleError(ResponseErrorHandler.java:63)
    at org.springframework.web.client.RestTemplate.handleResponse(RestTemplate.java:782)
    at org.springframework.web.client.RestTemplate.doExecute(RestTemplate.java:740)
    at org.springframework.web.client.RestTemplate.execute(RestTemplate.java:714)
    at org.springframework.web.client.RestTemplate.exchange(RestTemplate.java:602)
    at org.springframework.cloud.dataflow.configuration.metadata.container.DefaultContainerImageMetadataResolver.getImageBlob(DefaultContainerImageMetadataResolver.java:216)
    at
```

**Release versions:**
spring-cloud-dataflow-server:2.5.1.RELEASE

PR #4066, opened 2020-07-28
Fixes gh-4065: Added http client redirect strategy that removes authorization headers when detecting Amazon signed uris

aweigold:
Unit test spins up a Spring application as a Junit ExternalResource, as there is no way great way to test mocking the redirect with the resttemplate.  This Spring application also utilizes a signed SSL certificate as docker registry apis require TLS.

Updated jacoco maven plugin, due to build failure locally with the test application logging creating an error instead of a warning.


ilayaperumalg:
@aweigold Thank you for fixing the issue and looking forward to see more feedback/contributions!

tzolov:
@aweigold, thanks for revealing this faulty behavior! I was not aware that:
> following the redirect, as Amazon's API will return a 400 response when given more than one auth mechanism.
Is this general for all Amazon APIs or only for S3? 
Similarly is the `X-Amz-Credential` header by all AMZ apis or just S3? 

I am trying to understand the scope of this change and what side effects it might have

tzolov:
Thanks for your contribution @aweigold ! 
I've made few minor adjustments to streamline the code and improve the class naming. 
Squashed, rebased and merged at b550ecf6abd97f54fc54563a1f710fc3299a408a

aweigold:
> @aweigold, thanks for revealing this faulty behavior! I was not aware that:
> 
> > following the redirect, as Amazon's API will return a 400 response when given more than one auth mechanism.
> > Is this general for all Amazon APIs or only for S3?
> > Similarly is the `X-Amz-Credential` header by all AMZ apis or just S3?
> 
> I am trying to understand the scope of this change and what side effects it might have

I don't know offhand if the AMZ supports presigned urls for other apis, but doing so outside of the context of sharing an object in an S3 bucket seems very irregular and a security risk.  Doing a quick Google, it's pretty evident the common use case is S3.

IMHO, passing back presigned urls is a bad practice, as anyone who might capture that URL will have access to the resource until the credentials in the url expire.  I generally direct developers to proxy a stream from the S3 SDK, but I've seen this out in the wild (and obviously with Go Harbor) where you invoke the S3 API to generate a presigned url, to pass/redirect, back to the client.  Since the credentials in the url expire fairly quickly, it's a containable risk.

PR #4089, opened 2020-08-10
Stream Applications with app type should be separated by single pipes

johnlinp:
Using comma as separators is deprecated now.


ilayaperumalg:
@johnlinp Thanks for fixing the issue, merging.

PR #4093, opened 2020-08-10
Extend server monitoring dashboard info support

### LINKED ISSUE #4092: Add server support for Wavefront Dashboard UI buttons.
**Problem description:**
To support the UI https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/1509 issue we need to extend the SCDF server FeatureInfo and AboutResource to provide the Wavefront connectivity info.

tzolov:
- Allow support for Wavefront configurations, FeatureInfo#wavefrontEnabled, MonitoringDashboardInfo,
 - Unify the Monitoring Dashboard support:
     - mark the old Grafana classes and configurations as deprecated.
     - allow the MonitoringDashboardInfo to support both Grafana and Wavefront.
     - introduce the MonitoringDashboardType enum.
     - The spring.cloud.dataflow.monitoring-info.xxx extends and includes the old spring.cloud.dataflow.grafan-info.xxx
 - Update tests.

Resolves #4092


spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/resource/about/MonitoringDashboardInfo.java
```diff
@@ -0,0 +1,75 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.dataflow.rest.resource.about;
+
+/**
```
ilayaperumalg, line 19: @tzolov Could you add some Javadoc here that explains the essence of this type?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/MonitoringDashboardInfoProperties.java
```diff
@@ -0,0 +1,109 @@
...
+
+	/**
+	 *
+	 */
+	private MonitoringDashboardType dashboardType = MonitoringDashboardType.grafana;
+
+	/**
+	 *
+	 */
+	@Value("${management.metrics.export.wavefront.source:default-scdf-source}")
```
ilayaperumalg, line 59: I understand that currently we only have the `source` concept for wavefront, but having wavefront specific key here may be avoided by moving wavefront specific fields into a subtype?

spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/resource/about/MonitoringDashboardInfo.java
tzolov: will improve the javadocs coverage

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/MonitoringDashboardInfoProperties.java
tzolov: At the moment the source is the only diverging property. And it is hard to justify a separate class.  Furthermore for the 2.6.x we have the deprecated Grafana related classes. Once we move to 2.7 and remove the deprecated classes we can rethink should we have pre-dashboarding class. 
Alternative is to generalized even further and use the `Map<String, String> specific` property that would allow defining dynamically the specific properties. I actually like this approach better. But we should discuss it when moving to 2.7.x line.

ilayaperumalg:
LGTM, merging. Verified the changes with wavefront as well.

PR #4099, opened 2020-08-12
Extend micrometer configuration for SCDF

### LINKED ISSUE #4070: Instrument richer SCDF/Skipper metrics
Specifically, all the REST calls can be instrumented for traces and latency charting.

tzolov:
- enable HTTP requests percentiles-histogram.
 - enable JPA/Hibernate statistics collection.
 - add Micrometer common tags: application and application.version populated with the SCDF artifactId and version.
 - drop the spring.cloud.dataflow.x prefix for the http.server.request metrics (Has Wavefront implications).
 - create a custom spring.cloud.dataflo.server metrics of type LontTaskTimer to be used a Wavfront activation.
 - add new scdf-servers Grafana dashboard that can be used monitorint SCDF and Skipper servers.
 - update the docker-compose prometheus configurations to enable the SCDF and Skipper server metrics publishing. (TODO: update K8s manifests & documentation).
 - update the Grafana and Prometheus image versions.

 Resolves #4070

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/web/WebConfiguration.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/web/WebConfiguration.java
@@ -25,6 +25,9 @@ import javax.servlet.ServletContext;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
+import io.micrometer.core.instrument.LongTaskTimer;
+import io.micrometer.core.instrument.Metrics;
+import io.micrometer.core.instrument.Tags;
 import org.h2.tools.Server;
 import org.slf4j.LoggerFactory;
 
@@ -75,6 +78,7 @@ public class WebConfiguration implements ServletContextInitializer, ApplicationL
 	private String dataSourceUrl;
 
 	private Server server = null;
+	private LongTaskTimer.Sample longTaskSample;
 
 	public Server initH2TCPServer() {
 		logger.info("Starting H2 Server with URL: " + dataSourceUrl);
@@ -97,6 +101,11 @@ public class WebConfiguration implements ServletContextInitializer, ApplicationL
 
 	@Override
 	public void onStartup(ServletContext servletContext) {
+		LongTaskTimer longTaskTimer = LongTaskTimer
+				.builder("spring.cloud.dataflow.server").description("DataFlow duration timer")
+				.tags(Tags.empty()).register(Metrics.globalRegistry);
+		this.longTaskSample = longTaskTimer.start();
+
 		if (StringUtils.hasText(dataSourceUrl) && dataSourceUrl.startsWith("jdbc:h2:tcp://localhost:")) {
 			logger.info("Start Embedded H2");
 			initH2TCPServer();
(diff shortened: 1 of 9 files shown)


src/grafana/prometheus/docker/grafana/dashboards/scdf-servers.json
```diff
@@ -0,0 +1,5347 @@
+{
```
ilayaperumalg, line 1: Is this a generated file from a dashboard configuration?

spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
```diff
@@ -41,7 +48,9 @@ spring:
     initialize-schema: never
   jpa:
     properties:
-      hibernate.id.new_generator_mappings: true
+      hibernate:
+        id.new_generator_mappings: true
+        generate_statistics: true
```
ilayaperumalg, line 53: While this is good for metrics generation, do you think this will have cause any performance hit (given we set it here by default)?

src/grafana/prometheus/docker/grafana/dashboards/scdf-servers.json
tzolov: yes

spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
tzolov: You have a point
Not sure how significant, but the generate_statistics would defiantly have an impact on the performance.
We can move the configuration out from the defaults and document how to enable it when needed. If not enabled the dashboard charts will be empty. 
Same applies for the Skipper's configuration as well.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/web/WebConfiguration.java
```diff
@@ -97,6 +101,11 @@ private String getH2Port(String url) {
 
 	@Override
 	public void onStartup(ServletContext servletContext) {
+		LongTaskTimer longTaskTimer = LongTaskTimer
+				.builder("spring.cloud.dataflow.server").description("DataFlow duration timer")
```
ilayaperumalg, line 105: Not sure if this description will be displayed anywhere. If it gets displayed, please change the name `DataFlow` to `Spring Cloud Data Flow`

spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
tzolov: I've updated this and the skipper PR to remove the generate_staistics=true from the default configuration.

ilayaperumalg: Maybe you can add the documentation part to enable this as part of this PR as well?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/web/WebConfiguration.java
tzolov: done

spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
tzolov: This will be a separate issue: https://github.com/spring-cloud/spring-cloud-dataflow/issues/4102

chrisjs:
tested & LGTM merged in 4b1e4d03303168704274b7d1fe6086505e8baf15

PR #4105, opened 2020-08-12
Restart failed job now uses task arg to get platform

### LINKED ISSUE #4080: Restart Task IllegalStateException
Hi,

We're facing IllegalStateException when restarting task step. We have two different kubernetes namespaces. We can launch job on specified namespace. But when we tried to restart task, we are getting error like below.

   _{
        "logref": "IllegalStateException",
        "message": "Did not find platform for taskName=[abcdef] , taskId=[2923]"
    }_

We could see platformname environment variable on Openshift. But we couldn't see it on Spring Cloud Data Flow Task Execution Dashboard. 

SCDF couldn't store spring.cloud.data.flow.platformname variable in properties section.

It may be a bug. Can you review it?

SCDF version is 2.5.0. We could see platformName in local SCDF . But we couldn't see SCDF which is running on Openshift.

Thanks.

image

image

image

https://github.com/spring-cloud/spring-cloud-dataflow/blob/85e6f761e006b0201f6a66bdfb8e921724dd6d38/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskJobService.java#L194-L204

cppwfs:
We were expecting  the platform name  to be in the job param but, since the value is non identifying, it is excluded from the job param.
resolves #4080

### DIFF
--- a/spring-cloud-dataflow-classic-docs/src/test/java/org/springframework/cloud/dataflow/server/rest/documentation/JobExecutionsDocumentation.java
+++ b/spring-cloud-dataflow-classic-docs/src/test/java/org/springframework/cloud/dataflow/server/rest/documentation/JobExecutionsDocumentation.java
@@ -16,7 +16,7 @@
 
 package org.springframework.cloud.dataflow.server.rest.documentation;
 
-import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


dturanski:
LGTM; Merging

adrianriley:
Any chance this fix will be backported to 2.6.x?

ilayaperumalg:
@adrianriley Since this was a bug fix, this needs to be added into 2.6.x. We missed it out with the recent 2.6.1 Release though. Sorry about that. Thanks for bringing this up!

I have cherry-picked the changes and pushed into 2.6.x as ee541bde1. Until we do the next release, please use 2.6.2-SNAPSHOT if you want to try the changes.

ilayaperumalg:
> Since this was a bug fix, this needs to be added into 2.6.x.

A correction. This wasn't essentially a bug fix but given the criticality of the fix I believe this change potentially needs to be in 2.6.x

Issue #3866, opened 2020-04-09
SCDF Dashboard - Provide search/filter options in jobs, tasks, streams and apps view page


majorisit:
The current version of **scdf dashboard** is not flexible enough to **search/filter** deployed/running components using its name, id, status, or time. Please make enhancement with the current scdf dashboard to provide search capability at the column header level.  Currently, Its not flexible enough to search/filter scdf deployed components due to pagination.

![scdf_dashboard](https://user-images.githubusercontent.com/31349495/78922793-25a3d180-7a65-11ea-8c5a-c8a1d7785707.jpg)

siddhantsorann:
@sabbyanandan Hey, I'd like to give this a shot, as we would love to have this feature for our use case as well. Can you let me know how to go about it?

sabbyanandan:
@siddhantsorann: Excellent! @oodamien can help guide by showing you around the codebase, which hopefully can get you going.

sabbyanandan:
The backend is probably already available. If so, we may have to move this card to the UI repo. @ilayaperumalg/@jvalkeal: could one of you confirm?

siddhantsorann:
Hey, @sabbyanandan any updates on this? You can also let me know if there is some other ticket I can pick up which might be a good starting point.

sabbyanandan:
@ilayaperumalg @jvalkeal @oodamien: Could one of you help guide us here?

sabbyanandan:
I just also realize https://github.com/spring-cloud/spring-cloud-dataflow/issues/3564 could also benefit from this capability.

siddhantsorann:
Yeah, that also seems to have the same context. We, ourselves would love to have this filter as well. Hence eager to pick this up.

ilayaperumalg:
@siddhantsorann Thanks for the interest in this and sorry about the delay in getting back to you on this.

Here are the changes you would need to consider at the backend before you start designing at the UI (especially those queries related to searching batch job executions `from` and `to` etc.,) :

- Introduce the batch job retrieval methods that are necessary on these filtering cases in this [JobService] type. (https://github.com/spring-cloud/spring-cloud-dataflow/blob/497f6371111a59edd9e6fcba2b46eeb7f3d6ab65/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/JobService.java#L49)
[...]

PR #4116, opened 2020-08-21
Filtering Job executions

siddhantsorann:
This PR is for the [issue](https://github.com/spring-cloud/spring-cloud-dataflow/issues/3866). Currently raising it as a draft.

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/JobService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/JobService.java
@@ -349,4 +349,16 @@ public interface JobService {
 	 */
 	Collection<JobExecution> listJobExecutionsForJob(String jobName, BatchStatus status, int pageOffset, int pageSize)
 			throws NoSuchJobException;
+
+	/**
+	 * List the {@link JobExecutionWithStepCount JobExecutions} in descending order of
+	 * creation (usually close to execution order) without step execution data filtered by job
+	 * names matching query string.
+	 *
+	 * @param queryString search query string to filter job names
+	 * @param start the index of the first execution to return
+	 * @param count the maximum number of executions
+	 * @return a collection of {@link JobExecutionWithStepCount}
+	 */
+	Collection<JobExecutionWithStepCount> listFilteredJobExecutionsWithStepCount(String queryString, int start, int count);
 }
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/JobService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/JobService.java
@@ -355,10 +355,10 @@ public interface JobService {
 	 * creation (usually close to execution order) without step execution data filtered by job
 	 * names matching query string.
 	 *
-	 * @param queryString search query string to filter job names
+	 * @param jobName search query string to filter job names
 	 * @param start the index of the first execution to return
 	 * @param count the maximum number of executions
 	 * @return a collection of {@link JobExecutionWithStepCount}
 	 */
-	Collection<JobExecutionWithStepCount> listFilteredJobExecutionsWithStepCount(String queryString, int start, int count);
+	Collection<JobExecutionWithStepCount> listFilteredJobExecutionsWithStepCount(String jobName, int start, int count);
 }
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/JobExecutionThinController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/JobExecutionThinController.java
@@ -79,7 +79,7 @@ public class JobExecutionThinController {
 	 * Return a page-able list of {@link JobExecutionThinResource} defined jobs that
 	 * do not contain step execution detail.
 	 *
-	 * @param queryString search query string to filter job names
+	 * @param jobName search query string to filter job names
 	 * @param pageable page-able collection of {@code TaskJobExecution}s.
 	 * @param assembler for the {@link TaskJobExecution}s
 	 * @return a list of Task/Job executions(job executions do not contain step executions.
@@ -89,13 +89,13 @@ public class JobExecutionThinController {
 	@RequestMapping(value = "", method = RequestMethod.GET, produces = "application/json")
 	@ResponseStatus(HttpStatus.OK)
 	public PagedModel<JobExecutionThinResource> listJobsOnly(
-			@RequestParam(value = "search", required = false) String queryString,
+			@RequestParam(value = "name", required = false) String jobName,
 			Pageable pageable, PagedResourcesAssembler<TaskJobExecution> assembler) throws NoSuchJobExecutionException {
 		List<TaskJobExecution> jobExecutions;
-		if (StringUtils.isEmpty(queryString)) {
+		if (StringUtils.isEmpty(jobName)) {
 			jobExecutions = taskJobService.listJobExecutionsWithStepCount(pageable);
 		} else {
-			jobExecutions = taskJobService.listJobExecutionsWithStepCount(queryString, pageable);
+			jobExecutions = taskJobService.listJobExecutionsWithStepCount(jobName, pageable);
 		}
 		Page<TaskJobExecution> page = new PageImpl<>(jobExecutions, pageable, taskJobService.countJobExecutions());
 		return assembler.toModel(page, jobAssembler);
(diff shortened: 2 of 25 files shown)


pivotal-issuemaster:
@siddhantsorann Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-dataflow&pullRequestId=4116)!

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/JobExecutionThinController.java
```diff
@@ -86,9 +87,10 @@ public JobExecutionThinController(TaskJobService taskJobService) {
 	 */
 	@RequestMapping(value = "", method = RequestMethod.GET, produces = "application/json")
 	@ResponseStatus(HttpStatus.OK)
-	public PagedModel<JobExecutionThinResource> listJobsOnly(Pageable pageable,
-			PagedResourcesAssembler<TaskJobExecution> assembler) throws NoSuchJobExecutionException {
-		List<TaskJobExecution> jobExecutions = taskJobService.listJobExecutionsWithStepCount(pageable);
+	public PagedModel<JobExecutionThinResource> listJobsOnly(
+			@RequestParam(value = "q", required = false) String queryString,
```
cppwfs, line 91: Probably want to have something more descriptive than q.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/TaskJobService.java
```diff
@@ -155,13 +155,15 @@
...
 	 *
+	 *
+	 * @param queryString search query string to filter job names
 	 * @param pageable enumerates the data to be returned.
 	 * @return List containing {@link TaskJobExecution}s.
 	 *
 	 * @throws NoSuchJobExecutionException thrown if the job execution specified does not
 	 *     exist.
 	 */
-	List<TaskJobExecution> listJobExecutionsWithStepCount(Pageable pageable) throws NoSuchJobExecutionException;
```
cppwfs, line 164: Since we are at a minor version, we don't want to change an existing API. Meaning lets keep the original definition as an overloaded call to the listJobExecutionsWithStepCount(String queryString, Pageable pageable), in that it passes a `null` for the query string.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskJobService.java
```diff
@@ -106,10 +107,18 @@ public DefaultTaskJobService(JobService jobService, TaskExplorer taskExplorer,
 	}
 
 	@Override
-	public List<TaskJobExecution> listJobExecutionsWithStepCount(Pageable pageable) {
+	public List<TaskJobExecution> listJobExecutionsWithStepCount(String queryString, Pageable pageable) {
```
cppwfs, line 110: We'll need a unit test for this method as well.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/JobExecutionThinController.java
```diff
@@ -86,9 +87,10 @@ public JobExecutionThinController(TaskJobService taskJobService) {
 	 */
 	@RequestMapping(value = "", method = RequestMethod.GET, produces = "application/json")
 	@ResponseStatus(HttpStatus.OK)
-	public PagedModel<JobExecutionThinResource> listJobsOnly(Pageable pageable,
-			PagedResourcesAssembler<TaskJobExecution> assembler) throws NoSuchJobExecutionException {
-		List<TaskJobExecution> jobExecutions = taskJobService.listJobExecutionsWithStepCount(pageable);
+	public PagedModel<JobExecutionThinResource> listJobsOnly(
```
cppwfs, line 90: Probably want to have a unit test on this.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/JobService.java
```diff
@@ -349,4 +349,16 @@ StepExecution getStepExecution(Long jobExecutionId, Long stepExecutionId) throws
...
+	 * List the {@link JobExecutionWithStepCount JobExecutions} in descending order of
+	 * creation (usually close to execution order) without step execution data filtered by job
+	 * names matching query string.
+	 *
+	 * @param queryString search query string to filter job names
+	 * @param start the index of the first execution to return
+	 * @param count the maximum number of executions
+	 * @return a collection of {@link JobExecutionWithStepCount}
+	 */
+	Collection<JobExecutionWithStepCount> listFilteredJobExecutionsWithStepCount(String queryString, int start, int count);
```
cppwfs, line 363: I concerned that there maybe a case for sql injection here.

siddhantsorann: If I'm not wrong, we are using JDBCTemplate here which internally uses PreparedStatements which protects it from SQL Injections. So I think this should be safe.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/JobExecutionThinController.java
siddhantsorann: Since this was a draft PR I hadn't committed the unit tests, will do that.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/JobExecutionThinController.java
siddhantsorann: Any suggestions? How about "search"? I normally use "q" for search LIKE queries.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/TaskJobService.java
siddhantsorann: Oh, okay. But then how can we have 2 APIs exposed to the same GET Mapping "" with different params?

siddhantsorann:
> Thank you for your contribution.
> I added some comments.

@cppwfs Thanks for getting back, need a little more assistance with a couple of the comments, will raise a proper PR post that.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskJobService.java
siddhantsorann: Can you help me out with where the test cases have been added for other methods in this class? The DefaultTaskJobServiceTests class covers only the constructor in DefaultTaskJobService.

siddhantsorann:
@cppwfs I have made the required changes, can you check again please?

cppwfs:
LGTM just add the requests mentioned above.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/JobService.java
siddhantsorann: Changed that to jobName

siddhantsorann:
> LGTM just add the requests mentioned above. I look forward to the full solution. Thanks for the contribution and your patience with my review.

Hey, @cppwfs I have changed the queryString to jobName. Can you take a look again? Once this is merged, I'll raise a different PR for other filters as part of this issue, is that alright?

cppwfs:
I was reviewing the PR and noticed the following with the existing code.  
I can do the following http://localhost:9393/jobs/thinexecutions?name=%25myjob%25 and it provides the same features as this PR.  Thoughts?

siddhantsorann:
> I was reviewing the PR and noticed the following with the existing code in master.
> I can do the following http://localhost:9393/jobs/thinexecutions?name=%25myjob%25 and it provides the same features as this PR. Thoughts?

Oh, I wasn't aware of that. But in that case, we'll have to pass encoded % right? In that case, would we need this change?

cppwfs:
You are correct.  We probably don't need this PR.
But I want to thank you for you interest and investment in time in making SCDF a better product for the community.   
I apologize for taking so long in getting back to you on this PR.   
Please keep submitting PR's!

Again thanks soooo much!

Issue #4090, opened 2020-08-10
Application name should not contain colon symbol


johnlinp:
**Description:**
Colon symbol has a special meaning in both Stream DSL and Composed Task DSL, which is used for labeling. Therefore, when an application has a name that containing colon symbols, it cannot be used in Stream DSL or Composed Task DSL.

**Release versions:**
2.6.0

**Steps to reproduce:**
1. Register an application with the name "a:b".
```
dataflow:>app register "a:b" --type source --uri maven://some-group:some-artifact:some-version
Successfully registered application 'source:a:b'
```
2. Expected to see the registration fail, but it was successful.

sabbyanandan:
@johnlinp: Good catch! If you have cycles, please consider adding the validation for it. One of us can collaborate with you to get this wrapped both in the Shell and UI.

siddhantsorann:
@sabbyanandan Do you think this would be a good contribution to start off with? I've been looking into the filter one as well but I felt like maybe this would be a good ticket to ease into the code base?

ilayaperumalg:
The validation needs to happen at the [AppRegistry REST controller](https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppRegistryController.java#L223) where a new app registration is created.

siddhantsorann:
Thanks, @ilayaperumalg Yes, that is where I have added it. Is the current master unstable? I was unable to build it.

ilayaperumalg:
> Is the current master unstable? I was unable to build it.

I don't think so: https://build.spring.io/browse/SCD-BMASTER-4243

siddhantsorann:
It built now, was having an issue yesterday. Thanks.

siddhantsorann:
@ilayaperumalg Can you also let me know if we need to validate anything other than `Application Name cannot contain ':'` ?

PR #4117, opened 2020-08-22
Application name validation

siddhantsorann:
Raising this PR to validate that the application name should not contain some specific wildcards. Currently, we are only checking for ':'.
This PR is against this [issue](https://github.com/spring-cloud/spring-cloud-dataflow/issues/4090).

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppRegistryController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppRegistryController.java
@@ -225,6 +225,7 @@ public class AppRegistryController {
 			@RequestParam("uri") String uri, @RequestParam(name = "metadata-uri", required = false) String metadataUri,
 			@RequestParam(value = "force", defaultValue = "false") boolean force) {
 
+		validateApplicationName(name);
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppRegistryController.java
```diff
@@ -437,6 +438,20 @@ private void prefetchMetadata(List<AppRegistration> appRegistrations) {
...
+		char[] invalidWildCards = new char[]{':'};
+		StringBuilder invalidChars = new StringBuilder();
+		for (char invalidWildCard : invalidWildCards) {
+			if (name.contains(Character.toString(invalidWildCard))) {
+				invalidChars.append("'").append(invalidWildCard).append("'");
+			}
+		}
+
+		if (!invalidChars.toString().equals("")) {
+			throw new IllegalArgumentException("Application name: " + name + " cannot contain: " + invalidChars);
```
ilayaperumalg, line 451: It would be nice if we add single quotes around the `name` value. I will add it when merging.

ilayaperumalg:
@siddhantsorann Thank you for the contribution! The changes look good to me, merging.

ilayaperumalg:
Squashed and merged the PR as 1f3414b13

PR #4118, opened 2020-08-22
Extend server monitoring dashboard info support

### LINKED ISSUE #4071: Instrument SCDF/Skipper's datasource call-traces
Specifically, drilling capability to be able to see how the datasource is performing. The connection pool's health and things like that.

### LINKED ISSUE #4092: Add server support for Wavefront Dashboard UI buttons.

### LINKED ISSUE #4101: Port the SCDF/Skipper server metrics dashboard to Wavefront
Depends on: #4070

### LINKED ISSUE #4129: Allow applicationProperties to pass property placeholder (`${..}`) to the launched children applications
**Problem description:**
Spring Cloud Data Flow provides a mechanism for setting common properties to all [Stream](https://docs.spring.io/spring-cloud-dataflow/docs/current/reference/htmlsingle/#spring-cloud-dataflow-global-properties) and [Task](https://docs.spring.io/spring-cloud-dataflow/docs/current/reference/htmlsingle/#_common_application_properties) applications that are launched by it. This is  done via adding properties prefixed with `spring.cloud.dataflow.applicationProperties.stream` and `spring.cloud.dataflow.applicationProperties.task` when starting the server.

But if the common properties contain a placeholder (`${...}`) later will be evaluated and resolved in the SCDF's context not passed as a literal to the launched apps.  For example if we define the following common property:

`spring.cloud.dataflow.applicationProperties.stream.myName=Launched App name ${app.name}`

The SCDF will resolve/substitute the ${app.name} with its own name (lets say SCDF.NAME) and will pass to the launched app a property like this: `myName=Launched App name SCDF.NAME`. 

While allowing the the SCDF to resolve property placeholders in common properties definition should remain we must find a way to pass a placeholders to be resolved by the launched applications them selfs. 

**Solution description:**
Given that Spring Boot can not be stop from doing placeholder resolution on config properties, we should define a new placeholder signatures (for example `@{...}`), that can be parsed by SCDF when composing the common properties and converted into `${ ...}` before passed to the launched applications.

**Additional context:**
This slack topic is related to the discussion: https://vmware.slack.com/archives/C055WU98P/p1598775604072900

tzolov:
- Allow support for Wavefront configurations, MonitoringDashboardInfo.
 - Make the sub-type configurations per Grafana and Wavefront.
 - Unify the Monitoring Dashboard support:
    - remove deprecated GrafanaInfo, FeatureInfo#grafanaEnabled, FeatureInfo#wavefrontEnabled.
    - allow the MonitoringDashboardInfo to support both Grafana and Wavefront (spring.cloud.dataflow.monitoring-info.xxx).
    - introduce the MonitoringDashboardType enum.
 - Add server metrics property replication to stream and task, Monitoring Info inference. Class and property renaming. Tests
 - Update tests.
 - Introduce a stream-resource and task-resource common properties to allow loading common properties from YAML or properties files. The property placeholder (e.g. ${...}) in those files are passed to the launched apps unresolved.
 - Add default `META-INF/application-stream-common-properties-defaults.yml` and `META-INF/application-task-common-properties-defaults.yml` common stream and task properties that contain default metrics tags.
 - Add tests.

 Resolves #4071
 Resolves #4092
 Resolves #4129
 Resolves #4101


spring-cloud-dataflow-server-core/src/main/resources/application-init-cloud.yml
```diff
@@ -0,0 +1,5 @@
+spring:
```
markpollack, line 1: How does the file `application-init-cloud` get loaded?

spring-cloud-dataflow-server-core/src/main/resources/META-INF/application-stream-common-properties-defaults.yml
```diff
@@ -0,0 +1,30 @@
...
+  management:
+    metrics:
+      tags:
+        # Common Spring Cloud Data Flow stream app tags.
+        stream.name: ${spring.cloud.dataflow.stream.name:unknown}
+        application.name: ${vcap.application.application_name:${spring.cloud.dataflow.stream.app.label:unknown}}
+        application.type: ${spring.cloud.dataflow.stream.app.type:unknown}
+        instance.index: ${vcap.application.instance_index:${spring.cloud.stream.instanceIndex:0}}
+        application.guid: ${spring.cloud.application.guid:unknown}
+
```
markpollack, line 11: Isn't it still appropriate for CF deploys to have the following variable defined?

```
      organization.name: ${vcap.application.org_name:default}
      space.id: ${vcap.application.space_id:unknown}
      space.name: ${vcap.application.space_name:unknown}
      application.id: ${vcap.application.application_id:unknown}
      application.version: ${vcap.application.application_version:unknown}```

markpollack: and wouldn't these be needed for batch apps on CF as well?

tzolov: The `common:`  is applied to all platforms, including CF but also local K8s. The `application.name` and  `instance.index` metrics tags are common (and required) for all platforms, while the `organization.name`, `space.id`, `space.name`, `application.id`, `application.version` tags are only CF specific and therefore defined under the `cloudfoundry:` platform prefix.

ilayaperumalg:
Verified on local, CF and K8s using wavefront and the server dashboard integration including the application tag properties work fine.

There are some cases where the addition of new tag passed to wavefront but they don't get applied. This is something we need to investigate post M1.

Great work @tzolov ! merging.

PR #4131, opened 2020-09-02
Fix broken docker-compose-prometheus and docker-compose-influx scripts

tzolov:
- Fixing a broken docker-compose-prometheus and docker-compose-influx scripts. 
- Change the 8081 prometheus-rsocket-proxy to 8086 to a avoid conflicts with the McAfee daemons.


tzolov:
Covered by https://github.com/spring-cloud/spring-cloud-dataflow/pull/4133

PR #4133, opened 2020-09-03
Align with 2.7.x WF tile activation metrics

tzolov:
- Back port Server dashboard configurations.
- change the wavefront activation metrics to the spring.clodu.dataflow.server.*
- add SCDF application metrics tag to be the name of the jar.
- enable SCDF HTTP metrics percentiles.
- Upgrade to latest Prometheus Rsocket RPoxy 1.2.1
- Fix broken configurations in prometheus and influx dockcker-compose files.
- Update prometheus and Grafana container image versions.
- retrofit the scdf-servers.json dashboard for Grafana.
- Change the 8081 prometheus-rsocket-proxy to 8086 to a avoid conflicts with the McAfee daemons.

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/web/WebConfiguration.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/web/WebConfiguration.java
@@ -25,6 +25,9 @@ import javax.servlet.ServletContext;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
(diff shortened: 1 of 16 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/web/WebConfiguration.java
```diff
@@ -97,6 +102,11 @@ private String getH2Port(String url) {
 
 	@Override
 	public void onStartup(ServletContext servletContext) {
+		LongTaskTimer longTaskTimer = LongTaskTimer
+				.builder("spring.cloud.dataflow.server").description("Spring Cloud Data Flow duration timer")
```
ilayaperumalg, line 106: Instead of `duration timer`, can we something like `up time`?

ilayaperumalg:
LGTM, verified the changes on 2.6.x as well. Merging.

PR #4137, opened 2020-09-08
Add streams/definitions/*/applications to auth endpoints

### LINKED ISSUE #4130: Stream definition view broken if using OAuth authorization
The dataflow UI calls a `/streams/definitions/*/applications` endpoint which has no corresponding mapping in the authorization rules. The request fails with a 403 for this reason and the screen won't show.

https://github.com/spring-cloud/spring-cloud-dataflow/blob/2f425a1f850a6bfddc54415696b3b08d54134246/spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml#L158-L165

ilayaperumalg:
- Add ROLE_VIEW authorization to this endpoint

Resolves #4130


ilayaperumalg:
This needs to be back-ported into 2.6.x as well.

jvalkeal:
Backported per fce336e69439665896f8cfa30dcfd1f70ffbc294

PR #4141, opened 2020-09-11
Restrict stream creation if the stream name has underscore

### LINKED ISSUE #4079: Restrict stream creation if the stream name has underscore
Given the Kubernetes validation for the app label and SCDF adds a label with the `stream name` prefix to the application, we can restrict the stream creation if the name has underscore in it.

Fixing this will avoid the stream deployment failure into K8s due to the regex validation failure on app label:
```
a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?'
```

chrisjs:
Resolves #4079

### DIFF
--- a/spring-cloud-dataflow-classic-docs/src/test/java/org/springframework/cloud/dataflow/server/rest/documentation/JobExecutionsDocumentation.java
+++ b/spring-cloud-dataflow-classic-docs/src/test/java/org/springframework/cloud/dataflow/server/rest/documentation/JobExecutionsDocumentation.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 the original author or authors.
+ * Copyright 2017-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/stream/SkipperStreamDeployer.java
```diff
@@ -326,6 +331,14 @@ private void validateStreamDeploymentRequest(StreamDeploymentRequest streamDeplo
 				.findById(streamName)
 				.orElseThrow(() -> new NoSuchStreamDefinitionException(streamDeploymentRequest.getStreamName()));
 
+		if ("kubernetes".equals(platformInfo.getPlatformType())) {
```
ilayaperumalg, line 334: I understand that the platform info is **only** available at the time of stream deployment and hence we can't get this fixed at the stream definition controller. Maybe let's discuss if we can restrict the stream name with the below regex for all the platform types - that way this validation could move up at the stream creation itself (via Stream Definition Controller)

mminella: What's wrong with just making this a global rule and applying it to all platforms?

ilayaperumalg: I don't see any issue with that except some CF users might prefer the stream names with underscore in it ;)

mminella: For consistency and to be able to support the UI (where this validation is really needed), I'd think we want to just apply this rule globally.

chrisjs:
@ilayaperumalg updated per discussion. also had to extend to tasks as the name gets inserted into labels the same way

ilayaperumalg:
LGTM, merging.

PR #4143, opened 2020-09-14
CTR now does not have to be registered to schedule composed tasks.

### LINKED ISSUE #4122: Scheduling doesn't work if Composed Task Runner task app isn't explicitly registered
If the composed task runner isn't explicitly registered into app registry, the scheduling of task fails as it expects the app named `composed-task-runner` when scheduling a task.

cppwfs:
resolves #4122


ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Cherry-picked and merged as 02d3b096f into 2.6.x

PR #4144, opened 2020-09-15
Users can override driverClassName property

### LINKED ISSUE #4128: Unable to change task's spring.datasource.driverClassName
Despite changing the application.propeties driverClassName of the registered jar, and adding  `--spring.datasource.driverClassName=com.mysql.cj.jdbc.Driver` as an argument in the Dataflow interface, when lauching a task, still I get:
Screenshot 2020-08-30 at 17 22 06

Indeed dataflow-server (DATAFLOW_VERSION=2.6.0 SKIPPER_VERSION=2.5.0) is running as a docker container that was booted with a `docker-compose.yml` file environment set with `SPRING_DATASOURCE_DRIVER_CLASS_NAME=org.mariadb.jdbc.Driver`.

Nevertheless, I think it should be quite clear that if I am overriding the argument before lauching a specific task, that, well it should be overridden. 

This behaviour is easily reproducible by following:
https://dataflow.spring.io/docs/installation/local/docker/
to set up the environment, and for the Java application, this:
https://dataflow.spring.io/docs/batch-developer-guides/batch/spring-task/

cppwfs:
resolves #4128

url can also be overridden using properties.

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/TaskServiceUtils.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/TaskServiceUtils.java
@@ -128,8 +128,13 @@ public class TaskServiceUtils {
 			}
 			builder.setProperty("spring.datasource.username", dataSourceProperties.getUsername());
 		}
-		builder.setProperty("spring.datasource.url", dataSourceProperties.getUrl());
-		builder.setProperty("spring.datasource.driverClassName", dataSourceProperties.getDriverClassName());
+		if(!taskDefinition.getProperties().containsKey("spring.datasource.url")) {
+			builder.setProperty("spring.datasource.url", dataSourceProperties.getUrl());
+		}
+		if(!(taskDefinition.getProperties().containsKey("spring.datasource.driverClassName") ||
+				taskDefinition.getProperties().containsKey("spring.datasource.driver-class-name"))) {
+			builder.setProperty("spring.datasource.driverClassName", dataSourceProperties.getDriverClassName());
+		}
 		builder.setTaskName(taskDefinition.getTaskName());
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/TaskServiceUtils.java
```diff
@@ -128,8 +128,13 @@ public static TaskDefinition updateTaskProperties(TaskDefinition taskDefinition,
 			}
 			builder.setProperty("spring.datasource.username", dataSourceProperties.getUsername());
 		}
-		builder.setProperty("spring.datasource.url", dataSourceProperties.getUrl());
-		builder.setProperty("spring.datasource.driverClassName", dataSourceProperties.getDriverClassName());
+		if(!taskDefinition.getProperties().containsKey("spring.datasource.url")) {
+			builder.setProperty("spring.datasource.url", dataSourceProperties.getUrl());
+		}
+		if(!(taskDefinition.getProperties().containsKey("spring.datasource.driverClassName") ||
```
ilayaperumalg, line 134: Do you think we need to resolve this by Relaxed names as there are other combinations of property names users could come up with when specifying the task definition properties?

cppwfs: 👍

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
@cppwfs Do you think this needs to be back-ported? - though this is an enhancement, I think it would be a good fix for 2.6.x as well.

cppwfs:
@ilayaperumalg I agree with you.  I think this would be a good addition to 2.6.x.

ilayaperumalg:
Cherry-picked and pushed into 2.6.x as 34e480c66

PR #4146, opened 2020-09-15
Use tile service binding credentials as default property values

mheath:
Adds CfEnv as a dependency and uses CfEnv to get credentials from a `p-dataflow` service binding to set the SCDF URI property and Oauth2 client credentials properties.

This fixes the issues Loblaw is having when using CTR.

### DIFF
--- a/spring-cloud-dataflow-composed-task-runner/pom.xml
+++ b/spring-cloud-dataflow-composed-task-runner/pom.xml
@@ -39,7 +39,6 @@
 		<dependency>
 			<groupId>io.pivotal.cfenv</groupId>
 			<artifactId>java-cfenv</artifactId>
-			<version>${java-cfenv.version}</version>
 		</dependency>
 		<dependency>
 			<groupId>org.mockito</groupId>
--- a/spring-cloud-dataflow-composed-task-runner/pom.xml
+++ b/spring-cloud-dataflow-composed-task-runner/pom.xml
@@ -24,6 +24,7 @@
 	<packaging>jar</packaging>
 	<properties>
 		<plexus.utils.version>3.3.0</plexus.utils.version>
+		<java-cfenv.version>2.2.0.RELEASE</java-cfenv.version>
 	</properties>
 	<dependencies>
 		<dependency>
@@ -34,6 +35,11 @@
 		<dependency>
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-dataflow-rest-client</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>io.pivotal.cfenv</groupId>
+			<artifactId>java-cfenv</artifactId>
+			<version>${java-cfenv.version}</version>
 		</dependency>
 		<dependency>
 			<groupId>org.mockito</groupId>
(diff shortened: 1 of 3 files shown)


spring-cloud-dataflow-composed-task-runner/src/main/java/org/springframework/cloud/dataflow/composedtaskrunner/support/CfServiceBindingPropertySourceInitializer.java
```diff
@@ -0,0 +1,63 @@
...
+
+	private static final Logger log = LoggerFactory.getLogger(CfServiceBindingPropertySourceInitializer.class);
+
+	public static final String PROPERTY_SOURCE_NAME = "p-dataflow";
+
+	@Override
+	public void onApplicationEvent(ApplicationEnvironmentPreparedEvent event) {
+		final CfEnv cfEnv = new CfEnv();
+		try {
+			final CfCredentials credentials = cfEnv.findCredentialsByLabel("p-dataflow");
```
ilayaperumalg, line 48: Is this always going to be `p-dataflow` ? I wonder if someone changes the name (if that is allowed).

mheath: The label is not configurable so it will always be "p-dataflow".

ilayaperumalg:
@mheath This PR is submitted against `ctr-fix` branch and I think you want to push against `master` and we can subsequently back port it into `2.6.x`.

mheath:
Good catch on the branch. I've fixed that.

ilayaperumalg:
LGTM, will wait for others to review this as well.

spring-cloud-dataflow-composed-task-runner/pom.xml
```diff
@@ -24,6 +24,7 @@
 	<packaging>jar</packaging>
 	<properties>
 		<plexus.utils.version>3.3.0</plexus.utils.version>
+		<java-cfenv.version>2.2.0.RELEASE</java-cfenv.version>
```
jvalkeal, line 27: No version definition here, it's already in parent pom.

spring-cloud-dataflow-composed-task-runner/pom.xml
```diff
@@ -35,6 +36,11 @@
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-dataflow-rest-client</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>io.pivotal.cfenv</groupId>
+			<artifactId>java-cfenv</artifactId>
+			<version>${java-cfenv.version}</version>
```
jvalkeal, line 42: Dep handling in parent pom.

spring-cloud-dataflow-composed-task-runner/pom.xml
ilayaperumalg: This needs to be removed. I will remove while merging the PR.

ilayaperumalg:
Merged the changes into master.

ilayaperumalg:
Cherry picked and pushed the changes into 2.6.x as 002715a96

PR #4153, opened 2020-09-18
Re-use deployment properties post undeploy

### LINKED ISSUE #4058: Persist stream deployment settings
**Problem description:**
Sometimes we want to stop streams for a while and re-deploy them on the particular dates. Right now we are using deploy/undeploy features together. So we can deploy some stream on Monday, and undeploy it the next day, without destroying it, so the stream will stay in Data Flow, just without any running pods. The next Monday we want to deploy it again, but at that time we always need to provide "Generic Deployer" and "Deployment Platform" settings again. But "Applications Properties" will be stored in the database, do not need to enter them again.

**Solution description:**
It will be nice to store "Generic Deployer" and "Deployment Platform" settings as well as "Applications Properties". Some of the applications are using environment variables, and not application arguments, these envs only available in the "Deployment Platform" section.

If you guys agree with the needs of this feature, I can make a pull request.

ilayaperumalg:
- When a stream is deployed and subsequently undeployed, then we need to have an option to re-use the deployment properties when the stream gets re-deployed.
 - Add a new request parameter `reuse-deployment-properties` to specify if the stream's previously used deployment properties can be re-used.
   - We can't reuse the deployment properties as the same stream can be destroyed and re-deployed. Also, this needs to be come as an option

Resolves #4058


ilayaperumalg:
Created UI story https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/1555 to provide an option to reuse the deployment properties during stream deploy.

spring-cloud-dataflow-classic-docs/src/test/java/org/springframework/cloud/dataflow/server/rest/documentation/StreamDeploymentsDocumentation.java
```diff
@@ -123,6 +124,24 @@ public void unDeployAll() throws Exception {
...
 
+
+	@Test
+	public void info() throws Exception {
+		String json = "{\"app.time.timestamp.format\":\"YYYY\"}";
+		this.mockMvc.perform(
+				get("/streams/deployments/{timelog}?reuse-deployment-properties=true", "timelog")
+						.contentType(MediaType.APPLICATION_JSON)
+						.content(json))
+				.andDo(print())
```
cppwfs, line 135: We can remove the print() to clean up our logs during testing.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentControllerTests.java
```diff
@@ -190,12 +190,49 @@ public void testShowStreamInfo() {
...
+	@Test
+	public void testReuseDeploymentProperties() {
+		Map<String, String> deploymentProperties1 = new HashMap<>();
+		deploymentProperties1.put("test1", "value1");
+		Map<String, String> deploymentProperties2 = new HashMap<>();
+		deploymentProperties2.put("test2", "value2");
+		Map<String, Map<String, String>> streamDeploymentProperties = new HashMap<>();
+		streamDeploymentProperties.put("time", deploymentProperties1);
+		streamDeploymentProperties.put("log", deploymentProperties2);
+		Map<String, String> appVersions = new HashMap<>();
```
cppwfs, line 210: we can remove `appVersions` as it is not used by this test.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentControllerTests.java
```diff
@@ -190,12 +190,49 @@ public void testShowStreamInfo() {
...
+		appVersions.put("time", "1.0.0.BUILD-SNAPSHOT");
+		appVersions.put("log", "1.0.0.BUILD-SNAPSHOT");
+		StreamDefinition streamDefinition = new StreamDefinition("testStream1", "time | log");
+		StreamDeployment streamDeployment = new StreamDeployment(streamDefinition.getName(),
+				new JSONObject(streamDeploymentProperties).toString());
+		Map<StreamDefinition, DeploymentState> streamDeploymentStates = new HashMap<>();
+		streamDeploymentStates.put(streamDefinition, DeploymentState.undeployed);
+
+		StreamAppDefinition streamAppDefinition1 = new StreamAppDefinition("time", "time", ApplicationType.source, streamDefinition.getName(), new HashMap<>());
```
cppwfs, line 219: We are not using `streamAppDefinition1` or `streamAppDefinition2` so we can probably remove these.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentControllerTests.java
```diff
@@ -190,12 +190,49 @@ public void testShowStreamInfo() {
...
+
+		when(this.streamDefinitionRepository.findById(streamDefinition.getName())).thenReturn(Optional.of(streamDefinition));
+		when(this.streamService.info(streamDefinition.getName())).thenReturn(streamDeployment);
+		when(this.streamService.state(anyList())).thenReturn(streamDeploymentStates);
+		LinkedList<StreamAppDefinition> streamAppDefinitions = new LinkedList<>();
+		streamAppDefinitions.add(streamAppDefinition1);
+		streamAppDefinitions.add(streamAppDefinition2);
+		when(this.streamDefinitionService.redactDsl(any())).thenReturn("time | log");
+
+		StreamDeploymentResource streamDeploymentResource = this.controller.info(streamDefinition.getName(), true);
```
cppwfs, line 230: Can we add a test where the flag is set to false?

cppwfs:
* We probably want to add a new deploy function for stream operations that will take the `reuse-deployment-properties` flag.  
* We probably also want --reuseDeploymentProperties as an addition to the `stream deploy` command in the shell.   
* The RootControllerTests are currently failing in this PR.   To resolve add `?reuse-deployment-properties=false` at position 1079 in the root-controller-result.json file.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentControllerTests.java
ilayaperumalg: That is already being tested in the above test case `testShowStreamInfo` ;)

ilayaperumalg:
> We probably want to add a new deploy function for stream operations that will take the reuse-deployment-properties flag.
> We probably also want --reuseDeploymentProperties as an addition to the stream deploy command in the shell.

This specific feature can **only** be driven from the dashboard as the user would be displayed with the set of previously deployment properties and they can still choose a set of properties to go for/against the retrieved list. 

Having this on the `deploy` command level would apply all the deployment properties from the previous deployment which would be confusing as the user may not know what all the deployment properties used in the previous one.

cppwfs:
LGTM .  Rebased, Squashed, Merged.

cppwfs:
We have a build failure because when we attempt to do a stream update the `reuseDeploymentProperties` flag is null.   Maybe this should be a `boolean` instead of a `Boolean`.   Thoughts?   The line is https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/StreamDeploymentController.java#L200

PR #4171, opened 2020-09-28
Task name length validation

### LINKED ISSUE #4163: Limit task name length
**Description:**
We are using SCDF v2.4.0 on Kubernetes. If we create a task with the length of name > 52 characters, it gives an error saying "size must be between 0 and 63". From what I could figure out, this error is thrown by Kubernetes. As the task is launched in a pod and 11 characters are appended to the task name to create the pod name, it increase length of pod name to > 63 which matches the error description. This error might be confusing to some.

**Steps to reproduce:**
Run SCDF on K8s and create a task with name > 52 characters and try to launch it.

**Screenshots:**
Screenshot 2020-09-23 at 12 08 05 AM

To fix this maybe we could add a limit to the length of the task name. I could raise a PR for this if that makes sense.

siddhantsorann:
This PR adds a feature requested in issue #4163 

The following changes were made -

Added validation to make sure task length is less than 53 characters.

Fixes #4163


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
```diff
@@ -354,4 +355,13 @@ public TaskExecutionAwareTaskDefinition apply(TaskDefinition source) {
 			}
 		}
 	};
+
+	private void validateTaskName(String name) {
+		// Kubernetes does not support a pod name with length greater than 63 characters. It also
+		// adds 11 characters to the task name to create a pod, hence we need to ensure that task
+		// name added in SCDF has a length pf less than 52 characters.
+		if (name.length() > 52) {
```
sabbyanandan, line 363: It turned out to be a simple length check, nice. I wonder if we need to handle this a bit more gracefully to preserve backward compatibility. There can be users on local/cf with a lengthier naming convention, and when they upgrade to the latest minor release (with this fix), it would break. @ilayaperumalg @jvalkeal @tzolov: Do you have any opinions?

Lastly, however we end up deciding to address it, let's please also apply it to stream names.

cppwfs: We could add a validation step in the `KubernetesDeployerProoperties.deploy` method like what we do in `KuberetesScheduler.schedule`  method https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/blob/master/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/KubernetesScheduler.java#L72-L74.

cppwfs: The issue is that if we have more than one platform, we don't know which platform they will select until the task is launched.   i.e. it could be CF, Local or K8's.

siddhantsorann: Yeah, that makes sense. So you are saying that we don't validate it when a user has added a task... But do it when it launches which is K8s specific?

cppwfs: Correct.   Technically K8's already does this for us. Its just that the error message is cryptic for some users.   What I'm suggesting is that we make the error message more user friendly.

siddhantsorann: So should I make that change in the other repo then?

cppwfs: This topic brought up some good discussions in our last standup.     I'm chatting with some folks on this topic.   Again thanks for bringing this up and the contribution.

cppwfs:
Thank you for this contribution.  For now we will close this and update the issue with some additional notes to the existing issue   before the week is out.   
Again thank you for the contributions and issues you've been adding!!!!!!

PR #4179, opened 2020-10-03
Container Registry auth token URI auto-retrieval

### LINKED ISSUE #4178: Improve the Container Registry authorization token URI auto-retrieval
**Problem description:**
For stream and task applications loaded from private Container Registries, we need to provide DataFlow with  [RegistryConfiguration](https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/container/RegistryConfiguration.java) configuration to allow the server pull the metadata for those applications.
 
In case of the OAuth2 authorization the registry configuration look like this:
```
- spring.cloud.dataflow.container.registry-configurations[harbor].registry-host=demo.goharbor.io
- spring.cloud.dataflow.container.registry-configurations[harbor].authorization-type=dockeroauth2
- spring.cloud.dataflow.container.registry-configurations[harbor].user=***
- spring.cloud.dataflow.container.registry-configurations[harbor].secret=***
- spring.cloud.dataflow.container.registry-configurations[harbor].extra[registryAuthUri]=https://demo.goharbor.io/service/token?service=harbor-registry&scope=repository:{repository}:pull
```

The `registryAuthUri` configuration parameter sets the authorization token URI endpoint used by the API to obtain a valid token used for interacting with the repository.  Instead of setting this parameter explicitly there is [getDockerTokenServiceUri](https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/container/authorization/DockerConfigJsonSecretToRegistryConfigurationConverter.java#L106) method that can compute this property form the HTTP 401 response headers. 

Current `getDockerTokenServiceUri` implementation probes the `https://<registry-host>/v2/_catalog` to obtain HTTP 401 response.  For some registries this url could be less reliable to obtain OAuth2 Error HTTP response. Instead the `https://<registry-host>/v2/` should be used.

tzolov:
- Replace the probe url from https://<registry-host/v2/_catalog to https://<registry-host/v2/
  - Remove the base auth when hitting the probe registry url.
  - Improve the error handling. Check of potential NPE and throw AppMetadataResolutionException instead.
  - Reduce the logging messages in case of registry configuration problem. Don't print the exception traces for non debug level.
  - Minor code improvements.

  Resolves #4178

### DIFF
--- a/src/kubernetes/server/server-deployment.yaml
+++ b/src/kubernetes/server/server-deployment.yaml
@@ -25,6 +25,9 @@ spec:
           - name: database
             mountPath: /etc/secrets/database
             readOnly: true
+          - name: ctr-registry
+            mountPath: /workspace/runtime/secrets/ctr-registry
+            readOnly: true
         ports:
         - containerPort: 80
         livenessProbe:
@@ -64,7 +67,7 @@ spec:
         - name: SPRING_CLOUD_KUBERNETES_SECRETS_ENABLE_API
           value: 'false'
         - name: SPRING_CLOUD_KUBERNETES_SECRETS_PATHS
-          value: /etc/secrets
+          value: /etc/secrets,/workspace/runtime/secrets/ctr-registry
         - name: SPRING_CLOUD_DATAFLOW_SERVER_URI
           value: 'http://${SCDF_SERVER_SERVICE_HOST}:${SCDF_SERVER_SERVICE_PORT}'
           # Provide the Skipper service location
@@ -88,3 +91,6 @@ spec:
         - name: database
           secret:
             secretName: mysql
+        - name: ctr-registry
+          secret:
+            secretName: scdf-metadata-regcred
--- a/src/kubernetes/server/server-deployment.yaml
+++ b/src/kubernetes/server/server-deployment.yaml
@@ -15,82 +15,76 @@ spec:
         app: scdf-server
     spec:
       containers:
-      - name: scdf-server
-        image: springcloud/spring-cloud-dataflow-server:2.7.0-SNAPSHOT
-        imagePullPolicy: Always
-        volumeMounts:
-          - name: config
-            mountPath: /config
-            readOnly: true
-          - name: database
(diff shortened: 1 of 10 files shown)


src/kubernetes/server/server-deployment.yaml
```diff
@@ -64,7 +67,7 @@ spec:
         - name: SPRING_CLOUD_KUBERNETES_SECRETS_ENABLE_API
           value: 'false'
         - name: SPRING_CLOUD_KUBERNETES_SECRETS_PATHS
-          value: /etc/secrets
+          value: /etc/secrets,/workspace/runtime/secrets/ctr-registry
```
chrisjs: probably not needed with the above change

chrisjs (changes requested):

src/kubernetes/server/server-deployment.yaml
```diff
@@ -25,6 +25,9 @@ spec:
           - name: database
             mountPath: /etc/secrets/database
             readOnly: true
+          - name: ctr-registry
+            mountPath: /workspace/runtime/secrets/ctr-registry
```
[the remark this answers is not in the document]
tzolov: This was mistakenly committed. I removed it form the PR. The  server-deployment.yaml should be changed by this PR

src/kubernetes/server/server-deployment.yaml
tzolov: same, this was mistakenly committed and I removed it form the PR. The  server-deployment.yaml should be changed by this PR

chrisjs:
merged in 2f7ef64cacd075d6faf47dd5ee3cbe497b3dfeb4 & backported in 168cb9ebb81b1242894977a68707c133f35e65d9

PR #4194, opened 2020-10-09
Handle unresolved placeholders for metrics copy

### LINKED ISSUE #4193: Handle unresolved placeholders in Metrics replication processor
**Description:**

The [MetricsReplicationEnvironmentPostProcessor](https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/MetricsReplicationEnvironmentPostProcessor.java) is responsible to replicate any SCDF metrics configuration to the `s.c.d.applicaitonProperties.stream.` and `s.c.d.applicaitonProperties.task.` allowing later to use the same metrics endpoint without having to put those configuration manually. 

By default the `EnvironmentPostProcessor` and the `ConfigurableEnvironment` throws exception if the property value being copied is a placeholder expression like `${wavefront-api-secret}.  Later is expected as the EnvironmentPostProcessor is called before any placeholder is resolved. 

The solution is to set `ConfigurableEnvironment#setIgnoreUnresolvableNestedPlaceholders(true)` for the duration for property replication. 

Note: In this case we would like SCDF to resolve those properties before it sends them to the deployed Stream or Task apps!

**Release versions:**
This problem affects the 2.6.x and the master branches.

tzolov:
- Ignore the unresolvable placeholders in the MetricsReplicationEnvironmentPostProcessor. The check is reenabled after the processor is done.
 - Add tests

 Resolves #4193

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/MetricsReplicationEnvironmentPostProcessor.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/MetricsReplicationEnvironmentPostProcessor.java
@@ -22,6 +22,7 @@ import java.util.Properties;
 import java.util.function.Consumer;
 
 import io.micrometer.prometheus.rsocket.autoconfigure.PrometheusRSocketClientProperties;
+import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Cherry-picked and pushed into 2.6.x as 270c3a96153eed48039ff3565ab7e80110311dbb

Issue #4189, opened 2020-10-08
DefaultAppRegistryService.save needs to verify that app name is less than 255


cppwfs:
Currently if a user creates a name that is greater than 255 chars in length the message that is returned does not tell them what happened.  We need to check the app name is greater than 255 and return a meaningful message.   This will occur if using RESTful API or shell.

siddhantsorann:
Hey @cppwfs can I pick this up?

cppwfs:
Its yours!!!!   
Thanks @siddhantsorann !

siddhantsorann:
Thanks @cppwfs 
Can we have the check in AppRegistryController? We have a function there that validates the application name, can add this check there.

cppwfs:
sure.

PR #4195, opened 2020-10-13
App name length 255

siddhantsorann:
This PR adds a feature as part of #4189 

The following are the changes made - 
- Added validation to check if app name length is > 255 characters, returning a meaningful message in that case
- Added test case to check for failure if app name length is. > 255 characters.


jvalkeal:
Looks good, I created #4196 to change how errors are send back to client.

jvalkeal:
Merged per 52049bb69d872d50898d326a5e240caace622de6, thx!

PR #4203, opened 2020-10-16
Added validation for task name length depending on launch platform

### LINKED ISSUE #4190: Need to verify that task def name < 64 char when launching on CF or K8s
When user launches a task definition and the user selects a platform that is CF or K8s be sure to verify that the task definition name is less that 64 characters.   If not return a meaningful error message vs. the one returned from the platform.

siddhantsorann:
This PR fixes #4190

- Added validation when the task is launched depending on the platform launched in.
- Returns a meaningful message if the task was > 63 characters.


siddhantsorann:
@jvalkeal Is this alright?

jvalkeal:
Yes, that's now a clean PR.

jvalkeal:
lgtm

PR #4206, opened 2020-10-19
Lower the case of constraint foreign key

herman-d:
Context:
An error with error code [1769]; Foreign key 'TASK_METADATA_FK' references invalid column 'TASK_EXECUTION_ID' in referencing table 'task_execution_metadata'. was thrown when flyway executing migration script for SQL Server with Case Sensitive collation.

It was due to the script use upper case 'TASK_EXECUTION_ID' key referenced to a field in the task_execution_metadata table which is created in lower case 'task_execution_id'


pivotal-issuemaster:
@herman-d Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-dataflow&pullRequestId=4206)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-dataflow&pullRequestId=4206) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@herman-d Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-dataflow&pullRequestId=4206)!

jvalkeal:
Flyway is actually using https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/db/migration/sqlserver/V2__Add_Descriptions_And_OriginalDefinition.java

What is your sqlserver version and OS as we run some tests on MS's ubuntu version where all these has been fine and haven't seen any reports until now.

jvalkeal:
I assume sqlserver were installed to be case sensitive as it's insensitive on default. Also to be on a safe side, It should get changed for other db's as well.

jvalkeal:
Created an issue for this so that it's easier to find in a future. Looks like we should also modify postgresql/mysql scripts, but that's something I can "polish" during a merge.

jvalkeal:
Merged per 3b5c221eb9a088c9fc70680fcc5c33d380d219d5

PR #4207, opened 2020-10-19
Removed the type hint from the job params on job restarts.

### LINKED ISSUE #4204: Users can't restart failed batch job on CF Platform
When a user is launching a CTR  on CF with `increment-instance-enabled` set `true` (default for dataflow) or launch a Spring Batch Job that has an incrementer and that job fails.  If the user attempts restart the app, it will fail with the following exception:
```
 2020-10-15T10:24:09.31-0600 [APP/TASK/wctr8/0] ERR bash: -c: line 0: syntax error near unexpected token `('
   2020-10-15T10:24:09.31-0600 [APP/TASK/wctr8/0] ERR bash: -c: line 0: `JAVA_OPTS="-agentpath:$PWD/.java-buildpack/open_jdk_jre/bin/jvmkill-1.16.0_RELEASE=printHeapHistogram=1 -Djava.io.tmpdir=$TMPDIR -XX:ActiveProcessorCount=$(nproc) -Djava.ext.dirs=$PWD/.java-buildpack/container_security_provider:$PWD/.java-buildpack/open_jdk_jre/lib/ext -Djava.security.properties=$PWD/.java-buildpack/java_security/java.security $JAVA_OPTS" && CALCULATED_MEMORY=$($PWD/.java-buildpack/open_jdk_jre/bin/java-buildpack-memory-calculator-3.13.0_RELEASE -totMemory=$MEMORY_LIMIT -loadedClasses=18610 -poolType=metaspace -stackThreads=250 -vmOptions="$JAVA_OPTS") && echo JVM Memory Configuration: $CALCULATED_MEMORY && JAVA_OPTS="$JAVA_OPTS $CALCULATED_MEMORY" && MALLOC_ARENA_MAX=2 SERVER_PORT=$PORT eval exec $PWD/.java-buildpack/open_jdk_jre/bin/java $JAVA_OPTS -cp $PWD/. org.springframework.boot.loader.JarLauncher --dataflow-server-uri=https://dataflow-fb629132-add0-4cff-ac5e-c0965b82c29d.cfapps-31.slot-59.pez.vmware.com run.id(long)=1 --spring.cloud.data.flow.platformname=default --spring.cloud.task.executionid=92 --spring.cloud.data.flow.taskappname=composed-task-runner'
```
This is because dataflow retrieves the job parameters to relaunch the job the run.id (which is added by the incrementer) contains the type information in the key `run.id(long)` and the build pack fails to parse the `(` unless its triple escaped `\\\ ` .

cppwfs:
resolved #4204

The original intent was to use the entire key that was returned from batch.
However when relaunching a job from CF it will fail.  That is because the build pack requires a triple escape to use `(` character ie.`\\\(`
This also causes a problem when passing this value as a property when using shell entrypoint for k8s.

So since this is just a hint to batch, we will remove this info and let batch figure out the correct type.

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskJobService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskJobService.java
@@ -231,7 +231,7 @@ public class DefaultTaskJobService implements TaskJobService {
 					}
 				}
 				if(!existsFlag) {
-					result.add(String.format("%s(%s)=%s", key, jobParametersMap.get(key).getType().toString().toLowerCase(), jobParameters.getString(key)));
+					result.add(String.format("%s=%s", key,  jobParameters.getString(key)));
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


ilayaperumalg:
LGTM, merging.

ilayaperumalg:
I think this needs to be back ported into 2.6.x

ilayaperumalg:
Cherry-picked and pushed the commit into 2.6.x as 155b6859f76aa6abd2c4439eedda1bbbbfbe7181

PR #4211, opened 2020-10-21
Updated docs to discuss maximum char length for definition names.

### LINKED ISSUE #4191: Add doc to state max lengths for stream def names and task def names for each platform
Create a table similar to below 

Also mention that Stream apps that are deployed have a max len of 58 as shown below:

cppwfs:
resolves #4191


sabbyanandan (approved):
Let's go!

dturanski:
Closed via 7719d2762 . I added a brief note in tasks.doc.

PR #4230, opened 2020-10-31
Add validation for sort keys

### LINKED ISSUE #4229: Add validation for sort keys
We need to explicitly add validation for the sort keys being used in the SQL queries to avoid SQL injection.

ilayaperumalg:
Resolves #4229


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/JdbcDataflowTaskExecutionMetadataDao.java
```diff
@@ -118,7 +119,8 @@ public void save(TaskExecution taskExecution, TaskManifest manifest) {
 	@Override
 	public TaskManifest getLatestManifest(String taskName) {
 		Map<String, Order> sortKeys = new HashMap<>(1);
-		sortKeys.put("e.TASK_EXECUTION_ID", Order.DESCENDING);
+		// Make sure to add the sort keys into
+		sortKeys.put(SqlPagingQueryUtils.getSupportedKey("TASK_EXECUTION_ID"), Order.DESCENDING);
```
jvalkeal, line 123: I don't quite understand how this would cause any changes?

ilayaperumalg: Currently, the **only** place I see the sort parameter being used is at this code and hence added this to be part of the validation of sort keys.

ilayaperumalg:
Closing this PR after discussion (as this fix can be done at a different level to mitigate other sort parameters validation).

PR #4232, opened 2020-11-03
Add imagePullSecret for CTR

### LINKED ISSUE #4009: Add an "imagePullSecret" property for composedTaskRunner in TaskConfigurationProperties
**Problem description:**
We just added a property for composedTaskRunnerUri in TaskConfigurationProperties. Since this image might be in a private registry it would make sense to have an "imagePullSecret" property for this uri as well.

**Solution description:**
Add an "imagePullSecret" property for composedTaskRunner in TaskConfigurationProperties.

We could restructure the configuration to match:

```
spring:
  cloud:
    dataflow:
      task:
        composedTaskRunner: 
          uri: docker://registry.example.com/spring-cloud-dataflow-composed-task-runner
          imagePullSecret: example-regcred
```

It is possible to set "imagePullSecret" for all task apps using:

```
spring:
  cloud:
    dataflow:
      task:
        platform:
          kubernetes:
            accounts:
              default:
                imagePullSecret: example-regcred
```

Since CTR is now part of the distribution and separate from any task apps that are registered it is likely that they are in different registries or use different secrets. It would be better to have a specific entry for the secret to use for the CTR image. 

**Additional context:**

chrisjs:
Resolves #4009

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/TaskConfigurationProperties.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/TaskConfigurationProperties.java
@@ -96,7 +96,9 @@ public class TaskConfigurationProperties {
 	@Deprecated
 	public boolean isUseUserAccessToken() {
 		logDeprecationWarning();
-		return this.composedTaskRunnerConfigurationProperties.isUseUserAccessToken();
+
+		return this.composedTaskRunnerConfigurationProperties.isUseUserAccessToken() == null ? false :
+				this.composedTaskRunnerConfigurationProperties.isUseUserAccessToken();
 	}
 
 	@Deprecated
@@ -168,7 +170,9 @@ public class TaskConfigurationProperties {
 	}
 
 	void setComposedTaskRunnerConfigurationProperties(ComposedTaskRunnerConfigurationProperties
-															  composedTaskRunnerConfigurationProperties) {
-		this.composedTaskRunnerConfigurationProperties = composedTaskRunnerConfigurationProperties;
+							composedTaskRunnerConfigurationProperties) {
+		if (composedTaskRunnerConfigurationProperties != null) {
+			this.composedTaskRunnerConfigurationProperties = composedTaskRunnerConfigurationProperties;
+		}
 	}
 }
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/TaskConfigurationProperties.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/TaskConfigurationProperties.java
@@ -16,32 +16,34 @@
 
 package org.springframework.cloud.dataflow.server.service.impl;
 
-import javax.validation.constraints.NotBlank;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.cloud.dataflow.core.DataFlowPropertyKeys;
+import org.springframework.util.StringUtils;
 import org.springframework.validation.annotation.Validated;
-
 
 /**
  * Properties used to define the behavior of tasks created by Spring Cloud Data Flow.
  *
  * @author Glenn Renfro
  * @author David Turanski
+ * @author Chris Schaefer
  */
 @Validated
 @ConfigurationProperties(prefix = TaskConfigurationProperties.TASK_PREFIX)
 public class TaskConfigurationProperties {
+	private static final Logger LOGGER = LoggerFactory.getLogger(TaskConfigurationProperties.class);
 
 	public static final String TASK_PREFIX = DataFlowPropertyKeys.PREFIX + "task";
 
+	@Deprecated
 	public static final String COMPOSED_TASK_RUNNER_NAME = "composed-task-runner";
 
-	/**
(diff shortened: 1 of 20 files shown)


chrisjs:
below are the manual test cases i went through. i did these for both regular CTR launches and scheduled ones and they seem to work.

pre-reqs:
---
* k8s secret (or however you wanna create it):

```
kubectl create secret docker-registry regcred --docker-server=docker.io --docker-username=YOUR_DOCKER_USER --docker-password=YOUR_DOCKER_PW --docker-email=YOUR_DOCKER@EMAIL
```

* a CTR image in a private repo that gets referenced below in replacement of PRIVATE_REPO

test cases:
---
NOTE: `SPRING_CLOUD_DATAFLOW_TASK_COMPOSEDTASKRUNNER_URI` is currently committed in this PR as `SPRING_CLOUD_DATAFLOW_TASK_COMPOSED_TASK_RUNNER_URI` has been deprecated.

* "new" property namespace in src/kubernetes/server/server-deployment.yaml without image auth

```
        - name: SPRING_CLOUD_DATAFLOW_TASK_COMPOSEDTASKRUNNER_URI
          value: 'docker://springcloud/spring-cloud-dataflow-composed-task-runner:2.7.0-SNAPSHOT'
```

* "old" property namespace in src/kubernetes/server/server-deployment.yaml without image auth:

```
        - name: SPRING_CLOUD_DATAFLOW_TASK_COMPOSED_TASK_RUNNER_URI
          value: 'docker://springcloud/spring-cloud-dataflow-composed-task-runner:2.7.0-SNAPSHOT'
```

* "new" property namespace in src/kubernetes/server/server-deployment.yaml with image auth (need to make secret and private image):

```
        - name: SPRING_CLOUD_DATAFLOW_TASK_COMPOSEDTASKRUNNER_URI
          value: 'docker://PRIVATE_REPO/spring-cloud-dataflow-composed-task-runner:2.7.0-SNAPSHOT'
        - name: SPRING_CLOUD_DATAFLOW_TASK_COMPOSEDTASKRUNNER_IMAGE_PULL_SECRET
          value: 'regcred'

```

now use the CM instead so - remove the above env-vars from from src/kubernetes/server/server-deployment.yaml

* update src/kubernetes/server/server-config.yaml without image auth:

```
    spring:
      cloud:
        dataflow:
          task:
            composedTaskRunner:
              uri: 'docker://springcloud/spring-cloud-dataflow-composed-task-runner:2.7.0-SNAPSHOT'
```

* update src/kubernetes/server/server-config.yaml with image auth:

```
    spring:
      cloud:
        dataflow:
          task:
            composedTaskRunner:
              uri: 'docker://PRIVATE_REPO/spring-cloud-dataflow-composed-task-runner:2.7.0-SNAPSHOT'
              imagePullSecret: regcred
```

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/TaskConfigurationProperties.java
```diff
@@ -16,32 +16,34 @@
...
 
+	@Deprecated
 	public static final String COMPOSED_TASK_RUNNER_NAME = "composed-task-runner";
 
-	/**
-	 * The task application uri to be used for the composed task runner.
-	 */
-	@NotBlank
-	private String composedTaskRunnerUri = "composed-task-runner";
+	@Autowired
```
cppwfs, line 45: In some of our code we make the assumption that composedTaskRunnerConfigurationProperties can be null.    With that assumption should we mark this Autowired as optional.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/TaskConfigurationProperties.java
```diff
@@ -146,4 +155,20 @@ public boolean isUseKubernetesSecretsForDbCredentials() {
...
+
+	private void logDeprecationWarning(String newMethodName) {
+		String callingMethodName = Thread.currentThread().getStackTrace()[2].getMethodName();
+		LOGGER.warn(this.getClass().getName() + "." + callingMethodName
+				+ " is deprecated. Please use " + ComposedTaskRunnerConfigurationProperties.class.getName() + "."
+				+ (newMethodName != null ? newMethodName : callingMethodName));
+	}
+
+	void setComposedTaskRunnerConfigurationProperties(ComposedTaskRunnerConfigurationProperties
+															  composedTaskRunnerConfigurationProperties) {
```
cppwfs, line 171: May want to reformat this code.    Tabbing is not lining up.   Your friendly neighborhood nitpick.

chrisjs:
updated! the tabs should be just splendid now :100:

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/service/impl/TaskServiceUtilsTests.java
```diff
@@ -309,6 +309,48 @@ public void testImagePullSecretNullCTRProperties() {
 				taskDeploymentProperties.containsKey("deployer.composed-task-runner.kubernetes.imagePullSecret"));
 	}
 
+	@Test
+	public void testUseUserAccessTokenFromCTRProps() {
```
cppwfs, line 313: We should also have a test that  test if the user sets access token to false.   i.e. composedTaskRunnerConfigurationProperties.setUseUserAccessToken(false);

chrisjs: added

cppwfs:
LGTM

cppwfs:
Rebased Squashed merged.

PR #4237, opened 2020-11-03
Validate sort keys for task executions

### LINKED ISSUE #4229: Add validation for sort keys

jvalkeal:
- Now simply does a check for sort field that
  it is what can possibly work and alignes i.e.
  what we can request from UI.
- Fixes #4229

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskExecutionController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskExecutionController.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2019 the original author or authors.
+ * Copyright 2016-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -43,6 +43,7 @@ import org.springframework.cloud.task.repository.TaskExplorer;
 import org.springframework.data.domain.Page;
 import org.springframework.data.domain.PageImpl;
 import org.springframework.data.domain.Pageable;
+import org.springframework.data.domain.Sort;
 import org.springframework.data.web.PagedResourcesAssembler;
 import org.springframework.hateoas.PagedModel;
 import org.springframework.hateoas.server.ExposesResourceFor;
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskExecutionController.java
```diff
@@ -87,6 +88,9 @@
 
 	private final TaskSanitizer taskSanitizer = new TaskSanitizer();
 
+	private static final List<String> allowedSorts = Arrays.asList("TASK_EXECUTION_ID", "TASK_NAME", "START_TIME",
+			"END_TIME", "EXIT_CODE");
```
mminella, line 92: In SCT, I included the ability to sort by any field in the table...Are these the only fields we ever allow the user to sort by in the UI?

jvalkeal: Yeah, I took those from there. If there's other fields which work we could add those as well.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskExecutionController.java
```diff
@@ -87,6 +88,9 @@
 
 	private final TaskSanitizer taskSanitizer = new TaskSanitizer();
 
+	private static final List<String> allowedSorts = Arrays.asList("TASK_EXECUTION_ID", "TASK_NAME", "START_TIME",
```
cppwfs, line 91: Do we want to support lowercase?

jvalkeal: Ah yeah, I think we should

cppwfs:
LGTM    Thank You!

cppwfs:
Rebased, Squashed, Merged

ilayaperumalg:
Back ported the changes into 2.6.x as 8e00d35c850dd5f1be24ab078d28595d7c700046

PR #4241, opened 2020-11-09
Enable running ITs with k8s on GEK minikube

### LINKED ISSUE #4239: Adjust IT tests to allow running against k8s
**Description:**

The DataFlowIT tests similarly to the ATs use the apps URL to retrieve their logs. 
Apparently the way how the app urls are provided by the k8s deployer has changed and later has been reflected in the ATs. 
The ITs though still attempt the older app url retrieval, which is not applicable anymore. 

The new approach relays on the `url` attribute exposed by the k8s deployer in case of LB infra.

tzolov:
- Fix the app url resolution using the k8s deployer's "url" attribute. Later requires a LB to be exposed.
  - Add README instructions to run the IT test agains remote k8s/GKE or local minikube. Add instructions to install `metallb`.

  Resolves #4239

### DIFF
--- a/spring-cloud-dataflow-server/src/test/java/org/springframework/cloud/dataflow/integration/test/DataFlowIT.java
+++ b/spring-cloud-dataflow-server/src/test/java/org/springframework/cloud/dataflow/integration/test/DataFlowIT.java
@@ -742,9 +742,9 @@ public class DataFlowIT {
 				.put("app.*.management.endpoints.web.exposure.include", "*")
 				.put("app.*.spring.cloud.streamapp.security.enabled", "false");
 
-		// TODO this misterious 80 port set for K8s comes from the ATs and i'm not sure why. But without it partition tests fail.
 		if (this.runtimeApps.getPlatformType().equalsIgnoreCase(RuntimeApplicationHelper.KUBERNETES_PLATFORM_TYPE)) {
 			propertiesBuilder.put("app.*.server.port", "80");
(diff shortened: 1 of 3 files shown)


chrisjs (commented):
added a couple suggestions

spring-cloud-dataflow-server/src/test/java/org/springframework/cloud/dataflow/integration/test/DataFlowIT.java
```diff
@@ -742,9 +742,9 @@ public void streamWithConfigServer() {
 				.put("app.*.management.endpoints.web.exposure.include", "*")
 				.put("app.*.spring.cloud.streamapp.security.enabled", "false");
 
-		// TODO this misterious 80 port set for K8s comes from the ATs and i'm not sure why. But without it partition tests fail.
 		if (this.runtimeApps.getPlatformType().equalsIgnoreCase(RuntimeApplicationHelper.KUBERNETES_PLATFORM_TYPE)) {
 			propertiesBuilder.put("app.*.server.port", "80");
```
chrisjs, line 746: re: port 80 comment above, likely at some point somebody simply wanted to use port 80. the current AT's use 8080 which was only changed from 80 since there were some issues running in containers that had non-privileged access. at anyrate, no action needed as its simply a port to use. id be curious to know why partitions would fail without this, actually looks like in k8s deployer at least, if no `server.port` is defined, it defaults to 8080 anyway -- but feel free to ping on slack if wanting to discuss

chrisjs:
merged in 847eacb0598d230bb203afaaf4a41b125014c572

PR #4249, opened 2020-11-12
Add doc on how to set CTR properties via dataflow.

cppwfs:
resolves 4248


jvalkeal:
I think we need a bit more overhaul to section https://docs.spring.io/spring-cloud-dataflow/docs/2.7.0-SNAPSHOT/reference/htmlsingle/#spring-cloud-dataflow-composed-tasks

It mentions `spring.cloud.dataflow.task.composed.task.runner.uri` which I believe is wrong where I think `composed.task.runner` should be without dots.

Same story for `dataflow.server.uri` as these dots just confuse reader.

It also mentions `spring.cloud.dataflow.server.uri` and I'm not sure if that even used if we now have that ctr uri property.

cppwfs:
@jvalkeal PR updated.  Great catches.
The `spring.cloud.dataflow.server.uri` is used to specify the location of the dataflow instance that CTR is sending the commands.

jvalkeal:
Merged per 4accae7bdd77303db1fd2268743728a031ed17fb

PR #4270, opened 2020-12-01
Expose compose task runner options

### LINKED ISSUE #4257: Expose compose task runner options
As ctr app is now defined within a server configuration and not as a registered app we don't have a direct way from UI to ask its properties. This relates to #1515 where for ctr task it'd be nice to show possible options for it.

We should add a new rest endpoint which exposes options similarly what we can do with registered apps.

jvalkeal:
- Add boot metadata to ctr jar
- Add controller /tasks/ctr/options to expose
  those from a server to get used by UI.
- Fixes #4257

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowControllerAutoConfiguration.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowControllerAutoConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2019 the original author or authors.
+ * Copyright 2016-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -77,6 +77,7 @@ import org.springframework.cloud.dataflow.server.controller.StreamDefinitionCont
 import org.springframework.cloud.dataflow.server.controller.StreamDeploymentController;
 import org.springframework.cloud.dataflow.server.controller.StreamLogsController;
 import org.springframework.cloud.dataflow.server.controller.StreamValidationController;
+import org.springframework.cloud.dataflow.server.controller.TaskCtrController;
 import org.springframework.cloud.dataflow.server.controller.TaskDefinitionController;
 import org.springframework.cloud.dataflow.server.controller.TaskExecutionController;
 import org.springframework.cloud.dataflow.server.controller.TaskLogsController;
@@ -241,6 +242,12 @@ public class DataFlowControllerAutoConfiguration {
 					streamService,
 					appRegistry,
 					metadataResolver, appRegistryFJPFB, streamDefinitionService, appRegistrationAssemblerProvider);
+		}
+
+		@Bean
+		public TaskCtrController tasksCtrController(ApplicationConfigurationMetadataResolver metadataResolver,
+				TaskConfigurationProperties taskConfigurationProperties, AppResourceCommon appResourceCommon) {
+			return new TaskCtrController(metadataResolver, taskConfigurationProperties, appResourceCommon);
 		}
 
 		@Bean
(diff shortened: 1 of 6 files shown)


jvalkeal:
Right when starting from empty maven repo we don't yet have ctr jar with metadata when server-core is tested.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskCtrController.java
```diff
@@ -0,0 +1,75 @@
...
+			TaskConfigurationProperties taskConfigurationProperties,
+			ComposedTaskRunnerConfigurationProperties composedTaskRunnerConfigurationProperties,
+			AppResourceCommon appResourceCommon) {
+		this.metadataResolver = metadataResolver;
+		this.taskConfigurationProperties = taskConfigurationProperties;
+		this.composedTaskRunnerConfigurationProperties = composedTaskRunnerConfigurationProperties;
+		this.appResourceCommon = appResourceCommon;
+	}
+
+	@RequestMapping(value = "/options", method = RequestMethod.GET)
```
cppwfs, line 59: Probably need to add something like `root.add(WebMvcLinkBuilder.linkTo(WebMvcLinkBuilder.methodOn(TaskCtrController.class).options()).withRel("tasks/ctr/options"));` to https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/RootController.java#L176

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowControllerAutoConfiguration.java
```diff
@@ -243,6 +245,15 @@ public AppRegistryController appRegistryController(
 					metadataResolver, appRegistryFJPFB, streamDefinitionService, appRegistrationAssemblerProvider);
 		}
 
+		@Bean
+		public TaskCtrController tasksCtrController(ApplicationConfigurationMetadataResolver metadataResolver,
```
cppwfs, line 249: Would this fit better in the TaskEnabledConfiguration of the DataFlowControllerAutoConfiguration?   In that the CTR will not be active if tasks are not active.   Thoughts?

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/TaskCtrControllerTests.java
```diff
@@ -0,0 +1,78 @@
...
+import static org.mockito.ArgumentMatchers.any;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
+import static org.springframework.test.web.servlet.result.MockMvcResultHandlers.print;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.jsonPath;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest(classes = TestDependencies.class)
+@AutoConfigureTestDatabase(replace = Replace.ANY)
+public class TaskCtrControllerTests {
```
cppwfs, line 50: @author tag.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/TaskCtrControllerTests.java
```diff
@@ -0,0 +1,78 @@
...
+		p.setId("oauth2-client-credentials-scopes");
+		p.setName("oauth2-client-credentials-scopes");
+		List<ConfigurationMetadataProperty> props = Arrays.asList(p);
+		Mockito.when(metadataResolver.listProperties(any())).thenReturn(props);
+	}
+
+	@Test
+	public void testOptions() throws Exception {
+		mockMvc.perform(get("/tasks/ctr/options").accept(MediaType.APPLICATION_JSON))
+			.andDo(print())
```
cppwfs, line 74: Probably can remove the print()

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskCtrController.java
jvalkeal: I think we want to keep this away from a "public eye" as its solely made for UI.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowControllerAutoConfiguration.java
jvalkeal: Yeah, you're right.

cppwfs:
LGTM

cppwfs:
Rebased, squashed, Merged.

PR #4274, opened 2020-12-02
Remove remains of ctr app name

### LINKED ISSUE #4273: Remove defunc composedTaskRunnerName task launch shell option
When ctr app went from registered app to server setting, this option were left behind into shell and is no use today.

jvalkeal:
- Remove ctr app from task shell launch
- Remove its remains from TaskTemplate and tests
- Fixes #4273


cppwfs:
Shouldn't we deprecate this first?   Thoughts?

jvalkeal:
It should have been removed at a same time when ctr app moved from being a registered app. That option in a shell doesn't do anything.

cppwfs:
LGTM

cppwfs:
Rebased Merged.

PR #4276, opened 2020-12-02
add stream log to java DSL

### LINKED ISSUE #4275: Add stream log and stream logs by app name support to java DSL
**Problem description:**
According  to [API docs](https://docs.spring.io/spring-cloud-dataflow/docs/current/reference/htmlsingle/#api-guide-resources-stream-logs-by-app-name) there's an API to get the stream log and log for a specific app within a stream.

This is  missing from the java DSL 

**Solution description:**
Add endpoints for java DSL

N/A

**Additional context:**
N/A

DemianTinkiel:
Fixes #4275

### DIFF
--- a/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/StreamTemplate.java
+++ b/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/StreamTemplate.java
@@ -59,6 +59,9 @@ public class StreamTemplate implements StreamOperations {
 
 	private static final String VALIDATION_REL = "streams/validation";
 
+	private static final String LOGS_REL = "streams/logs/{streamName}/{appName}";
+
+
 	private static final String VALIDATION_RELATION_VERSION = "1.7.0";
 
 	private final RestTemplate restTemplate;
@@ -73,6 +76,8 @@ public class StreamTemplate implements StreamOperations {
 
 	private final Link validationLink;
 
+	private final Link logsLink;
+
 	private final String dataFlowServerVersion;
 
 	StreamTemplate(RestTemplate restTemplate, RepresentationModel<?> resources, String dataFlowServerVersion) {
@@ -81,6 +86,7 @@ public class StreamTemplate implements StreamOperations {
 		Assert.notNull(resources.getLink(DEFINITION_REL), "Definition relation is required");
 		Assert.notNull(resources.getLink(DEPLOYMENTS_REL), "Deployments relation is required");
 		Assert.notNull(resources.getLink(DEPLOYMENT_REL), "Deployment relation is required");
+		Assert.notNull(resources.getLink(LOGS_REL), "Logs relation is required");
 
 		if (VersionUtils.isDataFlowServerVersionGreaterThanOrEqualToRequiredVersion(
 				VersionUtils.getThreePartVersion(dataFlowServerVersion),
@@ -95,6 +101,7 @@ public class StreamTemplate implements StreamOperations {
 		this.definitionLink = resources.getLink(DEFINITION_REL).get();
 		this.deploymentLink = resources.getLink(DEPLOYMENT_REL).get();
 		this.validationLink = resources.getLink(VALIDATION_REL).get();
+		this.logsLink = resources.getLink(LOGS_REL).get();
 
 	}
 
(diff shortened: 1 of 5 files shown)


pivotal-issuemaster:
@DemianTinkiel Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-dataflow&pullRequestId=4276)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-dataflow&pullRequestId=4276) to manually synchronize the status of this Pull Request.

sabbyanandan:
Nice work, @DemianTinkiel! We will have a look shortly.

pivotal-issuemaster:
@DemianTinkiel Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-dataflow&pullRequestId=4276)!

spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/StreamTemplate.java
```diff
@@ -155,6 +162,16 @@ public void scaleApplicationInstances(String streamName,
 		restTemplate.postForObject(url, properties, Object.class);
 	}
 
+	@Override
+	public String streamExecutionLog(String streamName) {
+		return restTemplate.getForObject(logsLink.expand(streamName, null).getHref(), String.class);
```
ilayaperumalg, line 167: Have you tried this with an existing stream with applications deployed? I wonder if we can use the explicit `streams/logs/{streamName}` endpoint for this.

ilayaperumalg:
LGTM

DemianTinkiel:
Hi @ilayaperumalg @sabbyanandan  @tzolov , is there anything else I should do to get the PR merged in? Also if you don't mind me asking, when can I expect it released? I understand you have release cycles to adhere to, I just want to know when to start depending on it the "official way" instead of my own copy. Thanks!

tzolov:
Hi @DemianTinkiel, thanks for the very useful contribution! 

I've been reviewing but got puzzled by the: `Stream#logs(StreamApplication app)` syntax and its corresponding `streams/logs/{streamName}/{appName}` REST API. 

From what i can see the `{appName}` is not a `StreamApplication` but looks more like a runtime app name, (e.g. the runtime app names returned by the `/runtime/streams`).

IMO,  this is a problem of the `streams/logs/{streamName}/{appName}` API   <<< @ilayaperumalg 

Still until the ambiguity in the REST API is streamlined we should use a parameter like `String runtimeAppName` instead of `StreamApplication app)`. What do you think?

tzolov:
hold on @DemianTinkiel , 
i think we might be able to fix this within the Stream.getLogs(app) implementation itself. I believe we can retrieve the app's `deploymentId` (used by the REST API) form the StreamApplication name. I will give it a try.  
If it works the DSL will provide a proper abstraction,  despite the confusing rest api underneath.

DemianTinkiel:
Hi @tzolov, thanks for the review.  I don't mind making the changes to the `StreamTemplate ` 
but are you sure about
> Still until the ambiguity in the REST API is streamlined we should use a parameter like `String runtimeAppName` instead of `StreamApplication app)`. What do you think?

?
That would make `Stream#logs(StreamApplication)` the odd duck in relation to the other fluent APIs.

I see that there is `Stream#runtimeApps` (seems to only be used for testing?)`. Perhaps we can do something with that. Let me know what you prefer

tzolov:
> That would make Stream#logs(StreamApplication) the odd duck in relation to the other fluent APIs.
Indeed but this oddness comes from the existing `streams/logs/{streamName}/{appName}` implementation. The `{appName}` is in fact a `deploymentId`.  
The proper fix would be for the REST API impl to retrieve the `deploymentId` internally using the stream and app names. 
But as this will be a breaking API change i don't want to rush it. 

In the meantime (as i've suggested above) we can do the appName to deploymentId conversion inside the DSL logs implementation instead. So the user can do `stream.logs(new StreamApplication("transformer")`.  
I am working on this right now and will share the update shortly. 

The `Stream#runtimeApps` is unfortunately a yet another hack that surface the need of a `AppRuntime` abstraction in the DSL.

tzolov:
@DemianTinkiel check here the workaround i've added on top of your PR: https://github.com/tzolov/spring-cloud-dataflow/commit/40906bb4ad4d1218daaa2be58a51391d705c7d85

If you are happy with those changes i am ready to merge the PR.

DemianTinkiel:
@tzolov   looks good, thank you. :+1:

tzolov:
Updated, rebased and merged at 55936f24a280b14406370154aa87a0faca896149

@DemianTinkiel thank you for the contribution!

Issue #4303, opened 2020-12-29
Register cron expression based on timezone of SCDF Server.


23king:
Problem description:

The timezone between Cloud Platform's cron expression and Spring Cloud Data Flow Server's cron expression does not match.

Solution description:
In the k8s configMap, specify the timezone of the k8s platform. And when registering the schedule, modify the cron expression by referring to the timezone of the k8s platform.

apiVersion: v1
kind: ConfigMap
metadata:
  name: scdf-server
  labels:
    app: scdf-server
data:
  application.yaml: |-
    management:
      metrics:
        export:
          prometheus:
            enabled: true
            rsocket:
              enabled: true
              host: prometheus-proxy
              port: 7001
    spring:
      cloud:
        dataflow:
          metrics.dashboard:
            url: 'https://grafana:3000'
          task:
            platform:
              **timezone: UTC**
              kubernetes:
                accounts:
                  default:
                    limits:
                      memory: 1024Mi
      datasource:
        url: jdbc:mysql://${MYSQL_SERVICE_HOST}:${MYSQL_SERVICE_PORT}/mysql
        username: root
        password: ${mysql-root-password}
        driverClassName: org.mariadb.jdbc.Driver
        testOnBorrow: true
        validationQuery: "SELECT 1"

Description of alternatives:
N/A

Additional context:
I have already modified the code. We'll share your edits through Pull Requests.

jvalkeal:
Could you restructure this issue in a more consumable description so that we'd know what you are trying to accomplish. I'd very much like to see a use case, problem and then a possible discussion about a possible solution to a problem.

Unfortunately a related PR is contains unneeded changes and a change to a dependencies(which we don't accept easily), so I need to reject it for now.

23king:
AS-IS:
If the timezone of the scdf server is asia/seoul and the timezone of k8s is UTC,
When registering a schedule with cron, if it is registered as a timezone based on asia/seoul, the operation operates in the UTC timezone, which is the timezone of k8s.

TO-BE:
When the timezone of the scdf server is asia/seoul and the timezone of k8s is UTC
When cron schedule is registered as asia/seoul, cron expression is converted into UTC standard and registered.

It is the same content as issue #4111.

23king:
@jvalkeal  
Please give me feedback. I need this feature wherever I am using it.

sabbyanandan:
[...]

PR #4304, opened 2020-12-29
Fixes gh-4303

23king:
Fixes gh-4303
  add TaskPlatformConfigurationProperties.java (set platform timezone)
  add cronutils dependency (9.1.3)
  
  If the timzone of the platform exists, the cron expression is changed to the cron expression of the platform.
  and
  This changes the cron expression of the platform to the cron expression of the server even when importing the task.


jvalkeal:
Rejecting for now, see more on  #4303.

PR #4310, opened 2021-01-04
CTR will not attempt to connect to SCDF until it is required.

cppwfs:
This allows CTR to record its information to task execution table in the event that the host port is incorrect

Moved Relaxed names so that it can be shared


spring-cloud-dataflow-composed-task-runner/pom.xml
```diff
@@ -53,6 +53,10 @@
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter-batch</artifactId>
 		</dependency>
+		<dependency>
```
jvalkeal, line 56: Would adding this allow to drop `spring-cloud-dataflow-server-core` as a test dependency? I'd really like to get this app to get build before server core.

spring-cloud-dataflow-composed-task-runner/src/main/java/org/springframework/cloud/dataflow/composedtaskrunner/StepBeanDefinitionRegistrar.java
```diff
@@ -214,6 +224,20 @@ private ComposedTaskProperties composedTaskProperties() {
 		return collector.getTaskApps();
 	}
 
+	private String getPropertyValue(String key) {
```
jvalkeal, line 227: I'd advice not trying to use this legacy `RelaxedNames`. Btw in my PR #4314 I partially moved these property bindings into `ComposedTaskRunnerStepFactory` so that we would not need to do manual binding in this registrar where we really can autowire anything.

spring-cloud-dataflow-composed-task-runner/pom.xml
```diff
@@ -24,6 +24,7 @@
 	<packaging>jar</packaging>
 	<properties>
 		<plexus.utils.version>3.3.0</plexus.utils.version>
+		<spring.cloud.common.security>1.6.0-SNAPSHOT</spring.cloud.common.security>
```
jvalkeal, line 27: remove as this should already be in parent pom and only in parent pom.

spring-cloud-dataflow-composed-task-runner/pom.xml
cppwfs: Removed.

spring-cloud-dataflow-composed-task-runner/pom.xml
cppwfs: Removed

spring-cloud-dataflow-composed-task-runner/src/main/java/org/springframework/cloud/dataflow/composedtaskrunner/StepBeanDefinitionRegistrar.java
cppwfs: I had this in the comments on the 2nd commit.  But they are easily hidden.
```
Not sure how to remove the Relaxed names since the registrar creates its own BeanDefinitionBuilder using the ComposedTaskRunnerStepFactory class.
And thus needs to be given the properties at construction time.
```

cppwfs:
Rebased

jvalkeal:
Great now it actually completes when I "misconfigure" dataflow url but question about status:
```
2021-02-05 08:33:10.768  INFO 266232 --- [           main] o.s.batch.core.job.SimpleStepHandler     : Executing step: [ctr1-t1_0]
2021-02-05 08:33:10.870 ERROR 266232 --- [           main] o.s.batch.core.step.AbstractStep         : Encountered an error executing step ctr1-t1_0 in job ctr1

org.springframework.web.client.ResourceAccessException: I/O error on GET request for "http://localhost:9394": Connection refused
...
2021-02-05 08:33:10.872 ERROR 266232 --- [           main] o.s.batch.core.step.AbstractStep         : Exception in afterStep callback in step ctr1-t1_0 in job ctr1

java.lang.IllegalArgumentException: TaskLauncherTasklet did not return a task-execution-id.  Check to see if task exists.
...
2021-02-05 08:33:10.878  INFO 266232 --- [           main] o.s.batch.core.step.AbstractStep         : Step: [ctr1-t1_0] executed in 109ms
2021-02-05 08:33:10.888  INFO 266232 --- [           main] o.s.b.c.l.support.SimpleJobLauncher      : Job: [FlowJob: [name=ctr1]] completed with the following parameters: [{run.id=1}] and the following status: [FAILED] in 138ms
```
What we get:
```
dataflow:>task list
╔═════════╤══════════════════════════════╤═══════════╤═══════════╗
║Task Name│       Task Definition        │description│Task Status║
╠═════════╪══════════════════════════════╪═══════════╪═══════════╣
║ctr1-t1  │t1: timestamp                 │           │UNKNOWN    ║
║ctr1-t2  │t2: timestamp                 │           │UNKNOWN    ║
║ctr1     │t1: timestamp && t2: timestamp│           │COMPLETE   ║
╚═════════╧══════════════════════════════╧═══════════╧═══════════╝

dataflow:>task execution list --name ctr1
╔═════════╤══╤════════════════════════════╤════════════════════════════╤═════════╗
║Task Name│ID│         Start Time         │          End Time          │Exit Code║
╠═════════╪══╪════════════════════════════╪════════════════════════════╪═════════╣
║ctr1     │1 │Fri Feb 05 08:33:10 GMT 2021│Fri Feb 05 08:33:10 GMT 2021│0        ║
╚═════════╧══╧════════════════════════════╧════════════════════════════╧═════════╝
```
Should we get failed status?

cppwfs:
I'm going to give you the dreaded, works as designed answer.   
Since CTR is a batch application it captures the exception and records the batch job as FAILED.   Hence the user can go to the Job Executions tab and see that the batch job has failed and they can restart the job.   But as the task execution is concerned this was a successful boot execution, it just happens that a batch job failed.

Now if the user wants the task execution to fail because the CTR batch job fails they can set the `spring.cloud.task.batch.fail-on-job-failure` property to true.  So in this case when they go to the task execution page they will see an `exit code=1` (And a task definition of `ERROR`).   
Now we can create a new issue that states that anytime a CTR is launched we pass the `--spring.cloud.task.batch.fail-on-job-failure=true` and thus we can see a CTR failure show up as error. (and make it configurable).   Thoughts?

PR #4314, opened 2021-01-06
Overhaul task handling for versions and arguments

### LINKED ISSUE #4258: Explore options to support versions in tasks
It is a little cumbersome to always take default version when launching task as user would expect to be able to easily switch to different versions.

### LINKED ISSUE #4265: Support passing arguments to ctr tasks
Currently arguments with composed task are just passed to ctr itself and its `composed-task-arguments` option passes same arguments to all tasks. 

If having task `t1: timestamp && t2: timestamp` with name `ctr1` we could use similar format than in other parts:
```
app.*=--a=a
app.*=--d=d
app.t1=--b=b
app.t1=--e-e
app.t2=--c=c
app.t2=--f=f
```
`app.<task label>=` would be optional and if not present it'd be just a normal argument going into ctr app itself or to a task in case of simple task(non ctr).

Basically this new format means that if an argument start with `app.ddd=` it would be handled as an special option meant to ctr app itself. There's no need to think breaking changes as any these types of an arguments would have been discarded by ctr app itself.

jvalkeal:
- This commit is focusing two things, firstly able to
  use versions for task apps and secondly able to
  controll ctr tasks arguments better.
- Versions are in format of for simple:
  version.timestamp=2.1.1.RELEASE
- Versions are in format of for ctr:
  version.t1=2.0.0.RELEASE
  version.t2=2.1.1.RELEASE
- Arguments are in format of for simple:
  app.timestamp.0=--timestamp.format=yyyy
- Arguments are in format of for ctr:
  app.t1.0=--timestamp.format=yyyy
  app.t2.0=--timestamp.format=yyyyMM
- Fixes #4258
- Fixes #4265
- There's a few follow-up issues to hand after this, like
  shell is not yet changed and as we now support versions
  which cf is not ready to handle.
  
  You need to do maven install locally as crt app has changes.
  Use with UI PR spring-cloud/spring-cloud-dataflow-ui#1626

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/TaskServiceUtils.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/TaskServiceUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2020 the original author or authors.
+ * Copyright 2018-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -28,6 +28,7 @@ import org.springframework.boot.autoconfigure.jdbc.DataSourceProperties;
 import org.springframework.cloud.dataflow.core.TaskDefinition;
 import org.springframework.cloud.dataflow.core.TaskPlatformFactory;
 import org.springframework.cloud.dataflow.core.dsl.TaskApp;
+import org.springframework.cloud.dataflow.core.dsl.TaskAppNode;
 import org.springframework.cloud.dataflow.core.dsl.TaskNode;
 import org.springframework.cloud.dataflow.core.dsl.TaskParser;
 import org.springframework.cloud.dataflow.server.controller.VisibleProperties;
@@ -91,14 +92,19 @@ public class TaskServiceUtils {
 		Assert.notNull(taskDeploymentProperties, "taskDeploymentProperties must not be null");
 		Assert.notNull(taskNode, "taskNode must not be null");
 		String result = "";
+		Map<String, String> taskAppProperties = new HashMap<>();
 		for (TaskApp subTask : taskNode.getTaskApps()) {
 			result = updateProperties(taskNode, subTask, taskDeploymentProperties, result, "app");
 			result = updateProperties(taskNode, subTask, taskDeploymentProperties, result, "deployer");
 			result = updateVersionProperties(taskNode, subTask, taskDeploymentProperties, result, "version");
+			taskAppProperties.putAll(getTaskAppProperties(taskNode, subTask, taskDeploymentProperties, "app"));
 		}
 		if (result.length() != 0) {
 			taskDeploymentProperties.put("app.composed-task-runner.composed-task-properties", result);
 		}
+		taskAppProperties.entrySet().stream().forEach(e -> {
+			taskDeploymentProperties.put("app.composed-task-runner.composed-task-app-properties." + e.getKey(), e.getValue());
+		});
 		return taskDeploymentProperties;
 	}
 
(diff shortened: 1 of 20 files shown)


jvalkeal:
ctr task app has a test dep to core so it's getting build after, grin....

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/TaskServiceUtils.java
```diff
@@ -237,6 +238,29 @@ private static String updateProperties(TaskNode taskNode, TaskApp subTask, Map<S
...
+		String subTaskName1 = String.format("%s.%s-%s", prefix, taskNode.getName(),
+				(subTask.getLabel() == null) ? subTask.getName() : subTask.getLabel());
+		String subTaskName2 = String.format("%s.%s", prefix,
+				(subTask.getLabel() == null) ? subTask.getName() : subTask.getLabel());
+		// String subTaskName3 = subTask.getLabel() == null ? subTask.getName() : subTask.getLabel();
+		String subTaskName3 = subTask.getName();
+
+		String versionProperties = taskDeploymentProperties.entrySet().stream()
+				.filter(e -> e.getKey().startsWith(subTaskName1) || e.getKey().startsWith(subTaskName2))
+				// .map(e -> e.getKey() + "=" + e.getValue())
```
cppwfs, line 252: This can be removed.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/TaskServiceUtils.java
```diff
@@ -237,6 +238,29 @@ private static String updateProperties(TaskNode taskNode, TaskApp subTask, Map<S
 		return result;
 	}
 
+	private static String updateVersionProperties(TaskNode taskNode, TaskApp subTask, Map<String, String> taskDeploymentProperties,
+			String result, String prefix) {
```
cppwfs, line 242: Since updateVersionProperties only has one prefix we can probably make it a constant in the method.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionServiceTests.java
```diff
@@ -540,6 +592,42 @@ private void setupUpgradeForCommandLineArgsChange() throws IOException {
...
+			  when(taskLauncher.launch(any())).thenReturn("0");
+
+			  Map<String,String> deploymentProperties = new HashMap<>(1);
+
+			  this.taskExecutionService.executeTask(TASK_NAME_ORIG, deploymentProperties, Collections.singletonList("app.demo=--foo=bar"));
+			  TaskManifest lastManifest = this.dataflowTaskExecutionMetadataDao.getLatestManifest(TASK_NAME_ORIG);
+			  assertEquals(2, lastManifest.getTaskDeploymentRequest().getCommandlineArguments().size());
+			  assertEquals("--foo=bar", lastManifest.getTaskDeploymentRequest().getCommandlineArguments().get(0));
+
+			//   this.taskExecutionService.executeTask(TASK_NAME_ORIG, deploymentProperties, Collections.emptyList());
```
cppwfs, line 626: Remove these comments.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionServiceTests.java
```diff
@@ -372,6 +377,52 @@ public void testRestoreAppPropertiesV2() throws IOException {
...
+			assertEquals("default", lastManifest.getPlatformName());
+			assertEquals(1, lastManifest.getTaskDeploymentRequest().getDeploymentProperties().size());
+			assertEquals("1.0.1", lastManifest.getTaskDeploymentRequest().getDeploymentProperties().get("version.timestamp"));
+
+			verify(this.taskLauncher, never()).destroy(TASK_NAME_ORIG);
+		}
+
+		@Test
+		@DirtiesContext
+		public void testSavesRequestedVersionLabel() throws IOException {
```
cppwfs, line 405: may want to add a negative test where they ask for a version that does not exist.

jvalkeal: Prob a follow up issue. Not sure if we do fail fast even on a stream side and would eventually result a runtime failure.

jvalkeal:
btw, I did rebase and force push

cppwfs:
Not a bug in the code but noticed that the display page has problem displaying the result of `--composed-task-app-arguments.app.BBB.0=--timestamp.format=DD=YYYY` as `timestamp.format=DD=YYYY`.  It showed it as `timestamp.format=DD`.   The task execution param was correct.    i.e. this is a UI hiccup that we uncovered.

cppwfs:
LGTM

cppwfs:
Squash Merged.

PR #4324, opened 2021-01-14
Replaced scheduler. prefix with deployer. for scheduler deployer props

### LINKED ISSUE #4261: Update docs to remove the use of scheduler. for deployment properties.
Update docs to remove the use of the `scheduler.` prefix for setting deployment properties.   Users can use the `deployment.<taskname>.` prefix.

cppwfs:
resolves #4261


ilayaperumalg:
LGTM, merging.

ilayaperumalg:
@cppwfs Do you think we can back port this change as well?

PR #4337, opened 2021-01-20
Inject WF tracing properties to the stream apps

tzolov:
- inject wf.application.name and wf.service properties to the deployed streaming apps to enable the WF tracing support.

 Resolves #4333

PR #4347, opened 2021-01-25
Fixed test that is failing because of the following commit.

cppwfs:
https://github.com/spring-cloud/spring-cloud-dataflow/pull/4337/files


tzolov:
LGTM

PR #4349, opened 2021-01-27
Added --remove-data flag to task execution clean up shell command

### LINKED ISSUE #4247: Task execution cleanup initiated via the shell needs to remove task exec data
Currently when a user executes a `task execution cleanup --id 1`, the pod is removed as expected.   However the user is not given the option to delete the associated task execution entry from the task_execution table via the shell.   

We need to offer an optional `--remove-data` flag that will delete the task execution entry in the task_execution table along with the pods.

karthikbhat13:
Fixes #4247 

- Added --remove-data flag to task execution clean up shell command.


ilayaperumalg:
@karthikbhat13 Thanks for your contribution!

PR #4361, opened 2021-02-01
Allow streaming apps to use custom datasources

### LINKED ISSUE #4359: (Docker-Compose) Deployed applications should be able to override the datasource configuraitons inherided from the Skipper or Dataflow servers
Currently, for the local deployment, the Spring datasource configuration of thg Dataflow/Skipper server is passed to the deployed applications as part of the environment context.  

This affects only the maven/jar applications deployed in local mode.

Because the existing docker-compose configurations uses env. variables (e.g. `SPRING_DATASOURCE_URL`) instead of properties (like `--spring.datasource.url`) the deployed streaming apps are not able to configure customer per-app datasources.

tzolov:
- In the docker-compose configuration, replace the datasource env. variables (e.g. like SPRING_DATASOURCE_URL) by Spring properties to allow the apps to override it.

 Resolves #4359

### DIFF
--- a/spring-cloud-dataflow-server/docker-compose-postgres.yml
+++ b/spring-cloud-dataflow-server/docker-compose-postgres.yml
@@ -22,17 +22,17 @@ services:
 
   dataflow-server:
     environment:
-      - SPRING_DATASOURCE_URL=jdbc:postgresql://postgres:5432/dataflow
-      - SPRING_DATASOURCE_USERNAME=root
-      - SPRING_DATASOURCE_PASSWORD=rootpw
-      - SPRING_DATASOURCE_DRIVER_CLASS_NAME=org.postgresql.Driver
+      - spring.datasource.url=jdbc:postgresql://postgres:5432/dataflow
+      - spring.datasource.username=root
+      - spring.datasource.password=rootpw
+      - spring.datasource.driver-class-name=org.postgresql.Driver
     entrypoint: "./wait-for-it.sh -t 240 postgres:5432 -- java -jar /maven/spring-cloud-dataflow-server.jar"
 
   skipper-server:
     environment:
-      - SPRING_DATASOURCE_URL=jdbc:postgresql://postgres:5432/dataflow
-      - SPRING_DATASOURCE_USERNAME=root
-      - SPRING_DATASOURCE_PASSWORD=rootpw
-      - SPRING_DATASOURCE_DRIVER_CLASS_NAME=org.postgresql.Driver
+      - spring.datasource.url=jdbc:postgresql://postgres:5432/dataflow
+      - spring.datasource.username=root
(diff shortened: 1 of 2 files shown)


ilayaperumalg:
LGTM

spring-cloud-dataflow-server/docker-compose-postgres.yml
```diff
@@ -22,17 +22,17 @@ services:
 
   dataflow-server:
     environment:
-      - SPRING_DATASOURCE_URL=jdbc:postgresql://postgres:5432/dataflow
-      - SPRING_DATASOURCE_USERNAME=root
-      - SPRING_DATASOURCE_PASSWORD=rootpw
-      - SPRING_DATASOURCE_DRIVER_CLASS_NAME=org.postgresql.Driver
+      - spring.datasource.url=jdbc:postgresql://postgres:5432/dataflow
```
ilayaperumalg, line 25: @tzolov The way I understood was that, the properties under `environment` would always be ENV properties irrespective of the case. Can you help me understand how this works this way?

tzolov: @ilayaperumalg , when you deploy an app you can set apps's own `--spring.datasource.url` that would override the one coming form the Skipper. But if skipper uses env variables (e.g. `SPRING_DATASOURCE_URL`) then apparently children's property is ignored in favor of the inherited env variable.  I guess in boot the env. vars has precedence over properties. 
The above behaviors observed by the local deployers used by the docker-compose configurations.

jvalkeal: That's a bit weird as it's still coming from a boot's property resolver which resolves from environment and thus format in it should not make any difference.

ilayaperumalg: @jvalkeal Do you think this PR needs to change? Maybe there are some caveats we miss? What do you think @tzolov ?

jvalkeal: Lets say that environment variables have `SPRING_DATASOURCE_URL` and `spring.datasource.url` a `XXX_XXX_XXX takes precedence. Then this may be a symptom of how local deployer is setting properties into environment variables using dot style so it would override that but not the normal underscore type. Running apps in a same container is not so supported scenario being testing. Maybe local deployer could be setup not to inherit those database values automatically?

jvalkeal:
We should fix the issue on a local deployer as those environment inherit settings were wrongly removed a long time ago. That would be a proper fix and possible compose config could then choose what to inherit into local processes.

ilayaperumalg:
We have decided to apply the changes into `local` deployer. Hence, closing this here.

PR #4364, opened 2021-02-01
Updated to remove SCDF-Core from CTR build

### LINKED ISSUE #4363: Remove SCDF-Core dependency from SCDF-CTR

cppwfs:
resolves #4363


jvalkeal:
looks good. with it I'm able to add ctr as test dep to core and it gets build before.

PR #4383, opened 2021-02-10
Restore task app version from manifest

### LINKED ISSUE #4381: Use existing task resource from a manifest
With a previous task launch overhaul, while we did great things to allow user to define versions with a launch some things were left unimplemented:

- We always use version user selects in a UI form
- In shell we can just launch a task with no options so it silently default to default app

As this is a server side ticket, we should use the one from a manifest(if there's one) and use it if launch request doesn't define a version to use(and thus next launch would update manifest).

jvalkeal:
- If version is not explicitely passed in with launch,
  don't switch back to default version.
- This fixes default case with shell when you just relaunch,
  but doesn't change anything on a UI side where version is
  still the one defined in a form(not yet rehydrated from
  previous manifest).
- Fixes #4381


cppwfs:
LGTM

PR #4392, opened 2021-02-17
$END will not be validated as a app

### LINKED ISSUE #4387: Spring Cloud Data Flow fails to create composed tasks that contain  $END
A composed task definition of `timestamp '*'->t1: timestamp 'FOO'->$END` does not pass dataflow validation because it says task:$END is not a valid application.   
In the DefaultTaskService class where we saveStandardTaskDefinition (Line 110) for a composed task we need to gate it with a if statement like:
```
if(!composedTaskElementDefinition.getTaskName().contains("$END")) {
	saveStandardTaskDefinition(composedTaskElementDefinition);
}
```
This way $END is not treated like an app at task definition create time.

cppwfs:
resolves #4387


jvalkeal:
image
Would it be better to check if app name is `$END` instead of checking if task name ends with `$END`?

cppwfs:
The name that is created for a composed task `FOO` is `FOO-$END`.

jvalkeal:
Ah yeah you're right about app name, but my point of using `contains` having it in label fails to create a second sub-task in `t1: timestamp && t$END2: timestamp`

Issue #3355, opened 2019-07-18
Add support to search by "By version(s)" in the App Registry


eskuai:
**Problem description:**
As a user i imported a differents versions of apps.
I want to search apps by version

**Solution description:**
Add into applications web page a filter named , for example, "by version", to select all apps with this values.

**Description of alternatives:**
Into the table, column "Versions", a little switch showing all versions allowed.

eskuai:
It seems is not interesting ... closed

sabbyanandan:
Hi, @eskuai. This can be a good enhancement. I don't see anything wrong with it; it's just about time as to when we can get to it since we are a small team. If you've cycles, please feel free to start a PR, we can collaborate with you.

eskuai:
Hi @sabbyanandan 

I understand that you have a list of priorities into scdf 2.3.0 and more and I think it is more important to improve the "server" side ...  
I've never done a PR, actually my role is unix/aws system administrator, but i can write java, spring, i could wirte spring data repositories, etc..
if you send me more info , i will try., but client side is not my favour (i cant wrike nodejs)

siddhantsorann:
Hey, @sabbyanandan I'd like to pick this ticket up if this enhancement is still needed.

siddhantsorann:
Also, since we have multiple filters here have we thought about using predicates for this case?

siddhantsorann:
Since there are multiple filters and AppRegistration is a JPA entity here is how I think we can implement this without having too many if-else conditions - 
```
public Page<AppRegistration> findAllByTypeAndNameIsLikeAndVersionAndDefaultVersion(ApplicationType type,
      String name, String version, boolean defaultVersion, Pageable pageable) {
    CriteriaBuilder cb = entityManager.getCriteriaBuilder();
    CriteriaQuery<AppRegistration> cq = cb.createQuery(AppRegistration.class);
    Root<AppRegistration> appRegistrationRoot = cq.from(AppRegistration.class);
    final List<Predicate> predicates = new ArrayList<>();
    if (type != null) {
      predicates.add(cb.equal(appRegistrationRoot.get("type"), type));
    }
    if (StringUtils.hasText(name)) {
      predicates.add(cb.like(cb.lower(appRegistrationRoot.get("name")), name.toLowerCase()));
    }
    if (StringUtils.hasText(version)) {
      predicates.add(cb.equal(cb.lower(appRegistrationRoot.get("version")), version.toLowerCase()));
    }
[...]

PR #4395, opened 2021-02-17
Application Version Filter

siddhantsorann:
Closes - #3355 
Added parameter to API to filter using version.
Created DAO to use predicates to filter. This will avoid multiple repo methods and if-else conditions.

### DIFF
--- a/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/repository/AppRegistrationDao.java
+++ b/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/repository/AppRegistrationDao.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.dataflow.registry.repository;
+
+import org.springframework.cloud.dataflow.core.AppRegistration;
+import org.springframework.cloud.dataflow.core.ApplicationType;
+import org.springframework.data.domain.Page;
+import org.springframework.data.domain.Pageable;
+import org.springframework.lang.Nullable;
+
+/**
+ * DAO to access {@link org.springframework.cloud.dataflow.core.AppRegistration}. Contains
+ * predicate specific operations to make filtering based on optional parameters more
+ * efficient.
+ *
+ * @author Siddhant Sorann
+ */
+public interface AppRegistrationDao {
+
+	/**
+	 * Function to find all app registrations based on various optional parameters using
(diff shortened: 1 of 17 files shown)


spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/repository/AppRegistrationDao.java
```diff
@@ -0,0 +1,48 @@
...
+import org.springframework.lang.Nullable;
+
+/**
+ * DAO to access {@link org.springframework.cloud.dataflow.core.AppRegistration}. Contains
+ * predicate specific operations to make filtering based on optional parameters more
+ * efficient.
+ *
+ * @author Siddhant Sorann
+ */
+public interface AppRegistrationDao {
```
ilayaperumalg, line 32: Instead of creating a new DAO, can we add this method in [here](https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/repository/AppRegistrationRepository.java#L33)? or, do you see any issue with that approach?

ilayaperumalg: Did you try to address this via JPA repository specific implementation? If so, we would be keen to understand the issues you faced. In case if you would like to take a reference, you can check this repository implementation here in [Skipper code base](https://github.com/spring-cloud/spring-cloud-skipper/blob/805dcf99bfb6fc965fa23712eb814149df8538ea/spring-cloud-skipper-server-core/src/main/java/org/springframework/cloud/skipper/server/repository/jpa/ReleaseRepositoryImpl.java#L30) to see a pattern on how you can have repository specific implementation.

siddhantsorann (commented):
@ilayaperumalg I decided to use predicates for this to avoid multiple if conditions and JPA repo functions based on all possible combintations of filters being sent via the API. Since in this case we have 4 optional filters that would lead to about 24 such cases.

spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/repository/AppRegistrationDao.java
siddhantsorann: @ilayaperumalg I decided to use predicates for this to avoid multiple if conditions and JPA repo functions based on all possible combintations of filters being sent via the API. Since in this case we have 4 optional filters that would lead to about 24 such cases.

siddhantsorann: @ilayaperumalg I have made the changes in the implementation according to what you described above, please take a look.

ilayaperumalg: @siddhantsorann Thanks for the followup and there was nothing wrong in your earlier changes instead, I was trying to understand the reasonings and more importantly, it would be better to have the same approach in terms of customising the repositories in our codebase. I will review this PR and get back to you.

siddhantsorann: @ilayaperumalg I got it. This approach looks more cleaner and requires lesser changes so it's a better design.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Rebased, fixed conflicts and merged as 509acaea9.

Thanks @siddhantsorann for the PR!

PR #4397, opened 2021-02-18
Updated task dependency to 2.3.0.

cppwfs:
Also creeated new PlatformTransactionManager Bean to work around Task's introduction of springCloudTaskTransactionManager.

reesolves #4375


jvalkeal:
I'd say we just merge and see if we get any issues as tests passed.

PR #4415, opened 2021-03-03
Remove flyway transaction disable with postgres

### LINKED ISSUE #4370: Bad recovery in case of an error during DB schema creation could leave the DB in a permanent corrupted state
**Description:**
When DB schema creation fails after a certain point, SCDF fails to recover automatically, leaving the the DB in an unrecoverable state.

**Release versions:**
Observed on SCDF 2.5.3. Still exists in master if we analysed the issue correctly.

**Steps to reproduce:**
Our estimation of the repro steps following an investigation of the corrupted DB state -
1. Start SCDF using postgresql
2. Fail the flow at a rather advance step (after the table 'APP_REGISTRATION' was created, and after step 'Add Platform To AuditRecords' was executed on table 'flyway_schema_history_dataflow')
3. Restart SCDF
Expected result - DB setup should resume and complete successfully.
Actual result - SCDF appear to have dropped and re-created most of the tables, without properly applying post-init updates on them. 
For example, the table 'audit_records' got dropped and re-created without the column called 'platform_name'. At this point the flyway_schema_history_dataflow table already had the 'Add Platform To AuditRecords' state persisted, and so the column didn't get re-added.
End result is that SCDF fails to start due to many mismatches between the code and the DB tables.

**Additional context:**
We suspect the root cause is in https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/db/migration/AbstractBaselineCallback.java#L63
Relaying on the existence of the 'APP_REGISTRATION' table doesn't seem like a reliable method to perform such a destructive operation as appears in the following lines (Or - if there's a need to fully reset the DB at this point, need to ensure to reset the flyway state table as well).

jvalkeal:
- Previously with postgres driver there was an
  attempt not to use transactions as those always
  abort it even with query errors. This as a naive
  way would cause migration issues if something
  else goes wrong and database may be left in
  a wrong state.
- With postgres, move back to transactions and
  do a table exists check which should be more
  reliable which then seem to survive in tests
  where we manually break database and next
  launch attempt after "fixing" database looks
  to bring things in a good condition.
- These new tests are not part of this commit
  but coming in with a new integation tests
  what I've run locally.
- Fixes #4370

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/db/migration/AbstractBaselineCallback.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/db/migration/AbstractBaselineCallback.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 the original author or authors.
+ * Copyright 2019-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


jvalkeal:
This is mostly going to get tested with upcoming new db test infra.

PR #4419, opened 2021-03-04
Nest options in DataFlowITProperties

### LINKED ISSUE #4418: Nesting in DataFlowITProperties for tests
Prefix on a class level is `test.docker.compose` and we want to use this for new db tests as well. Have `test` as top-level and nest `test.docker.compose` so that we can also have namespace like `test.db`.

jvalkeal:
- Prepare to be able to add more sub-keys
  under `test` and simply nest `test.docker.compose`.
- Fixes #4418

### DIFF
--- a/spring-cloud-dataflow-server/src/test/java/org/springframework/cloud/dataflow/integration/test/DataFlowITProperties.java
+++ b/spring-cloud-dataflow-server/src/test/java/org/springframework/cloud/dataflow/integration/test/DataFlowITProperties.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 the original author or authors.
+ * Copyright 2020-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


spring-cloud-dataflow-server/src/test/java/org/springframework/cloud/dataflow/integration/test/DataFlowITProperties.java
```diff
@@ -21,78 +21,106 @@
...
-	}
-
-	public void setDataflowServerUrl(String dataflowServerUrl) {
-		this.dataflowServerUrl = dataflowServerUrl;
-	}
-
-	public String getPlatformName() {
-		return platformName;
-	}
+	private DockerProperties docker = new DockerProperties();
```
tzolov, line 27: I think we should find a better name Then `DockerProperties` as those properties has noting to do with docker or docker-compose. They are just IT/AT configuration properties.
the `docker.compose` suffix is unfortunate properties naming inherited from the initial DataFlowIT implementation.

PR #4420, opened 2021-03-04
CTR meta props will now work for docker images.

cppwfs:
CTR docker image now has a label that contains the viewable properties for CTR .

resolves 4405

### DIFF
--- a/spring-cloud-dataflow-composed-task-runner/pom.xml
+++ b/spring-cloud-dataflow-composed-task-runner/pom.xml
@@ -123,11 +123,11 @@
 			<plugin>
 				<groupId>io.fabric8</groupId>
 				<artifactId>docker-maven-plugin</artifactId>
-				<version>0.21.0</version>
+				<version>0.33.0</version>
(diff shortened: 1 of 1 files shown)


spring-cloud-dataflow-composed-task-runner/pom.xml
```diff
@@ -123,11 +123,11 @@
 			<plugin>
 				<groupId>io.fabric8</groupId>
 				<artifactId>docker-maven-plugin</artifactId>
-				<version>0.21.0</version>
+				<version>0.33.0</version>
 				<configuration>
 					<images>
 						<image>
-							<name>springcloud/${project.artifactId}</name>
+							<name>cppwfs/${project.artifactId}</name>
```
jvalkeal, line 130: wrong id

cppwfs: Thanks for catching this!

jvalkeal:
Should be ok

PR #4425, opened 2021-03-09
Implement database IT tests

### LINKED ISSUE #4424: Implement basic database tests
Integration tests for database scenarios has lived in a acc-tests repo and have grown to be a bit of a monster. To make things easier, do an attempt with test-containers to add simple scenarios as a foundation attempting to test mysql/postgres.

jvalkeal:
- New subsystem to build a "cluster" using test containers so that we can control
  when to bring things up/down when testing different scenarios.
- For now, just a base things like testing migration for just mysql/postgres.
- Lot of junit5 tags so that we can disect different tests.
- DataFlowIT has been tagged with `docker-compose`
- There's no default tags other that what we inherited into build from
  spring-cloud when we did our own parent. There tags `slow` and `docker`
  is excluded on default. (we probably need to revisit those).
- For example to just run new database tests, do:
  ./mvnw integration-test -Pfailsafe -pl spring-cloud-dataflow-server -Dgroups="database"
- To get current dep versions of dataflow/skipper, we rely on maven testresource resolving
  which creates `./spring-cloud-dataflow-server/target/test-classes/application-dbtests-shared.yml`
  and we can load that via boot tests using an activated profile.
- Fixes #4424

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -94,6 +94,7 @@
 		<joda-time.version>2.10.6</joda-time.version>
 		<aws-java-sdk-ecr.version>1.11.731</aws-java-sdk-ecr.version>
 		<commons-text.version>1.8</commons-text.version>
+		<testcontainers.version>1.15.2</testcontainers.version>
 	</properties>
 	<modules>
 		<module>spring-cloud-dataflow-container-registry</module>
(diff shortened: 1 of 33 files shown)


tzolov (commented):
@jvalkeal, this looks very good but it lacks any instructions how to use it. 
Can you add such to the README.adoc and update the obsoleted Docker-compose IT examples there.  Since the `maven-failsafe-plugin` is removed the existing snippets will not work.

tzolov:
Rebased, squashed and merged at 1fe64a211b70587dab9ab3a527217fcc65eee0ac

PR #4432, opened 2021-03-14
Let CTR ignore unknown job parameters from previous executions

hpoettker:
When users launch composed tasks in the UI, they can add arguments in a free text field. It does occasionally happen that users make small errors when doing so.

For example, when someone forgets the `--` in front of `composed-task-arguments=foo=bar`, the argument `foo=bar` will not be added to the composed tasks. Instead, the batch job created by the `ComposedRunnerJobFactory` will be started with a parameter `composed-task-arguments`, which is ignored by the job.

This behavior is okay, in my opinion. However, the mistyped arguments are currently taken over to every subsequent run of the composed task due to the internal logic of Spring Batch. This does not cause any direct issues as the arguments are always ignored. But it is a bit annoying. Currently, the only way to get rid of the unwanted parameter is to clean all executions that contain it or to manipulate the latest execution manually in the database.

That the implementation of a custom `JobParametersIncrementer` is the way to solve this problem is the result of the discussion here: https://github.com/spring-projects/spring-boot/pull/25533

### DIFF
--- a/spring-cloud-dataflow-composed-task-runner/src/main/java/org/springframework/cloud/dataflow/composedtaskrunner/ComposedRunnerJobFactory.java
+++ b/spring-cloud-dataflow-composed-task-runner/src/main/java/org/springframework/cloud/dataflow/composedtaskrunner/ComposedRunnerJobFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2020 the original author or authors.
+ * Copyright 2017-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -23,6 +23,9 @@ import java.util.Map;
 import java.util.UUID;
 
 import org.springframework.batch.core.Job;
+import org.springframework.batch.core.JobParameter;
+import org.springframework.batch.core.JobParameters;
+import org.springframework.batch.core.JobParametersBuilder;
 import org.springframework.batch.core.Step;
 import org.springframework.batch.core.configuration.annotation.JobBuilderFactory;
 import org.springframework.batch.core.job.builder.FlowBuilder;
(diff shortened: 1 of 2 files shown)


cppwfs:
But looking at this update would be a breaking change for the current contract on how composed tasks are launched with parameters, specifically clearing out the job parameters from the previous execution.    
I'm going to close this PR.   If you have further comments please let us know.   

Thanks.

hpoettker:
Hi @cppwfs , thank you for the review. The PR was not intended to break existing features. If it does, I agree that it should not be merged.

I currently don't see which feature would be broken by the PR. I'd still like to contribute a fix for the unwanted behavior that I described above. Can you be more specific about the breaking change in the PR? Or should I open an issue where I elaborate the problem in more detail?

In my understanding, the parameters that are cleared out by the PR have no effect any way. They are only seen by the CTR job which ignores them.

The PR only has an effect if `increment-instance-enabled=true` and in that case each launch of a composed task will still yield a new instance for the CTR job.

Apart from parameters to define the job instance, the composed task runner is only effected by arguments starting with `--` like e.g. `--interval-time-between-checks=1000`. The handling of these arguments it not influenced by the PR as these are not job parameters. It's also the current behavior that these arguments are not propagated from one execution to the next.

cppwfs:
Hello @hpoettker !
I guess the question is this.  What happens if I want to reprocess a failed Composed Task run?   This PR seems to prevent this.

hpoettker:
Hi @cppwfs , thanks for the quick response.

From a UI perspective, you mean if I click on _Restart the job_ in the job execution page of a failed CTR job? The PR has no effect on this.

If a failed CTR job is restarted, Dataflow's `DefaultTaskJobService::restartJobExecution` will look up the exact job parameters of the failed CTR job execution (including the `run.id`) and start the CTR with them. Inside the CTR, the logic from Spring Boot's `JobLauncherApplicationRunner::getNextJobParameters` will then kick in, which just restarts the existing job instance with the given parameters. The `JobParametersIncrementer` of the CTR job is not used in this scenario. Neither Spring Batch's `RunIdIncrementer` that is currently passed to the CTR job, nor the extension from this PR.

PR #4435, opened 2021-03-15
Use s-c-d-f-rest-client for all IT/AT connections

### LINKED ISSUE #4433: Allow Ts/ATs run against secured (oauth2) SCDF servers
**Problem description:**
For the purpose of SCDF CF tile ATs we need to ensure that ATs can run over secured (oauth2) connection!

Currently, the `ITs`&`ATs` use a custom, in-palace creation of the `DataFlowTemplate` instance used to interact with the SCDF servers. This code doesn't allow configuring an authenticated/authorized connections to the servers. 

**Solution description:**
Replace the inefficient, custom `DataFlowTemplate` factory implementations in favor or the `spring-cloud-dataflow-rest-client` related `DataFlowClientAutoConfiguration` utilities. Later offer extensive (spring compliant) mechanisms for building `DataFlowTemplate` and connecting to secured SCDF server.

tzolov:
- Replace the proprietary dataflowTemplate builder by s-c-d-f-rest-client#DataFlowClientAutoConfiguration
 - Remove the custom test.platform.connection.dataflowServerUrl property in favor or spring.cloud.dataflow.client.server-uri .
 - Use spring.cloud.dataflow.client.skipSslValidation to disable SSL validation.
 - Create a DataFlowOperationsITConfiguration configuration to be used by ITs/ATs for configuring the DataFlowTemplate and the RuntimeApplicationHelper instances.
 - Streamline the ITs/ATs implementation to ensure that all interactions with the SCDF server (including http-post and app/actuators) are funneled through the single dataflowTemplate + restTemplate.

 Resolves #4433

### DIFF
--- a/spring-cloud-dataflow-server/src/test/java/org/springframework/cloud/dataflow/integration/test/util/DockerComposeFactory.java
+++ b/spring-cloud-dataflow-server/src/test/java/org/springframework/cloud/dataflow/integration/test/util/DockerComposeFactory.java
@@ -169,6 +169,12 @@ public class DockerComposeFactory {
 	}
 
 	public static Path createTempDirectory() {
+
+		// Temp directory is not required when the docker-compose is not used.
+		if (DockerComposeFactoryProperties.isDockerComposeDisabled()) {
+			return null;
+		}
+
 		try {
 			Path tempDirPath = Files.createTempDirectory(null);
 			logger.info("Temp directory: " + tempDirPath);
--- a/spring-cloud-dataflow-server/src/test/java/org/springframework/cloud/dataflow/integration/test/util/DockerComposeFactory.java
+++ b/spring-cloud-dataflow-server/src/test/java/org/springframework/cloud/dataflow/integration/test/util/DockerComposeFactory.java
@@ -16,8 +16,6 @@
 
 package org.springframework.cloud.dataflow.integration.test.util;
 
-import java.io.IOException;
-import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Arrays;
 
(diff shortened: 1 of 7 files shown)


spring-cloud-dataflow-server/src/test/resources/application.properties
```diff
@@ -1,4 +1,4 @@
 spring.datasource.url=jdbc:h2:mem:testdb
-spring.cloud.dataflow.features.streams-enabled=false
-spring.autoconfigure.exclude=org.springframework.cloud.dataflow.rest.client.config.DataFlowClientAutoConfiguration
+spring.cloud.dataflow.client.skip-ssl-validation=true
+spring.cloud.dataflow.client.enable-dsl=true
```
ilayaperumalg, line 3: I see this property `enableDsl` is enabled here but I don't see any references of this property on how it is getting used in this code base?

tzolov: Not sure either. It comes from the spring-cloud-dataflow-rest-client. Maybe it is obsolete, as it doesn't seem to effect the behavior.

spring-cloud-dataflow-server/src/test/java/org/springframework/cloud/dataflow/integration/test/util/DockerComposeFactory.java
```diff
@@ -169,6 +169,12 @@ public static Extension startDockerCompose(Path tempFolder) {
 	}
 
 	public static Path createTempDirectory() {
+
+		// Temp directory is not required when the docker-compose is not used.
+		if (DockerComposeFactoryProperties.isDockerComposeDisabled()) {
```
ilayaperumalg, line 174: I wonder why we should call this method DockerComposeFactory when docker compose is not used?

tzolov: It had to be created externally we it can be removed in the @AfterAll. But i will try to replace it whit juni5 statement like this:
```java
@TempDir static Path tempYamlFolder;
```

spring-cloud-dataflow-server/src/test/resources/application.properties
jvalkeal: Probably property when dsl was external lib.

ilayaperumalg: ok, maybe we can remove it now?

ilayaperumalg:
LGTM

ilayaperumalg:
Merging.

Issue #4021, opened 2020-06-30
Launch Task IncorrectResultSizeDataAccessException: Incorrect result size: expected 1, actual 2 at


PatrickWe:
I use the SCDF-Server with a DB2 and my task is also using the same db2.
If I launch the Task the first and second time it works and i see the logs and the exit code = 0.
If I launch the Task the for the third time I got a exception:

In the Table TASK_EXECUTION_METADATA there are two Entries. If i delete one entry, after that i can start the third run and it runs sucessfull:

```
2020-06-30 11:50:34.355 ERROR 26828 --- [nio-9393-exec-4] o.s.c.d.s.c.RestControllerAdvice         : Caught exception while handling a request
 java.lang.IllegalStateException: Unable to generate query
         at org.springframework.cloud.dataflow.server.repository.JdbcDataflowTaskExecutionMetadataDao.getLatestManifest(JdbcDataflowTaskExecutionMetadataDao.java:153)
         at org.springframework.cloud.dataflow.server.service.impl.DefaultTaskExecutionService.executeTask(DefaultTaskExecutionService.java:263)
         at org.springframework.cloud.dataflow.server.service.impl.DefaultTaskExecutionService$$FastClassBySpringCGLIB$$422cda43.invoke(<generated>)
         at org.springframework.cglib.proxy.MethodProxy.invoke(MethodProxy.java:218)
         at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.invokeJoinpoint(CglibAopProxy.java:771)
         at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:163)
         at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.proceed(CglibAopProxy.java:749)
         at org.springframework.transaction.interceptor.TransactionAspectSupport.invokeWithinTransaction(TransactionAspectSupport.java:366)
         at org.springframework.transaction.interceptor.TransactionInterceptor.invoke(TransactionInterceptor.java:118)
         at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:186)
         at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.proceed(CglibAopProxy.java:749)
         at org.springframework.aop.framework.CglibAopProxy$DynamicAdvisedInterceptor.intercept(CglibAopProxy.java:691)
[...]

PR #4441, opened 2021-03-16
Fix paging query provider for DB2

hpoettker:
Fixes gh-4021.

`Db2PagingQueryProvider` currently does not work as expected as the SQL keyword `BETWEEN` is used where the end value should be excluded although `BETWEEN` is inclusive regarding both its start and end value,


cppwfs:
Kicked the tires and it looks good to me.

cppwfs:
Rebased Merged.   Thank you for your contribution!!!!

PR #4450, opened 2021-03-25
App cross version it at test

tzolov:
Add an IT/AT test that deploys http (3.x) + log (3.x) stream, then updates it to http (3.x) + log (2.x) stream and finally rollback  to http (3.x) + ver-log (3.x) stream.

### DIFF
--- a/spring-cloud-dataflow-server/src/test/java/org/springframework/cloud/dataflow/integration/test/DataFlowIT.java
+++ b/spring-cloud-dataflow-server/src/test/java/org/springframework/cloud/dataflow/integration/test/DataFlowIT.java
@@ -74,6 +74,7 @@ import org.springframework.cloud.dataflow.rest.resource.DetailedAppRegistrationR
 import org.springframework.cloud.dataflow.rest.resource.TaskExecutionResource;
 import org.springframework.cloud.dataflow.rest.resource.TaskExecutionStatus;
 import org.springframework.cloud.dataflow.rest.resource.about.AboutResource;
+import org.springframework.cloud.skipper.domain.SpringCloudDeployerApplicationManifestReader;
 import org.springframework.context.annotation.Import;
 import org.springframework.core.io.DefaultResourceLoader;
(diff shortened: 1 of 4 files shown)


spring-cloud-dataflow-server/src/test/java/org/springframework/cloud/dataflow/integration/test/DataFlowIT.java
```diff
@@ -379,6 +380,72 @@ public void streamPartitioning() {
...
+
+			runtimeApps.httpPost(stream.getName(), "http", "Test message One - " + RANDOM);
+			Awaitility.await().until(() -> stream.logs(app("ver-log")).contains("Test message One - " + RANDOM));
+
+			assertThat(stream.history().size()).isEqualTo(1L);
+
+			// UPDATE
+			logger.info("stream-app-cross-version-test: UPDATE");
+
+			stream.update(new DeploymentPropertiesBuilder().put("version.ver-log", "2.1.5.RELEASE").build());
```
sabbyanandan, line 418: Can any of these versions point to always the `master` or `snapshot` equivalent? That way, we know we are running the latest on top of the previously GA'd code. Otherwise, these versions would become stale very soon.

sabbyanandan (changes requested):

spring-cloud-dataflow-server/src/test/java/org/springframework/cloud/dataflow/integration/test/DataFlowIT.java
ilayaperumalg: Good point, but even if we switch to `SNAPSHOT` we need to manually edit to change the SNAPSHOT version as well. In the long run, we could make it more configurable.

ilayaperumalg:
LGTM

ilayaperumalg:
I am merging this for now and we can revisit the action items via separate issue: https://github.com/spring-cloud/spring-cloud-dataflow/issues/4451

PR #4468, opened 2021-04-05
Append group in stream DSL on app definition

### LINKED ISSUE #4463: Stream DSL is incorrect when overriding the input group
In case of the following scenario where the stream DSL is: `:SOURCE > splitter > :DEST` and when deploying the stream user overrides the input group of the app `splitter` by setting `app.splitter.spring.cloud.stream.bindings.input.group=test` instead of the default (SCDF sets stream name as the default group name). 

Upon the stream deployment, the reconstructed stream DSL has the `group` field added to the source destination like this: `:SOURCE --group=somevalue > splitter > :DEST`

This subsequently makes the Stream DSL not rendering at the pipeline view in the GUI as well.

ilayaperumalg:
- When the stream DSL has `input` group set, then the `group` property needs to be set on the app definition associated with it.
 - Update test

Resolves #4463


sobychacko:
LGTM. Merged upstream.

PR #4489, opened 2021-04-26
Added Sleuth to Dataflow [2.10]

marcingrzejszczak:
Depends on sleuth 3.1.0! 
First milestone is scheduled for ~August while the GE is ~Oct/Nov.


marcingrzejszczak:
Ga should be there by the end of the month

sabbyanandan:
@marcingrzejszczak @tzolov: What's the latest here? Are we submitting a PR against this?

marcingrzejszczak:
It's ready to be merged

ilayaperumalg:
Closing this now for due to inactivity. We can re-open if needed.

marcingrzejszczak:
I wrote that you can merge the code

ilayaperumalg:
@marcingrzejszczak Sorry again, similar to skipper, I closed it with a misunderstanding on the scope of this. Re-opening it and let's discuss further.

jvalkeal:
I'm closing sleuth related PR's in favour of possibly making a proper plan how tracing could be enabled.


Mostly there'd need to be a plan to bring it in centrally vi parent build pom, fully go through how tracing would get added to call stack, allow to turn it on/off, check how it actually work on a whole stack, study that nothing sensitive is not exposed, etc.

Finally seeing "then what" as we'd need to build something automated to represent things to a user.

onobc:
> fully go through how tracing would get added to call stack, 

By default it is completely transparent. When we invoke on technologies that are auto-configured such as [these](https://docs.spring.io/spring-cloud-sleuth/docs/current-SNAPSHOT/reference/html/integrations.html#sleuth-integration) we get traces/spans sent to the tracing system. Whatever tracing backend is being used, there is a UI available in some form that you can analyze the calls through the system. Once setup is an amazing tool to understand perf issue and debugging in general. 

> allow to turn it on/off, check how it actually work on a whole stack, 

There are properties that allow it to be disabled as a whole, or piecemeal by integration ([spring.sleuth.*](https://cloud.spring.io/spring-cloud-sleuth/reference/html/appendix.html))

> study that nothing sensitive is not exposed, etc.

This is always a good thing to check. I would be interested in the web traces on this since we pass properties via request params (although I don't think they are logged by default). 



### My Thoughts
One kink does come to my mind... Orgs almost always have their tracing backend chosen and in place (Sleuth has to fit into that). This means that Sleuth has to be configured to send traces to that backend. No big deal, it can do that via OpenZipkin, OpenTracing, oTEL, etc... However, it requires adding libs to the classpath. Something that I do not think we have an easy way to do w/ the default install of SCDF (AFAIK). 

The things I think would be good to add before releasing to the masses would be:
- docs on how to use/enable/disable (probably just mostly pointing to the Sleuth docs)
- add some auto-configuration sanity test that proves we can disable tracing (of course w/o re-testing all the things that sleuth already tests and guarantees it does). 
- docs on how to configure to use other tracing backends (the Sleuth docs cover in detail but this is mainly around how to update SCDF w/ the extra required libs). 

If we make a decision to ship this in 2.10 we should get the dependencies in sooner rather than later (ie this PR). However, if we make a decision to ship this > 2.10 we should probably close the PR. 

I think we should meet and discuss this together so we can make a decision on **if** and **when**. 

@jvalkeal @marcingrzejszczak @ilayaperumalg

Issue #4493, opened 2021-04-29
Align docker-compose with pack changes


jvalkeal:
We're now trying to build images with cloud native via pack, instead from maven builds. Docker compose files needs to be changed as they are using entrypoints.

PR #4494, opened 2021-04-29
Align docker compose with cloud native images

jvalkeal:
- Try not to override entrypoint where possible
- Rely on restart policy
- Relates #4493


ilayaperumalg:
LGTM, verified by running the `docker-compose up` as well as checked the IT tests starting fine with this configuration setup.
Merging.

PR #4510, opened 2021-05-07
Add IT/AT for updating multi-instance apps streams

### LINKED ISSUE #4505: SCDF UI - getting error 500 at the UI when trying to open Tasks definition window (related to Issue #4358)
Hello,

We are getting error 500 when trying to open the UI at task definitions window.
This error raise when task attribute value containing a pipe character - " | ".
The task defines successfully via REST API with all attributes and their values, including quotes, e.g. /tasks/definitions/TASK_NAME --data-urlencode "definition=--param-name='abcd|efg'"

This issue solved backed than at master release (I verified it a ~2 month ago) but seems to reproduce again now **with new scdf version 2.7.2.** .
Please see details at https://github.com/spring-cloud/spring-cloud-dataflow/issues/4358

Best regards,
Shalom

tzolov:
Extend the lifecycle stream test to allow testing with multiple instances.

  Resolves #4505


jvalkeal:
@tzolov  Can you fix commit message as issue number should be 4504, this will close wrong issue when merged.

PR #4515, opened 2021-05-12
Allow CTR to skip ssl certificate validation

tzolov:
When the SCDF server is configured with HTTPS with self-signed certificate the CTR REST client will fail due to invalid certificate.

  - Add ComposedTaskProperties isSkipTlsCertificateVerification property to control the ssl-skip functionality.
  - Extend the RestTemplate factory in the TaskLauncherTasklet to configure ssl-skip when isSkipTlsCertificateVerification is true.
  - Fix the docker-compose-ssl.yaml conf. to skip the CTR ssl validation.


cppwfs (commented):
What is the best way to test this?

cppwfs:
LGTM, Rebased, Squashed, Merged.

PR #4534, opened 2021-05-26
Fix task execution cleanup

### LINKED ISSUE #4532: Cleanup all task execution doesn't work on postgres
The [logic](https://github.com/spring-cloud/spring-cloud-dataflow/blob/5de42b21279fb11221abd3529a116bcb6f5c6c8c/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/JdbcDataflowTaskExecutionDao.java#L213) to get all the task executions doesn't work on postgres.

### LINKED ISSUE #4533: Task execution cleanup doesn't work on the scheduled tasks
When the task is scheduled, then the action to run the task execution cleanup fails as it expects parent execution ID to be set.

ilayaperumalg:
- Fix the result set matching on the query. This is needed especially on Postgres
 - Fix the task cleanup to allow the task executions that don't have external execution IDs

Resolves #4532
Resolves #4533


cppwfs:
LGTM:
Tested against
DB2
Oracle
Mysql
Postgres
SqlServer

PR #4541, opened 2021-05-27
Enable K8s-LB only for externally accessible apps

### LINKED ISSUE #4499: Remove LB's not needed
https://github.com/spring-cloud/spring-cloud-dataflow/blob/9337fc031293ee83b58fd7655f8d7207164856b1/spring-cloud-dataflow-server/src/test/java/org/springframework/cloud/dataflow/integration/test/DataFlowIT.java#L821

This is somewhat hitting public ip quota on aws account for tests.

tzolov:
- Replace the wildcard with an explicit list of apps to be configured with LB.
 - Ensure that all http source apps and apps that need to provide actuator/logfile access are configured with LB.

 Resolves #4499


jvalkeal:
lgtm, lets try it.

PR #4546, opened 2021-05-27
K8s Task launcher to use task launcher account properties

### LINKED ISSUE #4186: TaskLauncherProperties not being mapped properly from platform account
While investigating https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/issues/398#issuecomment-705124766 it was found that properties set into the scdf configmap that should be populated into `TaskLauncherProperties` is not happening. 

For example:

```
  cloud:
    dataflow:
      task:
        platform:
          kubernetes:
            accounts:
              dev:
                limits:
                    memory: 1024Mi
                    cpu: 1
                backoffLimit: 1
```

Likely this is due to the glue between SCDF and deployer, for example:

In SCDF `KubernetesTaskPlatformFactory.createLauncher` there is:

```		
         KubernetesTaskLauncher kubernetesTaskLauncher = new KubernetesTaskLauncher(
				kubernetesProperties, kubernetesClient, containerFactory);
```

Yet on the deployer side, in `KubernetesTaskLauncher` that constructor would be:

```
       @Autowired
	public KubernetesTaskLauncher(KubernetesDeployerProperties properties,
			KubernetesClient client, ContainerFactory containerFactory) {
		this(properties, new KubernetesTaskLauncherProperties(), client, containerFactory);
	}
```

which creates a new `KubernetesTaskLauncherProperties` so calling `taskLauncherProperties.getBackoffLimit();` in say for example `KubernetesTaskLauncher.getBackoffLimit(..)` is just going to return null...

ilayaperumalg:
- Introduce Platform task launcher properties to include K8s task launcher properties
 - Construct the task launcher using this explicit property
 - Update test

Resolves #4186

### DIFF
--- a/spring-cloud-dataflow-platform-kubernetes/src/main/java/org/springframework/cloud/dataflow/server/config/kubernetes/KubernetesPlatformTaskLauncherProperties.java
+++ b/spring-cloud-dataflow-platform-kubernetes/src/main/java/org/springframework/cloud/dataflow/server/config/kubernetes/KubernetesPlatformTaskLauncherProperties.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright 2017-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.dataflow.server.config.kubernetes;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.cloud.dataflow.core.AbstractPlatformProperties;
+import org.springframework.cloud.deployer.spi.kubernetes.KubernetesTaskLauncherProperties;
+
+/**
+ * @author Ilayaperumal Gopinathan
+ */
+@ConfigurationProperties("spring.cloud.dataflow.task.platform.kubernetes")
+public class KubernetesPlatformTaskLauncherProperties extends AbstractPlatformProperties<KubernetesTaskLauncherProperties> {
+}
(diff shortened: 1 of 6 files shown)


spring-cloud-dataflow-platform-kubernetes/src/main/java/org/springframework/cloud/dataflow/server/config/kubernetes/KubernetesPlatformTaskLauncherProperties.java
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright 2017-2019 the original author or authors.
```
cppwfs, line 2: 2021

cppwfs:
LGTM
Created Dev platform with 2 limits 1 for memory the other for cpu a
1) Launched task using dev platform  and both limits showed up properly in the pod.  
2) Launched task using dev platform again overriding memory limit and it showed up properly in the pod. 
3) Launched task using default and its values showed up properly in the pod 
4) Launched task using default platform and overrode its values and they showed up properly in the pod.

ilayaperumalg:
Thanks @cppwfs . While the above covers the regression part, we need to be testing with properties such as backoffLimit, restartPolicy that are part of platform account properties.

spring-cloud-dataflow-platform-kubernetes/src/test/java/org/springframework/cloud/dataflow/server/config/kubernetes/KubernetesTaskPlatformFactoryTests.java
```diff
@@ -62,6 +62,36 @@ public void kubernetesTaskPlatformNoScheduler() {
 
 	@Test
 	public void kubernetesTaskPlatformWithScheduler() {
+		KubernetesPlatformProperties platformProperties = new KubernetesPlatformProperties();
+		KubernetesDeployerProperties deployerProperties = new KubernetesDeployerProperties();
+		deployerProperties.getLimits().setMemory("5555Mi");
+		KubernetesTaskLauncherProperties taskLauncherProperties = new KubernetesTaskLauncherProperties();
+		taskLauncherProperties.setBackoffLimit(5);
```
cppwfs, line 69: Where do we test for the backofflimit?

ilayaperumalg: Good catch. We aren't testing backoffLimit in this test case rather the next one does. I kept this test as it used to be (except adding the new properties arg)

cppwfs:
Rebased, Squashed, Merged

PR #4550, opened 2021-06-01
Add a better exception message on invalid task config.

### LINKED ISSUE #4549: CTR return a meaningful exception if scdf is misconfigured
When CTR attemps to utilize task operations but the /tasks/definitions endpoint can't be reached it currently throws a NPE.   CTR should return a meaningful message if it can't create a taskOperations object.

cppwfs:
resolves #4549


jvalkeal:
lgtm

PR #4557, opened 2021-06-03
Handle short style task deployment properties

jvalkeal:
- Add handling for deployment properties where ctr name is not present
  and same time handle wildcard as well.
- Fixes #4551

PR #4565, opened 2021-06-07
Handle task deployment properties via ctr options

### LINKED ISSUE #4551: Deployment properties for child CTR apps are not being set
When launching CTR (ctr0601B) with the following graph `1: timestamp && 2: timestamp` the deployment properties as specified for the child applications are not making it to the child app.   For example:
When using the following properties as defined by builder, will not appear in the child execution of the ctr0601B-1 task:
```
deployer.1.local.debug-address=5555
deployer.1.local.debug-port=5555
deployer.1.local.debug-suspend=n
deployer.1.local.delete-files-on-exit=false
```
However if I update it to the following it will work:
```
deployer.ctr0601B.1.local.debug-address=5555
deployer.ctr0601B.1.local.debug-port=5555
deployer.ctr0601B.1.local.debug-suspend=n
deployer.ctr0601B.1.local.delete-files-on-exit=false
```

jvalkeal:
- Essentially reverts previous PR #4557 as it was a bad
  fix and broke more that it fixed.
- This change now tries to pass in task deployment properties
  via crt composed-task-app-properties option.
- Property key part is now base64 encoded with prefix base64_ because
  boot with its binding will silenly remove special characters from
  map keys and we could not use wildcards like deployer.*.local.working-directories-root
- Encoded values are then decoded on ctr side.
- One onnoying this is that these base64 encoded values are visible in a manifest but
  never directly interact with a user.
- Fixes #4551

### DIFF
--- a/spring-cloud-dataflow-composed-task-runner/src/main/java/org/springframework/cloud/dataflow/composedtaskrunner/ComposedTaskRunnerStepFactory.java
+++ b/spring-cloud-dataflow-composed-task-runner/src/main/java/org/springframework/cloud/dataflow/composedtaskrunner/ComposedTaskRunnerStepFactory.java
@@ -125,7 +125,7 @@ public class ComposedTaskRunnerStepFactory implements FactoryBean<Step> {
 
 		taskLauncherTasklet.setArguments(argumentsToUse);
 
-		log.debug("dedoced composed-task-app-properties {}", composedTaskProperties.getComposedTaskAppProperties());
+		log.debug("decoded composed-task-app-properties {}", composedTaskProperties.getComposedTaskAppProperties());
 
 		Map<String, String> propertiesFrom = Base64Utils
 				.decodeMap(this.composedTaskProperties.getComposedTaskAppProperties()).entrySet().stream()
--- a/spring-cloud-dataflow-composed-task-runner/src/main/java/org/springframework/cloud/dataflow/composedtaskrunner/ComposedTaskRunnerStepFactory.java
+++ b/spring-cloud-dataflow-composed-task-runner/src/main/java/org/springframework/cloud/dataflow/composedtaskrunner/ComposedTaskRunnerStepFactory.java
@@ -23,12 +23,16 @@ import java.util.Map;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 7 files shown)


cppwfs:
Probably been like this the whole time:
*  Launch `t1:timestamp && t2:timestamp` with the following deployer properties.
```
deployer.*.local.java-cmd=java
deployer.*.local.working-directories-root=/Users/glennrenfro/tmp/janne
deployer.t1.local.working-directories-root=/Users/glennrenfro/tmp/janne/two
```
* The updated t1 property is ignored in this case: i.e. CTR, t1, and t2 write to the `/Users/glennrenfro/tmp/janne` directory.   I would have expected t1 to written to the `deployer.t1.local.working-directories-root=/Users/glennrenfro/tmp/janne/two` directory.

cppwfs:
Probably been like this the whole time, but here is the scenario:
*  Launch `t1:timestamp && t2:timestamp` with the following deployer properties.
```
deployer.*.local.java-cmd=java
deployer.*.local.working-directories-root=/Users/glennrenfro/tmp/janne
deployer.t1.local.working-directories-root=/Users/glennrenfro/tmp/janne/two
```
* The updated t1 property is ignored in this case: i.e. CTR, t1, and t2 write to the `/Users/glennrenfro/tmp/janne` directory.   I would have expected t1 to written to the `deployer.t1.local.working-directories-root=/Users/glennrenfro/tmp/janne/two` directory.

cppwfs:
*  Launch `t1:timestamp && t2:timestamp` with the following deployer properties.
* Use builder for t1 to add the `working-directories-root` with the following value `/Users/glennrenfro/tmp/janne/`
* Use builder for t2 to add the `working-directories-root` with the following value `/Users/glennrenfro/tmp/janne/two`
* Launch the task
Result is :
* t1 has a `Platform Properties` of 
` deployer.t1.local.working-directories-root: deployer.t1.local.working-directories-root=/Users/glennrenfro/tmp/janne/`
* t2 has a `Platform Properties` of 
` deployer.t2.local.working-directories-root: deployer.t2.local.working-directories-root=/Users/glennrenfro/tmp/janne/`

cppwfs:
LGTM

cppwfs:
Squashed Merged.

PR #4567, opened 2021-06-08
Task ctr execution doesn't handle wildcard in arguments

### LINKED ISSUE #4566: Task ctr execution doesn't handle wildcard in arguments
Essentially same issue as for properties fixed in #4551. When user defines global field
with arguments, it's silentry dropped.

jvalkeal:
- Add handling of wildcard and base64 encode passed keys with ctr.
- Fixes #4566


cppwfs:
LGTM

PR #4576, opened 2021-06-11
Complete task task resource cleanup on task deletion.

### LINKED ISSUE #4574: Task cleanup should clean slate any and all footprint in TAS
> By design the task destroy expects the task deployment entry to be available for the associated task launch. The task deployment entry would only be created if the task is launched with the associated platform resources and in this case, the entry for the task deployment is missing. 

The log:

```
“TaskLauncher.destroy not invoked for task bmytask. Did not find a previously launched task to destroy”
```

> suggests that the corresponding task deployment wasn’t created because of the failed task launch much before in the lifecycle. But, when the same task is tried to be destroyed it expects the task deployment to exist to cleanup the associated task launch resources. 

In conclusion, a task "cleanup" should clean things up entirely, and not leave any residue behind regardless of whether a task was launches previously successfully or not.

Maybe relates to https://github.com/spring-cloud/spring-cloud-dataflow/issues/4548.

cppwfs:
If there is no task deployment when a task definition is deleted SCDF will delete all app resources for the task definition name across all platforms.
If there is a task deployment for that task definition, then it will only delete those app resources associated with that task deployment.
This is to resolve the case where a task launch failed on the platform and thus no task deployment record is created.  Hence we don't know what platform the launch attempt occured.

resolves #4574

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskDeleteService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskDeleteService.java
@@ -17,6 +17,7 @@
 package org.springframework.cloud.dataflow.server.service.impl;
 
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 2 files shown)


jvalkeal:
lgtm

PR #4577, opened 2021-06-11
Handle cases where an invalid platform is specified.

### LINKED ISSUE #4575: NPE task launches from shell and no platform selected
If multiple platforms are available and the user launches a task from the shell without specifying a platform (When multiple platform errors are available) SCDF server throws NPE
```
java.lang.NullPointerException: null
	at org.springframework.cloud.dataflow.server.service.impl.DefaultTaskExecutionService.validateTaskName(DefaultTaskExecutionService.java:416)
	at org.springframework.cloud.dataflow.server.service.impl.DefaultTaskExecutionService.executeTask(DefaultTaskExecutionService.java:269)
	at org.springframework.cloud.dataflow.server.service.impl.DefaultTaskExecutionService$$FastClassBySpringCGLIB$$422cda43.invoke(<generated>)
```
Return a meaningful exception

cppwfs:
Now SCDF will not throw a NPE

resolves #4575


jvalkeal:
lgtm

PR #4579, opened 2021-06-14
Align docker compose to the build-pack images

tzolov:
- Set the docker compose `COMPOSE_HTTP_TIMEOUT=300` or IT to prevent timeout with the new Skipper image.
 - Replace the default docker mount point for SCDF and Skipper form `/root` to `/home/cnb`, As documented already: https://dataflow.spring.io/docs/installation/local/docker/#accessing-the-host-file-system


jvalkeal:
lgtm

PR #4583, opened 2021-06-15
Handle ctr metadata manually

### LINKED ISSUE #4582: Remove graph property from available CTR property list on task exec page
image

As shown above user can update the graph that is launched via this property.   This property should not be displayed, as this is handled by the SCDF definition itself.

jvalkeal:
- Essentially remove spring-boot-configuration-processor
  from ctr and take existing generated spring-configuration-metadata.json
  and put it in sources.
- Modify spring-configuration-metadata.json to remove graph property.
- Remove spring-cloud-dataflow-apps-metadata-plugin and properties-maven-plugin
  from ctr build as those not needed as paketo is already generating
  label.
- Fixes #4582

### DIFF
--- a/spring-cloud-dataflow-composed-task-runner/pom.xml
+++ b/spring-cloud-dataflow-composed-task-runner/pom.xml
@@ -18,11 +18,6 @@
 			<scope>compile</scope>
 		</dependency>
 		<dependency>
-			<groupId>org.springframework.boot</groupId>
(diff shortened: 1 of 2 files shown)


cppwfs (commented):
I can fix these on merge if it is ok with you.

cppwfs:
I'll fix these on merge along with the removal of @link.

PR #4590, opened 2021-06-21
CF task execs are allowed if previous exec did not record external exec id

### LINKED ISSUE #4587: NPE launching tasks where previous launch has no external exec id
Scenario:
When Dataflow creates a task-execution instance and then attempts to launch the task on CF and the app fails to start, we are left with a Task Execution record with no external id.   
For CF this is problematic because when someone restarts the task, dataflow checks to see if the task is still running on CF using the external execution id.   However since dataflow doesn't have the external execution  id a null is passed to the CF-API which then throws a NPE.
The line exhibiting the behavior is shown below:
https://github.com/spring-cloud/spring-cloud-dataflow/blob/main/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java#L535
Suggested solution is if there is a null in the external-execution-id, then we should allow the task to launch, because we assume it never started.    
This needs to be back ported to 2.7.x.

cppwfs:
resolves #4587


jvalkeal:
lgtm

PR #4610, opened 2021-07-09
polish: statically import JUnit Jupiter assertions

jkschneider:
Co-authored-by: Moderne <team@moderne.io>

Not a big deal, but hope it helps :)


jvalkeal:
Yeah, we can put this in just because you made a PR.

We use mixed set of static/non-static imports and it's just what we've defined for a checkstyle(which we took from spring-cloud). Having said that we're in a long running process to convert everything to assertj from junit/hamcrest, but that's for another day.

PR #4623, opened 2021-07-14
Updated multipleTaskAppVersionTest to use IT timestamp  resource

### LINKED ISSUE #4622: Update IT's multipleTaskAppVersionTest to use url
Currently the `multipleTaskAppVersionTest` uses the platform to determine what resource type to use when testing rollback for timestamp.   This does not work as some platforms can use both `maven` and `docker`.    The test needs to look at the resource type of the timestamp that it is registered at IT start up time to determine the proper older version to register.

cppwfs:
resolves #4622

Instead of using platform to determine which resource type to use for the test it looks at the timestamp's resource type that was registered at the beginning of the integration test.


jvalkeal:
lgtm, lets try it.

PR #4626, opened 2021-07-16
Uses parent task's platform to stop task if it has no platform.

### LINKED ISSUE #3857: NPE thrown when a user attempts to cancel a partitioned worker task
A user will get a NPE at line https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java#L707
when attempting to cancel a partitioned worker through dataflow.   This is because the worker was not launched by dataflow but rather the deployer of the manager task.   So to resolve the issue, if we find that there is no platform associated with a task, see if the task has a parent and then use the parent's platform instead.   If the task does not have a parent, then throw a new exception other than NPE.

cppwfs:
This covers scenarios where a task was launched by another task (i.e. remote partitioned jobs)
However, this will work only on K8's and CF.
Based on how dataflow launches tasks using the local deployer this will not work for child apps launched from a task running using the local deployer.

resolves #3857


jvalkeal:
lgtm

PR #4631, opened 2021-07-19
Tests updated to handle invalid data

### LINKED ISSUE #4630: Definition Delete IT tests fail
In cases where there is task execution data from previous tests where the task execution name is null, these tests will fail.    The asserts check to see if the deleted execution is present, and they do this by looking at the task execution name.   If the name is null then the assert throws a NPE.

cppwfs:
resolves #4630


jvalkeal:
lgtm.

Btw, same thing but I always use nullSafeEquals from frameworks ObjectUtils.

PR #4635, opened 2021-07-21
Add Single Step Job module to Data Flow

### LINKED ISSUE #4346: Add spring-cloud-dataflow-single-step-batch-job module
As we continue to explore the exposition of the single-step-batch-job autoconfiguration's in SCDF's Dashboard, it would require building a new application (similar to CTR), and the source of it would reside under the `spring-cloud-dataflow-single-step-batch-job` module.

**Acceptance:**
- The `spring-cloud-dataflow-single-step-batch-job` module includes the source of the Boot app with all the available autoconfiguration's
- New module is included in the build process
- New module would produce an uber-jar
- New module/app is shipped as part of the project release process
- Includes README / docs / tests

Depends on #4375.

cppwfs:
resolves #4346

This PR implements the following:
* Creates a module for the Single Step Batch Job
* Creates the tests for the Single Step Batch Job
* No documentation is available as we in the process of determining how to register app modules within data flow


pom.xml
```diff
@@ -126,6 +127,11 @@
 				<type>pom</type>
 				<scope>import</scope>
 			</dependency>
+			<dependency>
```
jvalkeal, line 130: I think this is not needed and anyway version in it was wrong.

spring-cloud-dataflow-single-step-batch-job/pom.xml
```diff
@@ -0,0 +1,99 @@
...
+				<exclusion>
+					<groupId>org.junit.vintage</groupId>
+					<artifactId>junit-vintage-engine</artifactId>
+				</exclusion>
+			</exclusions>
+		</dependency>
+		<dependency>
+			<groupId>org.codehaus.plexus</groupId>
+			<artifactId>plexus-utils</artifactId>
+			<version>${plexus.utils.version}</version>
```
jvalkeal, line 82: No versions hidded in child poms. Check a bit deeper as server jar has 3.2.1(probably dependency from deployer maven stuff).

spring-cloud-dataflow-single-step-batch-job/src/main/docker/assembly.xml
```diff
@@ -0,0 +1,38 @@
+<!--
```
jvalkeal, line 1: Docker in maven, bye bye

ilayaperumalg:
Given that we have tasklauncher sink app being part of SCDF along with the single step batch job module, it is worth investigating a common way to register the applications (both stream/task) that come with SCDF.

jvalkeal:
Well yes we should find a common way. We need to brainstorm a good way to handle those things.

pom.xml
cppwfs: This is required in dependency management else the version has to specified in the SingleStepBatchJob pom.    This dependency is not a part of the task starter.

jvalkeal: Ah I didn't realise it's part of a spring-cloud-task.

jvalkeal:
Was looking what gets build into jar and there's some mixed versions from deployer:
```
     8821  2020-11-25 04:07   BOOT-INF/lib/spring-cloud-deployer-autoconfigure-2.5.0.jar
    13429  2021-07-28 09:10   BOOT-INF/lib/spring-cloud-deployer-resource-support-2.7.0-SNAPSHOT.jar
```

cppwfs:
@jvalkeal updated the deployer versions in task master.

jvalkeal:
Actually you could replace explicit deployer deps in dataflow main pom with:
```
			<dependency>
				<artifactId>spring-cloud-deployer-dependencies</artifactId>
				<version>${spring-cloud-deployer.version}</version>
				<type>pom</type>
				<scope>import</scope>
			</dependency>
```
That manages all versions(not sure why we didn't use bom as it's been there)

jvalkeal:
Looks to be in state where it can get merged as it helps with next steps.

PR #4659, opened 2021-08-17
Fix tasklauncher it

tzolov:
- Move the tasklaunncher app registration out of the IT test into the platform provisioning configs. (docker-compose files for IT and installation configs for K8s, CF)
   - Make the test conditional on the dataflow-tasklaunncher is pre-registered.

### DIFF
--- a/spring-cloud-dataflow-server/src/test/java/org/springframework/cloud/dataflow/integration/test/DataFlowIT.java
+++ b/spring-cloud-dataflow-server/src/test/java/org/springframework/cloud/dataflow/integration/test/DataFlowIT.java
@@ -242,7 +242,7 @@ public class DataFlowIT {
 
 		// Maven app without metadata
 		dataFlowOperations.appRegistryOperations().register("maven-app-without-metadata", ApplicationType.sink,
-				"maven://org.springframework.cloud.stream.app:file-sink-kafka:2.1.1.RELEASE", null, true);
+				"maven://org.springframework.cloud.stream.app:file-sink-kafka:3.0.1", null, true);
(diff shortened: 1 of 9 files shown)


spring-cloud-dataflow-server/src/test/java/org/springframework/cloud/dataflow/integration/test/DataFlowIT.java
```diff
@@ -753,6 +756,52 @@ public void namedChannelDirectedGraph() {
...
+	public void dataflowTaskLauncherSink() {
+		logger.info("dataflow-tasklauncher-sink-test");
+
+		if (!runtimeApps.isAppRegistered("dataflow-tasklauncher", ApplicationType.sink)) {
+			logger.info("dataflow-tasklauncher-sink-test: SKIP - no dataflow-tasklauncher app registered!");
+		}
+
+		Assumptions.assumeTrue(runtimeApps.isAppRegistered("dataflow-tasklauncher", ApplicationType.sink),
+				"dataflow-tasklauncher-sink-test: SKIP - no dataflow-tasklauncher app registered!");
+
```
dturanski, line 769: Why do we need 2 checks here?

tzolov: The `Assumptions.assumeTrue` is the real condition that will skip the test if the apps is not registered. But could not figure out how to make it to print its message in the logs. So, the preceding  `if/then->log` block is just to log the same message for skip condition. It is important to log this message to distinct successful test run from the case when it was skipped.
It will be nice if we can avoid this repetition (used in few other tests as well), but I still don't know how.

PR #4662, opened 2021-08-20
Validate K8s specific stream deployment/task launch

### LINKED ISSUE #4640: Spring Data flow server 2.7.2 is not recognizing  stream name with underscore ("_") character
Recently we are trying to migrate from 2.4.1 to 2.7.2 version spring data flow server.
When we try to deploy existing stream name,  **SCDF_test**,  Unable to deploy the stream. We are getting below error.

o.s.c.d.s.c.RestControllerAdvice : Caught exception while handling a request: Stream name must consist of alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name', or 'abc-123')
2021-07-27T10:12:33.017+05:30 [RTR/1] [OUT] dataflow-server.acp..com - [2021-07-27T04:42:32.990442450Z] "POST /streams/definitions HTTP/1.1" 400 835 281

 we have lot of streams are running with underscore("_") character in production environment.
Please can you help on this without renaming all streams to resolve this issue?.  Otherwise Its very tedious process.

ilayaperumalg:
- Instead of restricting the stream creation for all the platforms, restrict the invalid stream/task name only after when they are about to be deployed/launched.
   Since SCDF is aware of the underlying platform only at the time of deployment, we go with this approach of validation

 - Also, to provide compatibility with the previous deployments on local and CF, we don't enforce this validation on non K8s environments

Resolves #4640

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultSchedulerService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultSchedulerService.java
@@ -23,7 +23,9 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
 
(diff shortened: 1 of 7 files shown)


jvalkeal:
lgtm

PR #4671, opened 2021-09-01
Expose ssbj option groups

### LINKED ISSUE #4670: Expose options groups for ssbj
`single step batch job` is a monolith task app having set of readers/writers where combination of valid options depends on what reader and writer is "enabled" at any given time. order to "hint" UI we need to expose these set of options.

jvalkeal:
- single step batch job is a monolith task app having set of readers/writers
  where combination of valid options depends on what reader and writer is
  "enabled" at any given time. order to "hint" UI we need to expose
  these set of options.
- Using same idea how app can define inbount/outbound port names
  by properties file in META-INF, define a set of properties which
  maps into arbitrary grouping id with set of options in it.
- File name is `dataflow-configuration-option-groups.properties` which can contain
  sections named `org.springframework.cloud.dataflow.configuration-properties.option-groups.<xxx>.<ddd>`
  where `<xxx>.<ddd>` becomes grouping id and value of it is a delimited list of options.
- Fixes #4670

For docker needed labels can be added with pack using something like:
```
pack build \
  --path spring-cloud-dataflow-single-step-batch-job/target/spring-cloud-dataflow-single-step-batch-job-2.9.0-SNAPSHOT.jar \
  --builder gcr.io/paketo-buildpacks/builder:base \
  --env BP_JVM_VERSION=16 ghcr.io/jvalkeal/springcloud/spring-cloud-dataflow-single-step-batch-job:2.9.0-SNAPSHOT \
  --env BP_IMAGE_LABELS="$(unzip -p spring-cloud-dataflow-single-step-batch-job/target/spring-cloud-dataflow-single-step-batch-job-2.9.0-SNAPSHOT.jar META-INF/dataflow-configuration-option-groups.properties | sed -e ':a' -e 'N' -e '$!ba' -e 's/\\\n//g' -e 's/ //g' -e 's/\n/ /')"
```

### DIFF
--- a/spring-cloud-dataflow-classic-docs/src/test/java/org/springframework/cloud/dataflow/server/rest/documentation/AppRegistryDocumentation.java
+++ b/spring-cloud-dataflow-classic-docs/src/test/java/org/springframework/cloud/dataflow/server/rest/documentation/AppRegistryDocumentation.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2020 the original author or authors.
+ * Copyright 2016-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


BoykoAlex:
Currently my primary concern is if DSL is pasted into the UI where various type reader properties are mixed (i.e. AMQP, Kafka, flat file) the UI won't be able to determine the proper reader type... Shall we have an artificial property IO Type with value hints  File, Kafka, AMQP, JDBC? What are your thoughts about this corner case?

I understand that user using UI (no DSL) will be guided to have only relevant options. The issue is only for incoming DSL they may have IO options from different types mixed.

cppwfs:
LGTM
Will KafkaItemReader, KafkaItemWriter, AmqpItemReader, and AmqpItemWriter props on merge

cppwfs:
Rebased, Polished, Squashed, Merged

PR #4674, opened 2021-09-03
Update tasklauncher-sink README

dturanski:


spring-cloud-dataflow-tasklauncher/spring-cloud-dataflow-tasklauncher-function/pom.xml
```diff
@@ -17,6 +17,10 @@
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-dataflow-rest-client</artifactId>
 		</dependency>
+		<dependency>
```
jvalkeal, line 20: Was it meant to add this if issue is just to update readme?

dturanski: Yes. That is needed to expose the client connection properties in metadata

PR #4681, opened 2021-09-13
Adding support for ignoreExitMessage

### LINKED ISSUE #2874: Need to add feature for ignore-exit-message
As discussed here: https://github.com/spring-cloud-task-app-starters/composed-task-runner/issues/61  This feature had a PR sometime ago (before it was migrated to dataflow, but was never merged.   Need to update and migrate it over.

cppwfs:
resolves #2874

### DIFF
--- a/spring-cloud-dataflow-composed-task-runner/src/main/java/org/springframework/cloud/dataflow/composedtaskrunner/TaskLauncherTasklet.java
+++ b/spring-cloud-dataflow-composed-task-runner/src/main/java/org/springframework/cloud/dataflow/composedtaskrunner/TaskLauncherTasklet.java
@@ -30,8 +30,8 @@ import org.springframework.batch.core.scope.context.ChunkContext;
 import org.springframework.batch.core.step.tasklet.Tasklet;
 import org.springframework.batch.item.ExecutionContext;
 import org.springframework.batch.repeat.RepeatStatus;
-import org.springframework.boot.context.properties.source.ConfigurationProperty;
-import org.springframework.boot.context.properties.source.ConfigurationPropertyName;
+import org.springframework.boot.context.properties.bind.Bindable;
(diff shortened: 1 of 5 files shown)


spring-cloud-dataflow-composed-task-runner/src/main/java/org/springframework/cloud/dataflow/composedtaskrunner/TaskLauncherTasklet.java
```diff
@@ -170,6 +176,10 @@ public RepeatStatus execute(StepContribution contribution,
 			this.executionId = taskOperations.launch(tmpTaskName,
 					this.properties, args);
 
+			Boolean ignoreExitMessage = isIgnoreExitMessage(args, this.properties);
```
jvalkeal, line 179: Should we try to use boot's binder to handle parsing a property as this looks quite low level to handle camelCase and kebab-case and args on top of that?

cppwfs:
Updated and rebased

jvalkeal:
Binding values manually like boot does is something like:
https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/blob/main/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/ImagePullPolicy.java#L44
https://github.com/spring-cloud/spring-cloud-deployer-kubernetes/blob/main/src/main/java/org/springframework/cloud/deployer/spi/kubernetes/DeploymentPropertiesResolver.java#L652

cppwfs:
I'm not sure the solution above handles our arguments scenario.

cppwfs:
Updated per code review

cppwfs:
Rebased and signed commit.  Ready for review.

jvalkeal:
Looks good, lets try it.

PR #4685, opened 2021-09-14
Replacing runtimeIncrementer with ctr specific incrementer

### LINKED ISSUE #4684: CTR throws DuplicateKeyException when same CTR instance is run concurrently
Composed Task Runner throws a DuplicateKeyException for constraint `job_inst_un`, when multiple instances of the same composed task definition are run simultaneosly.   This is because the runIdIncrementer with the `ISOLATION_REPEATABLE_READ` transaction isolation allow multiple reads of the same job parameters.  
To resolve this we need to create a new incrementer that will use UUID instead of a sequence.

cppwfs:
resolves #4684

### DIFF
--- a/spring-cloud-dataflow-composed-task-runner/src/main/java/org/springframework/cloud/dataflow/composedtaskrunner/ComposedRunnerJobFactory.java
+++ b/spring-cloud-dataflow-composed-task-runner/src/main/java/org/springframework/cloud/dataflow/composedtaskrunner/ComposedRunnerJobFactory.java
@@ -31,6 +31,7 @@ import org.springframework.batch.core.configuration.annotation.JobBuilderFactory
 import org.springframework.batch.core.job.builder.FlowBuilder;
 import org.springframework.batch.core.job.builder.FlowJobBuilder;
 import org.springframework.batch.core.job.flow.Flow;
+import org.springframework.batch.core.launch.support.RunIdIncrementer;
 import org.springframework.beans.factory.FactoryBean;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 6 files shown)


spring-cloud-dataflow-composed-task-runner/src/main/java/org/springframework/cloud/dataflow/composedtaskrunner/ComposedRunnerJobFactory.java
```diff
@@ -352,4 +356,13 @@ private Flow getTaskAppFlow(TaskAppNode taskApp) {
 
 		return new FlowBuilder<Flow>(beanName).from(currentStep).end();
 	}
+
+	public static class UuidIncrementer implements JobParametersIncrementer {
+
+		@Override
+		public JobParameters getNext(JobParameters parameters) {
+			JobParameters params = (parameters == null) ? new JobParameters() : parameters;
+			return new JobParametersBuilder(params).addString(CTR_KEY, UUID.randomUUID().toString()).toJobParameters();
```
jvalkeal, line 365: So this doesn't cause any trouble that key is suddenly changed from `run.id` to `ctr.id` and type from long to string?

cppwfs: Not for batch or CTR.   However, if users are somehow useing the run.id property it could be an issue for them.   We could add a property that allows users to enable the use of UUID versus the default run id incrementer.   Thoughts?

cppwfs:
Updated with new option.

PR #4686, opened 2021-09-15
[Do Not Merge] Validate the registered apps existence 

### LINKED ISSUE #1497: Consider adding some URI filter/validation to AppRegistration when registering a new app

tzolov:
- Add isMetadataResourceExists(Resource app) method to the App configuration metadata resolver APIs - later verifies that the metadata artifact being a metadata jar, application jar or docker image exists physically.

 - Extend the AppRegistryController #register and #registerAll methods with an optional, artefact-validation boolean flag. For backward compatibility the artefact-validation defaults to false.

 - The AppRegistryController#register() method with artefact-validation=true will fail with NonExistingApplicationArtefactException if the URIs are pointing to non-existing artefacts. The application registration is aborted.

 - The new NonExistingApplicationArtefactException exception is added to the RestControllerAdvice

 - The AppRegistryController#register() method with artefact-validation=true will store all valid registration but will remove those that faile to resovle their metadata URIs. The error message lists the failing app's URIs.

 - REST client's AppRegistryOperations#{register,registerimportFromResource,registerAll} methods are extended with an optional artefactValidation parameter.

 - Add some initial tests.

 Resolves #1497

### DIFF
--- a/spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/ApplicationConfigurationMetadataResolver.java
+++ b/spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/ApplicationConfigurationMetadataResolver.java
@@ -62,6 +62,18 @@ public abstract class ApplicationConfigurationMetadataResolver {
 	 */
 	public abstract List<ConfigurationMetadataProperty> listProperties(Resource metadataResource, boolean exhaustive);
 
+	/**
+	 * Verifies if the app metadata resource exists.
+	 * @param app application to check existence for.
+	 * @return Returns true if the app resource exists or false otherwise.
+	 */
+	public abstract boolean isMetadataResourceExists(Resource app);
+
+	/**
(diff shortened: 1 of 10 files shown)


markpollack:
let's speak to christian and see if this is still relevant.

onobc:
I am closing this one, it has been in the same limbo for quite some time.

PR #4696, opened 2021-09-22
Replaced camel case with kabob case in option groups.

### LINKED ISSUE #4692: Use kebab-case format for ssbj options
In `spring-cloud-dataflow-single-step-batch-job/src/main/resources/META-INF/dataflow-configuration-option-groups.properties` all options refs should be with `kebab-case` as that's the boot's way to give out id's for options.

Like `spring.rabbitmq.virtualHost` should be `spring.rabbitmq.virtual-host`.

cppwfs:
resolves #4692


jvalkeal:
afaik, lgtm

PR #4709, opened 2021-10-05
(DO NOT MERGE) Enable inherit logging for ITs

tzolov:
- Introduce a new TEST_DOCKER_COMPOSE_APPS_INHERIT_LOGGING env. variable
   and corresponding test.docker.compose.apps.inherit.logging property to allow inherit the streaming and task logs
   into the logs of the Data Flow or Skipper servers instead. Later are IT persisted artifacts.
 - Inherit logging is disabled by default.
 - Uses the local deployer inherit property: https://docs.spring.io/spring-cloud-dataflow/docs/current/reference/htmlsingle/#configuration-local-deployer

### DIFF
--- a/src/docker-compose/docker-compose.yml
+++ b/src/docker-compose/docker-compose.yml
@@ -19,6 +19,10 @@ version: '3'
 #                             That means you can reach the application's actuator endpoints from your host machine.
 #                             The deployed stream applications that run in their own docker containers (e.g. docker:// registered apps),
 #                             can be reached on the ports they expose.
+#
+# - APPS_INHERIT_LOGGING - When enabled it redirects the streaming and task applications logging to the logs of to
+#                          the Data Flow or Skipper logs instead.
+
 services:
   mysql:
     image: mysql:5.7.25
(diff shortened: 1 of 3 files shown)


src/docker-compose/docker-compose.yml
```diff
@@ -19,6 +19,10 @@ version: '3'
 #                             That means you can reach the application's actuator endpoints from your host machine.
 #                             The deployed stream applications that run in their own docker containers (e.g. docker:// registered apps),
 #                             can be reached on the ports they expose.
+#
+# - APPS_INHERIT_LOGGING - When enabled it redirects the streaming and task applications logging to the logs of to
```
cppwfs, line 23: Reword it to something like:  When enabled, streaming and task applications log entries are also written to the Spring Cloud Data Flow or Skipper Server logs.

tzolov: @cppwfs please ignore this PR in favor of this instead: https://github.com/spring-cloud/spring-cloud-dataflow/pull/4711

I just found out that we can not safely enable inherit logging for the streaming tests as this messes up how the test validate the stream logs. A better approach is to allow gradual per-stream or per-task inherit logging support configured with plain docker compose file composition (this is what the 4711) does

tzolov:
Replaced by  #4711

PR #4711, opened 2021-10-05
Enable inherit logging for ITs (2)

### LINKED ISSUE #4710: Allow collecting the stream and task application logs for Integration tests.
**Problem description:**
In order to debug sporadically failing Integration Tests (ITs) we need to be able to collect the logs of the streaming and task pipeline apps. 
Currently the ITs collect the logs of all docker containers started by the docker-compose (e.g. Data Flow , Skipper, MySQL, Kafka, ...). Later are retained as build artifacts. 
We do not though retain the logs for the running applications part for the tests nor for the Composed Task Runner!

**Solution description:**
The Data Flow already provides a [inherit logging](https://docs.spring.io/spring-cloud-dataflow/docs/current/reference/htmlsingle/#configuration-local-deployer) feature for the Local Deployer. This allows the logs of the running task and stream applications to be inlined into the parent's (e.g. DataFlow or Skipper) logs that are kept as build artifacts. 

It is required to allow enabling/disabling the inherit logging declaratively for the ITs.

tzolov:
Additional docker-compose file to set the inherit-logging on.
   One can use the -Dtest.docker.compose.paths to add it to the ITs list of docker-compose files.

   Resolves #4710

### DIFF
--- a/spring-cloud-dataflow-server/src/test/resources/docker-compose-inherit-logging.yml
+++ b/spring-cloud-dataflow-server/src/test/resources/docker-compose-inherit-logging.yml
@@ -1,0 +1,13 @@
+version: '3'
+
+# Enable inherit-logging for all stream and task applications. This allows the logs for those apps (including CRT)
+# to be inlined within the logs of the parent (e.g. Data Flow or Skipper) applications.
+# The parent logs are often preserved by the CI as build artefacts, this would allow collect and preserver
+# the apps logs as well.
+services:
+  dataflow-server:
+    environment:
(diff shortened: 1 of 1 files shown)


spring-cloud-dataflow-server/src/test/resources/docker-compose-inherit-logging.yml
```diff
@@ -0,0 +1,13 @@
+version: '3'
+
+# Enable inherit-logging for all stream and task applications. This allows the logs for those apps (including CRT)
```
cppwfs, line 3: - (including CTR)

- preserver -> preserve

PR #4723, opened 2021-10-14
Updated to support chunking of large deletes of Task Executions

### LINKED ISSUE #4722: Support chunking of large task execution deletes
Currently all deletes of a task execution occur in one single event.   The user should be able to configure data flow to break these large deletes into chunks.

cppwfs:
resolves #4722


jvalkeal:
lgtm

PR #4772, opened 2021-11-25
Update TaskSchedulerDocumentation.java

w0wka91:
Add platform parameter


spring-cloud-dataflow-classic-docs/src/test/java/org/springframework/cloud/dataflow/server/rest/documentation/TaskSchedulerDocumentation.java
```diff
@@ -67,6 +67,7 @@ public void createSchedule() throws Exception {
 				.andDo(this.documentationHandler.document(
 						requestParameters(
 								parameterWithName("scheduleName").description("The name for the created schedule"),
+								parameterWithName("platform").description("The name of the platform the task is launched"),
```
cppwfs, line 70: Thank you for your contribution.   
Currently your solution will not build.   Please add `.param("platform", "default")` to the param list above  between lines lines 63-64.   

You can build your solution by executing `mvn clean package -Pfull -Dcheckstyle.skip=true` in the spring-cloud-data-flow-classic-docs module.

cppwfs:
Fixed compile error on merge.   Rebased, Squashed, Merged

Thank you for your contribution.

PR #4790, opened 2021-12-15
Break circular dependency between shell beans

### LINKED ISSUE #4783: Shell breaks with circular dependency
Boot from `2.6.x` is now enforcing not to have circular deps.
```
***************************
APPLICATION FAILED TO START
***************************

Description:

The dependencies of some of the beans in the application context form a cycle:

   scriptCommands (field private org.springframework.shell.core.JLineShellComponent org.springframework.shell.commands.ScriptCommands.shell)
      ↓
   shell defined in class path resource [org/springframework/cloud/dataflow/shell/autoconfigure/BaseShellAutoConfiguration.class]
┌─────┐
|  configCommands (field private org.springframework.web.client.RestTemplate org.springframework.cloud.dataflow.shell.command.ConfigCommands.restTemplate)

```

Need to think how to fix this in `spring-shell` or optionally use `spring.main.allow-circular-references` in shell app itself.

hpoettker:
Resolves #4783.

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/TaskConfiguration.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/TaskConfiguration.java
@@ -129,7 +129,7 @@ public class TaskConfiguration {
 	 */
 	@Profile({ "local", "default" })
 	@Bean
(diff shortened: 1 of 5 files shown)


hpoettker:
In the first commit, I missed the second `spring.main.allow-circular-references` that was set in the tests. Now I also see the problem with `spring-shell`.

Issue #4700, opened 2021-09-27
Database schema of Data Flow Server cannot be changed


uunilohi:
We tried to change the database schema of Data Flow Server (2.8.2). With the configuration below, the tables are created to a database schema called "scdf_dataflow" as expected.

```
spring:
  batch:
    table-prefix: scdf_dataflow.batch_
  cloud
    task:
      table-prefix: scdf_dataflow.task_
  flyway:
    default-schema: scdf_dataflow
    schemas: scdf_dataflow
  jpa:
    properties:
      hibernate:
        default_schema: scdf_dataflow
```
However when navigating to "Job executions" in Dashboard, an exception is thrown:
```
2021-09-27 08:25:20.367 ERROR 1 --- [nio-8080-exec-9] o.s.c.d.s.c.RestControllerAdvice         : Caught exception while handling a request
org.springframework.jdbc.BadSqlGrammarException: StatementCallback; bad SQL grammar [SELECT E.JOB_EXECUTION_ID, E.START_TIME, E.END_TIME, E.STATUS, E.EXIT_CODE, E.EXIT_MESSAGE, E.CREATE_TIME, E.LAST_UPDATED, E.VERSION, I.JOB_INSTANCE_ID, I.JOB_NAME, (SELECT COUNT(*) FROM BATCH_STEP_EXECUTION S WHERE S.JOB_EXECUTION_ID = E.JOB_EXECUTION_ID) as STEP_COUNT FROM BATCH_JOB_EXECUTION E, BATCH_JOB_INSTANCE I WHERE E.JOB_INSTANCE_ID=I.JOB_INSTANCE_ID ORDER BY JOB_EXECUTION_ID DESC LIMIT 20]; nested exception is org.postgresql.util.PSQLException: ERROR: relation "batch_job_execution" does not exist
  Position: 277
	at org.springframework.jdbc.support.SQLErrorCodeSQLExceptionTranslator.doTranslate(SQLErrorCodeSQLExceptionTranslator.java:239)
	at org.springframework.jdbc.support.AbstractFallbackSQLExceptionTranslator.translate(AbstractFallbackSQLExceptionTranslator.java:70)
	at org.springframework.jdbc.core.JdbcTemplate.translateException(JdbcTemplate.java:1541)
	at org.springframework.jdbc.core.JdbcTemplate.execute(JdbcTemplate.java:393)
[...]

PR #4800, opened 2022-01-20
Fix issue #4700 (Database schema of Data Flow Server cannot be changed)

aritzbastida:
Fix issue #4700 (Database schema of Data Flow Server cannot be changed)

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfiguration.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfiguration.java
@@ -16,6 +16,8 @@
 
 package org.springframework.cloud.dataflow.server.config;
 
+import java.util.Optional;
+
 import javax.persistence.EntityManager;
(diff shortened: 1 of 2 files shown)


jvalkeal:
Unfortunately supporting supporting tables like this is much more involved than just trying to use prefix from `BatchProperties`. Dataflow owns all the tables and possible solution should integrate to flyway layer and ddl files we keep for those who can't give servers rights to create tables.

aritzbastida:
I agree, in fact, even if we do accept this PR, the problem still remains in all SCDF-specific tables.
Right now, there are three different strategies in SCDF code related to "table prefix":
* Spring Cloud Task: Table prefix is considered (%PREFIX%) from TaskProperties. For example: JdbcDataflowTaskExecutionDao.
* Spring Batch: Table prefix is considered (%PREFIX%), but hardcoded to its default value (BATCH_). For example: JdbcDataflowJobExecutionDao.
* SCDF-specific: No table prefix in SQL queries. For example: JdbcDataflowTaskExecutionMetadataDao.

aritzbastida:
Closing this PR, as it does not fully solve the problem. The analysis can go on in issue #4700.

Issue #4803, opened 2022-01-28
Rendering errors in Job Executions view when there are jobs not tracked to tasks


aritzbastida:
If there is a job execution not related to Spring Cloud Task, the rendering of Job Executions view will fail with the following message: "_No corresponding taskExecutionId for jobExecutionId xx. This indicates that Spring Batch application has been executed that is not a Spring Cloud Task_". No results are shown in the table (even for other job executions correctly matched to a task).

Currently, we are sharing our JOB_EXECUTIONS table with other (legacy) Spring Batch jobs, launched via Spring Batch Admin (and will upgrade them to SCDF sometime soon).

As Spring Cloud Task is a prerrequisite for SCDF (a task _need not_ be a batch, but a batch _must_ be a task), I think that it would be just simpler to filter out job executions not related to a task. All legacy jobs would then be invisible to SCDF, instead of making it fail.

I am sharing PR https://github.com/spring-cloud/spring-cloud-dataflow/pull/4804 with the proposal.

aritzbastida:
Closing this issue, as the PR was rejected. Maintaining backward-compatibility with legacy jobs (not integrated with Spring Cloud Task) is not a supported use case.

PR #4804, opened 2022-01-28
Filter job executions not integrated with Spring Cloud Task

aritzbastida:
(cherry picked from commit 96aa4207e63e00e411b334cd71a8037aec284916)

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/JdbcSearchableJobExecutionDao.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/JdbcSearchableJobExecutionDao.java
@@ -49,7 +49,7 @@ import org.springframework.util.Assert;
  */
 public class JdbcSearchableJobExecutionDao extends JdbcJobExecutionDao implements SearchableJobExecutionDao {
 
-	private static final String GET_COUNT = "SELECT COUNT(1) from %PREFIX%JOB_EXECUTION";
+	private static final String GET_COUNT = "SELECT COUNT(1) from %PREFIX%JOB_EXECUTION E, TASK_TASK_BATCH B WHERE B.JOB_EXECUTION_ID = E.JOB_EXECUTION_ID";
 
 	private static final String GET_COUNT_BY_JOB_NAME = "SELECT COUNT(1) from %PREFIX%JOB_EXECUTION E, %PREFIX%JOB_INSTANCE I "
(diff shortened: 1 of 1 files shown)


onobc:
Hi @aritzbastida . Thanks for the PR. I am going to take a look shortly. One thing though, I can't see to find the commit that you cherry-picked from. Can you double-check and/or give me some more info on the commit? 

Thanks

aritzbastida:
Hi! I cherry-picked from a local branch; that's why you cannot see commit 96aa420. That message was auto-generated by Git (and I did not edit it). It can be a bit misleading, sorry. 

However, the PR is self-contained. As you can see, it just contains one commit (61ef421) and one file modification, compared to branch 2.9.x.

BTW, I also opened a related issue (https://github.com/spring-cloud/spring-cloud-dataflow/issues/4803). Is that redundant with this? How should I proceed when I want to open an issue with a related PR?

onobc:
> Hi! I cherry-picked from a local branch; that's why you cannot see commit 96aa420. That message was auto-generated by Git (and I did not edit it). It can be a bit misleading, sorry.

No problem. Thanks for the clarification.

> BTW, I also opened a related issue (#4803). Is that redundant with this? How should I proceed when I want to open an issue with a related PR?

So in Spring projects in general, if you are submitting a PR there is no need to also submit an issue. If you find an issue but only want to report it, then you can just file an issue. 

I will take a look at the code proposal later today (as soon as I free up) but one concern I have is that I see no tests updated etc.. Can you see if there is an existing test(s) that we can add/update a test to cover this case? 

Thanks

aritzbastida:
Ok, so I won't open an issue next time. Thanks for the info! 

I just checked that this code does not break any (additional) unit test in spring-cloud-dataflow-server-core module, but did not write a new one on my own. In fact, I don't see any test for "org.springframework.cloud.dataflow.server.batch" package!

onobc:
Hi @aritzbastida 

First off, thanks for the proposal and your efforts. 

I spoke w/ the team today in regards to this issue. Our stance is that we do not support this use case. We do understand that having the executions view throw an error is not ideal.

Currently Spring Cloud Data Flow only supports Spring Boot applications that utilize Spring Cloud Task when launching tasks.  This allows SCDF to support many types of tasks across multiple platforms including those that utilize Spring Batch.    

Again, we do appreciate your contributions and efforts.

aritzbastida:
Ok, thanks for considering the PR! I will close the related issue (#4803) as well.

onobc:
Thank you @aritzbastida

Issue #4802, opened 2022-01-28
Remove dependency of com.h2database:h2 for production deployments


michael-wirth:
**Problem description:**
The class `org.springframework.cloud.dataflow.server.config.web.WebConfiguration` initialized an embedded h2 database server if the datasource URL starts with "jdbc:h2:tcp://localhost:". As there is a direct dependency on the `org.h2.tools.Server` class it is not possible run SCDF without the h2 database driver if its needed or not.

https://github.com/spring-cloud/spring-cloud-dataflow/blob/main/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/web/WebConfiguration.java#L107

**Solution description:**
Decouple the logic into it's own Configuration file and use the `ConditionalOnClass(org.h2.tools.Server.class)` annotation

**Description of alternatives:**

**Additional context:**

chringwer:
+1 Starts showing up in security scans due to [CVE-2021-42392](https://nvd.nist.gov/vuln/detail/CVE-2021-42392)

altfatterz:
+1

PR #4806, opened 2022-02-01
fix: remove h2 dependency

michael-wirth:
Fixes gh-4802


spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfigurationTests.java
```diff
@@ -88,7 +89,8 @@ public void setup() {
 				DataFlowServerConfiguration.class, PropertyPlaceholderAutoConfiguration.class,
 				RestTemplateAutoConfiguration.class, HibernateJpaAutoConfiguration.class, WebConfiguration.class,
 				SchedulerConfiguration.class, JacksonAutoConfiguration.class, SimpleTaskAutoConfiguration.class,
-				ResourceLoadingAutoConfiguration.class, ComposedTaskRunnerConfigurationProperties.class);
+				ResourceLoadingAutoConfiguration.class, ComposedTaskRunnerConfigurationProperties.class,
+				H2ServerConfiguration.class);
```
onobc, line 93: I don't think this is needed since DataFlowServerConfiguration brings it in. 

**[ALREADY-IN-CODE]** Neither does the WebConfiguration.

spring-cloud-dataflow-shell/pom.xml
```diff
@@ -13,6 +13,11 @@
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-dataflow-shell-core</artifactId>
 		</dependency>
+		<dependency>
```
onobc, line 16: There are not tests in this module. I think we can remove this here.

onobc (changes requested):
@michael-wirth I left a few comments for you to address, but other than that - looks good. Nice work.

spring-cloud-dataflow-shell/pom.xml
michael-wirth: done

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfigurationTests.java
michael-wirth: I removed the WebConfiguration and H2ServerConfiguration

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/web/H2ServerConfiguration.java
```diff
@@ -0,0 +1,88 @@
...
+public class H2ServerConfiguration
+		implements ServletContextInitializer, ApplicationListener<ContextClosedEvent> {
+
+	private static final Logger logger = LoggerFactory
+			.getLogger(H2ServerConfiguration.class);
+
+	@Value("${spring.datasource.url:#{null}}")
+	private String dataSourceUrl;
+
+	private Server server = null;
```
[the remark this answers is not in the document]
michael-wirth: done. This was copied from the original WebConfiguration.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/web/H2ServerConfiguration.java
```diff
@@ -0,0 +1,88 @@
+/*
+ * Copyright 2015-2019 the original author or authors.
```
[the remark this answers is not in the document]
michael-wirth: Done

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/web/H2ServerConfiguration.java
```diff
@@ -0,0 +1,88 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionalOnWebApplication;
+import org.springframework.boot.web.servlet.ServletContextInitializer;
+import org.springframework.context.ApplicationListener;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.event.ContextClosedEvent;
+import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
+
+/**
+ * @author Wirth Michael
```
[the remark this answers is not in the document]
michael-wirth: Done

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/web/H2ServerConfiguration.java
```diff
@@ -0,0 +1,88 @@
...
+ */
+@Configuration
```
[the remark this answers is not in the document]
michael-wirth: Done

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/web/H2ServerConfiguration.java
```diff
@@ -0,0 +1,88 @@
...
+
+	private static final Logger logger = LoggerFactory
+			.getLogger(H2ServerConfiguration.class);
+
+	@Value("${spring.datasource.url:#{null}}")
+	private String dataSourceUrl;
+
+	private Server server = null;
+
+	public Server initH2TCPServer() {
```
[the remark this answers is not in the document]
michael-wirth: I think this was supposed to be a `@Bean`.
The [test](https://github.com/spring-cloud/spring-cloud-dataflow/blob/main/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfigurationTests.java#L150) seems to verify if a bean with name `initH2TCPServer has been created.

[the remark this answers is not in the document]
michael-wirth: the test doesn't seem to work at all. one test is ignored the other tests are don't work as expected as the H2 Server is never initialized in the test.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfiguration.java
```diff
@@ -70,7 +71,8 @@
  */
 @EnableSpringDataWebSupport
 @Configuration
-@Import({ CompletionConfiguration.class, FeaturesConfiguration.class, WebConfiguration.class })
+@Import({ CompletionConfiguration.class, FeaturesConfiguration.class, WebConfiguration.class,
```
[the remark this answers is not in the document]
michael-wirth: Done

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/web/H2ServerConfiguration.java
```diff
@@ -0,0 +1,88 @@
...
+
+	private String getH2Port(String url) {
+		String[] tokens = StringUtils.tokenizeToStringArray(url, ":");
+		Assert.isTrue(tokens.length >= 5, "URL not properly formatted");
+		return tokens[4].substring(0, tokens[4].indexOf("/"));
+	}
+
+	@Override
+	public void onStartup(ServletContext servletContext) {
+		if (StringUtils.hasText(dataSourceUrl)
```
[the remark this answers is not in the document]
michael-wirth: I agree the initialization of the H2Server requires a redesign. The related tests aren't working properly because the test requires to be run in WebApplication to start properly.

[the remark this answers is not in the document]
onobc: > The related tests aren't working properly because the test requires to be run in WebApplication to start properly.

Are they failing?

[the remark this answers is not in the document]
onobc: I see - the test that you linked in previous comment is passing but not verifying anything.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/web/H2ServerConfiguration.java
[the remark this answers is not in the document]
onobc: Ahh, so this worked before because `proxyBeanMethods` was true. This allows manual calls to bean methods (like `initH2TCPServer`). And in these cases the method does not need to be annotated w/ `@Bean`. There is no clear indication that the method produces a bean. This is the other reason besides perf improvement that we get rid of `proxyBeanMethods` - it leads to unclear bean paths etc..

So a couple of options:

1. Go back to `proxyBeanMethods = true` and make method the method public 
2. Modify it more to use an `Optional<Server>` reference and make the method a bean method (this gets ugly fast)
3. Go w/ auto-config conditional approach I outlined [here](https://github.com/spring-cloud/spring-cloud-dataflow/pull/4806#discussion_r797756988). 

I would do 1) or 3). I think we should do 1) in this PR and create an issue for rework to do 3) later. 

Wdyt?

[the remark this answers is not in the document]
michael-wirth: Thanks for the suggestions. The H2 server initialization is working as it is implemented now. 
But the tests can't be fixed that way. [The tests are disabled for 4 years already.](https://github.com/spring-cloud/spring-cloud-dataflow/blame/main/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfigurationTests.java#L107)

The **WebConfiguration** and the **H2ServerConfiguration** are only initialized on the `@ConditionalOnWebApplication` condition. 
The positive test was ignored and the other negative tests always succeeded because the H2 Server was never ever initialized. It requires a web application context to get the H2 server started, but the tests run without a web server.

I don't think it is worth going back to a former state if it doesn't fix the tests at all.
I think it would be easier to startup the H2 Server in an `ApplicatonEventlistener` or a `BeanFactoryPostProcessor` to decouple from the web application condition.

I prepared a POC [here](https://github.com/michael-wirth/spring-cloud-dataflow/blob/bugfix%2F4802-fix-tests/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/H2ServerConfiguration.java)

Wdyt?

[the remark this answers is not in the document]
onobc: > I don't think it is worth going back to a former state if it doesn't fix the tests at all.

I was not sure now much time/effort you could put into this but I do agree that "fixing it" would be the preferred approach.

> The WebConfiguration and the H2ServerConfiguration are only initialized on the @ConditionalOnWebApplication condition.

I did not catch that H2 config was still gated on COWA in my initial review. Agreed, it is not needed. 

Your POC looks good. However, I would prefer doing it via a custom condition that simple does not the let config load. 

Also, a couple of other notes/improvements:

1. We can replace the manual call of "server.stop" w/ `@Bean(destroyMethod = "stop")`. Or even better yet, do nothing for this, as Spring will infer the destroy method of "shutdown" [automatically](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/annotation/Bean.html#destroyMethod--).

2. I am in favor of splitting the test out into H2 and Web variants and rewrite whatever we need to. Somewhat of the "If I was going to write brand new tests for this, what would it look like?" For the HC config test we could leverage ApplicationContextRunner (setting properties and filtering out classes).

Thoughts?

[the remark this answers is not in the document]
michael-wirth: - updated the H2ServerConfiguration with the following annotations
```java
@Configuration(proxyBeanMethods = false)
@ConditionalOnClass(Server.class)
@ConditionalOnProperty(value = "spring.dataflow.embedded.database.enabled", havingValue = "true", matchIfMissing = true)
@ConditionalOnExpression("'${spring.datasource.url:#{null}}'.startsWith('jdbc:h2:tcp://localhost:')")
public class H2ServerConfiguration {
```
- removed the ApplicationListener and just provide a `public void close()` method to stop the H2 server
- updated the tests to use the `ApplicationContextRunner` (didn't know this feature, thanks!)

I'm not sure if there is another solution to initialize the H2 server than with a `BeanFactoryPostProcessor`. 
Is there a way to force a bean to be initialized before the `dataSource` bean and `Flyway` are initialized? The `entityManagerFactory` bean is of type `LoadTimeWeaverAware` and initialized, and with it all the datasource connections with it, right after all the listeners and beanfactorypostprocessor were initialized. 

Is there a way to express that a bean should be initialized "before" another bean?

As an alternative, I could just think of using an `ApplicationListener` and register it in the `META-INF/spring.factories` file.

Thoughts?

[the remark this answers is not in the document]
onobc: Hi Michael,

> updated the tests to use the ApplicationContextRunner (didn't know this feature, thanks!)

You are welcome. I was happy when it was shown to me as well (one of my favorite tools). 

> Is there a way to force a bean to be initialized before the dataSource bean and Flyway are initialized? 

So the creation/init order is determined by the container based mostly on bean references to one another. When beans do not have a direct reference on one another, [@DependsOn](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/annotation/DependsOn.html) can be used on a bean to tell the container that this bean requires the other bean. However, it can only be specified in one direction (in our case we can't annotate DataSource, only the bean we control). 

SpringBoot has the concept of [@AutoConfigureBefore/After](https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/autoconfigure/AutoConfigureBefore.html) which is used to control the order of bean definition amongst auto-configs (aka make sure that there is a bean A auto-configured and defined/available before bean B is auto-configured). However, this ordering is orthogonal to bean creation/init. Also, in this case it is not that the DatasourceAutoConfiguration "needs" a bean from H2SC, we just need to be sure that the DB server is up and available before all the DB-needy things start trying to invoke it.

#### PostConstruct
Just using the default container-determined loading and putting the start call in a `@PostConstruct` may be fine. Did you try that approach? 

#### AutoConfig + BFPP (your current proposal)
**However**, even if it works with `@PostConstruct`, since this is really being treated as "the DB server for the app", it makes sense to get a stronger guarantee that the server is up before the things that need it. Your current approach probably works. However,
1. I would have the BFPP implement `@PriorityOrdered` to ensure its invokes earlier than any other BFPP that we are unaware of related to DB-needy things. 
2. another thing to check: I am not sure if close() will be auto-called later for beans added directly to the BF

#### AutoConfig + ApplicationListener<ApplicationPreparedEvent>
Another approach would be, as you said, instead create an `ApplicationListener<ApplicationPreparedEvent>` to issue the start(). This would work nicely w/ the caveat that there is no easy way to handle conditional classes w/ spring.factories loaded listeners. The listener would load regardless and it would have code in it that referenced `Server.class` and would bomb when it was not on the classpath. There are ways around this by having a helper class that is only invoked if the class is available. It gets overly complicated quickly. 

#### ApplicationContextInitializer only
And yet another approach. In the past when I have done things like this, I have favored doing them entirely in an `ApplicationContextInitializer`.  The beauty of this approach is its self-contained and it is easy to set the execution precedence. The non-beauty of it is the same caveat w/ conditional classes exists here as well as having to inspect the properties that we are currently doing in the auto-config conditionals.

### Summary
No clear cut winner. I would dig into `PostConstruct` approach first and then go w/ `AutoConfig + BFPP (your current proposal)` if that does not work.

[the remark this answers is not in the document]
onobc: @michael-wirth 

I went ahead and tried some of the above approaches. What I found was that the JPA EM kicks in the datasource creation **very early** (in attempt to invoke sql init scripts). 

#### Auto-Config + BFPP 
✅ This is your latest proposal and it works well. It loads before the DatasourceAutoConfiguration needs it. As for my original concerns on adding PriorityOrdered I don't think we need this after all. Also, I believe close is being called as expected. 

#### Post-Construct
❌ This did not work as the JPA EM always wins in init order.

#### AutoConfig + ApplicationListener
❌ This did not work as I was unable to get the auto-config to even define before the datasource autoconfig loaded and failed w/o backing connection. 

#### ACI
❓ I skipped this as it would require overly complicated code to make conditional on class and property logic. 


I will next spend some time reviewing your changes to the tests.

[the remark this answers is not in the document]
michael-wirth: @onobc 

I found a way to update the "dependsOn" property of the `dataSource` bean definition in the BFPP and then initialize the H2Server as a normal bean. Please see https://github.com/michael-wirth/spring-cloud-dataflow/pull/2/files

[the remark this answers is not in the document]
onobc: Ahhh crafty! Thanks for the continued digging on this one - it has been fun. I think we are getting close.

So I verified locally that this works well too. To help us decide between the 2 approaches....

## BFPP w/ bean create and server start
- ✅ no knowledge of bean inter-relationships
- ❌ bean not defined in the config 

## BFPP w/ DependsOn
- ✅ bean defined in the config
- ❌ relies on knowledge of bean inter-relationships (is there any other beans that would need this same treatment?)
- ❌ brittle on bean name (add the following config to your app and it will break)
  ```java
	@Bean
	DataSource michaelsSuperSpecialCustomDatasource(DataSourceProperties properties) {
		return properties.initializeDataSourceBuilder().type(HikariDataSource.class).build();
	}
  ````

On the ❌ items in the 2nd option, it can be remedied by applying the same treatment to all DS beans in the context. It does start to feel more intrusive at this point. 

Did I miss anything in the pro/cons? Which direction are you leaning? 

**[SIDE_NOTE]**
ℹ️ If we go w/ the bean approach we can omit the shutdownHandler as the container will infer the `shutdown` method on the `Server`. I think even in the 1st approach we can rely on this to be called.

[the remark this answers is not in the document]
michael-wirth: @onobc 

I think that sums it up. 
I found another solution with BFPP and @Bean annotation, but instead of defining `dependsOn` properties the H2 server can be instantiated by just calling `beanFactory.getBean("initH2TCPServer");`

Hope that is a step to the final solution.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/H2ServerConfiguration.java
```diff
@@ -0,0 +1,81 @@
...
+		String[] tokens = StringUtils.tokenizeToStringArray(url, ":");
+		Assert.isTrue(tokens.length >= 5, "URL not properly formatted");
+		return tokens[4].substring(0, tokens[4].indexOf("/"));
+	}
+
+	static class H2ServerBeanFactoryPostProcessor implements BeanFactoryPostProcessor {
+
+		@Override
+		public void postProcessBeanFactory(ConfigurableListableBeanFactory beanFactory) throws BeansException {
+			beanFactory.getBean("initH2TCPServer");
```
onobc, line 78: Ahh yes. This does the trick. We want something to run very early and ask for the H2Server bean so that it gets initialized. Nice work.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/H2ServerConfiguration.java
```diff
@@ -0,0 +1,81 @@
...
+		}
+	}
+
+	private String getH2Port(String url) {
+		String[] tokens = StringUtils.tokenizeToStringArray(url, ":");
+		Assert.isTrue(tokens.length >= 5, "URL not properly formatted");
+		return tokens[4].substring(0, tokens[4].indexOf("/"));
+	}
+
+	static class H2ServerBeanFactoryPostProcessor implements BeanFactoryPostProcessor {
```
onobc, line 74: I think a comment that captures the BFPP sole purpose may be helpful, something like: 
```java
/**
 * A {@link BeanFactoryPostProcessor} whose sole job is to ensure that the H2 server is up and running before any 
 *  datasource initialization is attempted. It does this by requesting the H2Server bean which then in turn starts up 
 *  the server. 
 * /

File: spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/H2ServerConfiguration.java [hunk 12]
```diff
@@ -0,0 +1,81 @@
...
+
+	private static final Logger logger = LoggerFactory.getLogger(H2ServerConfiguration.class);
+
+	@Bean
+	public H2ServerBeanFactoryPostProcessor h2ServerBeanFactoryPostProcessor() {
+		return new H2ServerBeanFactoryPostProcessor();
+	}
+
+	@Bean
+	public Server initH2TCPServer(@Value("${spring.datasource.url}") String dataSourceUrl) {
```
onobc, line 55: Wdyt about renaming this bean to something like `h2TcpServer`?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/H2ServerConfiguration.java
```diff
@@ -0,0 +1,81 @@
...
+		logger.info("Starting H2 Server with URL: " + dataSourceUrl);
+		try {
+			Server server = Server.createTcpServer("-ifNotExists", "-tcp",
+					"-tcpAllowOthers", "-tcpPort", getH2Port(dataSourceUrl)).start();
+			server.setShutdownHandler(server::stop);
```
onobc, line 60: We can remove this altogether as the container will call `Server.shutdown` which in turn calls `stop`.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/H2ServerConfiguration.java
```diff
@@ -0,0 +1,81 @@
...
+			return server;
+		}
+		catch (SQLException e) {
+			throw new IllegalStateException(e);
+		}
+	}
+
+	private String getH2Port(String url) {
```
onobc, line 68: It may be cleaner to just inline this now that its only called in a single place.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/web/H2ServerConfiguration.java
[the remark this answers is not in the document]
onobc: I like where you landed on this - looks good. I left some minor comments. I will review the tests tomorrow and then we can get this one closed out. Thank you.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/H2ServerConfiguration.java
michael-wirth: Unfortunately, the default `shutdown` method calls `stopAll` instead of `stop`.
But `stopAll` can only stop services, which have been started by the `main` method.
When starting a server by the `createTcpServer` method it doesn't assign the running instance to [the `tcp` variable](https://github.com/h2database/h2database/blob/master/h2/src/main/org/h2/tools/Server.java#L324).
I don't know if the H2 library has been designed this way or if it is a bug.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/H2ServerConfiguration.java
michael-wirth: I inlined it and replaced it with a Regex expression

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/H2ServerConfiguration.java
michael-wirth: I agree, I didn't want to make too many changes to the original code and didn't want to break anything in case somebody is tracking that bean.

Done

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/H2ServerConfiguration.java
michael-wirth: done

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfigurationTests.java
```diff
@@ -51,118 +47,102 @@
...
+						SchedulerConfiguration.class, JacksonAutoConfiguration.class, SimpleTaskAutoConfiguration.class,
+						ResourceLoadingAutoConfiguration.class, ComposedTaskRunnerConfigurationProperties.class);
 	}
 
 	/**
 	 * Verify that embedded server starts if h2 url is specified with default properties.
 	 */
 	@Test
-	@Ignore
 	public void testStartEmbeddedH2Server() {
```
onobc, line 89: I am going to share w/ you some of the beauty of the ApplicationContextRunner. One thing that it lets us do is really break apart the rest of the system and focus solely on the tiny piece of autoconfig we are testing. You will notice this current test has to bring in the kitchen sink to startup as it tests the H2Server startup + Dataflow server. 

Take a look at this instead: 
```java
/**
 * Tests for {@link H2ServerConfiguration}.
 *
 * @author Michael Wirth
 */
public class H2ServerConfigurationTests {

	private ApplicationContextRunner runner = new ApplicationContextRunner()
			.withConfiguration(AutoConfigurations.of(PropertyPlaceholderAutoConfiguration.class, H2ServerConfiguration.class));

	@Test
	void serverStartsWhenUrlIsH2AndEmbeddedPropertyTrue() {
		runner.withPropertyValues(
						"spring.datasource.url=jdbc:h2:tcp://localhost:19092/mem:dataflow",
						"spring.dataflow.embedded.database.enabled=true")
				.run(context -> assertThat(context)
						.hasNotFailed()
						.hasBean("h2TcpServer")
						.getBean("h2TcpServer", Server.class)
						.satisfies((server) -> {
							assertThat(server.isRunning(false)).isTrue();
							assertThat(server.getPort()).isEqualTo(19092);
						}));
	}

	@Test
	void serverStartsWhenUrlIsH2AndEmbeddedPropertyNotSet() {
		runner.withPropertyValues(
						"spring.datasource.url=jdbc:h2:tcp://localhost:19092/mem:dataflow")
				.run(context -> assertThat(context)
						.hasNotFailed()
						.hasBean("h2TcpServer")
						.getBean("h2TcpServer", Server.class)
						.satisfies((server) -> {
							assertThat(server.isRunning(false)).isTrue();
							assertThat(server.getPort()).isEqualTo(19092);
						}));
	}

	@Test
	void serverStopsWhenContextClosed() {
		runner.withPropertyValues(
						"spring.datasource.url=jdbc:h2:tcp://localhost:19092/mem:dataflow")
				.run(context -> {
					assertThat(context).hasNotFailed().hasBean("h2TcpServer");
					Server server = context.getBean("h2TcpServer", Server.class);
					assertThat(server.isRunning(false)).isTrue();
					context.close();
					assertThat(server.isRunning(false)).isFalse();
				});
	}

	@Test
	void serverDoesNotStartByDefault() {
		runner.run(context -> assertThat(context).hasNotFailed().doesNotHaveBean("h2TcpServer"));
	}

	@Test
	void serverDoesNotStartWhenUrlIsNotH2() {
		// TODO
	}

	@Test
	void serverDoesNotStartWhenEmbeddedPropertyFalse() {
		// TODO
	}

	@Test
	void serverDoesNotStartWhenUrlIsH2ButInvalidForm() {
		// TODO (may have more than one of these - 1 for port bad, 1 for wrong parts in url,, etc..)
	}

	@Test
	void serverDoesNotStartWhenH2NotOnClasspath() {
		// look at withFilteredClassLoader(new FilteredClassloader(Server.class))
	}
}
```

Of course testing H2 + the full app is needed but we can have a single (+) and (-) h2 case in the DF server tests.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfigurationTests.java
```diff
@@ -51,118 +47,102 @@
...
-		environment = new StandardEnvironment();
-		propertySources = environment.getPropertySources();
-	}
-
-	@After
-	public void teardown() {
-		if (context != null) {
-			context.close();
-		}
+		contextRunner = new ApplicationContextRunner()
```
onobc, line 74: ℹ️ you will notice in my example below that the ACR is defined as member var rather than setup. The reason is that it's API always returns a new instance (it is designed as such).

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfigurationTests.java
```diff
@@ -51,118 +47,102 @@
...
-		throw exceptionResult;
+	@Test
+	public void testDoNotStartEmbeddedH2Server() {
+		contextRunner.withPropertyValues(
+						"spring.datasource.url=jdbc:h2:tcp://localhost:19092/mem:dataflow",
+						"spring.dataflow.embedded.database.enabled=false",
+						"spring.jpa.database=H2"
+				)
+				.run(context -> {
+					assertNotNull(context.getStartupFailure());
```
onobc, line 116: There is a rich assert model you can use for startup failures :
```java
					assertThat(context).hasFailed()
							.getFailure()
							.isInstanceOf(BeanCreationException.class)
							.hasMessage("....");
```
There are many `hasMessage**` variants to use etc..

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfigurationTests.java
onobc: If later you find yourself wanting to make the above assertions more DRY, you can do things like this: 

```java
	@Test
	void serverStartsWhenUrlIsH2AndEmbeddedPropertyTrue() {
		runner.withPropertyValues(
						"spring.datasource.url=jdbc:h2:tcp://localhost:19092/mem:dataflow")
				.run(assertServerStarted(19092, null));
	}

	@Test
	void serverStartsPlusCheckiingOtherFancyThings{
		runner.withPropertyValues(
                                           "spring.datasource.url=jdbc:h2:tcp://localhost:19092/mem:dataflow")
				.run(assertServerStarted(19092, (server) -> {
                                                assertThat(server).blahBlahBlah().youGetTheIdea();
                                           ));
	}

	private ContextConsumer<AssertableApplicationContext> assertServerStarted(int port, Consumer<Server> serverConsumer) {
		return (context) -> {
			assertThat(context).hasNotFailed().hasBean("h2TcpServer");
			Server server = context.getBean("h2TcpServer", Server.class);
			assertThat(server.isRunning(false)).isTrue();
			assertThat(server.getPort()).isEqualTo(port);
			if (serverConsumer != null) {
				serverConsumer.accept(server);
			}
		};
	}
```

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/H2ServerConfiguration.java
onobc: Looks good. I will try to not suggest that we should take advantage of SpEL support for regex and `matches` operation. Because the next iteration of that would be to share the pattern which would make the SpEL very "not pretty" :)

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/H2ServerConfiguration.java
onobc: Any reason to not use `@Bean(destroyMethod = "stop")` ?

michael-wirth: No, I didn't think about this.

I changed it.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/H2ServerConfiguration.java
```diff
@@ -0,0 +1,94 @@
...
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * Autostart an embedded H2 database server.
+ *
+ * @author Michael Wirth
+ */
+@Configuration(proxyBeanMethods = false)
+@ConditionalOnClass(Server.class)
+@ConditionalOnProperty(value = "spring.dataflow.embedded.database.enabled", havingValue = "true", matchIfMissing = true)
```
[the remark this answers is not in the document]
michael-wirth: Done

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/H2ServerConfiguration.java
```diff
@@ -0,0 +1,94 @@
...
+	@Bean
+	public H2ServerBeanFactoryPostProcessor h2ServerBeanFactoryPostProcessor() {
+		return new H2ServerBeanFactoryPostProcessor();
+	}
+
+	@Bean
+	public Server h2TcpServer(@Value("${spring.datasource.url}") String dataSourceUrl) {
+		logger.info("Starting H2 Server with URL: " + dataSourceUrl);
+
+		String port;
```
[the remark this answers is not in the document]
michael-wirth: Done

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfigurationTests.java
michael-wirth: Cools testing feature - thanks for the advise

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfigurationTests.java
michael-wirth: I updated the orginal test

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfigurationTests.java
```diff
@@ -51,118 +42,46 @@
...
-			context.close();
-		}
-	}
-
-	/**
-	 * Verify that embedded server starts if h2 url is specified with default properties.
-	 */
-	@Test
-	@Ignore
-	public void testStartEmbeddedH2Server() {
```
onobc, line 108: It would be a good idea to add 1 test that starts up the H2 server and make sure all is well.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfigurationTests.java
onobc: 👍🏻 you are more than welcome. It is definitely one of my favorites.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfigurationTests.java
onobc: 👍🏻 you are more than welcome. It is definitely one of my favorites.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfigurationTests.java
michael-wirth: I re-attached the `testStartEmbeddedH2Server` and the `testDoNotStartEmbeddedH2Server` again

onobc (approved):
Nice work Michael - thanks again for the contribution. I will get this merged sometime today.

kumardh007:
can someone let me know in which release version this fix is available ?

michael-wirth:
Looks like this change is planed for 2.10 Release

kumardh007:
Thank you for the information Michael.

PR #4821, opened 2022-02-11
Add h2 dependency back in

onobc:
This adds the h2 dependency back into the server core module. 

Removing it broke the local mode that is outlined in the Getting Started guides. We attempted to upgrade to h2 `2.1.210` but it does not play well w/ the rest of the ecosystem yet. 

The tradeoff is that SCDF 2.x will continue to get CVE alerts when scanned. 

We may attempt to exclude it before we GA though.

### DIFF
--- a/spring-cloud-dataflow-server-core/pom.xml
+++ b/spring-cloud-dataflow-server-core/pom.xml
@@ -120,6 +120,12 @@
 			</exclusions>
 		</dependency>
 		<dependency>
+			<groupId>com.h2database</groupId>
+			<artifactId>h2</artifactId>
+			<optional>true</optional>
+			<scope>provided</scope>
+		</dependency>
+		<dependency>
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter-test</artifactId>
 			<scope>test</scope>
@@ -137,11 +143,6 @@
 			<groupId>org.skyscreamer</groupId>
 			<artifactId>jsonassert</artifactId>
 			<scope>test</scope>
-		</dependency>
-		<dependency>
-			<groupId>com.h2database</groupId>
-			<artifactId>h2</artifactId>
-			<optional>true</optional>
 		</dependency>
 		<dependency>
 			<groupId>com.oracle.ojdbc</groupId>
--- a/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/config/H2ServerConfigurationTests.java
+++ b/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/config/H2ServerConfigurationTests.java
@@ -16,8 +16,6 @@
 
 package org.springframework.cloud.dataflow.server.config;
 
-import java.net.BindException;
-
 import org.h2.tools.Server;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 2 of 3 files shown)


spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/config/H2ServerConfigurationTests.java
```diff
@@ -112,16 +110,6 @@ void serverDoesNotStartWhenUrlIsH2ButInvalidForm() {
 						.hasMessage("Port value out of range: 99999"));
 	}
 
-	@Test
```
onobc, line 115: **[UNRELATED]** This is not about adding/removing h2 but rather removing a test that does not work on all systems. This fails when run on my mac os. This test is really testing H2Server.tcpServer creation which should be heavily tested already by H2. Removing as we have no logic in place to check reserved ports.

spring-cloud-dataflow-server-core/pom.xml
```diff
@@ -141,7 +141,6 @@
 		<dependency>
 			<groupId>com.h2database</groupId>
 			<artifactId>h2</artifactId>
-			<optional>true</optional>
```
jvalkeal, line 144: Could we keep this optional in core and add it back in `spring-cloud-dataflow-server` pom?

onobc: I am sure we can. This would have a benefit of being propagated transitively to fewer other libs. This makes sense and is a better place to put it, but is there another benefit or angle that this provides (rebuild the server w/o the h2 locally ??).

onobc: Also, just to clarify, it was originally in core.

spring-cloud-dataflow-server-core/pom.xml
```diff
@@ -119,6 +119,12 @@
 				</exclusion>
 			</exclusions>
 		</dependency>
+		<dependency>
```
onobc, line 122: We want the following:
1. h2 to be optional
2. h2 to be available during test.

The first thought would be this: 
```xml
<dependency>
  <groupId>com.h2database</groupId>
    <artifactId>h2</artifactId>
    <optional>true</optional>
</dependency>

<dependency>
  <groupId>com.h2database</groupId>
  <artifactId>h2</artifactId>
  <scope>test</scope>
</dependency>
```
That does not work as maven does not like the 2x declaration of the dependency. 

This does not work either (the scope only of test immediately limits it to only test availability)
```xml
<dependency>
  <groupId>com.h2database</groupId>
  <artifactId>h2</artifactId>
  <optional>true</optional>
  <scope>test</scope>
</dependency>
```
Where we landed on optional=true and scope=provided DOES work because 
> [^1] The provided dependencies are available only at compile time and in the test classpath of the project. These dependencies are also not transitive.

[^1]: https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#dependency-scope

spring-cloud-dataflow-server-core/pom.xml
onobc: @jvalkeal - done

jvalkeal:
looking good now

PR #4829, opened 2022-02-15
fix: also possible to use only non-key-value argument(s)

LOG-INFO:
ex) `--testArgument`

### DIFF
--- a/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
+++ b/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
@@ -149,7 +149,7 @@ public final class DeploymentPropertiesUtils {
 	 */
 	public static List<String> parseArgumentList(String s, String delimiter) {
 		ArrayList<String> pairs = new ArrayList<>();
-		if (s != null && s.contains("=")) {
+		if (s != null) {
 			// get raw candidates as simple comma split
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/util/DeploymentPropertiesUtils.java
```diff
@@ -149,7 +149,7 @@ private DeploymentPropertiesUtils() {
 	 */
 	public static List<String> parseArgumentList(String s, String delimiter) {
 		ArrayList<String> pairs = new ArrayList<>();
-		if (s != null && s.contains("=")) {
+		if (s != null) {
```
cppwfs, line 152: Thank you for your contribution.   This solution causes a side effect to the current functionality. Here is a test with the expected results based on current behavior with your changes:
```
props = DeploymentPropertiesUtils.parseArgumentList("boo foo=bar xxx3", " ");
assertTrue(props.contains("boo"));
assertTrue(props.contains("foo=bar xxx3"));
```
But the result of the parse produces:
* boo
* foo=bar
* xxx3

Currently, the code looks at embedded blanks as part of the value.   With the solution provided this would use the spaces as a delimiter and split a value into multiple keys.

cppwfs:
After chatting with the team, introducing arguments that have a `key` but no `value` would be a breaking change.   This can be seen in the tests and in the discussion above.   The recommended strategy is to have a key/value flag.   So instead of having an argument with just a key ie. `--foo`, it could be written as `--foo=true`.

PR #4838, opened 2022-02-24
Use managed spring-shell dependency from spring-cloud-dataflow-build

onobc:
~⚠️ **DO NOT MERGE YET:** Depends on https://github.com/spring-cloud/spring-cloud-dataflow-build/pull/65/files~

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -73,8 +73,6 @@
 
 		<spring-cloud-common-security-config.version>1.8.0-SNAPSHOT</spring-cloud-common-security-config.version>
 
-		<spring-shell.version>2.1.0-SNAPSHOT</spring-shell.version>
-
 		<jacoco-maven-plugin.version>0.8.5</jacoco-maven-plugin.version>
 		<sonar-maven-plugin.version>3.0.2</sonar-maven-plugin.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -195,11 +193,6 @@
 				<artifactId>spring-cloud-deployer-cloudfoundry</artifactId>
 				<version>${spring-cloud-deployer-cloudfoundry.version}</version>
 			</dependency>
-			<dependency>
```
jvalkeal, line 198: We don't get dependency from parent, we just get dependency management. We still need to define it but without version.

onobc: Yep, this was doing `<dependencyManagement>` for spring-shell in SCDF. Now that we have lifted up higher to SCDB we can remove it here. The actual dependency is still specified in [spring-cloud-dataflow-shell-core/pom.xml](https://github.com/spring-cloud/spring-cloud-dataflow/blob/a66193047b72e76162bcc2422cf4bbd00aa0858d/spring-cloud-dataflow-shell-core/pom.xml#L22).

jvalkeal: Ah stupid me, I was looking wrong file.

PR #4843, opened 2022-03-01
CTR launched from Schedules now will support parent id.

### LINKED ISSUE #4832: CTR does not propagate the argument `spring.cloud.task.parent-execution-id`  to child task in K8s scheduling.
**Description:**
1. CTR does not propagate the argument `spring.cloud.task.parent-execution-id`  to child task in K8s scheduling. And result in that the system table's column `task_execution.parent_execution_id` is always null;
2. It dose propagate this argument when manual launching in scdf-ui in K8s mode;
3. I also tried it in Local mode, It's OK too.

**Release versions:**
2.9.2, k8s(with Schedule enabled)

**Steps to reproduce:**
1. Create a task: `step1: timestamp && step2: timestamp`;
2. Launch the task and view execution details in scdf-ui, you can find `spring.cloud.task.parent-execution-id` in child task's execution details;
3. Create a schedule with `0/1 * * * *` for the task;
4. Wait to the task triggered, and check the child task's execution details, you'll missing the argument: `spring.cloud.task.parent-execution-id`

cppwfs:
Resolves #4832

### DIFF
--- a/spring-cloud-dataflow-composed-task-runner/src/main/java/org/springframework/cloud/dataflow/composedtaskrunner/TaskLauncherTasklet.java
+++ b/spring-cloud-dataflow-composed-task-runner/src/main/java/org/springframework/cloud/dataflow/composedtaskrunner/TaskLauncherTasklet.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2021 the original author or authors.
+ * Copyright 2017-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -166,9 +166,11 @@ public class TaskLauncherTasklet implements Tasklet {
 				}
 				args = cleansedArgs;
 			}
-			if(this.taskProperties.getExecutionid() != null) {
-				args.add("--spring.cloud.task.parent-execution-id=" + this.taskProperties.getExecutionid());
+			String parentTaskExecutionId = getParentTaskExecutionId(contribution);
+			if(parentTaskExecutionId != null) {
+				args.add("--spring.cloud.task.parent-execution-id=" + parentTaskExecutionId);
 			}
+
 			if(StringUtils.hasText(this.composedTaskProperties.getPlatformName())) {
 				properties.put("spring.cloud.dataflow.task.platformName", this.composedTaskProperties.getPlatformName());
 			}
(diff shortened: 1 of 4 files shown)


spring-cloud-dataflow-composed-task-runner/src/main/java/org/springframework/cloud/dataflow/composedtaskrunner/TaskLauncherTasklet.java
```diff
@@ -214,6 +216,17 @@ else if (taskExecution.getExitCode() != 0) {
 		return RepeatStatus.CONTINUABLE;
 	}
 
+	public String getParentTaskExecutionId(StepContribution stepContribution) {
+		Long result = null;
+		if (this.taskProperties.getExecutionid() != null) {
+			result = this.taskProperties.getExecutionid();
+		}
+		else if (stepContribution != null) {
+			result = this.taskExplorer.getTaskExecutionIdByJobExecutionId(stepContribution.getStepExecution().getJobExecutionId());
```
onobc, line 225: Just to play devil's advocate... in the case where `getJobExecutionId` returns `null`, the underlying DAO will issue:
```sql
SELECT TASK_EXECUTION_ID FROM %PREFIX%TASK_BATCH WHERE JOB_EXECUTION_ID = null
```
which seems like it could return other unwanted results.

onobc: ~which seems like it could return other unwanted results.~

Actually I think we are ok here because it is using `queryForObject` which will throw `IncorrectResultSizeDataAccessException` if there are more than 1 result. Hence if there is a `null` one and this succeeds it must be the `null` belonging to this task.

spring-cloud-dataflow-composed-task-runner/src/test/java/org/springframework/cloud/dataflow/composedtaskrunner/ComposedTaskRunnerConfigurationNoPropertiesTests.java
```diff
@@ -83,6 +83,6 @@ public void testComposedConfiguration() throws Exception {
 		assertThat(composedTaskProperties.getTransactionIsolationLevel()).isEqualTo("ISOLATION_REPEATABLE_READ");
 
 		Assert.notNull(job.getJobParametersIncrementer(), "JobParametersIncrementer must not be null.");
-		verify(taskOperations).launch("AAA", new HashMap<>(0), new ArrayList<>(0));
+		verify(taskOperations).launch("AAA", new HashMap<>(0), Collections.singletonList("--spring.cloud.task.parent-execution-id=1"));
```
onobc, line 86: Do you mind adding a test (any location is fine w/ me) that verifies the case when it comes up null?

onobc (commented):
@cppwfs LGTM. One minor ask around adding another test case.

onobc:
Closed via https://github.com/spring-cloud/spring-cloud-dataflow/commit/7a9bdaae52b1e4ba13f9323cede29dba92a0282c

PR #4855, opened 2022-03-15
Springdoc integration

onobc:
This is a continuation of @klopfdreh original PR [here](https://github.com/spring-cloud/spring-cloud-dataflow/pull/4836). 

@klopfdreh I squased/rebased your PR on 2.9.x and then cherry-picked to a userbranch off of main. I will close your original PR that is in 2.9.x and we can continue to work from here. 

I think we are good on testing now. 

**Here is what is remaining:**
- [x] code review my recent commit
- [x] add lightweight docs on "How to Use Springdoc" (including enabling/disabling/config options (point to springdoc docs) 
- [x] why is default api now showing up? 

@klopfdreh if you can do the above that would be great. Otherwise I will get around to it next weekend. 

**Steps to see enable**
1. Start SCDF server with Springdoc enabled properties
  ```shell
  -Dspringdoc.api-docs.enabled=true -Dspringdoc.swagger-ui.enabled=true
  ```
  OR
  ```yaml
  springdoc:
    api-docs:
      enabled: true
    swagger-ui:
      enabled: true
  ```
2. Navigate to http://localhost:9393/v3/api-docs (you will see the openapi definition)
3. Navigate to  http://localhost:9393/swagger-ui/index.html
  > ⚠️ the Swagger UI will be empty as the "Explore" bar is empty. 
4. Type in "/v3/api-docs/" and click "Explore"
5. Start exploring the APIs

NOTE: I set `springdoc.swagger-ui.disable-swagger-default-url: true` as that was loading the Swagger Petstore in the Swagger UI. I then tried to set `springdoc.swagger-ui.url: /v3/api-docs/` but it does not see

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/SpringDocAutoConfiguration.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/SpringDocAutoConfiguration.java
@@ -16,11 +16,19 @@
 
 package org.springframework.cloud.dataflow.server.config;
 
-import org.springdoc.core.SpringDocConfigProperties;
-import org.springdoc.core.SwaggerUiConfigProperties;
+import javax.annotation.PostConstruct;
 
-import org.springframework.beans.factory.annotation.Qualifier;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springdoc.core.SpringDocConfigProperties;
+import org.springdoc.core.SpringDocConfiguration;
+import org.springdoc.core.SwaggerUiConfigProperties;
+import org.springdoc.webmvc.ui.SwaggerConfig;
+
+import org.springframework.boot.autoconfigure.AutoConfigureAfter;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.web.servlet.FilterRegistrationBean;
 import org.springframework.cloud.dataflow.server.support.SpringDocJsonDecodeFilter;
 import org.springframework.context.annotation.Bean;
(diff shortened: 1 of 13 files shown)


spring-cloud-dataflow-server-core/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
onobc, line 1: This allows us to mock the final class `WebSecurity` to validate the customizer sets the proper ignore paths.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/config/SpringDocAutoConfigurationTests.java
```diff
@@ -0,0 +1,195 @@
...
+	void defaultsAreInSyncWithSpringdoc() {
+		contextRunnerWithSpringDocEnabled()
+				.run((context) -> {
+					SpringDocConfigProperties springDocConfigProps = context.getBean(SpringDocConfigProperties.class);
+					SwaggerUiConfigProperties swaggerUiConfigProps = context.getBean(SwaggerUiConfigProperties.class);
+					assertThat(springDocConfigProps.getApiDocs().getPath()).isEqualTo(Constants.DEFAULT_API_DOCS_URL);
+					assertThat(swaggerUiConfigProps.getPath()).isEqualTo(Constants.DEFAULT_SWAGGER_UI_PATH);
+					assertThat(swaggerUiConfigProps.getConfigUrl()).isEqualTo(
+							Constants.DEFAULT_API_DOCS_URL + AntPathMatcher.DEFAULT_PATH_SEPARATOR + Constants.SWAGGGER_CONFIG_FILE);
+					assertThat(swaggerUiConfigProps.getValidatorUrl()).isEqualTo("https://validator.swagger.io/validator");
```
klopfdreh, line 143: Just a question: Is there no Constant for this, too?

klopfdreh (commented):
Almost no nitpicks. Just two questions. Glad to see that all is covered with tests, now 👍

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/config/SpringDocAutoConfigurationTests.java
onobc: Good eye! Yes there is https://github.com/springdoc/springdoc-openapi/blob/378edcf142d9b2c5af637339b926f98687e55650/springdoc-openapi-common/src/main/java/org/springdoc/core/Constants.java#L61 and I attempted to use that. However, you will see it references https://github.com/springdoc/springdoc-openapi/blob/378edcf142d9b2c5af637339b926f98687e55650/springdoc-openapi-common/src/main/java/org/springdoc/core/Constants.java#L51 which contains SpEL which will fail a basic equality check here. The other alternative would be to evaluate the constant but it seemed more hassle than it was worth so I went w/ a pragmatic hybrid approach.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/SpringDocAutoConfiguration.java
```diff
@@ -0,0 +1,133 @@
...
+
+/**
+ * Makes SpringDoc public available without any authentication required by initializing a {@link WebSecurityCustomizer} and
+ * applying all path of SpringDoc to be ignored. Also applies a filter registration bean to unescape JSON content for the
+ * SpringDoc frontend.
+ *
+ * @author Tobias Soloschenko
+ */
+@Configuration(proxyBeanMethods = false)
+@ConditionalOnClass({ SpringDocConfigProperties.class, SwaggerUiConfigProperties.class })
```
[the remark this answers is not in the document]
onobc: Agreed, in its current manifestation we can live w/o this. This just future proofs against this scenario:

1. User installs SCDF and all is happy - they are Springdoc'ing
2. A major CVE comes up in Springdoc.
3. User does not want to wait for SCDF update release so decides to exclude the springdoc lib entirely 
4. SCDF starts up and does not explode when auto-config evaluation (eventhough enabled set to false). 

It is also questionable if 3 can even be done (or done easily). I decided to just act as-if and add this protection. This has bitten me many times on many apps over the years so I always approach the auto-config scenarios w/ the devil's advocate hat.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/config/SpringDocAutoConfigurationTests.java
klopfdreh: Yes right! 👍

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/SpringDocAutoConfiguration.java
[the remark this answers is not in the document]
klopfdreh: Thanks for the explanation! 👍

klopfdreh:
Hey @onobc - for me it will take some time when I am able to continue, so I would be very pleased if someone can take over the open points.

onobc:
> Hey @onobc - for me it will take some time when I am able to continue, so I would be very pleased if someone can take over the open points.

Sounds good @klopfdreh . I can handle the remaining todos. However, if you could test drive it via the Swagger UI a bit that would be helpful. I want to be sure its working as expected etc.. Were you able to get the system up and running via the command line?

klopfdreh:
With the command line I was able to run the server. `/v3/api-docs` is available, but somehow the swagger-ui does not scan the RestController endpoints anymore - I guess it has something to do with the ordering by some other configurations that might be started a bit later. But this is only an assumption.

klopfdreh:
We also should add this into our defaults, because we are directly using this:
```yaml
springdoc:
  webjars:
    prefix: /webjars
```

onobc:
Thanks again for the contribution @klopfdreh . I added the docs [here](https://github.com/spring-cloud/spring-cloud-dataflow/commit/c9a1d246c912ec8afdcb2dfa049aa4f81c22e490). When you get a chance give it a test drive.

klopfdreh:
Looks good 👍:
Bildschirmfoto 2022-03-22 um 08 55 09

PR #4868, opened 2022-04-02
Polish "Set spring.main.allow-circular-references=false" - using PR for CI PR run

onobc:


onobc:
Was using this simply to verify CI after polish.

PR #4870, opened 2022-04-05
Remove deprecated dependency spring-security-oauth2

corneil:
Remove deprecated dependency: `org.springframework.security.oauth:spring-security-oauth2`

Tests that use the dependency are also deleted


spring-cloud-dataflow-docs/src/main/asciidoc/configuration-local.adoc
```diff
@@ -581,35 +581,12 @@ uaac user get <username>
...
-allows you to easily create your own (very basic) OAuth2 Server with the following simple annotations:
+https://spring.io/projects/spring-security/[Spring Security]. It
+allows you to easily create your own OAuth2 Server by configuring the SecurityFilterChain.
 
-* `@EnableResourceServer`
-* `@EnableAuthorizationServer`
+Samples can be found at:
+https://docs.spring.io/spring-security/reference/samples.html[Spring Security Samples]
 
-NOTE: In fact the UAA uses Spring Security OAuth2 under the covers, thus the basic endpoints
```
onobc, line 590: We should check the other docs to make sure this is not in them as well. 
- [ ] skipper ref docs
- [ ] scdf ref docs (that is actually this doc here)
- [ ] dataflow.spring.io (I doubt it's in there but we should check)
- [ ] scdf samples (this doc is no longer published outward so no need to check there)

onobc (commented):
@corneil looks good, same concern as [here](https://github.com/spring-cloud/spring-cloud-skipper/pull/1041#pullrequestreview-931957318) though.

spring-cloud-dataflow-docs/src/main/asciidoc/configuration-local.adoc
corneil: I searched the docs and only one referenced the deprecated library. I changed the docs to refer to latest Spring Security and Samples.

corneil:
> @corneil looks good, same concern as [here](https://github.com/spring-cloud/spring-cloud-skipper/pull/1041#pullrequestreview-931957318) though.

Tests referring to the deprecated Spring Security OAuth2 library are invalid. The integration tests takes care of all auth needs.

corneil:
Resolves #4875 and [https://github.com/spring-cloud/spring-cloud-common-security-config/issues/93](https://github.com/spring-cloud/spring-cloud-common-security-config/issues/93)

Issue #4807, opened 2022-02-04
Ensure reference doc is up-to-date and valid


onobc:
1. [x] Remove all reference to stream/task-app-starters and replace w/ stream-applications
2. [x] Replace all links pointing to `master` w/ `main`
3. [x] Make sure all links to libs and jars are valid

onobc:
**5.1. Application and Server Properties**

> Properties to be applied for all deployed Tasks are defined in the src/kubernetes/server/server-config-[binder].yaml file and for Streams in src/kubernetes/skipper/skipper-config-[binder].yaml. Replace [binder] with the messaging middleware you are using — for example, rabbit or kafka.

**PROBLEM**: invalid path/name of files.

onobc:
**5. Getting Started - Kubernetes**

**PROBLEM**: Why the detail here but not w/ the other Local/CF? Those just point to the Microsite.

onobc:
**7.1. Wagon**

**QUESTION:** Any updates needed here wrt to the recent CVE ?

onobc:
**8. Security**

**QUESTION:** Is the following still valid?

> While you can theoretically choose any OAuth provider in conjunction with Spring Cloud Data Flow, we recommend using the CloudFoundry User Account and Authentication (UAA) Server.
> 
> Not only is the UAA OpenID certified and is used by Cloud Foundry, but you can also use it in local stand-alone deployment scenarios. Furthermore, the UAA not only provides its own user store, but it also provides comprehensive LDAP integration.

onobc:
8.1. Enabling HTTPS
8.2. Authentication by using OAuth 2.0
8.3. Customizing Authorization
8.4. Setting up UAA Authentication

**TODO** Validate the steps work as expected

onobc:
**9.2 Database**

**TODO:** Scan through the entire section, still valid?

onobc:
**17.1.1. Register Supported Applications and Tasks**

**Spring Cloud Stream App Starters**

> The following table includes the dataflow.spring.io links to the available Stream Application Starters based on Spring Cloud Stream 2.1.x and Spring Boot 2.1.x:


**OLD**

[...]

PR #4881, opened 2022-04-10
Remove old repositories links

ipanimnim:
Archived repositories that are confusing


onobc:
Thanks for the contribution @ipanimnim . We are in the process of updating all fo the the docs by the 2.10 release timeframe. Some of that work is capture [here](https://github.com/spring-cloud/spring-cloud-dataflow/issues/4807). Feel free to add to that issue or create another if you find other docs that are out of date.

PR #4882, opened 2022-04-13
Add segment creation immediate to all create-table for oracle

dturanski:
This avoids Oracle "Cannot serialize transaction" errors on startup. See https://oracle-base.com/articles/11g/segment-creation-on-demand-11gr2.  Not sure what this does on 10g.  Note that spring-batch 5.x does the same thing.


onobc:
@jvalkeal to verify if this works on Oracle or not.

onobc:
Closing as per https://github.com/spring-cloud/spring-cloud-dataflow/issues/5355

PR #4883, opened 2022-04-14
Make dataflow server uri a test property

dturanski:
tasklauncher test fails on CF now that I fixed the tasklauncher-sink app registration issue


dturanski:
Should get from dataflowClientProperties and also use authentication...

onobc:
@dturanski the changes make sense. I see its still in Draft and you have this comment:

> Should get from dataflowClientProperties and also use authentication...

Is that something you are going to do in this PR?

Issue #4436, opened 2021-03-15
Refactor junit5 and assertj


jvalkeal:
This issue tracks a progress of moving fully on junit5 and assertj.

- [ ] spring-cloud-dataflow-audit/src/test/java/org/springframework/cloud/dataflow/server/audit/service
- [ ] spring-cloud-dataflow-autoconfigure/src/test/java/org/springframework/cloud/dataflow/autoconfigure/local
- [ ] spring-cloud-dataflow-classic-docs/src/test/java/org/springframework/cloud/dataflow/server/rest/documentation
- [ ] spring-cloud-dataflow-completion/src/test/java/org/springframework/cloud/dataflow/completion
- [ ] spring-cloud-dataflow-composed-task-runner/src/test/java/org/springframework/cloud/dataflow/composedtaskrunner
- [ ] spring-cloud-dataflow-composed-task-runner/src/test/java/org/springframework/cloud/dataflow/composedtaskrunner/properties
- [ ] spring-cloud-dataflow-composed-task-runner/src/test/java/org/springframework/cloud/dataflow/composedtaskrunner/support
- [ ] spring-cloud-dataflow-composed-task-runner/src/test/java/org/springframework/cloud/dataflow/composedtaskrunner/configuration
- [ ] spring-cloud-dataflow-configuration-metadata/src/test/java/org/springframework/cloud/dataflow/container/registry/authorization
- [ ] spring-cloud-dataflow-configuration-metadata/src/test/java/org/springframework/cloud/dataflow/container/registry/authorization/support
- [ ] spring-cloud-dataflow-configuration-metadata/src/test/java/org/springframework/cloud/dataflow/configuration/metadata
- [ ] spring-cloud-dataflow-container-registry/src/test/java/org/springframework/cloud/dataflow/container/registry/authorization
- [ ] spring-cloud-dataflow-container-registry/src/test/java/org/springframework/cloud/dataflow/container/registry/authorization/support
- [ ] spring-cloud-dataflow-core-dsl/src/test/java/org/springframework/cloud/dataflow/core/dsl
- [ ] spring-cloud-dataflow-core/src/test/java/org/springframework/cloud/dataflow/core
- [ ] spring-cloud-dataflow-platform-cloudfoundry/src/test/java/org/springframework/cloud/dataflow/server/config/cloudfoundry
[...]

PR #4638, opened 2021-07-24
refactor: JUnit 5 migration

jkschneider:
Hi @jvalkeal. This is a continuation of our effort to lift you up to JUnit 5. I'd like to make a few notes about this one though. 

* There are still two JUnit 4 `ExternalResource` implementations left in the code that will need to be migrated to JUnit 5. That should be all you have to do. The tests won't pass completely until these are migrated.
* We can generate this patch off of the latest main commit at any time with very little effort. So if you don't want to do this in the next couple days, just ping me when you're ready and we can regenerate a fresh commit off of the latest head, leaving you with just the `ExternalResource` migrations. Mainly I want you to see that the bulk of the heavy lifting is done, and you just have a couple things left to do.
* I've intentionally left the assertions as JUnit assertions to limit the scope of change somewhat. As soon as we get this merged, I can follow it up immediately with a PR to migrate to AssertJ. Hopefully this plan makes sense.

See https://github.com/spring-cloud/spring-cloud-dataflow/issues/4436.

cc / @cppwfs As promised last week ;)

Co-authored-by: Moderne <team@moderne.io>

PR #4884, opened 2022-04-15
refactor: Migrate JUnit asserts to AssertJ

Devamparikh:
Relates to #4436
As @onobc described.
Let me know if I'm doing it correct or not.


Devamparikh:
Hey, Why is there no action by maintainers? Can please somebody take a look at it?

onobc:
Hi @Devamparikh ,

Sorry for the confusion, but what I meant in the original description was that the existing PR #4436 needs to be re-evaluated and possibly added to. I do not plan on merging that PR and then a follow on one, but rather a single PR that converts all the things to JUnit5. 

As for the 2 files you changed, they still reference JUnit4 imports. 

What I would do is pull the branch down from the original PR and use it as a starting point for the migration work. If you decide you do not want to deal w/ this large widespread PR that is fine too. There are plenty of other tickets you could work on as well if. you are interested.

Thanks

markpollack:
Will use #4638 has the basis for changing to junit5.  Closing for now.

PR #4929, opened 2022-05-26
Fix app uri for kubernetes

corneil:
Fixing the app uri when platform is kubernetes


jvalkeal:
Think it looks good

spring-cloud-dataflow-server/src/test/java/org/springframework/cloud/dataflow/integration/test/DataFlowIT.java
```diff
@@ -2054,8 +2058,13 @@ private void setDefaultVersionForTimestamp(String version) {
     }
 
     private void registerTasks() {
-        registerTask("testtimestamp", "maven://io.spring:timestamp-task", CURRENT_VERSION_NUMBER);
-        registerTask("testtimestamp-batch", "maven://io.spring:timestamp-batch-task", CURRENT_VERSION_NUMBER);
+        if (System.getProperty("PLATFORM_TYPE", "local").equals("kubernetes")) {
```
jvalkeal, line 2061: Ah Hold on, `local` needs to be `kubernetes`.

jvalkeal: It needs to be `RuntimeApplicationHelper.KUBERNETES_PLATFORM_TYPE`. There was some handling for local as well.

PR #4955, opened 2022-06-22
Sanitize audit data

corneil:


spring-cloud-dataflow-audit/src/main/java/org/springframework/cloud/dataflow/audit/service/DefaultAuditRecordService.java
```diff
@@ -38,74 +41,101 @@
...
-			Instant toDate) {
-		return this.auditRecordRepository.findByActionTypeAndOperationTypeAndDate(operations, actions, fromDate, toDate,
-				pageable);
-	}
-
-	@Override
-	public Optional<AuditRecord> findById(Long id) {
-		return this.auditRecordRepository.findById(id);
-	}
+    private static final Logger logger = LoggerFactory.getLogger(DefaultAuditRecordService.class);
```
onobc, line 48: It would be helpful next time to put the formatting in a separate commit as its hard to tell what has changed.

spring-cloud-dataflow-audit/src/main/java/org/springframework/cloud/dataflow/audit/service/DefaultAuditRecordService.java
```diff
@@ -38,74 +41,101 @@
...
+
+    private final AuditRecordRepository auditRecordRepository;
+
+    private final ObjectMapper objectMapper;
+    private final ArgumentSanitizer sanitizer;
```
onobc, line 53: **[NIT]** spacing inconsistent w/ other member vars. I am ok either way w/ or w/o the space in between the vars as long as they are consistent.

spring-cloud-dataflow-audit/src/main/java/org/springframework/cloud/dataflow/audit/service/DefaultAuditRecordService.java
```diff
@@ -38,74 +41,101 @@
...
+        return this.populateAndSaveAuditRecord(auditOperationType, auditActionType, correlationId, dataAsString, platformName);
+    }
+
+    private Map<String, Object> sanitizeMap(Map<String, Object> data) {
+        final Map<String, Object> result = new HashMap<>();
+        data.entrySet()
+                .stream()
+                .map(this::sanitize)
+                .forEach(entry -> result.put(entry.getKey(),entry.getValue()));
+        return result;
```
onobc, line 110: Now that we have to use `forEach`, I would instead use the one directly on the map and then make `sanitize` just take a k/v. It was only using `Map.Entry` to take advantage of the method reference in the `map` on the stream really. This really simplifies it down to this now:
```java
Map<String, Object> result = new HashMap<>();
data.forEach((k, v) -> result.put(k, sanitize(k, v)));
return result;
```

spring-cloud-dataflow-audit/src/test/java/org/springframework/cloud/dataflow/server/audit/service/DefaultAuditRecordServiceTests.java
```diff
@@ -33,6 +35,8 @@
 import org.springframework.data.domain.PageRequest;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
```
onobc, line 38: Wdyt about yanking this old Junit4 out in favor of Junit5 and AssertJ assertions?

onobc: I am ok doing that in a later PR as well.

onobc (changes requested):
@corneil looks good. I have a few suggestions. Feel free to merge once they are addressed (or not). 

Also, if you do merge, please do a squash/rebase merge (I still see the extra commits in the PR). 

Thanks

spring-cloud-dataflow-audit/src/test/java/org/springframework/cloud/dataflow/server/audit/service/DefaultAuditRecordServiceTests.java
corneil: Will revisit the tests and upgrade to JUnit 5 at some point.

spring-cloud-dataflow-audit/src/main/java/org/springframework/cloud/dataflow/audit/service/DefaultAuditRecordService.java
corneil: Done

spring-cloud-dataflow-audit/src/test/java/org/springframework/cloud/dataflow/server/audit/service/DefaultAuditRecordServiceTests.java
onobc: In the meantime let's not pull in `org.junit.*` classes and instead use AssertJ or Jupiter APIs.

PR #4960, opened 2022-06-27
Respect 'manifest' task def controller parameter

onobc:
Do not execute SQL to load manifest when parameter is 'false'.

Fixes https://github.com/pivotal/scdf-pro/issues/123


cppwfs:
LGTM

PR #4982, opened 2022-07-04
Update database versions.

corneil:
Updated msqlserver JDBC url to include encrypt=false.
Updated MariaDB to 10.5.
Updated SQL Server to 2019-CU16.


corneil:
I've tried to find an image tag for MS SQL Server that works.
* `2019-CU16-ubuntu-20.04` reports `FlywayException: Unsupported Database: Microsoft SQL Server 15.0`
* `2022-CTP2.0-ubuntu-20.04` reports `FlywayException: Unsupported Database: Microsoft SQL Server 16.0`

PR #4986, opened 2022-07-06
remove executable boot build

corneil:
Remove executable = true from boot plugin as it breaks the packeto packaging.


corneil:
wrong target

Issue #4997, opened 2022-07-14
Composed Task Runner does not use clientId and clientSecret in client credentials request


klopfdreh:
**Description:**

Currently we are trying to get the composed task runner to work with client credentials flow to authenticated against our OAuth2 server and use this bearer token to start two described task in the compose task DSL at our SCDF server. (e.g. `task-a && task-b`) We are setting the required properties as environment variables like described here: https://dataflow.spring.io/docs/feature-guides/batch/composed-task/#client-credentials

Also we applied the environment variable "DEBUG" and set it to "ALL" in the composed task runner to get some information about the OAuth2 request

The composed task runner is also starting like expected, but the clientId and clientSecret are not in the post body of the request against the token endpoint. The TokenUri and the Scopes are used correctly as they are placed into the post request.

Because of the relaxed binding the spring application of the composed task runner should take the following properties as environment variables under consideration.

***properties as environment variables***
oauth2ClientCredentialsTokenUri > OAUTH2_CLIENT_CREDENTIALS_TOKEN_URI
oauth2ClientCredentialsScopes > OAUTH2_CLIENT_CREDENTIALS_SCOPES_0 up to OAUTH2_CLIENT_CREDENTIALS_SCOPES_5
oauth2ClientCredentialsClientId > OAUTH2_CLIENT_CREDENTIALS_CLIENT_ID
oauth2ClientCredentialsClientSecret > OAUTH2_CLIENT_CREDENTIALS_CLIENT_SECRET

***log of the composed task runner***
{"timestamp":"2022-07-13T15:36:18.321+0200","level":"DEBUG","thread":"main","logger":"org.springframework.web.client.RestTemplate","message":"HTTP POST https://tokenuri/token/","context":"default"}
{"timestamp":"2022-07-13T15:36:18.322+0200","level":"DEBUG","thread":"main","logger":"org.springframework.web.client.RestTemplate","message":"Accept=[application/json, application/*+json]","context":"default"}
{"timestamp":"2022-07-13T15:36:18.323+0200","level":"DEBUG","thread":"main","logger":"org.springframework.web.client.RestTemplate","message":"Writing [{grant_type=[client_credentials], scope=[scope1 scope2]}] as \"application/x-www-form-urlencoded;charset=UTF-8\"","context":"default"}
[...]

PR #4998, opened 2022-07-14
feat: CTR client authentication method property

klopfdreh:
Resolves: #4997

### DIFF
--- a/spring-cloud-dataflow-composed-task-runner/src/test/java/org/springframework/cloud/dataflow/composedtaskrunner/properties/ComposedTaskPropertiesTests.java
+++ b/spring-cloud-dataflow-composed-task-runner/src/test/java/org/springframework/cloud/dataflow/composedtaskrunner/properties/ComposedTaskPropertiesTests.java
@@ -28,6 +28,7 @@ import org.springframework.boot.test.context.runner.ApplicationContextRunner;
 import org.springframework.cloud.dataflow.core.Base64Utils;
 import org.springframework.core.env.StandardEnvironment;
 import org.springframework.core.env.SystemEnvironmentPropertySource;
+import org.springframework.security.oauth2.core.ClientAuthenticationMethod;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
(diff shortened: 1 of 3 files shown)


spring-cloud-dataflow-composed-task-runner/src/test/java/org/springframework/cloud/dataflow/composedtaskrunner/properties/ComposedTaskPropertiesTests.java
```diff
@@ -121,6 +122,23 @@ public void testComposedTaskAppArguments() {
 				});
 	}
 
+	@Test
+	public void testAssignmentOfOauth2ClientCredentialsClientAuthenticationMethod(){
+		this.contextRunner
+				.withInitializer(context -> {
+					Map<String, Object> map = new HashMap<>();
+					map.put("oauth2ClientCredentialsClientAuthenticationMethod", "POST");
+					context.getEnvironment().getPropertySources().addLast(new SystemEnvironmentPropertySource(
```
onobc, line 131: I am pretty sure you copied the intiializer/properties part from the test above but a couple of points...

* I think we can instead simply use `ApplicationContextRunner.withSystemProperties`

However, if we did continue using the initializer approach:

* I don't think we would need to use `SystemEnvironmentPropertySource` and instead could use a simple `MapPropertySource`
* I am not sure why the `addLast` method is used rather than `addFirst` as we want these props to take precedence (unless the `SystemEnvironmentPropertySource` has a higher built-in order/priority.

klopfdreh: Hey @onobc - thanks for the review! I used withSystemProperties, because I just want to check if the property is converted into the enum correctly. 👍

onobc (approved):
Thanks again @klopfdreh . Another great contribution! Much appreciated.

klopfdreh:
Hey @onobc - I also changed the documentation in a separate PR. Are you able to review and merge this, too?

https://github.com/spring-io/dataflow.spring.io/pull/452

onobc:
@klopfdreh thanks for the contribution. 

I just merged. It should be available on dataflow.spring.io shortly.

onobc:
@klopfdreh it is now available on the site (https://scdfmicrosite.azurewebsites.net/docs/2.10.0.SNAPSHOT/feature-guides/batch/composed-task/#client-credentials)

klopfdreh:
@onobc thanks a lot! 👍

klopfdreh:
Hey @onobc - I guess I forgot one place to change with this PR.

https://github.com/spring-cloud/spring-cloud-dataflow/blob/450027466c43fdd1c20dd8503a97fced205bf336/spring-cloud-dataflow-composed-task-runner/src/main/java/org/springframework/cloud/dataflow/composedtaskrunner/StepBeanDefinitionRegistrar.java#L226

I guess we have to add something like
```java
String oauth2ClientCredentialsClientAuthenticationMethodAsString = getPropertyValue("oauth2-client-credential-client-authentication-method")
properties.setOauth2ClientCredentialsClientAuthenticationMethod(oauth2ClientCredentialsClientAuthenticationMethodAsString != null ? ClientAuthenticationMethod.valueOf(oauth2ClientCredentialsClientAuthenticationMethodAsString) : null);
```

I am going to create a PR for this, soon.

klopfdreh:
See https://github.com/spring-cloud/spring-cloud-dataflow/pull/5058

PR #5025, opened 2022-07-21
Chunking large deletes of StepExecutions and JobExecutions

blueaadvark:
#4931  Chunking for large deletes of StepExecutions and JobExecutions

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskDeleteService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskDeleteService.java
@@ -320,6 +320,18 @@ public class DefaultTaskDeleteService implements TaskDeleteService {
 
 		int chunkSize = getTaskExecutionDeleteChunkSize(this.dataSource);
 
+		final Set<Long> stepExecutionIds = findStepExecutionIds(jobExecutionIds, chunkSize);
+
+		final AtomicInteger numberOfDeletedBatchStepExecutionContextRows = new AtomicInteger(0);
+		if (!stepExecutionIds.isEmpty()) {
+			deleteBatchStepExecutionContextByStepExecutionIds(stepExecutionIds, chunkSize, numberOfDeletedBatchStepExecutionContextRows);
+		}
+
+		deleteStepAndJobExecutionsByJobExecutionId(jobExecutionIds, chunkSize, auditData, numberOfDeletedBatchStepExecutionContextRows);
+	
+	}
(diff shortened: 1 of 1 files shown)


corneil:
We do have tests that cover this method.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskDeleteService.java
```diff
@@ -262,44 +264,7 @@ public void deleteTaskExecutions(Set<Long> taskExecutionIds) {
 		auditData.put("Deleted Job Execution IDs", StringUtils.collectionToDelimitedString(jobExecutionIds, ", "));
 
 		if (!jobExecutionIds.isEmpty()) {
-			final Set<Long> stepExecutionIds = dataflowJobExecutionDao.findStepExecutionIds(jobExecutionIds);
```
onobc, line 265: Ahh, ok. This is just applying the same chunking treatment to the step executions+contexts that was previously added to this method in https://github.com/spring-cloud/spring-cloud-dataflow/pull/4723.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskDeleteService.java
```diff
@@ -262,44 +264,7 @@ public void deleteTaskExecutions(Set<Long> taskExecutionIds) {
 		auditData.put("Deleted Job Execution IDs", StringUtils.collectionToDelimitedString(jobExecutionIds, ", "));
 
 		if (!jobExecutionIds.isEmpty()) {
```
onobc, line 266: @blueaadvark wdyt about pulling the call to `getTaskExecutionDeleteChunkSize(this.dataSource)` up here from https://github.com/spring-cloud/spring-cloud-dataflow/pull/5025/files#diff-ec79e846595460e3b9bcc697304fcee7dc73a8fa171c9ba48dacc21eda82986cR280 and then we can pass into this newly added method to avoid calling it 2x?

onobc: I went ahead and took care of this doing merge.

onobc:
Thanks for the contribution @blueaadvark . Sorry it took a while to get back around to final review/merge. 

Closed via https://github.com/spring-cloud/spring-cloud-dataflow/commit/52081a6a38335fd82b616c56842671bdaa494e91

PR #5044, opened 2022-08-05
Instructions for updating container with extra dependencies.

corneil:


corneil:
@onobc, @markpollack  Can you take a look at this.

onobc:
This seems like it should work @corneil . Did you test it on the Oracle driver? 

The previous advice (for stream-applications anyways) is to pull the repo locally and [rebuild w/ adjusted deps](https://docs.spring.io/stream-applications/docs/current/reference/html/#_patching_pre_built_applications). This is less intrusive, which is nice.

corneil:
This obviously only addresses the k8s problem.
For CF we need to drop the executable from Spring Boot plugin so that dependencies can be added to application jars before publishing them in a private Maven Repo.

I believe that private Maven repo and private Container registry should be standard recommendations.

klopfdreh:
Just a question for testing - is there any way to test .bat files with GitHub actions?

corneil:
> Just a question for testing - is there any way to test .bat files with GitHub actions?

We will need to create a GitHub actions workflow targeting a Windows runner.

Issue #4887, opened 2022-04-19
Add support for MySQL 8 in OSS


onobc:
### Technical steps
- [x] Upgrade SCDF to Boot 2.7.x (allows differentiating MySQL / MariaDB schemas)
- [x] Split the MariaDB / MySQL schemas
- [x] Run SCDF manually w/ the following combinations:
   - [x] MariaDB driver + MariaDB 
   - [x] MariaDB driver + MySQL 
   - [x] MariaDB driver + MariaDB 
   - [x] MariaDB driver + MySQL 
- ~[ ]~ ~Make sure we can ship MySQL driver (@onobc discussing w/ OSM on next steps)~

onobc:
MySQL license info: https://www.mysql.com/about/legal/licensing/oem/

onobc:
In regards to...

> - [ ] Make sure we can ship MySQL driver (@onobc discussing w/ OSM on next steps)

I reached out to our OSM contact and they felt upon cursory review that we should reach out to our BU counsel (Krystia P. and Padma Choudry) who would then be able to further guide us on next steps. I will reach out them next week.

onobc:
I believe we can use the same approach I used for [MySQL 5.7](https://github.com/spring-cloud/spring-cloud-dataflow/pull/4894) on this and avoid upgrading SB.

onobc:
Closed via https://github.com/spring-cloud/spring-cloud-dataflow/commit/8ab2b1e97f7cf6cba7cfb3313caf037bb127f908

PR #5071, opened 2022-08-31
Add support for MySQL 5/8 via MariaDB driver

onobc:
* Use FlywayVendorReplacingEnvironmentPostProcessor to replace {vendor} in spring.flyways.locations
* Add back in the MySQL schema
* Use vendor (mysql) not driver in flyway configuration customizer

See #4887

> ⚠️ ⚠️ **depends on** https://github.com/spring-cloud/spring-cloud-dataflow-common/pull/22


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/db/migration/mysql/V1__Initial_Setup.java
```diff
@@ -0,0 +1,321 @@
+/*
+ * Copyright 2019-2022 the original author or authors.
```
cppwfs, line 2: Nitpick but it would be 2022-2022.    
I had to pick on something ;-)

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/db/migration/mysql/V4__Add_Step_Name_Indexes.java
```diff
@@ -0,0 +1,45 @@
...
+import org.springframework.cloud.dataflow.common.flyway.SqlCommand;
+import org.springframework.cloud.dataflow.common.flyway.SqlCommandsRunner;
+
+/**
+ * This migration class adds index for STEP_NAME on BATCH_STEP_EXECUTION.
+ *
+ * @author Glenn Renfro
+ * @author Chris Bono
+ *
+ * @since 2.7
```
cppwfs, line 32: @since 2.9.   Nitpick ;-)

spring-cloud-dataflow-server-core/src/main/resources/schemas/mysql/V1-skipper.sql
```diff
@@ -0,0 +1,269 @@
+create table if not exists hibernate_sequence (
```
cppwfs, line 1: I know this is how it is setup.   Just seems odd that the skipper migration is in dataflow and not in skipper.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/config/FlywayVendorReplacingEnvironmentPostProcessorTests.java
```diff
@@ -0,0 +1,76 @@
...
+ * in order to verify the {@link FlywayVendorReplacingEnvironmentPostProcessor} properly
+ * handles the '{vendor} token.
+ *
+ * @author Chris Bono
+ */
+@Testcontainers(disabledWithoutDocker = true)
+public class FlywayVendorReplacingEnvironmentPostProcessorTests {
+
+	@Container
+	private static final MariaDBContainer<?> MARIADB_CONTAINER = new MariaDBContainer<>("mariadb:10.4")
```
cppwfs, line 41: We are using Mariadb test container because of licensing?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/db/migration/mysql/V1__Initial_Setup.java
onobc: Yep - fair enough. All of these files were actually here since 2019 and recently deleted like 6 months ago. I just resurrected these from git so was unsure if we should leave the start dates like they were? Thoughts?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/db/migration/mysql/V4__Add_Step_Name_Indexes.java
onobc: Ditto as the above outdated copyright range.

spring-cloud-dataflow-server-core/src/main/resources/schemas/mysql/V1-skipper.sql
onobc: Yeh it is a bit odd but it expects that SCDF "goes first" and if it does then it must provide the schemas.

Issue #5079, opened 2022-09-07
Update ref docs on supported DB and drivers


onobc:
Need to update the database section in both Dataflow and Skipper ref docs to reflect reality wrt supported databases, drivers, and how to use each one.

onobc:
Closed via https://github.com/spring-cloud/spring-cloud-skipper/commit/2cf13cd13d26d0973a6a5fd214d5d2eca6c407e3

PR #5082, opened 2022-09-08
Update database section of ref guide

onobc:
This handles the Dataflow ref guide updates - Skipper to follow.

### Point1
I have created a common database adoc and included in both places that were previously talking about the details of which database/driver is supported etc.. 

### Point2
This will not be the final suggested mechanism for users to extend the classpath as @corneil has another way that is less intrusive and will be addressed by [spring-cloud-dataflow#5083](https://github.com/spring-cloud/spring-cloud-dataflow/issues/5083).

### Point3
I have attached generated PDF for your viewing pleasure while reviewing in case you do not have the IDEA asciidoc plugin etc.. These will NOT be checked in. 
- [configuration-kubernetes.pdf](https://github.com/spring-cloud/spring-cloud-dataflow/files/9528541/configuration-kubernetes.pdf)
- [configuration-local.pdf](https://github.com/spring-cloud/spring-cloud-dataflow/files/9528542/configuration-local.pdf)

See #5079

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/configuration-database.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/configuration-database.adoc
@@ -1,0 +1,65 @@
+
+[[configuration-database-overview]]
+A relational database is used to store stream and task definitions as well as the state of executed tasks.
+Spring Cloud Data Flow provides schemas for *MariaDB*, *MySQL*, *Oracle*, *PostgreSQL*, *Db2*, *SQL Server*, and *H2*. The schema is automatically created when the server starts.
+
+NOTE: The JDBC drivers for *MariaDB*, *MySQL* (via the _MariaDB_ driver), *PostgreSQL*, *SQL Server* are available without additional configuration. To use any other database you need to put the corresponding JDBC driver jar on the classpath of the server as described <<#add-custom-driver,here>>.
+
+To configure a database the following properties must be set:
+
+* `spring.datasource.url`
+* `spring.datasource.username`
+* `spring.datasource.password`
+* `spring.datasource.driver-class-name`
+
+The `username` and `password` are the same regardless of the database. However, the `url` and `driver-class-name` vary per database as follows.
+
+[frame="none"]
+[cols="a,a,a,a"]
+[cols="10%,30%,20%,10%"]
+|===
+|[.small]#Database#|[.small]#spring.datasource.url#|[.small]#spring.datasource.driver-class-name#|[.small]#Driver included#
+
+|[.small]#MariaDB 10.4+#
+|[.small]#jdbc:mariadb://${db-hostname}:${db-port}/${db-name}#
+|[.small]#org.mariadb.jdbc.Driver#
+|[.small]#Yes#
+
+|[.small]#MySQL 5.7#
+|[.small]#jdbc:mysql://${db-hostname}:${db-port}/${db-name}?permitMysqlScheme#
+|[.small]#org.mariadb.jdbc.Driver#
+|[.small]#Yes#
+
+|[.small]#MySQL 8.0+#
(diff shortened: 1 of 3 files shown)


spring-cloud-dataflow-docs/src/main/asciidoc/configuration-database.adoc
```diff
@@ -0,0 +1,65 @@
...
+* `spring.datasource.url`
+* `spring.datasource.username`
+* `spring.datasource.password`
+* `spring.datasource.driver-class-name`
+
+The `username` and `password` are the same regardless of the database. However, the `url` and `driver-class-name` vary per database as follows.
+
+[frame="none"]
+[cols="a,a,a,a"]
+[cols="10%,30%,20%,10%"]
```
markpollack, line 19: sexy table!

spring-cloud-dataflow-docs/src/main/asciidoc/configuration-database.adoc
```diff
@@ -0,0 +1,65 @@
...
+Spring Cloud Data Flow provides schemas for *MariaDB*, *MySQL*, *Oracle*, *PostgreSQL*, *Db2*, *SQL Server*, and *H2*. The schema is automatically created when the server starts.
+
+NOTE: The JDBC drivers for *MariaDB*, *MySQL* (via the _MariaDB_ driver), *PostgreSQL*, *SQL Server* are available without additional configuration. To use any other database you need to put the corresponding JDBC driver jar on the classpath of the server as described <<#add-custom-driver,here>>.
+
+To configure a database the following properties must be set:
+
+* `spring.datasource.url`
+* `spring.datasource.username`
+* `spring.datasource.password`
+* `spring.datasource.driver-class-name`
```
markpollack, line 13: should we show a sample, just one, as in the case of the code snippits that were removed below showing `java -jar sdcdf.jar --spring.datasource.url` or is that overkill?

onobc: It is still there, but below this section.

spring-cloud-dataflow-docs/src/main/asciidoc/configuration-database.adoc
onobc: You can get an idea of what it looks like w/ the attached PDF. Also , note the sexy footnote on the MySQL8 url info

spring-cloud-dataflow-docs/src/main/asciidoc/configuration-database.adoc
```diff
@@ -0,0 +1,65 @@
...
+|[.small]#No#
+
+|[.small]#Oracle#
+|[.small]#jdbc:oracle:thin:@${db-hostname}:${db-port}/{db-name}#
+|[.small]#oracle.jdbc.OracleDriver#
+|[.small]#No#
+|===
+
+==== H2
+When no other database is configured and the *H2* driver has been added to the server classpath then
```
cppwfs, line 60: Noob question, Is there a case where we do not include the H2 driver?   If they don't specify a datasource dataflow defaults to H2.

onobc: So....

The `H2` config was split into its own config and made optional in #4802 as it was lighting up CVE scans. 

We quickly realized the missing dep. broke running locally w/ `java -jar` as well as many examples that do the same in the docs. Because of this I made the dependency non-optional [here](https://github.com/spring-cloud/spring-cloud-dataflow/commit/7525254469ed52fa75b86ccc5c814c9e0c99de97).

As of now the dependency is still provided. We need to decide what we are going to do in this area (provide it or not). This doc is written as if it will be removed. Lets leave it like that for now and we will re-address the docs when we make a decision on H2. 

Sound good @cppwfs ?

onobc: I'm going to merge this and we can follow on w/ whatever we need to.

PR #5086, opened 2022-09-09
Added docs for REST API Version to SCDF Version

cppwfs:
its a brief discusion on RESTFul API Version
* Where to find it
* Maps between the REST API version and SCDF Version


onobc (approved):
Looks good @cppwfs ! Thanks for adding this.

PR #5126, opened 2022-10-21
Deprecates attributes of the StepExecutionHistory

cppwfs:
These attributes are not actively being used and cause confusion.

* The reason that the StepExecutionHistory was not deprecated as a whole is that duration and count are used in the percent complete feature for steps offered by the UI.
* Also added some documentation to the StepExecutionHistory class.
* Notifying folks of a deprecation via the RESTFul API has to be done through documentation for now.
* jThere was an option do so via the schema of the JSON however, based on my reading the plugin in Jackson that allows us to add a schema will soon no longer be supported.

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/api-guide.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/api-guide.adoc
@@ -2769,6 +2769,7 @@ include::{snippets}/job-step-executions-documentation/step-progress/curl-request
 
 include::{snippets}/job-step-executions-documentation/step-progress/http-response.adoc[]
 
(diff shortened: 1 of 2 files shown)


onobc (approved):
Looks good @cppwfs . Thank you.

klopfdreh:
Is there a link to the UI ticket which removes the count attribute for step history?

klopfdreh:
> Is there a link to the UI ticket which removes the count attribute for step history?

Any hints on this @onobc / @cppwfs?

cppwfs:
@klopfdreh Thanks for catching this.  I've created a ticket.

Issue #5111, opened 2022-10-10
Colon (port) in docker registry URI brokes detection of WWW-Authorization URL in DockerConfigJsonSecretToRegistryConfigurationConverter


Nirvandil:
**Description:**
Hi there. 
I'm using scdf in k8s. 
Having something like this in .dockerjson secret
`{
        "auths": {
                "registry.somedomain.com:5050": {
                        "username": "kubernetes-puller",
                        "password": "somePassword"
                }
        }
}`
leads to [org.springframework.cloud.dataflow.container.registry.authorization.DockerConfigJsonSecretToRegistryConfigurationConverter#getDockerTokenServiceUri](https://github.com/spring-cloud/spring-cloud-dataflow/blob/2.9.6/spring-cloud-dataflow-container-registry/src/main/java/org/springframework/cloud/dataflow/container/registry/authorization/DockerConfigJsonSecretToRegistryConfigurationConverter.java#L100) fail to detect right URL. 
Especially, it passes **registry.somedomain.com:5050** (with port!) as  **host** variable to UriComponentsBuilder, so later it detected as IPv6 address in URI.java:
![image](https://user-images.githubusercontent.com/14275342/194900720-c5e0f471-8977-4271-ac82-cb0d5c106bdd.png)
which leads to adding brackets to the URI and then (on build URI) throwing exception because it's invalid IPv6 address obviously. This exception then is swallowed here [org/springframework/cloud/dataflow/container/registry/authorization/DockerConfigJsonSecretToRegistryConfigurationConverter.java:212 ](https://github.com/spring-cloud/spring-cloud-dataflow/blob/28230a0ef308e9a2f58b075f8646679e326501d2/spring-cloud-dataflow-container-registry/src/main/java/org/springframework/cloud/dataflow/container/registry/authorization/DockerConfigJsonSecretToRegistryConfigurationConverter.java#L211) silently returning Optional.empty(). 
**Release versions:**
2.9.6
**Steps to reproduce:**
Just add custom docker registry with port in his address and see DockerConfigJsonSecretToRegistryConfigurationConverter detects it as _basicauth_ instead of _dockeroauth2_

corneil:
@Nirvandil We've identified this is a bug and we will need to ensure the name is handled properly.

onobc:
Closed via https://github.com/spring-cloud/spring-cloud-dataflow/commit/167132d00f758d9fa15fe4b0f135801d3e669710

PR #5130, opened 2022-10-25
Colon (port) in docker registry URI cause error

corneil:
Reproduce #5111


corneil:
Merged from different branch

Issue #5127, opened 2022-10-25
Security issues in transitive dependency of snakeyaml-1.27 (SPRING-CLOUD-DATAFLOW 2.9.6)


obaSAP:
Hello,

We've found the following security issue in the transitive dependency of snakeyaml-1.27 in version 2.9.6:
SCDF 2.9.6 transitively consumes snakeyaml-1.27 spring-boot-starter:jar:2.4.2.
https://www.mend.io/vulnerability-database/CVE-2022-25857
Would appreciate your kind assistance with fixing the issue.

Thanks & BR,
Omri.

benaomri:
Today we upgraded SCDF version to 2.10.2
But still there are issues
![image](https://user-images.githubusercontent.com/63296602/229356427-eed28822-e32e-4d69-b0f2-9fc56e10217b.png)

obaSAP:
@onobc Hi,
Will this also be solved in version 2.7.11?

onobc:
HI @obaSAP ,
We have already updated the usage and version of snake yaml in both 2.10 and 2.9 SCDF. However, 2.9 has not been released w/ the changes. We will discuss w/ the team what our plan is for 2.9 release. We are targeting a 2.10.3 release next week.

cppwfs:
Upgrading to the latest release of SCDF will resolve this issue.  If closing in error please add comment.  Thank you.

PR #5134, opened 2022-10-28
[Issue] #5127 Ensure Jettison is at least 1.5.1

corneil:


hpoettker:
I haven't checked but it may well be that Jettison is only pulled in transitively through Spring Batch. If that's the case then upgrading Spring Batch to 4.3.7 will take care of this and jettison Jettison: https://github.com/spring-projects/spring-batch/releases/tag/4.3.7

Issue #5167, opened 2022-12-21
Local Install Fails for mysql


robertmcnees:
While installing Dataflow locally for the first time following this [this guide](https://dataflow.spring.io/docs/installation/local/docker/) I initially tried mysql as a database provider.  The documentation references this as an option and [seems valid](https://github.com/spring-cloud/spring-cloud-dataflow/tree/main/src/docker-compose).

However when executing the command:
`docker-compose -f docker-compose.yml -f docker-compose-rabbitmq.yml -f docker-compose-mysql.yml up`

the application failed to start and produced this error:

```
skipper              | org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'flyway' defined in class path resource [org/springframework/boot/autoconfigure/flyway/FlywayAutoConfiguration$FlywayConfiguration.class]: Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.flywaydb.core.Flyway]: Factory method 'flyway' threw exception; nested exception is java.lang.RuntimeException: Driver org.mariadb.jdbc.Driver claims to not accept jdbcUrl, jdbc:mysql://mysql:3306/dataflow
```

Changing the database to mariadb and executing the same command 
`docker-compose -f docker-compose.yml -f docker-compose-rabbitmq.yml -f docker-compose-mariadb.yml up`
was successful.

Should the local install for Dataflow support mysql as a database provider?

Full stack trace for the error:
```
dataflow-server      | 2022-12-21 18:17:52.083  INFO 1 --- [           main] o.s.b.w.embedded.tomcat.TomcatWebServer  : Tomcat initialized with port(s): 9393 (http)
dataflow-server      | 2022-12-21 18:17:52.094  INFO 1 --- [           main] o.a.coyote.http11.Http11NioProtocol      : Initializing ProtocolHandler ["http-nio-9393"]
dataflow-server      | 2022-12-21 18:17:52.094  INFO 1 --- [           main] o.apache.catalina.core.StandardService   : Starting service [Tomcat]
dataflow-server      | 2022-12-21 18:17:52.095  INFO 1 --- [           main] org.apache.catalina.core.StandardEngine  : Starting Servlet engine: [Apache Tomcat/9.0.69]
[...]

PR #5168, opened 2022-12-22
Fix for mysql when using docker compose to run local

robertmcnees:
The guide for running dataflow locally does not work when choosing mysql as the database provider.  Full details in issue #5167


onobc (approved):
LGTM @robertmcnees 
Thanks for finding the issue and most of all for submitting a fix for it - much appreciated.

PR #5192, opened 2023-01-20
Change private repositories

corneil:
Change from `springsource-docker-private-local.jfrog.io` to `dev.registry.pivotal.io/p-scdf-for-kubernetes`


onobc (approved):
Thanks @corneil - LGTM

PR #5198, opened 2023-01-25
Fix repo.spring.io references on 2.9.x branch

### LINKED ISSUE #5197: Remove all uses of repo.spring.io/release (2.9.x branch)
Need to remove all invalid repo.spring.io references through all the things. 

Just like we did in the main branch (#5193)

corneil:
Fixes #5197


.settings.xml
```diff
@@ -23,23 +23,15 @@
...
-					<url>https://repo.spring.io/libs-milestone-local</url>
-					<snapshots>
-						<enabled>false</enabled>
-					</snapshots>
-				</repository>
-				<repository>
-					<id>spring-releases</id>
-					<name>Spring Releases</name>
-					<url>https://repo.spring.io/release</url>
+					<url>https://repo.spring.io/milestone</url>
```
onobc, line 34: Unlike `libs-milestone-local` that contains 3rd party deps, when we switch to `milestone` we have to get our 3rd party deps from somewhere... Maven Central. AFAIK we pretty much need to add MC everywhere that we do this replacement. That is my understanding. Is that correct though?

corneil (commented):
OK

PR #5201, opened 2023-01-27
Updates the prometheus proxy version in docker-compose yml

cppwfs:
This allows apps that terminate regardless of the reason to record their entire suite of metrics data.  This is important for especially for ephemeral apps.

### DIFF
--- a/src/docker-compose/docker-compose-prometheus.yml
+++ b/src/docker-compose/docker-compose-prometheus.yml
@@ -22,7 +22,7 @@ services:
       #- SPRING_APPLICATION_JSON={"spring.jpa.properties.hibernate.generate_statistics":true}
 
   prometheus-rsocket-proxy:
-    image: micrometermetrics/prometheus-rsocket-proxy:1.0.0
+    image: micrometermetrics/prometheus-rsocket-proxy:1.5.0
     container_name: prometheus-rsocket-proxy
     expose:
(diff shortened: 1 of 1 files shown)


src/docker-compose/docker-compose-prometheus.yml
```diff
@@ -22,7 +22,7 @@ services:
       #- SPRING_APPLICATION_JSON={"spring.jpa.properties.hibernate.generate_statistics":true}
 
   prometheus-rsocket-proxy:
-    image: micrometermetrics/prometheus-rsocket-proxy:1.0.0
+    image: micrometermetrics/prometheus-rsocket-proxy:1.5.0
```
onobc, line 25: Looks like a good thing to bump up. What all has changed since `1.0.0`?

onobc: Ahh ok, https://github.com/micrometer-metrics/prometheus-rsocket-proxy/releases looks like some bug fixes as well as updates from 1.0.0 w/ `GA release built on top of RSocket 1.0.0 and Spring Boot 2.3.0.` to 1.5.0 w/ `Upgrade to Spring Boot 2.7.2 `. 

This is good.

onobc (approved):
LGMT @cppwfs  Thanks for the update!

PR #5206, opened 2023-02-01
Updated version for prometheus proxy and updated delete script

cppwfs:
Script now removes cluster roles, cluster role bindings, and service account for prometheus if user sets PROMETHEUS to true. Updated proxy to 1.5 from 0.11.0


onobc (approved):
LGTM - thx for the update!

PR #5208, opened 2023-02-02
Removed invalid deprecation on StepExecutionHistory.getReadCount

### LINKED ISSUE #5207: StepExecutionHistory.getReadCount is incorrectly marked for deprecation
Currently the StepExecutionHistory.getReadCount  is marked for deprecation as part of an effort to remove confusing information about step execution history.  That effort's goal is to provide the basic units of measure to users such that they can calculate statistics in the ways they see fit.  However, SCDF does need the existing read count for its calculations of  batch execution percent complete.

cppwfs:
resolves #5207


onobc (approved):
LGTM - thx for the update!

PR #5210, opened 2023-02-03
Updated docs on `k8s/use-*.sh` scripts.

corneil:
Updated ref docs for local k8s/use-*.sh scripts.


onobc (approved):
LGTM - thx for the update @corneil

Issue #5239, opened 2023-03-03
Shell / CLI support for Task 3.0 


onobc:
The CLI command defined in the class AppRegistryCommands will require updates for several of its commands:

- [x] `Register` - add bootVersion argument with a default value of 2.  Only values of 2 or 3 are accepted, otherwise error.
- [x] `Info` - needs to list boot version
- [x] ~`defaultApplication` - The argument gets converted to QualifiedApplicationName which does not include “bootVersion”~ not needed to invoke method - apps must already exist
- [x] ~`importFromResource` - probably needs work because one would also indicate a boot3 application in there.~ (done on the backend)

#### `importFromResource` notes
Import of apps via property files, e.g. with the addition of a .bootVersion key value pair.
```
source.cdc-debezium=maven://org.springframework.cloud.stream.app:cdc-debezium-source-rabbit:3.2.1
source.cdc-debezium.metadata=maven://org.springframework.cloud.stream.app:cdc-debezium-source-rabbit:jar:metadata:3.2.1
source.cdc-debezium.bootVersion=3 
```

onobc:
Closed via https://github.com/spring-cloud/spring-cloud-dataflow/commit/c8707c61de9bcad693c8e03017072d95c36cc385

PR #5240, opened 2023-03-03
Add AppBootVersion to 'app register' shell/client

onobc:
- Introduces the AppBootVersion enum
- Hooks into 'app register' shell command
- Passes boot version into AppRegistryController

See #5239

### Steps to exercise

- build code in the branch w/ `./mvnw clean install -DskipTests`
- start server w/ `java -jar spring-cloud-dataflow-server/target/spring-cloud-dataflow-server-2.10.2-SNAPSHOT.jar`
- start shell w/ `java -jar spring-cloud-dataflow-shell/target/spring-cloud-dataflow-shell-2.10.2-SNAPSHOT.jar`
- in shell register boot2 app w/o boot version
  ```
  dataflow>app register --type task --force --name timestamp --uri maven://org.springframework.cloud.task.app:task-timestamp:3.2.0
  ```
- in shell register boot2 app w/ boot version
  ```
  dataflow>app register --type task --force --name timestamp --uri maven://org.springframework.cloud.task.app:task-timestamp:3.2.0 --bootVersion 2
  ```
- in shell register boot3 app w/ boot version
  ```
  dataflow>app register --type task --force --name timestamp --uri maven://org.springframework.cloud.task.app:task-timestamp:3.2.0 --bootVersion 3
  ```

The output from the shell is as follows for boot2
```
Successfully registered application 'task:timestamp'
```

The output from the shell is as follows for boot3
```
Successfully registered application 'task:timestamp' (boot 3)
```


corneil:
I would have 
```java
enum BootSchemaVersion {
   BOOT2("2"),
   BOOT3("3");
  private String version;
  BootSchemaVersion(String version) {
    this.version = version
  }
  public static BootSchemaVersion defaultVersion = BOOT2;
  public static BootSchemaVersion fromString(String version) {
    for(BootSchemaVersion item :  BootSchemaVersion.values()) {
        if(version.equals.item.version) {
         return item;
       }
      if(version.equalsIgnoreCase.item.name()) {
        return item;
      }
  }
  return defaultVersion
}
```

Then
```java
class SchemaVersionTarget {
    private String name;
    private BootSchemaVersion schemaVersion;
    private String taskPrefix;
   private String batchPrefix;
   private String dataSource;
  
   // getters and setters and default constructor for values

  public BootSchemaVersion getSchemaVersion() {
    return schemaVersion == null ? BootSchemaVersion.defaultVersion : schemaVersion;
  }
  public static SchemaVersionTarget createDefault(BootSchemaVersion schemaVersion) {
      return SchemaVersionTarget(
          schemaVersion.name().toLowerCase(), 
          schemaVersion, 
          schemaVersion.name() + "_TASK_", 
          schemaVersion.name()+"_BATCH_", null
      );
  }
}
```

```java
List<SchemaVersionTarget> getVersionTargets() {
    return Arrays.stream(BootSchemaVersion.values())
        .map(SchemaVersionTarget::createDefault)
}
```

Everywhere you have `AppBootVersion` I will use `SchemaTargetVersion`

The Resource for SchemaTargetVersions which is the list and default version. UI will contain the text representation in en.json.

```java
public class SchemaTargetVersions extends RepresentationModel<SchemaTargetVersions> {
    String defaultSchema;
    List<SchemaVersionTarget> schemaTargets;
}
```

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/AppBootSchemaVersion.java
```diff
@@ -0,0 +1,54 @@
...
+	private String bootVersion;
+
+	AppBootSchemaVersion(String bootVersion) {
+		this.bootVersion = bootVersion;
+	}
+	public static AppBootSchemaVersion defaultVersion = BOOT2;
+	public static AppBootSchemaVersion fromBootVersion(String bootVersion) {
+		return Arrays.stream(AppBootSchemaVersion.values())
+				.filter((bv) -> bv.bootVersion.equals(bootVersion) || bv.name().equalsIgnoreCase(bootVersion))
+				.findFirst().orElse(defaultVersion);
```
onobc, line 43: I don't think passing back a valid version (the default) if someone passed in something invalid version (eg.  `fromBootVersion('GiGo')` is the best thing to do. It may be masking an error in the app.

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/AppBootSchemaVersion.java
```diff
@@ -0,0 +1,54 @@
...
+
+	BOOT2("2"),
+	BOOT3( "3");
+
+	private String bootVersion;
+
+	AppBootSchemaVersion(String bootVersion) {
+		this.bootVersion = bootVersion;
+	}
+	public static AppBootSchemaVersion defaultVersion = BOOT2;
```
onobc, line 39: I would prefer moving the "default" property to the actual enum. That way anywhere in the system (ui etc..) you can easily see if the boot version you are dealing w/ is the default.

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/AppBootSchemaVersion.java
```diff
@@ -0,0 +1,54 @@
...
+	public static AppBootSchemaVersion fromBootVersion(String bootVersion) {
+		return Arrays.stream(AppBootSchemaVersion.values())
+				.filter((bv) -> bv.bootVersion.equals(bootVersion) || bv.name().equalsIgnoreCase(bootVersion))
```
onobc, line 42: I also favor the stricter "fromBootVersion" where "from<THING>" the "<THING>" is the "bootVersion" property on the enum.

spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/service/DefaultAppRegistryService.java
```diff
@@ -399,7 +399,7 @@ public List<AppRegistration> importAll(boolean overwrite, Resource... resources)
 			String[] typeName = lineSplit[0].split("\\.");
 			if (typeName.length < 2 || typeName.length > 3) {
 				throw new IllegalArgumentException("Invalid format for app key '" + lineSplit[0]
-						+ "'in file. Must be <type>.<name> or <type>.<name>.metadata or <type>.<name>.bootVersion");
```
onobc: This will be part of the change when you rewrite this section

PR #5274, opened 2023-03-27
Docker test will ignore case.

cppwfs:
Currently this test will fail.   With this update it matches the other tests.


onobc (approved):
LGTM. Thx @cppwfs

PR #5319, opened 2023-04-27
Simplify stream app lookup

onobc:
- Improve findStreamContainingAppOf() to short circuit early if no app exists in the stream


corneil (approved):

corneil (approved):

onobc:
Will bring back to life if we need this

PR #5338, opened 2023-05-10
Resolves by labeled app properties are not resolved

### LINKED ISSUE #5337: App properties fail to resolve for CTR tasks using labels with boot entry point
When a user specifies app properties at launch time for tasks that use labels will not have their values resolved when using in local or any entry point style other than `exec`.
For example:

```
a: timestamp && b: timestamp
```
If the user specifies `app.a.foo=bar` the property is passed to the `a` timestamp as a deployment property as `app.a.foo=bar` and thus not passed to the boot app in the pod.   

However, it does work if the timestamp is registered under two different tasks i.e.:
```
task-a && task-b
```

cppwfs:
resolves #5337

The code that handled identifying what properties should be moved from deployment properties
to app properties assumed the following format `app.label.taskname.key`.   But the format is actually `app.label.key`.

### DIFF
--- a/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/service/impl/TaskServiceUtilsTests.java
+++ b/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/service/impl/TaskServiceUtilsTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2020 the original author or authors.
+ * Copyright 2018-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/service/impl/TaskServiceUtilsTests.java
```diff
@@ -183,6 +183,22 @@ public void testExtractAppProperties() {
 		assertThat(result.get("test")).isEqualTo("baz");
 	}
 
+	@Test
+	public void testExtractAppLabelProperties() {
+		Map<String, String> taskDeploymentProperties = new HashMap<>();
+		taskDeploymentProperties.put("app.mylabel.foo", "bar");
+		taskDeploymentProperties.put("test.none", "boo");
```
onobc, line 190: **[TinyNit]** Having the `name = test` and the property name `test` is confusing. I think using something like:
```
name = myappName
label = myappLabel
prop name = myProp
```
would make it clear.

onobc (approved):
@cppwfs looks good. I added a "tinynit" improvement. 

LGTM

cppwfs:
Resolved issues
Squashed 
Merged

PR #5344, opened 2023-05-16
feat: filter for description and dslText

klopfdreh:


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
```diff
@@ -151,36 +153,41 @@ public void destroyAll() {
 	 * Return a page-able list of {@link TaskDefinitionResource} defined tasks.
 	 *
 	 * @param pageable page-able collection of {@code TaskDefinitionResource}
-	 * @param search optional findByTaskNameContains parameter
+	 * @param taskName optional findByTaskNameContains parameter
```
onobc, line 156: I agree, `search` is not a good name for this parameter, especially now that another filter parameter has been added. However, this will be a breaking change to the controller API. 

Let's keep the original endpoint, mark it deprecated, and have it simply call the newly added method, something like: 
```java
	@RequestMapping(value = "", method = RequestMethod.GET)
	@Deprecated
	public PagedModel<? extends TaskDefinitionResource> list(Pageable pageable,
			@RequestParam(required = false) String taskName,
			@RequestParam(required = false) boolean manifest,
			@RequestParam(required = false) String dslText,
			PagedResourcesAssembler<TaskExecutionAwareTaskDefinition> assembler) {
		return list(pageable, taskName, null, dslText, manifest, assembler);
	}
```

The other option would be to keep a single endpoint impl method and add back in the `search` param and javadoc the param as deprecated. 

Wdyt? 

In fact, I was suprised that tests did not break because of this change. When I ran `TaskControllerTests` (that looks like the test that checks this task def. controller) locally it does fail. @corneil can you please take a look at why https://github.com/spring-cloud/spring-cloud-dataflow/actions/runs/4995230765/jobs/8946846252?pr=5344 is not running all tests? It runs some, but you will not see `Starting TaskControllerTests` in the build logs like you do when running local.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
```diff
@@ -151,36 +153,41 @@ public void destroyAll() {
...
 
 		final Page<TaskDefinition> taskDefinitions;
-		if (search != null) {
-			if (dslText != null) {
-				throw new TaskQueryParamException(new String[] {"search", "dslText"});
-			} else {
-				taskDefinitions = repository.findByTaskNameContains(search, pageable);
-			}
+
+		if (Stream.of(taskName, description, dslText).filter(Objects::nonNull).count() > 1L) {
```
onobc, line 174: Clever use!

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
```diff
@@ -151,36 +153,41 @@ public void destroyAll() {
...
+				taskDefinitions = repository.findByDescriptionContains(description, pageable);
 			} else {
-				taskDefinitions = repository.findAll(pageable);
+				taskDefinitions = repository.findByDslTextContains(dslText, pageable);
 			}
 		}
 
-		final Map<String, TaskDefinition> taskDefinitionMap = taskDefinitions.stream()
-				.collect(Collectors.toMap(TaskDefinition::getTaskName, Function.identity()));
+		final Map<String, TaskDefinition> taskDefinitionMap = taskDefinitions.stream().collect(Collectors.toMap(TaskDefinition::getTaskName, Function.identity()));
```
onobc, line 190: **[TinyNit]** Do you mind breaking this back into 2 lines. Did the formatter do this?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
```diff
@@ -151,36 +153,41 @@ public void destroyAll() {
...
-				taskDefinitions = repository.findByDslTextContains(dslText, pageable);
+
+		if (taskName == null && dslText == null && description == null) {
+			taskDefinitions = repository.findAll(pageable);
+		} else {
+			if (taskName != null) {
+				taskDefinitions = repository.findByTaskNameContains(taskName, pageable);
+			} else if (description != null) {
+				taskDefinitions = repository.findByDescriptionContains(description, pageable);
 			} else {
```
onobc, line 185: **[NIT]** I think it would be a bit simpler if we instead call `findAll` in the `else` and have a single if/else branch. Wdyt?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/TaskDefinitionRepository.java
```diff
@@ -32,7 +32,7 @@ public interface TaskDefinitionRepository extends KeyValueRepository<TaskDefinit
 
 	Page<TaskDefinition> findByTaskNameContains(String taskName, Pageable pageable);
 
-	Page<TaskDefinition> findByTaskNameContainsAndDslTextContains(String taskName, String dslText, Pageable pageable);
```
onobc, line 35: Unfortunately I think we should this other (seemingly unused) method and just mark it as `@deprecated`. It is not directly related to this PR and leaving it would reduce any risk for this proposal.

onobc (changes requested):
Thanks for the quick turnaround @klopfdreh . I have a few points to address.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskDefinitionController.java
klopfdreh: I re-added the old method and marked it as deprecated. The old method delegates to the new one and the local tests should not fail any longer.

klopfdreh: I saw that two methods is not possible due to rest restrictions and ambiguous mapping. I changed the implementation to only deprecated the parameter. I think the UI should not be affected by this change as it hands over the parameters to the backend.

onobc: Yeh, I wondered in the back of my head whether or not the method would be ambiguous (due to mappings).

onobc (changes requested):
Thanks for the updates. Everything looks good. 

One request though..
I do see that the new api/params are verified in the UI PR you added, but I think we should add a test (or 2) here in the `TaskControllerTests` for the new param. Really, I would ❤️ to break the test out into its own `TaskDefinitionControllerTests` but that is a bit larger scope than what is in this ticket and I would not ask (nor expect you) to include that here. Would you mind adding a test for the new param though?

klopfdreh:
Sure no problem. I am going to add a test. 👍

klopfdreh:
Little typo in the commit message, but the fix is important 😀

corneil:
How about a simple `findByNameContainsOrDescriptionContainsOrDslTextContains`?

klopfdreh:
> How about a simple `findByNameContainsOrDescriptionContainsOrDslTextContains`?

I thought so too, but I thought there is restriction, because there was already a find method for taskName and dslText, but it was not used here.

onobc (approved):
Great work @klopfdreh . Thanks for the contribution!

PR #5351, opened 2023-05-19
Add boot version to `app info` shell command

onobc:
- Update related tests to Junit5

See #5239

### DIFF
--- a/spring-cloud-dataflow-docs/pom.xml
+++ b/spring-cloud-dataflow-docs/pom.xml
@@ -12,7 +12,7 @@
 	<description>Spring Cloud Data Flow Docs</description>
 	<properties>
 		<main.basedir>${basedir}/..</main.basedir>
-		<docs.resources.version>0.2.1.RELEASE</docs.resources.version>
+		<docs.resources.version>0.2.5</docs.resources.version>
 		<maven-javadoc-plugin.version>3.4.1</maven-javadoc-plugin.version>
 	</properties>
 	<dependencies>
--- a/spring-cloud-dataflow-shell-core/src/test/java/org/springframework/cloud/dataflow/shell/AbstractShellIntegrationTest.java
+++ b/spring-cloud-dataflow-shell-core/src/test/java/org/springframework/cloud/dataflow/shell/AbstractShellIntegrationTest.java
@@ -16,10 +16,10 @@
 
 package org.springframework.cloud.dataflow.shell;
 
-import org.junit.AfterClass;
-import org.junit.BeforeClass;
-import org.junit.Rule;
-import org.junit.rules.TestName;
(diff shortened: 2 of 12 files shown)


spring-cloud-dataflow-docs/pom.xml
```diff
@@ -12,7 +12,7 @@
 	<description>Spring Cloud Data Flow Docs</description>
 	<properties>
 		<main.basedir>${basedir}/..</main.basedir>
-		<docs.resources.version>0.2.1.RELEASE</docs.resources.version>
+		<docs.resources.version>0.2.5</docs.resources.version>
```
onobc, line 15: **[UNRELATED]** `0.2.1.RELEASE` not available in snapshots nor central, `0.2.5` is what we need to be using which is in fact on central.

spring-cloud-dataflow-shell-core/src/test/java/org/springframework/cloud/dataflow/shell/AbstractShellIntegrationTest.java
```diff
@@ -16,10 +16,10 @@
 
 package org.springframework.cloud.dataflow.shell;
 
-import org.junit.AfterClass;
-import org.junit.BeforeClass;
-import org.junit.Rule;
-import org.junit.rules.TestName;
+import org.junit.jupiter.api.AfterAll;
```
onobc, line 19: **[UNRELATED]** I converted the app registry test to Junit5 and therefore had to do the same across the test inheritance hieararchy.

spring-cloud-dataflow-shell-core/src/test/java/org/springframework/cloud/dataflow/shell/command/ConfigCommandTests.java
```diff
@@ -77,7 +77,7 @@ public class ConfigCommandTests {
 
 	@Before
 	public void setUp() {
-		MockitoAnnotations.initMocks(this);
+		MockitoAnnotations.openMocks(this);
```
onobc, line 80: **[UNRELATED]** `initMocks` is deprecated

spring-cloud-dataflow-shell/pom.xml
```diff
@@ -30,6 +30,13 @@
 			<plugin>
 				<groupId>org.springframework.boot</groupId>
 				<artifactId>spring-boot-maven-plugin</artifactId>
+				<executions>
+					<execution>
+						<goals>
+							<goal>repackage</goal>
```
onobc, line 36: **[UNRELATED]** this was not being called during `clean install` which resulted in non-executable shell jar.

corneil (approved):

PR #5361, opened 2023-05-31
Add support to retrieve metadata on task properties

### LINKED ISSUE #5229: Spring Cloud Task Properties to be shown for CTR
`spring.cloud.task` are not currently shown in the available list of properties for CTR.   These are important in that sometimes a user needs to set the `spring.cloud.task.closecontext-enabled` to true when using splits.

cppwfs:
Tested it locally but found an issue in UI deleting property when navigating between builder and freetext
where the task properties were not reconstituted on the UI.   

Resolves #5229


onobc (approved):
Thanks @cppwfs - LGTM!

Issue #5357, opened 2023-05-31
Spring Cloud Data Flow Server Prometheus metrics are not showing the correct values


klopfdreh:
**Description:**
Currently there is an issue with the Prometheus metrics of SCDF server. For example `http_server_requests_seconds_max` of any path is showing the value 0.0 even if I navigate through the UI. 

**Release versions:**
2.10.3

**Custom apps:**
-

**Steps to reproduce:**
Setup `spring-cloud-dataflow-server` with `prometheus-rsocket-proxy` and see `/metrics/connected` endpoint.

**Screenshots:**
![image](https://github.com/spring-cloud/spring-cloud-dataflow/assets/980773/3a6cf4ee-a9da-4ec6-a687-e4a774efe109)

Note: We created our own artifact. based on `https://github.com/spring-cloud/spring-cloud-dataflow/tree/v2.10.3/spring-cloud-dataflow-server` That is the reason why there is a 1.0.63 mentioned.

**Additional context:**
The metrics are provided, but the count somehow is not working

klopfdreh:
I got the dasbhoard from here: https://grafana.com/grafana/dashboards/9933-streams/ and this might be changed so that the application is check that it starts with a pattern so that you can name the application with `myservername-1` and `myservername-2` or `myservername-randomidentifier`

klopfdreh:
The dashboard should be adjusted so that it use `=~` in the metrics.

Variable Value: `SERVER_APPLICATION_NAME=myservername.*` (the `.*` is important to match all pods)
Env-Variable: `MY_POD_NAME` = myservername-3h35f2t3d-rcg8d

Example:
```
"expr": "process_uptime_seconds{application=~\"${SERVER_APPLICATION_NAME}\"}",
```

application.yml
```yml
management:
  metrics:
    tags:
      application: ${MY_POD_NAME}
```

SCDF deployment `env`-variables:
```yml
            - name: MY_POD_NAME
              valueFrom:
                fieldRef:
                  apiVersion: v1
                  fieldPath: metadata.name
```

klopfdreh:
Hope this helps for a kubernetes setup with more than 1 replica. 👍

klopfdreh:
Other than that you could create selection to choose between the servers in the dashboard.

onobc:
We should implement @klopfdreh suggested fix (or something similar) in:

1. Dashboard(s) we provide in SCDF repo
[...]

PR #5362, opened 2023-06-01
fix: enable ctr for prometheus reporting

klopfdreh:
Currently the CTR does not report to prometheus-rsocket-proxy - even if the environment variables are given from the SCDF server. With this fix the CTR is reporting to the prometheus-rsocket-proxy like all other task applications

### DIFF
--- a/spring-cloud-dataflow-build/spring-cloud-dataflow-build-dependencies/pom.xml
+++ b/spring-cloud-dataflow-build/spring-cloud-dataflow-build-dependencies/pom.xml
@@ -27,7 +27,7 @@
 		<commons-io.version>2.7</commons-io.version>
 		<commons-text.version>1.10.0</commons-text.version>
 		<postgresql.version>42.4.3</postgresql.version>
-		<prometheus-rsocket.version>1.5.0</prometheus-rsocket.version>
+		<prometheus-rsocket.version>1.5.1</prometheus-rsocket.version>
 		<java-cfenv.version>2.3.0</java-cfenv.version>
 		<spring-cloud-services-starter-config-client.version>3.5.4</spring-cloud-services-starter-config-client.version>
 		<kubernetes-fabric8-client.version>5.12.4</kubernetes-fabric8-client.version>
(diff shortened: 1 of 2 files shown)


klopfdreh:
@cppwfs - I think there are still references of the build which points to the spring repository instead of central.

image

cppwfs:
This issue resolves #5357 .

klopfdreh:
> This issue resolves #5357 .

No this issue is not the same.

This PR solves the issue that the CTR sends the metrics to the Prometheus Proxy.

The issue linked by you occurs if you scale up the replicas of SCDF to 2 or higher. When you do so there are two instances of the SCDF server reporting the values to prometheus with the same application. So one overrides the values of the other. In this case the dashboard needs to be adjusted. In the issue itself I mentioned how to fix this.

cppwfs:
Whoops!   You are correct.   I'll remove my comment from before to remove confusion.

corneil:
The docresources is still and issue. I thought the changes was in. 
Creating a PR for it now.

cppwfs:
Probably need to add an application.properties where  Prometheus is disabled by default.    i.e. 
`management.metrics.export.prometheus.rsocket.enabled=false`
`management.metrics.export.prometheus.enabled=false`
This way users who do use Prometheus can enable it when they need it.

klopfdreh:
As far as I know those properties are disabled by default, but given when ever Prometheus RSocket Proxy is enabled in the SCDF.

I think separate properties are not needed.

onobc:
They are disabled by default [here](https://github.com/spring-cloud/spring-cloud-dataflow/blob/844b32e544b60d83d454754f3f25d533738de820/spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml#L18).

onobc (approved):
LGTM @klopfdreh - thanks for another excellent contribution!

onobc:
Closed via https://github.com/spring-cloud/spring-cloud-dataflow/commit/5df193a1538275ca9cf9510fdc112bd134da384f

PR #5366, opened 2023-06-04
Remove support for legacy config props metadata files

onobc:
This commit removes the handling of the config props metadata files named:
- spring-configuration-metadata-whitelist.properties
- dataflow-configuration-metadata-whitelist.properties

See https://github.com/spring-cloud/stream-applications/issues/252


onobc:
Closing in favor of #5367

PR #5381, opened 2023-06-13
DB Driver section of reference document updated

### LINKED ISSUE #5380: [Docs] Fix section on adding custom driver in SCDF ref guide
Step 3) of https://docs.spring.io/spring-cloud-dataflow/docs/current/reference/htmlsingle/#add-custom-driver is invalid - no need to "build" anymore. Needs to be updated.

cppwfs:
Updates the section of the Dataflow reference docs that discusses the addition of database drivers and then building the dataflow project with a link to the section on  how to extend the classpath to include the drivers.

resolves #5380

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/configuration-local.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/configuration-local.adoc
@@ -53,12 +53,6 @@ SPRING_DATASOURCE_DRIVER_CLASS_NAME=org.mariadb.jdbc.Driver
 java -jar spring-cloud-dataflow-server/target/spring-cloud-dataflow-server-{project-version}.jar
 ----
 
-[#add-custom-driver]
-==== Adding a Custom JDBC Driver
-You may require the addition of dependencies to Spring Cloud Data Flow that are not provided by default to support specific databases (for example, Oracle).
-Spring Cloud Data Flow provides a method to extend the classpath such that a user can add these drivers as need.
-The following section discusses how to  <<appendix-extend-classpath.adoc#extend-classpath, extend Spring Cloud Data Flow's classpath>>.
-
 [[configuration-local-rdbms-schema]]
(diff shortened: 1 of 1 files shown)


spring-cloud-dataflow-docs/src/main/asciidoc/configuration-local.adoc
```diff
@@ -55,43 +55,7 @@ java -jar spring-cloud-dataflow-server/target/spring-cloud-dataflow-server-{proj
...
-[source]
-----
-spring:
-  datasource:
-    url: jdbc:postgresql://localhost:5432/mydb
-    username: myuser
-    password: mypass
-    driver-class-name:org.postgresql.Driver
-----
+You may require the addition of dependencies to Spring Cloud Data Flow that are not provided by default to support specific databases (for example, Oracle).   Spring Cloud Data Flow provides a method to extend the classpath such that a user can add these drivers as need.   The following section discusses how to  <<appendix-extend-classpath.adoc#extend-classpath, extend Spring Cloud Data Flow's classpath>>.
```
onobc, line 58: Por favor, [one-sentence-per-line](https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line). 😄

onobc: I think we can reword this to be more direct. Wdyt about something like:

> By default, Dataflow includes database drivers for A, B, C on its classpath.
> 
> To use a database that is not included or to use a custom driver for one of the included databases, follow <<appendix-extend-classpath.adoc#extend-classpath,these steps>> to add the driver to the classpath.

What made me read into this a bit was the fact that we do not list the supported drivers and that will be a user's next question. 

Wdyt @cppwfs ?

onobc (approved):
Deleting duplicate info is always nice. LGTM @cppwfs - thank you.

PR #5368, opened 2023-06-05
feat: enable ctr annotation build custom configuration

klopfdreh:
Addresses: https://github.com/spring-cloud/spring-cloud-dataflow/issues/4597

PR #5402, opened 2023-07-17
feature: UnexpectedTaskExecutionException

klopfdreh:
It would be great if the CTR throws an `UnexpectedTaskExecutionException` in favor of `UnexpectedJobExecutionException` so that it is possible to create an `ApplicationListener` which can handle certain error causes.

Because the `TaskExecution` itself is not serializable, the fields are assigned to the `UnexpectedTaskExecutionException`.

If this code is fine for you it also needs to be merged into: https://github.com/spring-cloud/spring-cloud-dataflow/pull/5368

The `UnexpectedTaskExecutionException` also uses the `ExitCodeGenerator` interface so that the error code of a task which has failed is also forwarded through the CTR to its caller.

If the exit code is `null`, the `UnexpectedTaskExecutionException` returns `-1`

### DIFF
--- a/spring-cloud-dataflow-composed-task-runner/src/main/java/org/springframework/cloud/dataflow/composedtaskrunner/support/UnexpectedTaskExecutionException.java
+++ b/spring-cloud-dataflow-composed-task-runner/src/main/java/org/springframework/cloud/dataflow/composedtaskrunner/support/UnexpectedTaskExecutionException.java
@@ -31,6 +31,7 @@ import org.springframework.cloud.task.repository.TaskExecution;
 public class UnexpectedTaskExecutionException extends UnexpectedJobExecutionException implements ExitCodeGenerator {
 
 	private static final long serialVersionUID = 1080992679855603656L;
+
 	/**
 	 * The unique id associated with the task execution.
 	 */
--- a/spring-cloud-dataflow-composed-task-runner/src/main/java/org/springframework/cloud/dataflow/composedtaskrunner/support/UnexpectedTaskExecutionException.java
+++ b/spring-cloud-dataflow-composed-task-runner/src/main/java/org/springframework/cloud/dataflow/composedtaskrunner/support/UnexpectedTaskExecutionException.java
@@ -16,9 +16,7 @@
 
 package org.springframework.cloud.dataflow.composedtaskrunner.support;
 
-import java.util.ArrayList;
 import java.util.Date;
-import java.util.List;
 
 import org.springframework.batch.core.UnexpectedJobExecutionException;
 import org.springframework.boot.ExitCodeGenerator;
@@ -32,8 +30,7 @@ import org.springframework.cloud.task.repository.TaskExecution;
  */
 public class UnexpectedTaskExecutionException extends UnexpectedJobExecutionException implements ExitCodeGenerator {
 
-	private static final long serialVersionUID = 3494615323781500165L;
-
+	private static final long serialVersionUID = 1080992679855603656L;
 	/**
 	 * The unique id associated with the task execution.
 	 */
@@ -78,11 +75,6 @@ public class UnexpectedTaskExecutionException extends UnexpectedJobExecutionExce
 	 * Error information available upon the failure of a task.
 	 */
 	private String errorMessage;
-
-	/**
-	 * The arguments that were used for this task execution.
-	 */
-	private ArrayList<String> arguments;
 
 	/**
 	 * Constructs an UnexpectedTaskExecutionException with the specified
@@ -146,7 +138,6 @@ public class UnexpectedTaskExecutionException extends UnexpectedJobExecutionExce
 			externalExecutionId = taskExecution.getExternalExecutionId();
 			errorMessage = taskExecution.getErrorMessage();
 			exitMessage = taskExecution.getExitMessage();
-			arguments = new ArrayList<>(taskExecution.getArguments());
 		}
 	}
 
@@ -178,10 +169,6 @@ public class UnexpectedTaskExecutionException extends UnexpectedJobExecutionExce
 
 	public String getExitMessage() {
 		return this.exitMessage;
-	}
-
-	public List<String> getArguments() {
-		return this.arguments;
 	}
 
 	public String getErrorMessage() {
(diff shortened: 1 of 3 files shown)


klopfdreh:
@cppwfs - This PR is also mentioned for the time period after 2.11.0, but in general - does the change looks be fine for you?

cppwfs:
Thank you for the contribution.   LGTM.  But I want to test it once we get PR 5404 merged.

klopfdreh:
Great! Thanks in advance for testing it! 👍

klopfdreh:
@cppwfs / @onobc - due to the reason that this is only a small change and tests are already present - can you consider to integrate it in 2.11.1?

klopfdreh:
Sure - at any time 👍

onobc:
HI @klopfdreh ,

Just wanted to give you a heads up, we are shipping a very quick `2.11.1` that contains CVE fixes only. No sooner had we released than 2 new CVEs were discovered in libraries we use. We are doing planning for `2.11.2` early next week and will include this PR in the next release which should be out relatively quick.

klopfdreh:
Thanks for the information! 👍

onobc:
Hi @klopfdreh , do you mind rebasing against main?

spring-cloud-dataflow-composed-task-runner/src/main/java/org/springframework/cloud/dataflow/composedtaskrunner/support/UnexpectedTaskExecutionException.java
```diff
@@ -0,0 +1,199 @@
...
+ * @author Tobias Soloschenko
+ */
+public class UnexpectedTaskExecutionException extends UnexpectedJobExecutionException implements ExitCodeGenerator {
+
+	private static final long serialVersionUID = 3494615323781500165L;
+
+	/**
+	 * The unique id associated with the task execution.
+	 */
+	private long executionId;
```
onobc, line 40: Any reason not to just hold a reference to the TaskExecution since you are capturing all of its fields?

spring-cloud-dataflow-composed-task-runner/src/main/java/org/springframework/cloud/dataflow/composedtaskrunner/support/UnexpectedTaskExecutionException.java
```diff
@@ -0,0 +1,199 @@
...
+			executionId = taskExecution.getExecutionId();
+			parentExecutionId = taskExecution.getParentExecutionId();
+			exitCode = taskExecution.getExitCode();
+			taskName = taskExecution.getTaskName();
+			startTime = taskExecution.getStartTime();
+			endTime = taskExecution.getEndTime();
+			externalExecutionId = taskExecution.getExternalExecutionId();
+			errorMessage = taskExecution.getErrorMessage();
+			exitMessage = taskExecution.getExitMessage();
+			arguments = new ArrayList<>(taskExecution.getArguments());
```
onobc, line 149: Refresh my memory, are these arguments already obfuscated at this point?

spring-cloud-dataflow-composed-task-runner/src/main/java/org/springframework/cloud/dataflow/composedtaskrunner/support/UnexpectedTaskExecutionException.java
klopfdreh: I did so because of two reasons:
1. The TaskExecution itself is not serializable and I didn't wanted to change this
2. This way you can decide which information are available in the exception

spring-cloud-dataflow-composed-task-runner/src/main/java/org/springframework/cloud/dataflow/composedtaskrunner/support/UnexpectedTaskExecutionException.java
klopfdreh: I actually don't know if they are sanitized. It depends on what is returned by `this.taskExplorer.getTaskExecution(this.executionId)`

klopfdreh:
> Hi @klopfdreh , do you mind rebasing against main?

Done - let's see what the CI does and if all tests are running as expected.

I force-pushed the changes based on the current main.

spring-cloud-dataflow-composed-task-runner/src/main/java/org/springframework/cloud/dataflow/composedtaskrunner/support/UnexpectedTaskExecutionException.java
onobc: > This way you can decide which information are available in the UnexpectedTaskExecutionException

Right, but currently we are adding all fields. How about we have a compromise and start out w/ a handful of essential fields (not including the parameters :) ) maybe these:

These for identifying the task execution:
```
executionId = taskExecution.getExecutionId();
externalExecutionId = taskExecution.getExternalExecutionId();
taskName = taskExecution.getTaskName();
```

These for identifying the problem:
```
errorMessage = taskExecution.getErrorMessage();
exitCode = taskExecution.getExitCode();
exitMessage = taskExecution.getExitMessage();
```

Wdyt?

klopfdreh: For parameter I think this is to much you are right but I would include start and end time as well. 👍

onobc: If you can update the commit w/ this change I can merge. Otherwise I can make the change and merge in the next 24hrs. Either way works for me. Thanks @klopfdreh

klopfdreh: Hey @onobc - arguments are removed - the parentExecutionId I also kept in the class as this might be also relevant to get a hint from which parent the actual task has been started.

onobc (approved):
Thanks for the contribution @klopfdreh. Awesome work!

Issue #5408, opened 2023-07-21
Error during "Clean up all task/job executions"


klopfdreh:
**Description:**
We have almost 7000 task executions in our history.

When I tried to use "Clean up all task/job executions" there was an error on client side with a lot of "undefined"

I reloaded the ui, but all task executions and job executions were still present. In the "Task execution" the button "Clean up task execution" was not showing up for each task execution when I pressed on the three dots. 

Also there were a lot of logs

`{"timestamp":"2023-07-21T07:24:13.005+0200","level":"WARN","thread":"https-jsse-nio-8443-exec-8","logger":"org.springframework.cloud.deployer.spi.kubernetes.KubernetesTaskLauncher","message":"Cannot delete job for task \"fortschreibung-tgl-task-m68n655vv3\" (reason: job does not exist)","context":"default"}`

but I don't know if this affects the delete process.

**Release versions:**
Spring Cloud Data Flow Server 2.10.3
Spring Cloud Skipper: 2.9.3

**Custom apps:**
N/A

**Steps to reproduce:**
Create a loooot of task executions with batch jobs and clean them up.

**Screenshots:**
N/A

klopfdreh:
The "Clean up task execution" was missing because I didn't pressed at the three dots from the compose task. Other than that the issue of not being able to "Clean up all task/job executions" is still present.

Question regarding this: If the UI runs in a timeout - will the thread still be opened to proceed the call in the backend?

klopfdreh:
We increased the request timeout of our gateway, but the UI still shows errors. Beside the solution I provided in the PR we currently try to figure out how to solve the issue in the meantime. I just tried to increase the tomcat connection timeout.`server.tomcat.connection-timeout` as this might lead to the errors on the UI if the request is canceled on serverside at 30 seconds (the default configuration of the tomcat in spring). See https://docs.spring.io/spring-boot/docs/current/reference/html/application-properties.html#application-properties.server.server.tomcat.connection-timeout

klopfdreh:
I also had a look into the AngularJS implementation and it seems that there needs to be a configuration for this timeout adjusted as well: https://stackoverflow.com/questions/45938931/default-and-specific-request-timeout - this post is rather old but I guess there should be something in Angular as well.

PR #5412, opened 2023-07-24
feat: async task execution for cleanup - fix for #5408

klopfdreh:
Fixes: https://github.com/spring-cloud/spring-cloud-dataflow/issues/5408

To be able to clean up a lot of task executions the servlet requests needs to be committed. If the timeout on client side of the http request is too small this might cause an issue and the operation is rolled back, even if the task executions could be deleted.

This PR introduces Spring Async to Rest-Controllers so that the request is committed immediately and the server processes the request in the background.

Also you can disable the `DataflowAsyncConfiguration` to process the Request/Response as before.

As you can see here we have some task executions:
Bildschirmfoto 2023-07-24 um 07 26 59

Now we are going to delete them:
Bildschirmfoto 2023-07-24 um 07 38 03
Bildschirmfoto 2023-07-24 um 07 38 50

With `DataflowAsyncConfiguration` enabled you can see that the deletion is handled in the `ThreadPoolTaskExecutor`:
Bildschirmfoto 2023-07-24 um 07 27 29

Without the `DataflowAsyncConfiguration` the requests/responses are processed as before:
Bildschirmfoto 2023-07-24 um 07 43 53


klopfdreh:
If there are further operations that require a long time to be processed, just use `@Async` at the top of controller endpoints. 😃

klopfdreh:
@onobc - I hope it is ok to ask you directly as you helped me also with the integration of: https://github.com/spring-cloud/spring-cloud-dataflow/issues/5408. 

With this implementation the client does not rely on the finished response from server when all tasks / executions are cleaned up, but receives a direct response from the server. The server is cleaning up all tasks / executions in the background then.

onobc:
Not a problem w/ the direct ping @klopfdreh .

We are very busy making sure we hit the 2.11.0 updated date but I will try to get this reviewed and hopefully in for the release. If you don't see any activity on this in the next 48 hrs, please ping me here again. 

Thanks

klopfdreh:
As requested @onobc - ping! 😁

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskExecutionController.java
```diff
@@ -265,6 +266,7 @@ public void cleanup(@PathVariable("id") Set<Long> ids,
 	 */
 	@RequestMapping(method = RequestMethod.DELETE)
 	@ResponseStatus(HttpStatus.OK)
+	@Async
```
onobc, line 269: This needs to be an opt-in feature and here are a couple of options I can think of:

### Option 1
Add another API (eg. `cleanupAllAsync`). 
- ❌ UI will not pick this up and you must use this via REST API
- ➕ low risk of breaking other users
- ➕ UI does not give the false appearance that all has been cleaned up instantly
- ➕ can add CompletableFuture return value for better usage later from UI/callers etc..

### Option 2
Guard the auto-config w/ `enabled` property - hence w/o `@EnableAsync` the `@Async` is invisible.
- ❌ UI will give false appearance that all has been cleaned up instantly when async is enabled
- ❌ moderate risk of breaking/confusing other users
- ➕ for those users that opt-in, the UI does this async

I am in favor of the latter option. If you agree, please prototype/verify the absence of `@EnableAsync` does what I think it does.

spring-cloud-dataflow-server-core/src/main/resources/META-INF/spring.factories
```diff
@@ -5,7 +5,8 @@ org.springframework.boot.env.EnvironmentPostProcessor=\
 org.springframework.boot.autoconfigure.EnableAutoConfiguration=\
   org.springframework.cloud.dataflow.server.config.DataFlowServerAutoConfiguration,\
   org.springframework.cloud.dataflow.server.config.DataFlowControllerAutoConfiguration, \
-  org.springframework.cloud.dataflow.server.config.SpringDocAutoConfiguration
+  org.springframework.cloud.dataflow.server.config.SpringDocAutoConfiguration, \
+  org.springframework.cloud.dataflow.server.config.DataflowAsyncConfiguration
```
onobc, line 9: Lets simplify this by moving this config from auto-config to being imported by `org.springframework.cloud.dataflow.server.config.DataFlowServerConfiguration`. 

Also, let's leverage the Spring Boot [task executor builder](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#features.task-execution-and-scheduling) to avoid having to create another set of config props. I envision something like this: 
```java
@Configuration(proxyBeanMethods = false)
@ConditionalOnProperty(prefix = ASYNC_PREFIX, name = "enabled")
@EnableAsync
class DataflowAsyncConfiguration implements AsyncConfigurer {

	public static final String ASYNC_PREFIX = DataFlowPropertyKeys.PREFIX + "task.cleanup.async";

	@Bean(name = "taskCleanupExecutor")
	Executor taskCleanupExecutor(TaskExecutorBuilder taskExecutorBuilder) {
		return taskExecutorBuilder.threadNamePrefix("TaskCleanup-").build();
	}

	@Override
	public AsyncUncaughtExceptionHandler getAsyncUncaughtExceptionHandler() {
		return (throwable, method, objects) -> logger.error("Exception thrown in @Async Method " + method.getName(),
				throwable);
	}
}
```
We can add SCDF specific props later if we decide they are needed.

onobc (changes requested):
@klopfdreh this looks good. I have some requested changes (mainly around simplifying). Also, we need to add a test (or two) for whatever we add here. 

Thanks for the contribution. We are tight on time but I am able to give this another review in the next 24hrs.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskExecutionController.java
klopfdreh: I don’t know why it should be an opt in feature as the UI says: „X task will be deleted“ - it doesn’t matter if the user can interact directly after pressing ok (in case of async) or has to wait till the backend processed the operation.

spring-cloud-dataflow-server-core/src/main/resources/META-INF/spring.factories
klopfdreh: This is awesome! We can refactor it like this!

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskExecutionController.java
onobc: I hear you. That is a good point. My concern is 

1. w/o some sort of messaging in the UI that mentions that it will happen in the background I fear some users may start looking around and still see the executions and then try again, then run into an issue, then file bug report, etc.. 

2. whether opt-in or opt-out, we need some sort of "light switch" for the feature in case we run into unforeseen issues in production. 

Because it is late in the release cycle I am just being extra cautious. I do want to get this feature in, but I also am treading carefully.

klopfdreh: Good point!

As you prefer we can delay this feature as with „clean task executions after n days“ you can run multiple cleanups and shrink down the count of present executions in multiple steps, so there is no need to hurry up with this anymore.😃 

I am going to refactor and add the simplifications and we can think about a good way to implement this in a following release.

WDYT?

onobc: That would be great and I would appreciate that. I know you put your hard work and effort into this so I wanted to match you w/ whatever effort I could to get this in. Moving it to post 2.11 is the smart decision. Thank you.

spring-cloud-dataflow-server-core/src/main/resources/META-INF/spring.factories
klopfdreh: Refactored as requested 😁

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/TaskExecutionController.java
klopfdreh: Just one little question regarding `Option 2`. The current refactoring includes the `ConditionalOnProperty` to check if the functionality is enabled so you have to opt in to use this feature.

From my understanding `@EnableAsync` has to be used with `@Async` as the annotation on the method picks up the thread pool executor provided with `@EnableAsync` - but I am not 100% sure.

Here are some information regarding async implementation: https://spring.io/guides/gs/async-method/ and https://stackoverflow.com/a/53357076

If this is the case I would suggest to still let `@EnableAsync` be on the Configuration, because it is not used when the opt flag in is not set.

klopfdreh:
The async processing might be used also for other endpoints in addition to task cleanup. So I changed it to a more generic prefix

onobc:
Closed via https://github.com/spring-cloud/spring-cloud-dataflow/commit/79845a016614bb6e60b11cf35a01e18513d3a025

Issue #5409, opened 2023-07-21
Out of Memory Exception while clicking on "Job Execution Ids" in a task


klopfdreh:
**Description:**
With those 7000 task / job executions we had a Task and when I clicked on the number beside "Job Execution Ids" an "java.lang.OutOfMemoryError" occurred.

`{"timestamp":"2023-07-21T07:00:31.855+0200","level":"INFO","thread":"https-jsse-nio-8443-exec-1","logger":"org.hibernate.engine.internal.StatisticalLoggingSessionEventListener","message":"Session Metrics {\n    17814 nanoseconds spent acquiring 1 JDBC connections;\n    0 nanoseconds spent releasing 0 JDBC connections;\n    76242 nanoseconds spent preparing 1 JDBC statements;\n    2300334 nanoseconds spent executing 1 JDBC statements;\n    0 nanoseconds spent executing 0 JDBC batches;\n    0 nanoseconds spent performing 0 L2C puts;\n    0 nanoseconds spent performing 0 L2C hits;\n    0 nanoseconds spent performing 0 L2C misses;\n    73172 nanoseconds spent executing 1 flushes (flushing a total of 1 entities and 0 collections);\n    6864 nanoseconds spent executing 1 partial-flushes (flushing a total of 0 entities and 0 collections)\n}","context":"default"}
Terminating due to java.lang.OutOfMemoryError: Java heap space`

The server is configured with 4000MiB RAM and 1500m cores.

**Release versions:**
Spring Cloud Data Flow Server 2.10.3
Spring Cloud Skipper 2.9.3

**Custom apps:**
N/A

**Steps to reproduce:**
Create a lot of task / job executions and click on the number beside "Job Execution Ids" in a Task.

**Screenshots:**
N/A

**Additional context:**
N/A

cppwfs:
Not repeatable in 2.11.x.  Thank you @klopfdreh for verifying.

klopfdreh:
@cppwfs - sorry that I did not close this ticket. The error is indeed not reproducible. Thanks! 👍

PR #5413, opened 2023-07-24
fix: possible out of memory error

klopfdreh:
Possibly fixes: https://github.com/spring-cloud/spring-cloud-dataflow/issues/5409

As of https://dataschool.com/how-to-teach-people-sql/difference-between-where-and-on-in-sql/ some databases might join before filter which might cause out of memory exceptions because all task manifests are load.

> Different SQL languages may have different query plans based on filtering in the ON clause vs the WHERE clause, so test the performance on your database


klopfdreh:
Edit: This might only optimize the performance on DB server but not cause the JVM to raise an OutOfMemoryError.

I try to figure out what's the real cause and if I find something I am going to push it into this branch. `klopfdreh:fix/outofmemory`

corneil:
We can revisted once we have done testing on the support for Boot 3 workloads with new Spring Batch and Spring Cloud Task schemas.

klopfdreh:
> We can revisted once we have done testing on the support for Boot 3 workloads with new Spring Batch and Spring Cloud Task schemas.

Thanks a lot! 👍

klopfdreh:
@corneil the out of memory issue did not happened with 2.11.0 anymore. I am going to close this PR as it only changes the SQLs slightly.

Issue #5373, opened 2023-06-08
Unable to delete stream with corrupted package data


onobc:
**Steps to reproduce:**

1. Create stream 'foo' and deploy it
2. Manually delete its data from `skipper_package_metadata` and `skipper_package_file`
3. Create stream 'foobar' and deploy it
4. Delete stream 'foo'

         💥 💣 💥 💣 

          An error occurred while bulk deleting Streams.
          Please check the server logs for more details.

**Cause**
The delete operation attempts to also delete the release info in Skipper. 
The un-deploy operation only calls Skipper when the target has package info in Skipper.
It users a wildcard search of stream name to find package info in Skipper.

So even though there is no package info for 'foo', the results are found for 'foobar' and the operation proceeds and then fails and an exception is bubbled up, ultimately preventing the stream def from being deleted.
```
DefaultStreamService.deleleStream('foo')
	SkipperStreamDeployer.undeploy('foo')
		guard on delete block does a search and '%foo%' finds packages for 'foobar'
		SkipperClient.delete('foo') -> fails in StateMachine - so attempts now to manually delete packages
			SkipperClient.packageDelete('foo') -> fails as the package does not exist (exception bubbles up)
        **NEVER GETS CALLED**
        deleteStreamDef('foo') 
```

**What should happen**
The un-deploy should not find 'foobar' when looking for 'foo' and hence not call over to Skipper. 
This in turn will allow the delete of the stream def to proceed.

onobc:
Another instance of this has been reported on [here on Stackoverflow](https://stackoverflow.com/questions/76479127/spring-cloud-dataflow-stream-org-springframework-cloud-skipper-skipperexceptio). 

In general, we should allow a forced destroy even when the stream is in some corrupted state (regardless of the failure/corruption) reason. 

NOTE: It is also possible that https://github.com/spring-cloud/spring-cloud-dataflow/issues/5382 is just a random failure of the undeploy and may not have anything to do w/ the upgrade from 2.8 -> 2.9

PR #5418, opened 2023-07-28
Skipper undeploy attempts delete of other info when no package is found for stream

onobc:
Fixes GH-5373

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/stream/SkipperStreamDeployer.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/stream/SkipperStreamDeployer.java
@@ -473,7 +473,6 @@ public class SkipperStreamDeployer implements StreamDeployer {
                 this.skipperClient.packageDelete(streamName);
             };
         } else {
-            // Stream may have package data is corrupted/missing so we attempt to delete all but the package info
             logger.info("Can not find package named '{}' - bypassing Skipper delete.", streamName);
         }
     }
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/stream/SkipperStreamDeployer.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/stream/SkipperStreamDeployer.java
@@ -463,14 +463,18 @@ public class SkipperStreamDeployer implements StreamDeployer {
     }
 
     public void undeployStream(String streamName) {
-        Collection<PackageMetadata> packageMetadataResources = this.skipperClient.search(streamName, false);
-        if (!packageMetadataResources.isEmpty()) {
+        Collection<PackageMetadata> packages = this.skipperClient.search(streamName, false);
+        boolean packageExists = packages.stream().map(PackageMetadata::getName).anyMatch(s -> s.equalsIgnoreCase(streamName));
+        if (packageExists) {
             try {
                 this.skipperClient.delete(streamName, true);
             } catch (ReleaseNotFoundException e) {
                 logger.info("Release not found for {}. Deleting the package {}", streamName, streamName);
                 this.skipperClient.packageDelete(streamName);
-            }
+            };
+        } else {
+            // Stream may have package data is corrupted/missing so we attempt to delete all but the package info
+            logger.info("Can not find package named '{}' - bypassing Skipper delete.", streamName);
         }
     }
 
--- a/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/stream/SkipperStreamDeployerTests.java
+++ b/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/stream/SkipperStreamDeployerTests.java
@@ -66,9 +66,11 @@ import static junit.framework.TestCase.fail;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoMoreInteractions;
 import static org.mockito.Mockito.when;
 
 /**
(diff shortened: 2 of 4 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/stream/SkipperStreamDeployer.java
```diff
@@ -463,15 +463,20 @@ private List<Template> createGenericTemplate() {
...
+        if (packageExists) {
             try {
                 this.skipperClient.delete(streamName, true);
             } catch (ReleaseNotFoundException e) {
                 logger.info("Release not found for {}. Deleting the package {}", streamName, streamName);
                 this.skipperClient.packageDelete(streamName);
             }
+            return;
         }
+        // The assumption is the package data is corrupted/missing so we attempt to delete all but the package info
```
onobc, line 477: We should ultimately provide a boolean flag that can be passed in to decide whether to do this. It is only useful in the case where a customer has "corrupted" data and we want to force a delete regardless of what is going on.

Let's discuss @markpollack @corneil 

If we want to add a flag this is out of scope for `2.11.0` (IMO).

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/stream/SkipperStreamDeployerTests.java
```diff
@@ -499,40 +501,67 @@ public void testStateOfDefinedUndeployedStream() {
...
 		SkipperClient skipperClient = mock(SkipperClient.class);
 		StreamDefinitionRepository streamDefinitionRepository = mock(StreamDefinitionRepository.class);
-
 		SkipperStreamDeployer skipperStreamDeployer = new SkipperStreamDeployer(skipperClient,
 				streamDefinitionRepository, appRegistryService, mock(ForkJoinPool.class), new DefaultStreamDefinitionService());
-
 		StreamDefinition streamDefinition = new StreamDefinition("foo", "foo|bar");
-
-		when(skipperClient.search(eq(streamDefinition.getName()), eq(false))).thenReturn(new ArrayList<>());
+		PackageMetadata packageMetadata = new PackageMetadata();
```
onobc, line 511: These tests are too mocked out - only really verifying that the impl I have added is in fact what gets called. Ultimately we need to do the following in a full integration test:

- Create stream `foo`
- Create stream `foobar`
- Delete foo package data (for real or mocked out)
- Delete stream `foo`
- Verify stream `foo` deleted
- Create stream `foo` 
- Verify it creates stream w/o barfing about "release already exists".

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/stream/SkipperStreamDeployer.java
corneil: We probably don't want to fail a delete for something that doesn't exist. It is the same type of scenario like when multiple tasks are deleted from UI using forkJoin which runs mutiple deletes in parallel, some have children that get deleted when deleting parent and then when the child is tried and doesn't exists it fails. We don't need to fail if delete doesn't actually delete anything since another delete may have been done just before by another party.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/stream/SkipperStreamDeployerTests.java
corneil: We need a test that launches rabbit, skipper and database using containers and then using streamservice to deploy tests and jdbctemplate to delete the entries. Then service to perform delete

onobc: Exactly.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/stream/SkipperStreamDeployer.java
onobc: > We probably don't want to fail a delete for something that doesn't exist. 

Not sure where you are referring to. We are skipping the delete when the package does not exist which prevents the failing delete from happening. There are 2 changes above:

### Stop wildcard search from allowing streams w/o package metadata to be deleted
```java
        boolean packageExists = packages.stream().map(PackageMetadata::getName).anyMatch(s -> s.equalsIgnoreCase(streamName));
        if (packageExists) {
              // do same thing as before 'delete(name, true)'
        }
```
Before the above change, if package metadata was missing for "foo" (but there was a stream "foobar") then it would attempt the delete (including package data). The delete would fail, but not with the caught exception `ReleaseNotFoundException` and instead a `SkipperException` and hence it would bubble up to the caller and the stream def would not be deleted. Therefore the user was never able to delete the stream def from the UI.

The above is effectively an exact name (case insensitive) search so that delete will only work if the package data exists. This allows the user to delete the stream def and all "appears" fine.  It is not though, see next section.

### Allow stream to be re-created w/ same name
The 2nd part of the code change is to delete the stream when the package does not exist (excluding package metadata because we know it does not exist). 
```java
  logger.warn("Can not find package named '{}' - assuming package corrupted; attempting to delete release info", streamName);
        this.skipperClient.delete(streamName, false);
```
Note the `false` param. 

Without this change, the next issue is that when the user attempts to re-create the stream w/ same name (eg. "foo"), it fails w/ a "Release w/ name 'foo' already exists" because it did not remove the other data/state from Skipper as we skipped the delete during "undeploy".

Adding this other step clears deletes any other stream data and solves both of the cases. 

### My 2c
I think this is the wrong way to go. I think it would be better to provide a dedicated "cleanup stream" API that could be used in the case where a user has corrupted data (for whatever reason). The API would delete each piece of the stream in a safe manner and ensure the data is gone (all of it). This could be secured and under the "Misc" panel in the UI. Or shell access only??

I also think this is growing in scope and is not as important as getting `2.11.0` out.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/stream/SkipperStreamDeployer.java
```diff
@@ -463,14 +463,18 @@ private List<Template> createGenericTemplate() {
...
+        if (packageExists) {
             try {
                 this.skipperClient.delete(streamName, true);
             } catch (ReleaseNotFoundException e) {
                 logger.info("Release not found for {}. Deleting the package {}", streamName, streamName);
                 this.skipperClient.packageDelete(streamName);
-            }
+            };
+        } else {
+            // Stream may have package data is corrupted/missing so we attempt to delete all but the package info
```
cppwfs, line 476: Nitpick: package data `that` is

onobc: Good catch - actually the comment is outdate altogether - yanked.

PR #5422, opened 2023-08-02
feat: task cleanup before end time in days

klopfdreh:
Improvement:
* Cleanup task executions older than x days.

Fix:
* Cleanup in management is working now, after the boot 3 changes.

Fixes: https://github.com/spring-cloud/spring-cloud-dataflow/issues/5423
Needs to be integrated with: https://github.com/spring-cloud/spring-cloud-dataflow-ui/pull/1940


onobc:
@klopfdreh thanks again for another contribution! Do you mind rebasing and merging as there are currently conflicts and it is a bit difficult to see what has changed in main vs. in this PR (I did take a pass at that but realized it may be better if you could do that). I will keep an eye out and re-review once that is done. 

Thanks.

klopfdreh:
@onobc - okidoki - I am going to do so, tomorrow! 👍

klopfdreh:
@onobc - well `tomorrow` in 23 mins 😆

onobc:
Hi @klopfdreh ,
I just now got back around to look at the updates and I realized you did a merge. This still leaves the changes unclear once I pull them down to review. When I said "Do you mind rebasing and merging" I should have said "Do you mind rebasing/squashing and handling the 2 conflicting files".  Ideally there will be a single commit over the latest changes in main so that I can easily tell what you changed. 

Sorry for the back and forth. I will make this my top priority tomorrow am (or whenever you get the updates in). Thanks.

klopfdreh:
Ah ok. I created a single commit with all changes. Just to ensure to have a backup: https://github.com/klopfdreh/spring-cloud-dataflow/tree/feature/cleanupEndTimeBackup

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/TaskExecutionService.java
```diff
@@ -94,4 +94,14 @@ public interface TaskExecutionService {
...
 	 */
 	Integer getAllTaskExecutionsCount(boolean onlyCompleted, String taskName);
+
+	/**
+	 * Returns the count of all the task execution IDs with the option to include only the completed task executions.
+	 * @param onlyCompleted filter by completed task executions
+	 * @param taskName the task name, if null then retrieve all the tasks
+	 * @param days the days before the task should ended
+	 * @return the number of executions
+	 * @since 2.8
```
onobc, line 104: Should be `@since 2.11.0`.

onobc, line 104: Should be `@since 2.11.0`.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/TaskExecutionService.java
```diff
@@ -94,4 +94,14 @@ public interface TaskExecutionService {
 	 * @since 2.8
 	 */
 	Integer getAllTaskExecutionsCount(boolean onlyCompleted, String taskName);
+
+	/**
+	 * Returns the count of all the task execution IDs with the option to include only the completed task executions.
+	 * @param onlyCompleted filter by completed task executions
+	 * @param taskName the task name, if null then retrieve all the tasks
+	 * @param days the days before the task should ended
+	 * @return the number of executions
```
onobc, line 103: I know the other similar method was returning big integer, but I am pretty sure in the "no data" case it returns 0. I would just state that fact in the `@return` something like "never null, 0 if no data" etc..

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/TaskExecutionService.java
```diff
@@ -94,4 +94,14 @@ public interface TaskExecutionService {
 	 * @since 2.8
 	 */
 	Integer getAllTaskExecutionsCount(boolean onlyCompleted, String taskName);
+
+	/**
+	 * Returns the count of all the task execution IDs with the option to include only the completed task executions.
+	 * @param onlyCompleted filter by completed task executions
+	 * @param taskName the task name, if null then retrieve all the tasks
+	 * @param days the days before the task should ended
```
onobc, line 102: I think "days" is a bit ambiguous - especially since this method does not include the "BeforeEndDate" suffix like the other newly introduced methods in the DAO layer. 

Wdyt about adding renaming "days" to something like ...
```java
@param includeTasksEndedMinDaysAgo only include tasks that have ended at least this many days ago
```

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/TaskDeleteService.java
```diff
@@ -53,6 +53,15 @@ public interface TaskDeleteService {
 	 */
 	void cleanupExecutions(Set<TaskExecutionControllerDeleteAction> actionsAsSet, String taskName, boolean completed);
 
+	/**
+	 * Clean up the resources that resuled from running the task with the given name and actions.
+	 *
+	 * @param actionsAsSet the actions
+	 * @param taskName the task name
+	 * @param completed the completion state of the task executions
```
onobc, line 61: Missing the `@param days..` javadoc. Also, let's make it consistent w/ outcome of this comment.

spring-cloud-dataflow-aggregate-task/src/main/java/org/springframework/cloud/dataflow/aggregate/task/impl/AggregateDataFlowTaskExecutionQueryDao.java
```diff
@@ -143,18 +144,54 @@ public class AggregateDataFlowTaskExecutionQueryDao implements DataflowTaskExecu
 			+ "PARENT_EXECUTION_ID, SCHEMA_TARGET"
 			+ " from AGGREGATE_TASK_EXECUTION";
 
+	private static final String GET_EXECUTION_BY_NAME_COMPLETED_BEFORE_END_TIME = "SELECT TASK_EXECUTION_ID,"
```
onobc, line 147: These statements are 99.9% identical so my DRY alarm is going crazy. That being said, you took from the already existing pattern in this class so we should refactor all at once in a separate effort. The current impl (even before this PR) seemed like we are trying to implement JPA repository and we have these permutations of params and they all just result to a piece of info on the WHERE clause (typically). 

If you want to extract a base query that has everything up to the `WHERE` clause for your newly introduced statements only - that would be fine as well. 

I am going to create an issue to refactor the class as a whole.

spring-cloud-dataflow-aggregate-task/src/main/java/org/springframework/cloud/dataflow/aggregate/task/impl/AggregateDataFlowTaskExecutionQueryDao.java
```diff
@@ -143,18 +144,54 @@ public class AggregateDataFlowTaskExecutionQueryDao implements DataflowTaskExecu
...
 			+ "AGGREGATE_TASK_EXECUTION ";
 
+	private static final String TASK_EXECUTION_COUNT_BEFORE_END_TIME = "SELECT COUNT(*) FROM "
+			+ "AGGREGATE_TASK_EXECUTION WHERE END_TIME < :endTime";
+
 	private static final String TASK_EXECUTION_COUNT_BY_NAME = "SELECT COUNT(*) FROM "
 			+ "AGGREGATE_TASK_EXECUTION where TASK_NAME = :taskName";
 
+	private static final String TASK_EXECUTION_COUNT_BY_NAME_AND_BEFORE_END_TIME = "SELECT COUNT(*) FROM "
+			+ "AGGREGATE_TASK_EXECUTION where TASK_NAME = :taskName AND END_TIME < :endTime";
```
onobc, line 181: What happens when end time is null? If it is included in the result then we could cleanup still running tasks. If it is excluded then it does not make any sense to pass in a `onlyCompleted` flag.

spring-cloud-dataflow-aggregate-task/src/main/java/org/springframework/cloud/dataflow/aggregate/task/impl/AggregateDataFlowTaskExecutionQueryDao.java
```diff
@@ -310,6 +361,27 @@ public long getTaskExecutionCountByTaskName(String taskName) {
...
+		Long count;
+		if (StringUtils.hasText(taskName)) {
+			final SqlParameterSource queryParameters = new MapSqlParameterSource()
+					.addValue("taskName", taskName, Types.VARCHAR)
+					.addValue("endTime", endTime, Types.DATE);
+
+			try {
+				count = this.jdbcTemplate.queryForObject(TASK_EXECUTION_COUNT_BY_NAME_AND_BEFORE_END_TIME, queryParameters, Long.class);
+			} catch (EmptyResultDataAccessException e) {
+				count = 0L;
```
onobc, line 375: I am curious why the later `queryForObject` below is also not in a try/catch to enforce return 0 default? I see there are other cases in this code that are already doing the same.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskDeleteService.java
```diff
@@ -193,7 +199,20 @@ private void performCleanupExecution(long id, String schemaTarget) {
...
 	public void cleanupExecutions(Set<TaskExecutionControllerDeleteAction> actionsAsSet, String taskName, boolean completed) {
-		List<AggregateTaskExecution> tasks = this.taskExplorer.findTaskExecutionsByName(taskName, completed);
+		cleanupExecutions(actionsAsSet, taskName, completed, null);
+	}
+
+	@Override
+	@Transactional
+	public void cleanupExecutions(Set<TaskExecutionControllerDeleteAction> actionsAsSet, String taskName, boolean completed, Integer days) {
+		List<AggregateTaskExecution> tasks;
+		if (days != null) {
```
onobc, line 209: I think we could extract this code block to a DateUtils w/ a `numDaysAgoFromLocalMidnightToday(int n)`. It could also be used in the new method in DTES.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
```diff
@@ -200,44 +206,44 @@ public class DefaultTaskExecutionService implements TaskExecutionService {
...
-		TaskAppDeploymentRequestCreator taskAppDeploymentRequestCreator,
-		AggregateTaskExplorer taskExplorer,
-		DataflowTaskExecutionDaoContainer dataflowTaskExecutionDaoContainer,
-		DataflowTaskExecutionMetadataDaoContainer dataflowTaskExecutionMetadataDaoContainer,
-		DataflowTaskExecutionQueryDao dataflowTaskExecutionQueryDao,
-		OAuth2TokenUtilsService oauth2TokenUtilsService,
-		TaskSaveService taskSaveService,
-		TaskConfigurationProperties taskConfigurationProperties,
-		AggregateExecutionSupport aggregateExecutionSupport
+			PropertyResolver propertyResolver,
```
onobc, line 209: Please remove all the extra formatting. I am sure IntelliJ did it and you did not notice it.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/TaskDeleteService.java
onobc, line 61: I think `onlyCompleted` would be more accurate.

spring-cloud-dataflow-aggregate-task/src/main/java/org/springframework/cloud/dataflow/aggregate/task/AggregateTaskExplorer.java
```diff
@@ -97,7 +98,17 @@ public interface AggregateTaskExplorer {
...
 	 * @return list of task executions
 	 */
-	List<AggregateTaskExecution> findTaskExecutionsByName(String taskName, boolean completed);
+	List<AggregateTaskExecution> findTaskExecutions(String taskName, boolean completed);
+
+	/**
+	 * Get a list of executions for a task by name, completion status and end time.
+	 *
+	 * @param taskName  the name of the task to be searched
+	 * @param completed Indicator to find only completed tasks
```
onobc, line 107: The Javadoc does explain the meaning, but I think naming `onlyCompleted` may be more consistent w/ the other newly introduced APIs and also more readable/clear.

I will only comment that once here but it applies to multiple locations in this PR.

onobc (changes requested):
@klopfdreh thanks for the rebase+squash - it was super helpful. I have left a handful of comments to address. 

Also, I noticed you added no tests. Then I quickly realized you added no tests because there are no tests in this newly added module. Since they are not in place now I will not trouble w/ you that burden - we will [create the tests](https://github.com/spring-cloud/spring-cloud-dataflow/issues/5446) and include these new APIs in those tests.

spring-cloud-dataflow-aggregate-task/src/main/java/org/springframework/cloud/dataflow/aggregate/task/AggregateTaskExplorer.java
klopfdreh: I will also change this to the method above as it is called the same there. 👍

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
klopfdreh: Yes I pressed the shortcut for reformat for the whole class which leads to this noise.

spring-cloud-dataflow-aggregate-task/src/main/java/org/springframework/cloud/dataflow/aggregate/task/impl/AggregateDataFlowTaskExecutionQueryDao.java
klopfdreh: Thanks - fixed. 👍

spring-cloud-dataflow-aggregate-task/src/main/java/org/springframework/cloud/dataflow/aggregate/task/impl/AggregateDataFlowTaskExecutionQueryDao.java
klopfdreh: I refactored all statements to reuse the select `SELECT_CLAUSE` and later `GET_EXECUTIONS`

spring-cloud-dataflow-aggregate-task/src/main/java/org/springframework/cloud/dataflow/aggregate/task/impl/AggregateDataFlowTaskExecutionQueryDao.java
klopfdreh: The end time should never be null and is passed in from the calling methods which calculate the endTime. But you are right. I added `@NotNull` to the methods to ensure that the end time is always given.

onobc: Sorry, I meant the `END_TIME` on the record which would make `END_TIME < :endTime` (ie. null < '2023-08-30 23:59:59').

onobc:
Thanks for the updates @klopfdreh . I replied to one of your comments but it is late and I will continue the review in the morning.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskDeleteService.java
klopfdreh: Done.

spring-cloud-dataflow-aggregate-task/src/main/java/org/springframework/cloud/dataflow/aggregate/task/impl/AggregateDataFlowTaskExecutionQueryDao.java
klopfdreh: I fixed it accordingly as I already saw that you could only meant `END_TIME` and not `onlyCompleted` as this was not present in the query 👍

klopfdreh:
> @klopfdreh thanks for the rebase+squash - it was super helpful. I have left a handful of comments to address.
> 
> Also, I noticed you added no tests. Then I quickly realized you added no tests because there are no tests in this newly added module. Since they are not in place now I will not trouble w/ you that burden - we will [create the tests](https://github.com/spring-cloud/spring-cloud-dataflow/issues/5446) and include these new APIs in those tests.

Thanks a lot for helping me out and create tests. I have only a limited time to file in changes.

spring-cloud-dataflow-aggregate-task/src/main/java/org/springframework/cloud/dataflow/aggregate/task/impl/AggregateDataFlowTaskExecutionQueryDao.java
```diff
@@ -86,75 +88,74 @@ public class AggregateDataFlowTaskExecutionQueryDao implements DataflowTaskExecu
...
+	private static final String GET_EXECUTIONS_COMPLETED = GET_EXECUTIONS +
+			" where END_TIME IS NOT NULL";
+
+	private static final String GET_EXECUTION_BY_NAME_COMPLETED_BEFORE_END_TIME = GET_EXECUTIONS +
+			" where TASK_NAME = :taskName AND END_TIME IS NOT NULL AND END_TIME < :endTime";
+
+	private static final String GET_EXECUTIONS_COMPLETED_BEFORE_END_TIME = GET_EXECUTIONS +
+			" where END_TIME IS NOT NULL AND END_TIME < :endTime";
+
+	private static final String GET_EXECUTION_BY_NAME_BEFORE_END_TIME = GET_EXECUTIONS +
```
onobc, line 129: @klopfdreh thanks for the updates. This looks much better! 

One point of contention before I merge though, these queries that include `END_TIME < :endTime` but do not make sure `END_TIME IS NOT NULL`... since we are using the end date as the criteria, what do we expect those w null END_TIME will do? Be included or not?

klopfdreh: Yes I see  - we can remove `GET_EXECUTIONS_BEFORE_END_TIME` and `GET_EXECUTION_BY_NAME_BEFORE_END_TIME` and remove the completed parameter from `findTaskExecutionsBeforeEndTime` and simplify the if statement a bit. I am going to do so later. 👍

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
```diff
@@ -1030,9 +1036,18 @@ public Set<Long> getAllTaskExecutionIds(boolean onlyCompleted, String taskName)
...
-		return (int) (
-			onlyCompleted ? dataflowTaskExecutionQueryDao.getCompletedTaskExecutionCountByTaskName(taskName)
-				: dataflowTaskExecutionQueryDao.getTaskExecutionCountByTaskName(taskName)
-		);
+		return getAllTaskExecutionsCount(onlyCompleted, taskName, null);
+	}
+
+	@Override
+	public Integer getAllTaskExecutionsCount(boolean onlyCompleted, String taskName, Integer days) {
+		if (days != null) {
```
onobc, line 1044: @klopfdreh thanks for the updates to the cleanup path - looks great. I want to get this merged but I think the "what to do w/ END_TIME null when user specifies onlyCompleted='false'" still applies to the executions count path as well. 

I totally understand you have limited time - thank you for. your patience while we go through our back/forth :) .

Let's figure out what we want to do here and I am more than willing to make those changes and we can co-author this - unless you have the cycles to make another update. 

### Case 1
If `includeTasksEndedMinDaysAgo` is specified then that implies `onlyCompleted=true` - it has to be in order to compare the dates.
* the new API `getAllTaskExecutionsCount(boolean onlyCompleted, String taskName, Integer includeTasksEndedMinDaysAgo)` can have its `onlyCompleted` param removed and it would solve this case

### Case 2
If `includeTasksEndedMinDaysAgo` is NOT specified then `onlyCompleted=true|false` makes sense.
* the already existing API `getAllTaskExecutionsCount(boolean onlyCompleted, String taskName)` solves this one

Wdyt?

klopfdreh: You are completely right with the scenario if the END_TIME is null the onlyCompleted should still apply! Thanks your a lot for the time to review and for the hints!

Sadly I am busy by solving issues at our integration and I don’t know when I am able to revisit to make changes again.

I would appreciate it very much if you could assist me a bit with those changes.

As this feature very important - I guess also for a lot of users of SCDF - it would be very nice to see it integrated soon as your time allows it.

Thank you a lot for the help and the code reviews and of course for the assistance!

onobc: > Thanks your a lot for the time to review and for the hints!

You are more than welcome - anytime.

> Sadly I am busy by solving issues at our integration and I don’t know when I am able to revisit to make changes again.
> I would appreciate it very much if you could assist me a bit with those changes.
> 

No worries @klopfdreh . I will get it updated and integrated this weekend. ☮️ and thanks again.

onobc:
Verifying my commits on top of @klopfdreh in https://github.com/spring-cloud/spring-cloud-dataflow/pull/5453. 

If that passes the CI PR then we will merge it and close this one out a super ceded by it.

onobc:
Superceded by. https://github.com/spring-cloud/spring-cloud-dataflow/pull/5453

PR #5462, opened 2023-09-12
Spring Boot 3.x support doc needs section on CTR

cppwfs:
Added docs on CTR not requiring any special instructions for launching. A note was added to remind users to select the appropriate boot version during app registration


onobc (approved):
LGTM - thanks @cppwfs !

PR #5473, opened 2023-09-19
Readded param to method

cppwfs:
Added docs


onobc (approved):
LGTM - thanks @cppwfs

PR #5541, opened 2023-11-07
Add git and build info to AboutController

felipeg48:
Adding the new git and build info metadata to the AboutController class.
- [x] Actuator is always necessary
- [x] `build-info` is necessary
- [x] the `git-commit-id-maven-plugin` is necessary.


spring-cloud-dataflow-server-core/pom.xml
```diff
@@ -306,6 +306,44 @@
...
+				<groupId>org.springframework.boot</groupId>
+				<artifactId>spring-boot-maven-plugin</artifactId>
+				<executions>
+					<execution>
+						<goals>
+							<goal>build-info</goal>
+						</goals>
+						<configuration>
+							<additionalProperties>
+								<encoding.source>UTF-8</encoding.source>
```
onobc, line 319: What are these additional props needed for? These would be `build.encoding.source=UTF-8` and `build.encoding.reporting=UTF-8`.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AboutController.java
```diff
@@ -15,28 +15,20 @@
...
-import org.springframework.cloud.dataflow.rest.resource.about.AboutResource;
-import org.springframework.cloud.dataflow.rest.resource.about.Dependency;
-import org.springframework.cloud.dataflow.rest.resource.about.FeatureInfo;
-import org.springframework.cloud.dataflow.rest.resource.about.MonitoringDashboardInfo;
-import org.springframework.cloud.dataflow.rest.resource.about.MonitoringDashboardType;
-import org.springframework.cloud.dataflow.rest.resource.about.RuntimeEnvironment;
-import org.springframework.cloud.dataflow.rest.resource.about.RuntimeEnvironmentDetails;
-import org.springframework.cloud.dataflow.rest.resource.about.SecurityInfo;
-import org.springframework.cloud.dataflow.rest.resource.about.VersionInfo;
+import org.springframework.cloud.dataflow.rest.resource.about.*;
```
onobc, line 31: Please replace this w/ the specific imports. There is a setting in IDEA where you can disable this wildard imports.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AboutController.java
```diff
@@ -15,28 +15,20 @@
  */
 package org.springframework.cloud.dataflow.server.controller;
 
-import java.util.ArrayList;
```
onobc, line 18: Please move the j`ava.util.*` back up to here. The formatter will bark about this. The [code style instructions](https://github.com/spring-cloud/spring-cloud-dataflow/wiki/Working-with-the-Code#code-formatting) are in the Wiki as well - you may want to set those up.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AboutController.java
```diff
@@ -101,14 +98,21 @@ public class AboutController {
 
 	private DataflowMetricsProperties dataflowMetricsProperties;
 
+	private ObjectProvider<GitInfoContributor> gitInfoContributor;
+
+	private ObjectProvider<BuildInfoContributor> buildInfoContributor;
+
 	public AboutController(StreamDeployer streamDeployer, LauncherRepository launcherRepository, FeaturesProperties featuresProperties,
-			VersionInfoProperties versionInfoProperties, SecurityStateBean securityStateBean, DataflowMetricsProperties monitoringProperties) {
+			VersionInfoProperties versionInfoProperties, SecurityStateBean securityStateBean, DataflowMetricsProperties monitoringProperties,
```
onobc, line 106: Even though nobody is probably creating their own AboutController, technically this is a breaking API change. To be safe and to keep in good habits lets instead mark the current constructor deprecated and then create a new one that also includes the 2 new fields. Wdyt?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AboutController.java
```diff
@@ -315,4 +321,14 @@ else if (version.contains(".RC")) {
 		}
 		return result;
 	}
+
+	private Map getGitInfoFromEndpoints() {
```
onobc, line 325: I think this method is still named from the original impl when you were using the endpoints. I think it may be more clear if we name it something like "addGitAndBuildInfoIfAvailable" .. that way we only call the setter if there was actually info available. 

```java
	private void addGitAndBuildInfoIfAvailable(AboutResource aboutResource) {
		Info.Builder builder = new Info.Builder();
		gitInfoContributor.ifAvailable(c -> c.contribute(builder));
		buildInfoContributor.ifAvailable(c -> c.contribute(builder));
		Map<String, Object> details = builder.build().getDetails();
		if (!ObjectUtils.isEmpty(details)) {
			aboutResource.setGitAndBuildInfo(details);
		}
	}
```

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/AboutControllerTests.java
```diff
@@ -37,9 +38,7 @@
 import org.springframework.web.client.ResourceAccessException;
 import org.springframework.web.context.WebApplicationContext;
 
-import static org.hamcrest.Matchers.is;
-import static org.hamcrest.Matchers.emptyOrNullString;
-import static org.hamcrest.Matchers.not;
+import static org.hamcrest.Matchers.*;
```
onobc, line 41: Ditto (wildcard imports)

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/AboutControllerTests.java
```diff
@@ -49,8 +48,7 @@
 /**
  * @author Glenn Renfro
  */
-@ExtendWith(SpringExtension.class)
-@SpringBootTest(classes = TestDependencies.class)
+@SpringBootTest(classes = {TestDependencies.class, ProjectInfoAutoConfiguration.class, InfoContributorAutoConfiguration.class},properties = {"management.info.git.enabled=true"})
```
onobc, line 51: I did not look at this test close enough earlier, but I think lets remove the info auto configs from the top-level SBT and instead focus on these new features in the `AboutTests` at the very bottom.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/AboutControllerTests.java
```diff
@@ -343,7 +345,7 @@ public void testChecksumNoDefaults() throws Exception {
 	}
 
 	@ExtendWith(SpringExtension.class)
-	@SpringBootTest(classes = TestDependencies.class)
+	@SpringBootTest(classes = {TestDependencies.class, ProjectInfoAutoConfiguration.class, InfoContributorAutoConfiguration.class},properties = {"management.info.git.enabled=true"})
```
onobc, line 348: **[TinyNit]** spacing before/after the `{` and `}` and put the `properties` on its own line.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/AboutControllerTests.java
```diff
@@ -398,6 +400,8 @@ public void testAbout() throws Exception {
 					.andExpect(jsonPath("$.runtimeEnvironment.appDeployer.deployerName", is("skipper server")))
 					.andExpect(jsonPath("$.monitoringDashboardInfo.url", is("http://localhost:3001")))
 					.andExpect(jsonPath("$.monitoringDashboardInfo.dashboardType", is("GRAFANA")))
+					.andExpect(jsonPath("$.gitAndBuildInfo.git.commit").exists() )
```
onobc, line 403: Does `exists()` fail if the value is empty on the field? Maybe instead we can use a `hasText` or `notEmpty` matcher. Wdyt?

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/AboutControllerTests.java
```diff
@@ -398,6 +400,8 @@ public void testAbout() throws Exception {
...
+					.andExpect(jsonPath("$.gitAndBuildInfo.build.version").exists() )
 					.andExpect(jsonPath("$.monitoringDashboardInfo.refreshInterval", is(30)));
```
onobc, line 405: I think it would be good to add tests for the following cases:
- when no contributor info then no git and build info in the results (ie `doesNotExists()`)
- when only git contributor then only git info
- when only build contributor then only build info

spring-cloud-dataflow-server/pom.xml
```diff
@@ -296,6 +302,27 @@
 					</execution>
 				</executions>
 			</plugin>
+
+			<plugin>
```
onobc, line 306: Do we need these in scdf-server since they are already in scdf-server-core?

onobc (changes requested):
@felipeg48 great work! Thanks for the quick turnaround. I have requested some changes but they are of the minor and nit variety.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AboutController.java
felipeg48: Yes... I forgot about that, I've just changed that option... soon a new commit.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AboutController.java
felipeg48: Thanks... done...

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AboutController.java
felipeg48: Perfect! done!

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AboutController.java
felipeg48: Nice! done!

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/AboutControllerTests.java
felipeg48: Done!

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/AboutControllerTests.java
felipeg48: yes... better. Done

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/AboutControllerTests.java
felipeg48: Can we discuss this part?? 
- The git contributor will be missing if this code doesn’t have any `.git` folder or the `git-commit-id-maven-plugin` plugin is not present, and due this project has both, probably doesn’t make sense to have something that evaluates to none existing.
- The same for the build info.

**NOTE**: The extreme case will be when we manual delete both and test for that.... but they will be re-generated regardless.

WDYT??

spring-cloud-dataflow-server-core/pom.xml
felipeg48: Not really, I'll remove them...
Done

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/AboutControllerTests.java
onobc: The tests are mostly fail safes to make sure if actuator is excluded or the user disables eitehr of the contributors via props that we don't 💣 . 

I don't mind adding these other cases during a polish commit either though. Let's stick w/ a single test now and I will do that. No worries.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowControllerAutoConfiguration.java
```diff
@@ -38,6 +38,8 @@
 import org.springframework.cloud.common.security.core.support.OAuth2AccessTokenProvidingClientHttpRequestInterceptor;
 import org.springframework.cloud.common.security.core.support.OAuth2TokenUtilsService;
 import org.springframework.cloud.common.security.support.SecurityStateBean;
+import org.springframework.cloud.dataflow.aggregate.task.AggregateExecutionSupport;
```
onobc, line 41: These imports are still out of order. I would expect to see only green lines for the newly introduced contributor clasess.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AboutController.java
```diff
@@ -101,14 +98,32 @@ public class AboutController {
...
+		this.featuresProperties = featuresProperties;
+		this.versionInfoProperties = versionInfoProperties;
+		this.securityStateBean = securityStateBean;
+		this.dataflowMetricsProperties = monitoringProperties;
+	}
+
 	public AboutController(StreamDeployer streamDeployer, LauncherRepository launcherRepository, FeaturesProperties featuresProperties,
-			VersionInfoProperties versionInfoProperties, SecurityStateBean securityStateBean, DataflowMetricsProperties monitoringProperties) {
+			VersionInfoProperties versionInfoProperties, SecurityStateBean securityStateBean, DataflowMetricsProperties monitoringProperties,
+						   ObjectProvider<GitInfoContributor> gitInfoContributor,ObjectProvider<BuildInfoContributor> buildInfoContributor) {
```
onobc, line 118: **[TinyNit]** spacing

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowControllerAutoConfiguration.java
felipeg48: Lesson learned!! NEVER do `Optimize Imports` from IntelliJ, it messes up everything

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/AboutControllerTests.java
```diff
@@ -98,6 +99,10 @@ public void testListApplications() throws Exception {
 				.andExpect(jsonPath("$.featureInfo.streamsEnabled", is(true)))
 				.andExpect(jsonPath("$.featureInfo.tasksEnabled", is(true)))
 				.andExpect(jsonPath("$.featureInfo.schedulesEnabled", is(false)))
+
+				.andExpect(jsonPath("$.gitAndBuildInfo.git.commit").exists() )
```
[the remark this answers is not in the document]
felipeg48: Done!

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowControllerAutoConfiguration.java
onobc: Lol. Yeh, unless you have the code styles setup you do not want to do this.

onobc:
Great work @felipeg48 !! Thank you. 

Closed via https://github.com/spring-cloud/spring-cloud-dataflow/commit/6d729af01dcd31ab5adf9cbd6124a6cd22e9e15a

PR #5553, opened 2023-11-14
Add support for Job instance and execution dao tests [Do not merge till 5577 is merged]

cppwfs:
Updated MARIADB schema  to use sequences instead of tables for Boot 3 batch tables


spring-cloud-dataflow-server-core/src/main/resources/schemas/mariadb/V6-dataflow.sql
corneil, line 1: We should drop SEQ tables if they exists. We should also restart the sequence with the max from the table.

cppwfs: I can create a story to add this as a feature to our schemas.   WDYT?

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/batch/AbstractDaoTests.java
```diff
@@ -0,0 +1,71 @@
...
+		JdbcDatabaseDelegate containerDelegate = new JdbcDatabaseDelegate(dbContainer, "");
+		ScriptUtils.runInitScript(containerDelegate, "schemas/drop-table-schema-" + schemaName + ".sql");
+
+		getResourceFiles("schemas/" + schemaName).forEach(str -> {
+			if (str.contains("dataflow"))
+				ScriptUtils.runInitScript(containerDelegate, "schemas/" + schemaName + "/" + str);
+		});
+	}
+
+	private List<String> getResourceFiles(String path) throws IOException {
```
onobc, line 52: It feels like there should be some "utils" lib that would already do this for us.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/batch/AbstractJdbcJobSearchableExecutionDaoTests.java
```diff
@@ -0,0 +1,210 @@
...
+
+	protected JdbcTemplate jdbcTemplate;
+
+	protected TaskRepository taskRepository;
+
+	protected JdbcTaskBatchDao jdbcTaskBatchDao;
+
+	public void setupSearchableExecutionDaoTest(JdbcDatabaseContainer dbContainer, String schemaName,
+												DatabaseType databaseType) throws Exception {
+		this.dataSource = createDataSourceForContainer(dbContainer);
```
onobc, line 66: **[This Comment Applies to all concrete impls]**

The pattern I am seeing is that a concrete subclass provides a testcontainer and then calls a base method via  `@BeforeEach` that initializes the DB to clean slate before running each test. 

The current impl requires each concrete test to do these things:
```java
this.dataSource = createDataSourceForContainer(dbContainer);
this.jdbcTemplate = new JdbcTemplate(this.dataSource);
createDataFlowSchema(dbContainer, schemaName);
```

I propose that we pull up the DS and JDBCT to the `AbstractDaoTests` and then provide a base method called `prepareForTest` that can be called by the concrete tests. Something like this: 
```java
protected void prepareForTest(JdbcDatabaseContainer dbContainer, String schemaName) {
	this.dataSource = createDataSourceForContainer(dbContainer);
	this.jdbcTemplate = new JdbcTemplate(this.dataSource);
	createDataFlowSchema(dbContainer, schemaName);
}
```
Otherwise, all the ADT is really providing is 2 util methods that everyone has to call.

Wdyt?

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/batch/AbstractJdbcJobSearchableExecutionDaoTests.java
```diff
@@ -0,0 +1,210 @@
...
+		this.jobExecutionDao.setJdbcTemplate(new JdbcTemplate(this.dataSource));
+		jobExecutionDao.afterPropertiesSet();
+		this.stepExecutionDao = new JdbcStepExecutionDao();
+		this.stepExecutionDao.setJdbcTemplate(this.jdbcTemplate);
+		this.stepExecutionDao.setStepExecutionIncrementer(incrementerFactory.getIncrementer(databaseType.name(),
+			AbstractJdbcBatchMetadataDao.DEFAULT_TABLE_PREFIX + "STEP_EXECUTION_SEQ"));
+		this.stepExecutionDao.afterPropertiesSet();
+	}
+
+	@Test
```
onobc, line 91: Since these are newly added tests and using Junit5 let's take advantage of "package-protected" visibility modifiers on the test methods ("less is more" principle). Also, the names no longer require the annoying "test" prefix and instead can express more clearly what the test is doing. One naming format that seems to work well is `shouldDo<thing>When<condition>` or even just dropping the `test` prefix reads better IMO. Wdyt?

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/batch/JdbcJobSearchableExecutionMariadbDaoTests.java
```diff
@@ -0,0 +1,36 @@
...
+
+@Testcontainers
+public class JdbcJobSearchableExecutionMariadbDaoTests extends AbstractJdbcJobSearchableExecutionDaoTests {
+
+	@Container
+	private static final JdbcDatabaseContainer dbContainer = new MariaDBContainer("mariadb:10.9.3");
+
+	@BeforeEach
+	public void setup() throws Exception {
+		setupSearchableExecutionDaoTest(dbContainer, "mariadb", DatabaseType.MYSQL);
```
onobc, line 34: Is the `MYSQL` a mistake here? If not, please explain and then add that as a comment too :)

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/batch/AbstractJdbcJobSearchableInstanceDaoTests.java
```diff
@@ -0,0 +1,49 @@
...
+		createDataFlowSchema(dbContainer, schemaName);
+		jdbcSearchableJobInstanceDao = new JdbcSearchableJobInstanceDao();
+		jdbcSearchableJobInstanceDao.setJdbcTemplate(new JdbcTemplate(this.dataSource));
+		jdbcSearchableJobInstanceDao.afterPropertiesSet();
+	}
+
+	@Test
+	public void testCountJobInstances() {
+		assertThat(jdbcSearchableJobInstanceDao.countJobInstances(BASE_JOB_INST_NAME)).isEqualTo(0);
+		jdbcSearchableJobInstanceDao.createJobInstance(BASE_JOB_INST_NAME, new JobParameters());
```
onobc, line 46: What tests verify the `createJobInstance` API?

spring-cloud-dataflow-server-core/src/main/resources/schemas/mariadb/V6-dataflow.sql
onobc: I agree w/ @cppwfs and I think these changes need to be moved to a `V10-dataflow.sql` which would include the drop of the sequence tables and the add of the sequences.

onobc: @cppwfs at this point I would be in favor of excluding these changes from your PR and moving it to its own issue/PR.

cppwfs: I understand the scope creep.  I'll submit another PR with these changes.   Without them, the tests will fail.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/batch/JdbcJobSearchableExecutionMariadbDaoTests.java
onobc: > Until we migrate the Batch JobExecution API's to Boot 3.x/Batch 5.x

What ticket is that under? Do you mind adding a comment to this effect as well? 

Thanks

onobc:
Awesome work @cppwfs ! 

Closing via https://github.com/spring-cloud/spring-cloud-dataflow/commit/c98116a5d6feb01ccaa2ae9df6f69a66dfde4793

Issue #5560, opened 2023-11-23
Bump Spring Boot to 2.7.18


corneil:

ilayaperumalg:
Marking this as complete.

PR #5562, opened 2023-11-23
Bump spring boot 2.17.18

corneil:
#5560

### DIFF
--- a/spring-cloud-dataflow-parent/pom.xml
+++ b/spring-cloud-dataflow-parent/pom.xml
@@ -18,7 +18,7 @@
 		<javadoc.opts>-Xdoclint:none</javadoc.opts>
 		<maven-resources-plugin.version>3.3.1</maven-resources-plugin.version>
 		<spring-boot.version>2.7.18</spring-boot.version>
-		<spring.version>5.3.30</spring.version>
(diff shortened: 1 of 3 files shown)


spring-cloud-dataflow-parent/pom.xml
```diff
@@ -17,8 +17,8 @@
 		<java.version>1.8</java.version>
 		<javadoc.opts>-Xdoclint:none</javadoc.opts>
 		<maven-resources-plugin.version>3.3.1</maven-resources-plugin.version>
-		<spring-boot.version>2.7.17</spring-boot.version>
-		<spring.version>5.3.30</spring.version>
+		<spring-boot.version>2.7.18</spring-boot.version>
+		<spring.version>5.3.31</spring.version>
```
ilayaperumalg, line 21: Going through [this](https://github.com/spring-cloud/spring-cloud-dataflow/commit/d3379755f2536272fe9202c344f68636a8b16df9) change, I wonder if other dependencies need to be updated along with this?

ilayaperumalg: @corneil Merging this change for now with the issue created to track other dependencies [here](https://github.com/spring-cloud/spring-cloud-dataflow/issues/5564)

Issue #5571, opened 2023-11-28
JobTemplate.executionsLink uses wrong endpoint


Clemens-L:
**Description:**
`JobTemplate.executionsLink` is initialized using

https://github.com/spring-cloud/spring-cloud-dataflow/blob/d22a119c56c85d5f12e4374fd3c0b7353e79926a/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/JobTemplate.java#L96

while it should be 

```java
this.executionsLink = resources.getLink(EXECUTIONS_RELATION).get();
```

When calling `dataFlowOperations.jobOperations().executionList()` this leads to the error `org.springframework.cloud.dataflow.rest.client.DataFlowClientException: Failed to convert value of type 'java.lang.String' to required type 'long'; nested exception is java.lang.NumberFormatException: For input string: "{id}"`, as the wrong endpoint is used.

**Release versions:**

Bug is present in 2.11.1.

**Steps to reproduce:**

Try to call  `dataFlowOperations.jobOperations().executionList()` using the Dataflow REST client.

PR #5573, opened 2023-11-28
Fix the HATEOAS reference.

corneil:
#5571


onobc (approved):
LGTM, thanks @corneil

PR #5579, opened 2023-11-29
Support the use of LogCacheClient when retrieve logs from CF apps [DO NOT MERGE]

cppwfs:
This contains the code to use LogCacheClient when using the CloudFoudnryReleaseManager.

This code contains heavy logging and we still need to determine which guid is approprate for the java cf client


cppwfs:
Closing this PR in that the deployer is responsible for retrieving log information.

PR #5580, opened 2023-11-30
Configure Surefire to run Junit4 and Junit5 tests

onobc:
Several modules were configured to only run Junit 4 tests. As such, many tests were out-of-date. This commit adjusts to run both Junit 4 and 5 test as well as fix any broken/bit-rot tests.

### DIFF
--- a/spring-cloud-dataflow-server-core/pom.xml
+++ b/spring-cloud-dataflow-server-core/pom.xml
@@ -171,22 +171,12 @@
 			<scope>provided</scope>
 		</dependency>
 		<dependency>
-			<groupId>org.springframework.boot</groupId>
-			<artifactId>spring-boot-starter-test</artifactId>
-			<scope>test</scope>
-		</dependency>
-		<dependency>
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-task-core</artifactId>
 		</dependency>
 		<dependency>
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-task-batch</artifactId>
-		</dependency>
-		<dependency>
-			<groupId>org.skyscreamer</groupId>
-			<artifactId>jsonassert</artifactId>
-			<scope>test</scope>
 		</dependency>
 		<dependency>
 			<groupId>org.mariadb.jdbc</groupId>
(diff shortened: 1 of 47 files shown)


spring-cloud-dataflow-server-core/pom.xml
```diff
@@ -183,11 +178,6 @@
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-task-batch</artifactId>
 		</dependency>
-		<dependency>
```
cppwfs, line 186: I see we still use it, are we picking this up transitively from somewhere else?

spring-cloud-dataflow-server/src/test/java/org/springframework/cloud/dataflow/server/db/migration/MySQL57SmokeTest.java
```diff
@@ -26,6 +27,7 @@
  *
  * @author Corneil du Plessis
  */
+@Disabled("Will fix once PR is merged to run all tests")
```
cppwfs, line 30: Assumed it is this PR that is merged before activating the smoke tests.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/AppRegistryControllerTests.java
```diff
@@ -425,15 +425,15 @@ public void testListSingleApplication() throws Exception {
...
+				.andExpect(jsonPath("$.options[*]", hasSize(7)));
 	}
 
 	@Test
 	public void testListSingleApplicationExhaustive() throws Exception {
 		mockMvc.perform(get("/apps/source/time?exhaustive=true").accept(MediaType.APPLICATION_JSON))
 				.andExpect(status().isOk()).andExpect(jsonPath("name", is("time")))
 				.andExpect(jsonPath("type", is("source")))
-				.andExpect(jsonPath("$.options[*]", hasSize(2001)));
+				.andExpect(jsonPath("$.options[*]", hasSize(2022)));
```
cppwfs, line 436: Was this test not being executed and thus when the tests were updated to use the latest time or log their parameter count had changed?

cppwfs:
LGTM Great Work!!!!

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/AppRegistryControllerTests.java
onobc: Exactly... The more recent app has new properties (new version of boot and poller etc..)

spring-cloud-dataflow-server-core/pom.xml
onobc: `spring-boot-starter-test`

Issue #5520, opened 2023-10-19
Stopping a batch with Spring Boot 3/Spring Batch 5 doesn't work properly.


khaeghar:
**Description:**
When using the STOP function of SCDF, the SCDF fails with the following message:

```
2023-10-19 09:25:33.460 INFO [http-nio-8081-exec-6] [org.springframework.cloud.dataflow.server.service.impl.DefaultTaskJobService] getJobExecution:59:boot3
2023-10-19 09:25:48.931 WARN [http-nio-8081-exec-1] [org.springframework.cloud.dataflow.configuration.metadata.BootApplicationConfigurationMetadataResolver] Failed to retrieve properties for resource Docker Resource [docker:.../.../apabatch:] because of ContainerRegistryException: Could not find an Registry Configuration for: registry.san.dev.weu.azure.paas.cloudcenter.corp
2023-10-19 09:25:48.931 WARN [http-nio-8081-exec-1] [org.springframework.cloud.dataflow.configuration.metadata.BootApplicationConfigurationMetadataResolver] Failed to retrieve port names for resource Docker Resource [docker:.../.../apabatch] because of ContainerRegistryException: Could not find an Registry Configuration for: registry.san.dev.weu.azure.paas.cloudcenter.corp
2023-10-19 09:25:48.932 WARN [http-nio-8081-exec-1] [org.springframework.cloud.dataflow.configuration.metadata.BootApplicationConfigurationMetadataResolver] Failed to retrieve port names for resource Docker Resource [docker:.../.../apabatch] because of ContainerRegistryException: Could not find an Registry Configuration for: registry.san.dev.weu.azure.paas.cloudcenter.corp
2023-10-19 09:25:48.941 ERROR [http-nio-8081-exec-3] [org.springframework.cloud.dataflow.server.controller.RestControllerAdvice] Caught exception while handling a request
org.springframework.jdbc.BadSqlGrammarException: PreparedStatementCallback; bad SQL grammar [SELECT JOB_EXECUTION_ID, KEY_NAME, TYPE_CD, STRING_VAL, DATE_VAL, LONG_VAL, DOUBLE_VAL, IDENTIFYING from BOOT3_BATCH_JOB_EXECUTION_PARAMS where JOB_EXECUTION_ID = ?]; nested exception is org.postgresql.util.PSQLException: ERROR: column "key_name" does not exist
Position: 26
[...]

Issue #5533, opened 2023-10-31
Add Boot3 Support for SimpleJobService


cppwfs:
SimpleJobService needs to support Boot 2.x and 3.x applications.

PR #5581, opened 2023-11-30
Support stopping Boot3 jobs

cppwfs:
Add Boot Schema specific APIs to SimpleJobService

Tests use new DAO infrastructure for setting up DB Add Boot Schema specific APIs to SimpleJobService


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/AggregateJobQueryDao.java
```diff
@@ -54,4 +56,7 @@ public interface AggregateJobQueryDao {
 	JobInstanceExecutions getJobInstanceExecution(String jobName, long instanceId);
 
 	JobInstanceExecutions getJobInstanceExecutions(long id, String schemaTarget);
+
+	JobInstance getJobInstance(long id, String schemaTarget) throws NoSuchJobInstanceException;
```
onobc, line 60: I know we have passed the `schemaTarget` as string so far but I think it would be more robust and readable to simply pass in the `AppBootSchemaVersion`.  This became apparent to me when I saw the conversion to `"boot2|3"` in `SimpleJobService`.  As a user of the DAO how do I know what value to pass in here? The enum controls this.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
```diff
@@ -83,17 +85,25 @@ public class SimpleJobService implements JobService, DisposableBean {
 
 	private JobOperator jsrJobOperator;
 
+	private AggregateJobQueryDao aggregateJobQueryDao;
```
onobc, line 88: Since this an `<init>` param, it makes sense to make it immutable w/ `final`.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
```diff
@@ -336,17 +346,13 @@ public int countJobInstances(String name) {
 
 	@Override
 	public JobExecution getJobExecution(Long jobExecutionId) throws NoSuchJobExecutionException {
-		JobExecution jobExecution = jobExecutionDao.getJobExecution(jobExecutionId);
-		if (jobExecution == null) {
-			throw new NoSuchJobExecutionException("There is no JobExecution with id=" + jobExecutionId);
-		}
-		jobExecution.setJobInstance(Objects.requireNonNull(jobInstanceDao.getJobInstance(jobExecution)));
+		JobExecution jobExecution =  this.aggregateJobQueryDao.getJobExecution(jobExecutionId, this.schemaTarget).getJobExecution();
+		jobExecution.setJobInstance(Objects.requireNonNull(this.jobInstanceDao.getJobInstance(jobExecution)));
```
onobc, line 350: I think you want to use `this.getJobInstance` here instead of `this.jobIinstanceDao.getJobInstance` as that uses the newly added aggregate API.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/JdbcAggregateJobQueryDao.java
```diff
@@ -665,6 +681,21 @@ private List<TaskJobExecution> getTaskJobExecutionsByDate(Date startDate, Date e
 				endDate
 		);
 	}
+	private class JobInstanceExtractor implements ResultSetExtractor<List<JobInstance>> {
+
+		@Override
```
onobc, line 686: I think this can be simplified a bit, something like...
```java
		@Override
		public List<JobInstance> extractData(ResultSet rs) throws SQLException, DataAccessException {
			List<JobInstance> jobInstances = new ArrayList();
			while (rs.next()) {
				jobInstances.add(new JobInstance(rs.getLong("JOB_INSTANCE_ID"), rs.getString("JOB_NAME")));
			}
			return jobInstances;
		}
```
Wdyt?

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/batch/AbstractJdbcAggregateJobQueryDaoTests.java
```diff
@@ -0,0 +1,91 @@
...
+			this.jobServiceContainer);
+		jdbcSearchableJobInstanceDao = new JdbcSearchableJobInstanceDao();
+		jdbcSearchableJobInstanceDao.setJdbcTemplate(new JdbcTemplate(super.getDataSource()));
+		incrementerFactory = new MultiSchemaIncrementerFactory(super.getDataSource());
+		this.databaseType = databaseType;
+	}
+
+	@Test
+	void getJobInstancesForBoot3AndBoot2Instances() throws Exception {
+		assertThat(this.jdbcAggregateJobQueryDao.getJobInstance(1, "boot2")).isNull();
```
onobc, line 65: Another example of how could clear up the API, this would become
```java
assertThat(this.jdbcAggregateJobQueryDao.getJobInstance(1,  AppBootSchemaVersion.BOOT2)).isNull();
```

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/batch/AbstractJdbcAggregateJobQueryDaoTests.java
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	private void verifyJobInstance(long id, String schemaTarget, String suffix) throws Exception{
+		JobInstance  jobInstance = this.jdbcAggregateJobQueryDao.getJobInstance(id, schemaTarget);
+		assertThat(jobInstance).isNotNull();
+		assertThat(jobInstance.getJobName()).isEqualTo(BASE_JOB_INST_NAME + suffix );
+	}
+
+	private JobInstance createJobInstance(String suffix, AppBootSchemaVersion appBootSchemaVersion) {
+		String prefix = AbstractJdbcBatchMetadataDao.DEFAULT_TABLE_PREFIX;
```
onobc, line 80: It might be helpful to retrieve a `SchemaVersionTarget` from `org.springframework.cloud.dataflow.schema.service.impl.DefaultSchemaService#getTarget` which then includes the prefixes.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/batch/JdbcAggregateJobQueryMariadbDaoTests.java
```diff
@@ -0,0 +1,38 @@
...
+
+@Testcontainers
+public class JdbcAggregateJobQueryMariadbDaoTests extends AbstractJdbcAggregateJobQueryDaoTests{
+
+	@Container
+	private static final JdbcDatabaseContainer dbContainer = new MariaDBContainer("mariadb:10.9.3");
+
+	@BeforeEach
+	void prepareForTest() throws Exception {
+		super.prepareForTest(dbContainer, "mariadb", DatabaseType.MYSQL);
```
onobc, line 36: So I figured out why this has to be MYSQL for MariaDB incrementer factory... it is because on older version of MariaDB, the sequence is a table and the `org.springframework.cloud.dataflow.core.database.support.MultiSchemaIncrementerFactory#getIncrementerType` therefore will treat it as such and ultimately drop into the old incrementer factory which knows nothing about MariaDB and therefore must be MYSQL. 

I have created the 
```java
	protected String determineDatabaseType(DatabaseType databaseType) {
		return (databaseType != DatabaseType.MARIADB) ? databaseType.name() : DatabaseType.MYSQL.name();
	}
```
in the other abstract sibling `AbstractJdbcJobSearchableExecutionDaoTests`, do you mind pushing that function up to `AbstractDaoTests` and using it instead?
```java
super.prepareForTest(dbContainer, "mariadb", determineDatabaseType(DatabaseType.MARIADB));
```

onobc (changes requested):
Nice work @cppwfs . I have added a few comments to address. Thanks

cppwfs:
Resolves #5520

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
cppwfs: In this case we are using the JobExecutionId to find its `JobInstance`

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
```diff
@@ -412,7 +418,7 @@ public int countStepExecutionsForJobExecution(long jobExecutionId) {
 
 	@Override
 	public JobInstance getJobInstance(long jobInstanceId) throws NoSuchJobInstanceException {
-		JobInstance jobInstance = jobInstanceDao.getJobInstance(jobInstanceId);
+		JobInstance jobInstance = this.aggregateJobQueryDao.getJobInstance(jobInstanceId, this.schemaTarget);
```
corneil, line 421: The `jobInstanceDao` is support to already be one that was selected from the container.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobServiceFactoryBean.java
```diff
@@ -83,10 +88,24 @@ public class SimpleJobServiceFactoryBean implements FactoryBean<JobService>, Ini
 
 	private PlatformTransactionManager transactionManager;
 
+	private JobServiceContainer jobServiceContainer;
+
+	private SchemaService schemaService;
+
+	private AppBootSchemaVersion appBootSchemaVersion;
```
corneil, line 95: The SchemaTarget is better since it may contain different prefixes in the case where we have multiple databases or targets per database.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
```diff
@@ -83,17 +85,25 @@ public class SimpleJobService implements JobService, DisposableBean {
...
+	private AggregateJobQueryDao aggregateJobQueryDao;
+
 	private int shutdownTimeout = DEFAULT_SHUTDOWN_TIMEOUT;
+	String schemaTarget = "boot2";
 
 	public SimpleJobService(SearchableJobInstanceDao jobInstanceDao, SearchableJobExecutionDao jobExecutionDao,
 							SearchableStepExecutionDao stepExecutionDao, JobRepository jobRepository,
-							ExecutionContextDao executionContextDao, JobOperator jsrJobOperator) {
+							ExecutionContextDao executionContextDao, JobOperator jsrJobOperator, AggregateJobQueryDao aggregateJobQueryDao,
+							AppBootSchemaVersion appBootSchemaVersion) {
```
corneil, line 96: The SchemaTarget is known to SimpleJobServiceFactory and should be used my factory to provide all the specifics needed. The SimpleJobService should operate using the provided components.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/AggregateJobQueryDao.java
cppwfs: I'll add this as an issue to 5533, since we have to make this change because other calls use this same pattern.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/batch/AbstractJdbcAggregateJobQueryDaoTests.java
cppwfs: Agreed.   Will resolve this in #5533

onobc (approved):
LGTM @cppwfs - nice work!

I am going to go ahead and merge so that its available in the am. @corneil if there are any changes that you would like to see in this PR, we can get them in a subsequent PR (hope that is ok w/ you).

onobc (approved):
LGTM @cppwfs - thank you.

Issue #5582, opened 2023-12-01
Ensure docker-compose scripts are up-to-date and working 


onobc:
A possible issue was reported in https://github.com/spring-io/dataflow.spring.io/issues/502

> The docker-compose-dood.yml configuration is not installing the Docker CLI as expected. This deviation from the expected behavior is hindering the intended functionality of the Docker Compose setup.

- [x] check the above report and fix if it is broken
- [x] check the other docker compose scripts and make sure they are all up-to-date and working

corneil:
Seems like deployer is trying to read a stdout that isn't where it is expected.

PR #5587, opened 2023-12-04
Improve docker-compose script to handle docker compose and docker-compose.

corneil:
Related to #5582.

### DIFF
--- a/src/docker-compose/dc.sh
+++ b/src/docker-compose/dc.sh
@@ -6,12 +6,13 @@ fi
 SCDIR=$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")
 SCDIR=$(realpath $SCDIR)
 if [ "$1" == "" ]; then
-    echo "Usage is: "${BASH_SOURCE[0]}" <database> <broker> [options] [flags]"
+    echo "Usage is: "${BASH_SOURCE[0]}" <database> <broker> [options] [flags] [--no-wait]"
     echo "Where:"
     echo "    database: is one of mariadb, mysql or postgres"
     echo "    broker: is on of kafka or rabbitmq"
     echo "    options: one or more of ssl, zipkin, prometheus, influxdb, wavefront, ssl, dood, debug-dataflow, debug-skipper"
     echo "    flags: any docker-compose options"
+    echo "    --no-wait: Don't display command line and don't wait for input."
     echo "  This will invoke docker-compose with all the relevant files and provided options as well as the 'up' command"
     exit 1
 fi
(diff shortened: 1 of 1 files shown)


src/docker-compose/dc.sh
```diff
@@ -36,6 +38,16 @@ while [ "$1" != "" ]; do
     "postgres" | "postgresql")
         DATABASE=postgres
         ;;
+    "--no-wait")
+        NO_WAIT=true
+        ;;
+    "down" | "up" | "rm" | "kill" | "stop" | "start" | "run" | "restart" | "pull" | "pause" | "create" | "build")
+        if [ "$DC_CMD" != "" ];then
```
onobc, line 45: Is `DC_CMD` newly added here? Should it be in the usage?

src/docker-compose/dc.sh
```diff
@@ -6,19 +6,21 @@ fi
...
 SCDIR=$(realpath $SCDIR)
 if [ "$1" == "" ]; then
-    echo "Usage is: "${BASH_SOURCE[0]}" <database> <broker> [options] [flags]"
+    echo "Usage is: "${BASH_SOURCE[0]}" <database> <broker> [options] [flags] [--no-wait]"
     echo "Where:"
     echo "    database: is one of mariadb, mysql or postgres"
     echo "    broker: is on of kafka or rabbitmq"
     echo "    options: one or more of ssl, zipkin, prometheus, influxdb, wavefront, ssl, dood, debug-dataflow, debug-skipper"
     echo "    flags: any docker-compose options"
+    echo "    --no-wait: Don't display command line and don't wait for input."
```
onobc, line 15: Should we say what the default is?

onobc (changes requested):
@corneil the changes LGTM but do you mind updating the PR description to indicate how this fixes #5582? Also, you may want to move the commit to "See #5582" since there is another aspect of that issue - to check all of the docker compose scripts (unless you already did that and do in fact want the issue to be closed w/ this PR). Thanks

src/docker-compose/dc.sh
corneil: I decided that waiting for input on all executions is problematic and switched it around by adding a --wait option instead.

corneil:
> @corneil the changes LGTM but do you mind updating the PR description to indicate how this fixes #5582? Also, you may want to move the commit to "See #5582" since there is another aspect of that issue - to check all of the docker compose scripts (unless you already did that and do in fact want the issue to be closed w/ this PR). Thanks

This doesn't resolve the issue, it is only related.
The real issue is the the local launcher is expecting output files for stdout and stderr to exist in the dataflow or skipper container when they are on the host or other container filesystem.

src/docker-compose/dc.sh
onobc: I don't see how to use the script in the usage (the DC_CMD) anyways. Maybe I am missing something.

onobc (approved):
LGTM @corneil - nice work!

PR #5590, opened 2023-12-04
Retrieve logs using externalExecutionId for tasks

cppwfs:
In the past we had to use the task-app-name when retrieving logs.   
This process is no longer necessary as we can use external execution id.
Do not merge this change until https://github.com/spring-cloud/spring-cloud-deployer/pull/413 
has been merged

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
@@ -853,26 +853,6 @@ public class DefaultTaskExecutionService implements TaskExecutionService {
 	 */
 	@Override
 	public String getLog(String platformName, String taskId, String schemaTarget) {
-		Launcher launcher = this.launcherRepository.findByName(platformName);
-		// In case of Cloud Foundry, fetching logs by external execution Id isn't valid as the execution instance is destroyed.
-		// We need to use the task name instead.
-		if (launcher != null && launcher.getType().equals(TaskPlatformFactory.CLOUDFOUNDRY_PLATFORM_TYPE)) {
-			try {
-				TaskDeployment taskDeployment = this.taskDeploymentRepository.findByTaskDeploymentId(taskId);
-				if (taskDeployment == null) {
-					throw new IllegalArgumentException();
-				}
-				String taskName = taskDeployment.getTaskDefinitionName();
-				AggregateTaskExecution taskExecution = this.taskExplorer.getLatestTaskExecutionForTaskName(taskName);
-				if (taskExecution != null && !taskExecution.getExternalExecutionId().equals(taskId)) {
-					return "";
-				}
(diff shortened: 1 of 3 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionService.java
```diff
@@ -853,26 +853,6 @@ private boolean isAppDeploymentSame(TaskManifest previousManifest, TaskManifest
 	 */
 	@Override
 	public String getLog(String platformName, String taskId, String schemaTarget) {
-		Launcher launcher = this.launcherRepository.findByName(platformName);
```
onobc, line 856: I ❤️ all these deletes! 

This does make me realize that we should really firm up the javadocs on `DefaultTaskExecutionService.getLog` to clearly spell out what the taskId is expected to be. Before this change, it was `taskName` but the param is called `taskId`. 
```java
	@Override
	public String getLog(String platformName, String taskId, String schemaTarget) {
```

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskExecutionServiceTests.java
```diff
@@ -1067,9 +1067,11 @@ public void getCFTaskLog() {
 		public void getCFTaskLogByInvalidTaskId() {
 			String platformName = "cf-test-platform";
 			String taskDeploymentId = "12345";
-			this.launcherRepository.save(new Launcher(platformName, TaskPlatformFactory.CLOUDFOUNDRY_PLATFORM_TYPE, taskLauncher));
+			TaskLauncher taskLauncherCF = mock(TaskLauncher.class);
+			when(taskLauncherCF.getLog(any())).thenThrow(new IllegalArgumentException("could not find a GUID app id for the task guid id"));
+			this.launcherRepository.save(new Launcher(platformName, TaskPlatformFactory.CLOUDFOUNDRY_PLATFORM_TYPE, taskLauncherCF));
 			SchemaVersionTarget schemaVersionTarget = SchemaVersionTarget.defaultTarget();
-			assertEquals("Log could not be retrieved as the task instance is not running by the ID: 12345", this.taskExecutionService.getLog(platformName, taskDeploymentId, schemaVersionTarget.getName()));
+			assertEquals("Log could not be retrieved.  Verify that deployments are still available.", this.taskExecutionService.getLog(platformName, taskDeploymentId, schemaVersionTarget.getName()));
```
onobc, line 1074: I have been replacing any `assertEquals` w/ `AssertJ` `assertThat().isEqualTo` as I go through these tests.

onobc (approved):
@cppwfs LGTM. A couple of minor comments but 🟢 . 

One question on the PR description...

> This process is no longer necessary as we can use external execution id.

We "can" or we "must" ?

onobc (approved):
Thanks for the updates @cppwfs

onobc (approved):
LGTM @cppwfs - thanks!

PR #5594, opened 2023-12-06
Await for all audits to be recorded

cppwfs:
AuditRecordControllerTests ocassionally fails on the testRetrieveAuditRecordsFromGivenDateToNull test.

This is because it is expecting 2 audits but ocassionally gets 3 because a create audit for the last entry isn't recorded till after the endTime is recorded.

This commit adds a 30 second max wait for all create audits to complete.

### DIFF
--- a/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/AuditRecordControllerTests.java
+++ b/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/AuditRecordControllerTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2019 the original author or authors.
+ * Copyright 2018-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/AuditRecordControllerTests.java
```diff
@@ -127,6 +129,9 @@ public void setupMocks() throws Exception {
 		mockMvc.perform(post("/streams/definitions/").param("name", "myStream2").param("definition", "time | log")
 				.accept(MediaType.APPLICATION_JSON)).andExpect(status().isCreated());
 
+		Awaitility.await().atMost(Duration.ofMillis(30000)).until(() -> auditRecordRepository.count() == 7);
```
onobc, line 132: Oh my,  a 🪄 #️⃣ ... where does the 7 come from?

cppwfs: Great question.  I added a comment and a descriptive constant explaining where the 🪄 #️⃣ is derived. Basically its 4 app registration audits and 3 stream create audits.   @onobc Great Catch!

onobc (approved):
LGTM - thanks @cppwfs

Issue #5593, opened 2023-12-06
Update logback to handle CVE-2023-6378


onobc:
Need to update logback to deal w/ CVE-2023-6378.

We are on 1.2.12 - fix it to update to 1.2.13. 

**NOTE:** 1.2.13 does have the CVE fix, but Maven is slow to update this reported fact ([background](https://github.com/qos-ch/logback/issues/745)).

PR #5595, opened 2023-12-07
Override logback version to 1.13.14

onobc:
This commit overrides the logback version in order to fix CVE-2023-6378.

See #5593


spring-cloud-dataflow-build/spring-cloud-dataflow-build-dependencies/pom.xml
```diff
@@ -87,13 +88,6 @@
 				<type>pom</type>
 				<scope>import</scope>
 			</dependency>
-			<dependency>
```
onobc, line 90: Since our parent is SBD, having the direct import of SBD does nothing other than removes the ability to do easy version overrides such as https://github.com/spring-cloud/spring-cloud-dataflow/pull/5595/files#diff-7059f95906ccc5bf99ecd1a8702cabced0dc3f24f787c8d32281c624e4f33943R42.

spring-cloud-dataflow-parent/pom.xml
```diff
@@ -113,6 +114,22 @@
 				<artifactId>spring-security-oauth2-client</artifactId>
 				<version>${spring-security.version}</version>
 			</dependency>
+			<!-- Override Logback provided by Spring Boot -->
```
onobc, line 117: Because we do NOT inherit from SBD all we have to do is declare the dep. mgmt. above the SBD dep. mgmt ([background](https://docs.spring.io/spring-boot/docs/2.7.18/maven-plugin/reference/htmlsingle/#using.import)).

corneil (approved):

PR #5596, opened 2023-12-07
Remove use of security starter

jvalkeal:
- Remove spring-cloud-starter-common-security-config-web as a test-jar dependency as its know to cause issues with jdt based ide's.
- skipper/dataflow servers don't even need it and composed task runner can use spring-cloud-common-security-config-web as plain test scope.
- This should make project structure import into vscode work again.


corneil (approved):

onobc (approved):
LGMT @jvalkeal

PR #5599, opened 2023-12-08
Date Tests need to handle delayed audit records

cppwfs:
* Added another pause when setting up the data for the tests.
* Set the date tests to be a bit more robust to make sure that their data is present vs an absolute count check.


onobc (approved):
LGTM - thanks @cppwfs

PR #5604, opened 2023-12-14
Fix display of StepExecutionContext.

corneil:


onobc (approved):
LGTM - thanks @corneil

corneil:
@onobc 
There was an assertion missing in the one controller. 
I notice that repo.spring.io sometime just fails to serve files. A re-run usually fixes it.

onobc (approved):
LGTM

PR #5608, opened 2023-12-18
Use boot3 notation when restarting a batch job

cppwfs:
Following updates made:
1. Now supports Long type for boot3 app job parameters
2. Supports default format for boot 3 job parameters 
3. Added tests for both Boot2 and boot 3 job parameters.

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskJobService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskJobService.java
@@ -45,6 +45,7 @@ import org.springframework.cloud.dataflow.rest.job.JobInstanceExecutions;
 import org.springframework.cloud.dataflow.rest.job.TaskJobExecution;
 import org.springframework.cloud.dataflow.rest.job.support.JobUtils;
 import org.springframework.cloud.dataflow.schema.AggregateTaskExecution;
+import org.springframework.cloud.dataflow.schema.AppBootSchemaVersion;
 import org.springframework.cloud.dataflow.schema.SchemaVersionTarget;
 import org.springframework.cloud.dataflow.server.batch.JobExecutionWithStepCount;
 import org.springframework.cloud.dataflow.server.batch.JobService;
@@ -234,7 +235,7 @@ public class DefaultTaskJobService implements TaskJobService {
 			deploymentProperties.put(DefaultTaskExecutionService.TASK_PLATFORM_NAME, platformName);
 			taskExecutionService.executeTask(taskDefinition.getName(), deploymentProperties,
 					restartExecutionArgs(taskExecution.getArguments(),
-							taskJobExecution.getJobExecution().getJobParameters()));
+							taskJobExecution.getJobExecution().getJobParameters(), schemaTarget));
 		} else {
 			throw new IllegalStateException(String.format("Did not find platform for taskName=[%s] , taskId=[%s]",
 					taskExecution.getTaskName(), taskJobExecution.getTaskId()));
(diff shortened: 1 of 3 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskJobService.java
```diff
@@ -252,8 +253,10 @@ public void restartJobExecution(long jobExecutionId, String schemaTarget) throws
 	 * @return deduped list of arguments that contains the original arguments and any
 	 * identifying job parameters not in the original task execution arguments.
 	 */
-	private List<String> restartExecutionArgs(List<String> taskExecutionArgs, JobParameters jobParameters) {
+	private List<String> restartExecutionArgs(List<String> taskExecutionArgs, JobParameters jobParameters, String schemaTarget) {
 		List<String> result = new ArrayList<>(taskExecutionArgs);
+		final String BOOT3_VERSION = SchemaVersionTarget.createDefault(AppBootSchemaVersion.BOOT3).getName();
```
onobc, line 258: Please remove final. Also, upper case signifies a static constant.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskJobService.java
```diff
@@ -265,9 +268,27 @@ private List<String> restartExecutionArgs(List<String> taskExecutionArgs, JobPar
 					}
 				}
 				if (!existsFlag) {
-					result.add(String.format("%s(%s)=%s", key,
+					String param;
+					if (BOOT3_VERSION.equals(schemaTarget)) {
+						if (JobParameter.ParameterType.LONG.equals(jobParametersMap.get(key).getType())) {
```
onobc, line 273: I assume by this time that `jobParametersMap.get(key)` is always non-null?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskJobService.java
```diff
@@ -265,9 +268,27 @@ private List<String> restartExecutionArgs(List<String> taskExecutionArgs, JobPar
...
+							type = "java.lang.Long";
```
onobc, line 274: Might be less error-prone to use `Long.class.getSimpleName()`

onobc (changes requested):
LGMT @cppwfs - a few minor comments to address. Thanks for the updates.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskJobService.java
cppwfs: Yes.  At line jobParametersMap is a non modifiable map and would have thrown an NPE above this.   Good question

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskJobService.java
cppwfs: Will do.   FYI it is : `Long.class.getCanonicalName()`

onobc: In this case, `getCanonicalName() == getSimpleName()` as it is not an array, inner class, nested, etc..

onobc (approved):

PR #5617, opened 2023-12-19
Adding support for sorting by SchemaTarget.

### LINKED ISSUE #5615: Unable to sort tasks by Schema Target
**Description:**
Using Schema Target as a sort column causes exception.

corneil:
Fixes #5615 and https://github.com/spring-cloud/spring-cloud-dataflow-ui/issues/1948

### DIFF
--- a/spring-cloud-dataflow-aggregate-task/src/main/java/org/springframework/cloud/dataflow/aggregate/task/impl/AggregateDataFlowTaskExecutionQueryDao.java
+++ b/spring-cloud-dataflow-aggregate-task/src/main/java/org/springframework/cloud/dataflow/aggregate/task/impl/AggregateDataFlowTaskExecutionQueryDao.java
@@ -15,24 +15,8 @@
  */
 package org.springframework.cloud.dataflow.aggregate.task.impl;
 
-import java.sql.ResultSet;
-import java.sql.SQLException;
-import java.sql.Types;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Date;
-import java.util.HashSet;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-
-import javax.sql.DataSource;
-
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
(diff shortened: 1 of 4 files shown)


spring-cloud-dataflow-aggregate-task/src/main/java/org/springframework/cloud/dataflow/aggregate/task/impl/AggregateDataFlowTaskExecutionQueryDao.java
```diff
@@ -54,6 +38,20 @@
 import org.springframework.util.CollectionUtils;
 import org.springframework.util.StringUtils;
 
+import javax.sql.DataSource;
```
onobc, line 41: Either the import order was wrong before, or it is now. Please adjust.

onobc (changes requested):
@corneil I think your Intellij formatter is not setup correctly (import order issues in the PR).

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/TaskExecutionControllerTests.java
```diff
@@ -637,6 +637,13 @@ void sorting() throws Exception {
 				.andDo(print())
 				.andExpect(status().isOk());
 
+		mockMvc.perform(get("/tasks/executions").param("sort", "SCHEMA_TARGET").accept(MediaType.APPLICATION_JSON))
```
onobc, line 640: This is good that we have added a controller test for this, but do we have anything that actually tests the query and verifies the sort works properly? That is the part that I am more concerned with.

corneil: Added to `org.springframework.cloud.dataflow.server.repository.TaskExecutionExplorerTests`

corneil:
> @corneil I think your Intellij formatter is not setup correctly (import order issues in the PR).

Installed spring-java-format for IntelliJ.

onobc (approved):
LGTM @corneil - thanks for the updates. Great work!

PR #5619, opened 2023-12-20
Add support for running MSSQL Testcontainers on Mac ARM64

onobc:
This commit adds the ability for Mac ARM64 machines to run MSSQL Testcontainers based tests. The same treatment was added for Oracle previously.

Also, the container support is the final arbiter of container creation and now all of the DB smoke tests are refactored to use the new support interfaces.


onobc:
Reviewed w/ @cppwfs over Zoom.

PR #5645, opened 2024-01-17
⚠[DO NOT MERGE] Base Migration suggestions applied to SCDF ⚠

cppwfs:
This PR is to show the results of the recipes applied by the migrator.   This is only for evaluation.   
* Each commit shows a suggestion applied
* There are some suggestions we need to discuss and those I did not apply.
* Also you will see some the yaml that have been renamed to ymlbak.   The migrator would fail when scanning these files.   
  *  Will re-add them for merge if we decide to use the suggestions..


cppwfs:
this PR has been merged with the main-3 branch.  So closing.

PR #5652, opened 2024-01-25
Make SCDF and Skipper compilable for boot 3.x

cppwfs:
* Replaced SocketUtils with TestSocketUtils 
  * It was moved to the test package.
* Migrated httpclient to httpclient5 ** Removed use of httpclient 4.x dependencies added yesterday
* Updated Types in AuditRecord Entity to use JdbcTypeCode
* `DropAuthorizationHeaderRequestRedirectStrategy` requires a bit more scrutiny.  
    Some of the changes required were not exact matches, so had to research.   But there maybe better ways.   

I disabled spring-cloud-common-security-config to push forward .  It was re-added before PR was submitted.

### DIFF
--- a/spring-cloud-dataflow-configuration-metadata/pom.xml
+++ b/spring-cloud-dataflow-configuration-metadata/pom.xml
@@ -73,8 +73,8 @@
 				<artifactId>maven-compiler-plugin</artifactId>
 				<version>3.11.0</version>
 				<configuration>
-					<source>1.8</source>
-					<target>1.8</target>
+					<source>17</source>
+					<target>17</target>
 				</configuration>
 			</plugin>
 			<plugin>
--- a/spring-cloud-dataflow-configuration-metadata/pom.xml
+++ b/spring-cloud-dataflow-configuration-metadata/pom.xml
@@ -49,7 +49,7 @@
 		</dependency>
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
-			<artifactId>spring-boot-loader</artifactId>
+			<artifactId>spring-boot-loader-classic</artifactId>
 		</dependency>
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
--- a/spring-cloud-dataflow-configuration-metadata/src/test/java/org/springframework/cloud/dataflow/container/registry/authorization/S3SignedRedirectRequestServerResource.java
+++ b/spring-cloud-dataflow-configuration-metadata/src/test/java/org/springframework/cloud/dataflow/container/registry/authorization/S3SignedRedirectRequestServerResource.java
@@ -23,7 +23,7 @@ import org.slf4j.LoggerFactory;
 import org.springframework.boot.builder.SpringApplicationBuilder;
 import org.springframework.cloud.dataflow.container.registry.authorization.support.S3SignedRedirectRequestServerApplication;
 import org.springframework.context.ConfigurableApplicationContext;
-import org.springframework.util.SocketUtils;
+import org.springframework.test.util.TestSocketUtils;
 
 /**
  * @author Adam J. Weigold
@@ -43,7 +43,7 @@ public class S3SignedRedirectRequestServerResource extends ExternalResource {
     @Override
     protected void before() throws Throwable {
 
-        this.s3SignedRedirectServerPort = SocketUtils.findAvailableTcpPort();
+        this.s3SignedRedirectServerPort = TestSocketUtils.findAvailableTcpPort();
 
         logger.info("Setting S3 Signed Redirect Server port to " + this.s3SignedRedirectServerPort);
 
--- a/spring-cloud-dataflow-container-registry/src/main/java/org/springframework/cloud/dataflow/container/registry/ContainerImageRestTemplateFactory.java
+++ b/spring-cloud-dataflow-container-registry/src/main/java/org/springframework/cloud/dataflow/container/registry/ContainerImageRestTemplateFactory.java
@@ -16,6 +16,9 @@
 
 package org.springframework.cloud.dataflow.container.registry;
 
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.TrustManager;
+import javax.net.ssl.X509TrustManager;
 import java.security.KeyManagementException;
 import java.security.NoSuchAlgorithmException;
 import java.security.cert.X509Certificate;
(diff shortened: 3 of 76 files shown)


spring-cloud-dataflow-configuration-metadata/pom.xml
```diff
@@ -49,7 +49,7 @@
 		</dependency>
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
-			<artifactId>spring-boot-loader</artifactId>
+			<artifactId>spring-boot-loader-classic</artifactId>
```
onobc, line 52: What is the significance of this?

spring-cloud-dataflow-configuration-metadata/src/test/java/org/springframework/cloud/dataflow/container/registry/authorization/S3SignedRedirectRequestServerResource.java
```diff
@@ -23,7 +23,7 @@
 import org.springframework.boot.builder.SpringApplicationBuilder;
 import org.springframework.cloud.dataflow.container.registry.authorization.support.S3SignedRedirectRequestServerApplication;
 import org.springframework.context.ConfigurableApplicationContext;
-import org.springframework.util.SocketUtils;
+import org.springframework.test.util.TestSocketUtils;
```
onobc, line 26: 🎉 finally people are using it! 😉

spring-cloud-dataflow-container-registry/src/main/java/org/springframework/cloud/dataflow/container/registry/ContainerImageRestTemplateFactory.java
```diff
@@ -171,19 +177,25 @@ public void checkServerTrusted(java.security.cert.X509Certificate[] certs, Strin
...
 		// Install trust manager to SSL Context.
 		sslContext.init(null, trustAllCerts, new java.security.SecureRandom());
+		ConnectionSocketFactory sslsf = new SSLConnectionSocketFactory(sslContext, NoopHostnameVerifier.INSTANCE);
+
+		Registry<ConnectionSocketFactory> socketFactoryRegistry =
+			RegistryBuilder.<ConnectionSocketFactory> create()
+				.register("https", sslsf)
+				.register("http", new PlainConnectionSocketFactory())
+				.build();
+		final BasicHttpClientConnectionManager connectionManager = new BasicHttpClientConnectionManager(socketFactoryRegistry);
```
onobc, line 187: Please remove `final`.

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/AuditRecord.java
```diff
@@ -58,7 +59,7 @@ public class AuditRecord {
 	private String correlationId;
 
 	@Lob
-	@Type(type = "org.hibernate.type.TextType")
+	@JdbcTypeCode(Types.LONGVARCHAR)
```
onobc, line 62: I know we had a headache around LOB support specifically w/ Postgres. Hopefully the DB ITs will and ITs will ferret this out if it is a problem. The pain w/ the LOB support was that not only was it different across vendors, but also versions of a vendor.

spring-cloud-dataflow-container-registry/src/main/java/org/springframework/cloud/dataflow/container/registry/ContainerImageRestTemplateFactory.java
```diff
@@ -171,19 +177,25 @@ public void checkServerTrusted(java.security.cert.X509Certificate[] certs, Strin
 		SSLContext sslContext = SSLContext.getInstance("SSL");
 		// Install trust manager to SSL Context.
 		sslContext.init(null, trustAllCerts, new java.security.SecureRandom());
+		ConnectionSocketFactory sslsf = new SSLConnectionSocketFactory(sslContext, NoopHostnameVerifier.INSTANCE);
```
onobc, line 180: It took me a bit to figure out what was new here. My final analysis is that we just have to jump through another hoop to construct the required http client builder. 

I reworked this a bit to abstract out the "get http client builder" which hides the new extra unimportant confusing steps.  Wdyt about replacing this chunk w/ the following?:

```suggestion
		// Create a RestTemplate that uses custom request factory
		return initRestTemplate(
				httpClientBuilder(sslContext),
				withHttpProxy,
				extra);
	}

	private HttpClientBuilder httpClientBuilder(SSLContext sslContext) {
		// Register http/s connection factories
		Lookup<ConnectionSocketFactory> connSocketFactoryLookup = RegistryBuilder.<ConnectionSocketFactory> create()
				.register("https", new SSLConnectionSocketFactory(sslContext, NoopHostnameVerifier.INSTANCE))
				.register("http", new PlainConnectionSocketFactory())
				.build();
		return HttpClients.custom()
				.setConnectionManager(new BasicHttpClientConnectionManager(connSocketFactoryLookup));
	}
```

onobc (changes requested):
Nice work @cppwfs ! I left a couple of suggestions, but other than that - LGTM.

spring-cloud-dataflow-core/src/main/java/org/springframework/cloud/dataflow/core/AuditRecord.java
corneil: Running Postgres Smoke Tests locally will quickly identify if this is a valid change.

spring-cloud-dataflow-configuration-metadata/pom.xml
cppwfs: `org.springframework.boot.loader.archive.JarFileArchive` (The one dataflow uses) moved to this dependency.   There is another `org.springframework.boot.loader.launch.JarFileArchive`.  I chose the one that dataflow has used in the past.

onobc (approved):
LGTM - thanks for the updates @cppwfs

spring-cloud-skipper/spring-cloud-skipper-client/src/main/java/org/springframework/cloud/skipper/client/util/HttpClientConfigurer.java
```diff
@@ -75,12 +84,22 @@ public HttpClientConfigurer basicAuthCredentials(String username, String passwor
...
-
+		Lookup<ConnectionSocketFactory> connSocketFactoryLookup = RegistryBuilder.<ConnectionSocketFactory> create()
+			.register("https", new SSLConnectionSocketFactory(HttpUtils.buildCertificateIgnoringSslContext(), NoopHostnameVerifier.INSTANCE))
+			.register("http", new PlainConnectionSocketFactory())
+			.build();
+		httpClientBuilder.setConnectionManager(new BasicHttpClientConnectionManager(connSocketFactoryLookup));
 		return this;
 	}
-
+	private HttpClientBuilder httpClientBuilder(SSLContext sslContext) {
```
onobc, line 94: I don't think this is currently being used in this class.

onobc (approved):
@cppwfs LGTM - one minor suggestion. Other than that 🍏

spring-cloud-skipper/spring-cloud-skipper-client/src/main/java/org/springframework/cloud/skipper/client/util/HttpClientConfigurer.java
cppwfs: Good catch.  I was going to extract the code from a method and make it re-usable.  Then realized that it was only being used there.   Then forgot to remove the new method.

cppwfs:
Squashed and merged into main-3

PR #5653, opened 2024-01-31
Fixed checkstyle errors about unused imports and import orders.

corneil:
Only fixed import order and unused imports.


onobc (approved):
Thanks for the update @corneil - this does update all the things to the correct import order as described in [checkstyle.xml](https://github.com/spring-cloud/spring-cloud-dataflow/blob/1b8edd6b3848354d3fb320bc220a11a6d3a6b402/spring-cloud-dataflow-build/spring-cloud-dataflow-build-tools/src/main/resources/checkstyle.xml#L110). However, really the checkstyle import order needs to be updated to rename `javax` -> `jakarta` otherwise we will have to do this to every single file that we modify for the migration. 

I say go ahead and merge this and then we can come back and do the update in a single commit. I don't mind doing that.

corneil:
Not needed since last PR by Glenn resolved this as well.

PR #5657, opened 2024-02-01
Dataflow needs to be compilable using boot 3.x

cppwfs:
* Batch removed some methods that are still required by dataflow. 
  * Created dataflow version of those classes so that we can implement those methods
* Removed hibernate version from skipper (use boot's bom)
* Updated Skipper so that it is 3.0 compliant. 
* Update classic docs to be restful doc 3.0 compliant
* Disabled CTR, TaskLauncher, SingleStepBatch jobs for this effort.  Will need be addressed when worked on.

* Work to be done once merged:
  * Check to see if versions are up to date in modules
  * Remove versions when possible ( use boot's bom)
  * Deprecations were not updated in this process.  This still needs to be done.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -77,9 +77,9 @@
 		<module>spring-cloud-starter-dataflow-server</module>
 		<module>spring-cloud-starter-dataflow-ui</module>
 		<module>spring-cloud-dataflow-server</module>
-		<module>spring-cloud-dataflow-tasklauncher</module>
-		<module>spring-cloud-dataflow-single-step-batch-job</module>
-		<module>spring-cloud-dataflow-composed-task-runner</module>
+<!--		<module>spring-cloud-dataflow-tasklauncher</module>-->
+<!--		<module>spring-cloud-dataflow-single-step-batch-job</module>-->
+<!--		<module>spring-cloud-dataflow-composed-task-runner</module>-->
 		<module>spring-cloud-dataflow-test</module>
 		<module>spring-cloud-dataflow-dependencies</module>
 		<module>spring-cloud-dataflow-classic-docs</module>
--- a/pom.xml
+++ b/pom.xml
@@ -77,6 +77,7 @@
 		<module>spring-cloud-starter-dataflow-server</module>
 		<module>spring-cloud-starter-dataflow-ui</module>
 		<module>spring-cloud-dataflow-server</module>
+<!--		TODO: Boot3x followup -->
 <!--		<module>spring-cloud-dataflow-tasklauncher</module>-->
 <!--		<module>spring-cloud-dataflow-single-step-batch-job</module>-->
 <!--		<module>spring-cloud-dataflow-composed-task-runner</module>-->
--- a/spring-cloud-dataflow-server-core/pom.xml
+++ b/spring-cloud-dataflow-server-core/pom.xml
@@ -51,6 +51,10 @@
 			<artifactId>hibernate-micrometer</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>org.hibernate.orm</groupId>
+			<artifactId>hibernate-ant</artifactId>
+		</dependency>
+		<dependency>
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-dataflow-schema</artifactId>
 			<version>${dataflow.version}</version>
--- a/spring-cloud-dataflow-server-core/pom.xml
+++ b/spring-cloud-dataflow-server-core/pom.xml
@@ -51,10 +51,6 @@
 			<artifactId>hibernate-micrometer</artifactId>
 		</dependency>
 		<dependency>
-			<groupId>org.hibernate.orm</groupId>
-			<artifactId>hibernate-ant</artifactId>
-		</dependency>
-		<dependency>
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-dataflow-schema</artifactId>
 			<version>${dataflow.version}</version>
@@ -255,6 +251,11 @@
 		<dependency>
 			<groupId>org.testcontainers</groupId>
 			<artifactId>postgresql</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.hibernate.orm</groupId>
+			<artifactId>hibernate-ant</artifactId>
 			<scope>test</scope>
 		</dependency>
 	</dependencies>
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/JobRestartRuntimeException.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/JobRestartRuntimeException.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright 2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.dataflow.server.batch;
+
+public class JobRestartRuntimeException extends RuntimeException {
+
+	public JobRestartRuntimeException(Long jobExecutionId, Exception cause) {
+		super(String.format("JobExecutionId '%d' was not restarted.", jobExecutionId), cause);
+	}
+
+	public JobRestartRuntimeException(Long jobExecutionId) {
+		super(String.format("JobExecutionId '%d' was not restarted.", jobExecutionId));
+	}
+}
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/JobRestartRuntimeException.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/JobRestartRuntimeException.java
@@ -16,6 +16,7 @@
 
 package org.springframework.cloud.dataflow.server.batch;
 
+//TODO: Boot3x followup
 public class JobRestartRuntimeException extends RuntimeException {
 
 	public JobRestartRuntimeException(Long jobExecutionId, Exception cause) {
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
@@ -108,7 +108,8 @@ public class SimpleJobService implements JobService, DisposableBean {
 		this.executionContextDao = executionContextDao;
 		this.aggregateJobQueryDao = aggregateJobQueryDao;
 		this.schemaVersionTarget = schemaVersionTarget;
-		Assert.notNull(this.jobOperator = jobOperator, "jobOperator must not be null");
+		Objects.requireNonNull(jobOperator, "jobOperator must not be null");
+		this.jobOperator = jobOperator;
 	}
 
 	/**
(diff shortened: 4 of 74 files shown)


pom.xml
```diff
@@ -77,9 +77,9 @@
 		<module>spring-cloud-starter-dataflow-server</module>
 		<module>spring-cloud-starter-dataflow-ui</module>
 		<module>spring-cloud-dataflow-server</module>
-		<module>spring-cloud-dataflow-tasklauncher</module>
-		<module>spring-cloud-dataflow-single-step-batch-job</module>
-		<module>spring-cloud-dataflow-composed-task-runner</module>
+<!--		<module>spring-cloud-dataflow-tasklauncher</module>-->
```
onobc, line 80: Do you mind adding a standard boot3 comment here for easy finding later?

spring-cloud-dataflow-server-core/pom.xml
```diff
@@ -50,6 +50,10 @@
 			<groupId>org.hibernate.orm</groupId>
 			<artifactId>hibernate-micrometer</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.hibernate.orm</groupId>
+			<artifactId>hibernate-ant</artifactId>
```
onobc, line 55: Uhhh what is this?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/JobRestartRuntimeException.java
```diff
@@ -0,0 +1,28 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.dataflow.server.batch;
+
+public class JobRestartRuntimeException extends RuntimeException {
```
onobc, line 19: It may be helpful to mark these new temp placeholders w/ a comment or naming prefix in order to easily see what needs to be impled/removed/addressed etc.. wdyt?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
```diff
@@ -162,27 +162,28 @@ public JobExecution restart(Long jobExecutionId, JobParameters params) throws No
...
-				jobExecution = new JobExecution(jsrJobOperator.restart(jobExecutionId, new Properties()));
-			}
-		} else {
-			throw new NoSuchJobException(String.format("Can't find job associated with job execution id %s to restart",
-					String.valueOf(jobExecutionId)));
-		}
+        try {
+            jobExecution = new JobExecution(jobOperator.restart(jobExecutionId.longValue()));
+        }
+		  catch (JobParametersInvalidException | JobRestartException | NoSuchJobExecutionException |
```
onobc, line 168: Are these 4 exceptions of a common type? Should we just `catch (Exception ex)` instead?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
```diff
@@ -162,27 +162,28 @@ public JobExecution restart(Long jobExecutionId, JobParameters params) throws No
...
 	}
 
 	@Override
 	public JobExecution launch(String jobName, JobParameters jobParameters) throws NoSuchJobException {
 		JobExecution jobExecution;
 
-		if (jsrJobOperator != null) {
-			jobExecution = new JobExecution(jsrJobOperator.start(jobName, jobParameters.toProperties()));
-		} else {
+		if (jobOperator != null) {
```
onobc, line 180: **[NIT]** formatting looks wonky

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
```diff
@@ -256,14 +257,19 @@ public int countJobs() {
 	@Override
 	public int stopAll() {
 		Collection<JobExecution> result = jobExecutionDao.getRunningJobExecutions();
-		Collection<String> jsrJobNames = getJsrJobNames();
+		Collection<String> jsrJobNames = getJobNames();
```
onobc, line 260: Do you mind renaming the var to drop the `jsr` as well?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
```diff
@@ -256,14 +257,19 @@ public int countJobs() {
...
 		for (JobExecution jobExecution : result) {
-			if (jsrJobOperator != null && jsrJobNames.contains(jobExecution.getJobInstance().getJobName())) {
-				jsrJobOperator.stop(jobExecution.getId());
-			} else {
-				jobExecution.stop();
-				jobRepository.update(jobExecution);
+			if (jsrJobNames.contains(jobExecution.getJobInstance().getJobName())) {
+                try {
+                    jobOperator.stop(jobExecution.getId());
+                } catch (NoSuchJobExecutionException e) {
```
onobc, line 266: The top 2 catch can be consolidated via an `|` in the catch clause.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
```diff
@@ -256,14 +257,19 @@ public int countJobs() {
...
-		Collection<String> jsrJobNames = getJsrJobNames();
+		Collection<String> jsrJobNames = getJobNames();
 
 		for (JobExecution jobExecution : result) {
-			if (jsrJobOperator != null && jsrJobNames.contains(jobExecution.getJobInstance().getJobName())) {
-				jsrJobOperator.stop(jobExecution.getId());
-			} else {
-				jobExecution.stop();
-				jobRepository.update(jobExecution);
+			if (jsrJobNames.contains(jobExecution.getJobInstance().getJobName())) {
```
onobc, line 263: **[NIT]** I think if we invert this we can short circuit and throw the stop exception rather than have to indent the try/catch deeper.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
```diff
@@ -304,15 +309,15 @@ public JobExecution abandon(Long jobExecutionId) throws NoSuchJobExecutionExcept
 
 		logger.info("Aborting job execution: " + jobExecution);
 
-		Collection<String> jsrJobNames = getJsrJobNames();
+		Collection<String> jsrJobNames = getJobNames();
```
onobc, line 312: **[NIT]** goodbye jsr prefix

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/DataflowSqlPagingQueryUtils.java
```diff
@@ -0,0 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.dataflow.server.repository;
+
+import org.springframework.batch.item.database.support.AbstractSqlPagingQueryProvider;
+
+
+//TODO: Boot3x followup
```
onobc, line 22: 👍🏻

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskJobService.java
```diff
@@ -254,10 +254,12 @@ public void restartJobExecution(long jobExecutionId, String schemaTarget) throws
 	 * identifying job parameters not in the original task execution arguments.
 	 */
 	private List<String> restartExecutionArgs(List<String> taskExecutionArgs, JobParameters jobParameters, String schemaTarget) {
+		if(schemaTarget.equals(SchemaVersionTarget.createDefault(AppBootSchemaVersion.BOOT2).getName())) {
```
onobc, line 257: Would it be safer to invert this and ensure that the `schemaTarget` is boot3?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskJobService.java
```diff
@@ -268,27 +270,8 @@ private List<String> restartExecutionArgs(List<String> taskExecutionArgs, JobPar
...
-								jobParametersMap.get(key).getType() + " to known type of JobParameters");
-						}
-						param = String.format("%s=%s,%s", key, jobParametersMap.get(key).getValue(), type);
-					} else {
-						param = String.format("%s(%s)=%s", key,
-							jobParametersMap.get(key).getType().toString().toLowerCase(),
-							jobParameters.getString(key));
-					}
-					result.add(param);
+					type = jobParametersMap.get(key).getType().getCanonicalName();
```
onobc, line 273: I notice the key was `toLowerCase`ed before, do we still need to do that?

onobc (changes requested):
@cppwfs thanks for knocking this out. All my comments are small in nature.

spring-cloud-dataflow-server-core/pom.xml
cppwfs: This is used by SchemaGenerationTests for the `org.hibernate.tool.hbm2ddl.SchemaExport`.   I changed the scope of this to `test` and moved it under the tests in the dependencies.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskJobService.java
cppwfs: I'll add a boot3 comment.  Because we will be removing the Boot2 code.   But at this moment, until we remove it, I wanted to let folks this was a no no.

onobc: Agreed. It may be a wider net to say "If this is not boot3 BOOM" rather than "if this is boot2 BOOM"... however I think the inputs are constrained to one or the other so it really does not matter. Especially since we will be removing it. Disregard. Thx. :)

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskJobService.java
cppwfs: Lower case was for Batch 4 job params types .  For Boot3 we use the class name as the type for job params

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
cppwfs: Changed it to a catch for `Exception`

spring-cloud-dataflow-server-core/pom.xml
```diff
@@ -257,6 +253,11 @@
 			<artifactId>postgresql</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
```
onobc, line 256: 👍🏻

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
```diff
@@ -108,7 +108,8 @@ public SimpleJobService(SearchableJobInstanceDao jobInstanceDao, SearchableJobEx
 		this.executionContextDao = executionContextDao;
 		this.aggregateJobQueryDao = aggregateJobQueryDao;
 		this.schemaVersionTarget = schemaVersionTarget;
-		Assert.notNull(this.jobOperator = jobOperator, "jobOperator must not be null");
+		Objects.requireNonNull(jobOperator, "jobOperator must not be null");
+		this.jobOperator = jobOperator;
```
onobc, line 112: Even 1 step further yields...
```java
this.jobOperator = Objects.requireNonNull(jobOperator, "jobOperator must not be null");
```

onobc (approved):
Thanks for the updates @cppwfs - LGTM

cppwfs:
Made the last change that @onobc requested.   Squashed, Merged here: https://github.com/spring-cloud/spring-cloud-dataflow/commit/67d3bd1fd5fc712b96965c1e56a190715124c71e

Thanks for all the reviews and help along the way!

PR #5662, opened 2024-02-06
Update scdf common deps (see below)

onobc:
* Remove duplicated maven-source-plugin from `spring-cloud-dataflow-common/**`

* Update Awaitility and Semver libs

This commit removes the dependency management for `com.jayway.awaitility` in
favor of the Spring Boot managed version at the new `org.awaitility`
coordinates.

* Also, the Semver lib is updated to 0.10.2

### DIFF
--- a/spring-cloud-dataflow-common/pom.xml
+++ b/spring-cloud-dataflow-common/pom.xml
@@ -18,8 +18,7 @@
 	</parent>
 
 	<properties>
-		<jayway-awaitility.version>1.7.0</jayway-awaitility.version>
-		<java-semver.version>0.9.0</java-semver.version>
+		<java-semver.version>0.10.2</java-semver.version>
 		<joda-time.version>2.12.7</joda-time.version>
 	</properties>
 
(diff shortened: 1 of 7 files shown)


spring-cloud-dataflow-common/pom.xml
```diff
@@ -18,8 +18,7 @@
 	</parent>
 
 	<properties>
-		<jayway-awaitility.version>1.7.0</jayway-awaitility.version>
```
onobc, line 21: This managed by boot and its group has changed from `com.jayway.awaitility` to `org.awaitility`

spring-cloud-dataflow-common/pom.xml
```diff
@@ -50,19 +44,6 @@
 			</dependency>
 		</dependencies>
 	</dependencyManagement>
-	<build>
```
onobc, line 53: This is all specified in parent builds. I wanted to remove it here first in a small slice and if all goes well move to the rest of SCDF in a separate subsequent code proposal.

cppwfs:
LGTM 
Squash and merge

cppwfs:
Squashed and merged.   Thanks for the great contribution!

PR #5665, opened 2024-02-07
Fix the docker compose tests

onobc:
With these changes the following commands now succeed:

```
cd spring-cloud-dataflow-common
../mvnw clean install 
```

### DIFF
--- a/spring-cloud-dataflow-common/spring-cloud-dataflow-common-test-docker/src/test/java/org/springframework/cloud/dataflow/common/test/docker/compose/execution/CommandTests.java
+++ b/spring-cloud-dataflow-common/spring-cloud-dataflow-common-test-docker/src/test/java/org/springframework/cloud/dataflow/common/test/docker/compose/execution/CommandTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2019 the original author or authors.
+ * Copyright 2018-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


spring-cloud-dataflow-common/spring-cloud-dataflow-common-test-docker/src/test/java/org/springframework/cloud/dataflow/common/test/docker/compose/execution/DockerComposeTests.java
```diff
@@ -148,106 +142,90 @@ public void call_docker_compose_with_the_follow_flag_when_the_version_is_at_leas
...
+                .withMessage("No container with name 'db' found");
     }
 
     @Test
-    public void pass_concatenated_arguments_to_executor_on_docker_compose_exec()
-            throws IOException, InterruptedException {
-        when(executedProcess.getInputStream()).thenReturn(toInputStream("docker-compose version 1.7.0rc1, build 1ad8866"));
-        compose.exec(options("-d"), "container_1", arguments("ls"));
-        verify(executor, times(1)).execute("exec", "-T", "-d", "container_1", "ls");
+    void failIfDockerComposeVersionIsPriorTo17OnDockerComposeExec() {
```
cppwfs, line 190: Nitpick but it threw me reading `PriorTo17`.  In this case we could change the name to `failIfDockerComposeIsNotMinimumVersion`.   Then add a javadoc specifying that we only support 1.7 and above.

spring-cloud-dataflow-common/spring-cloud-dataflow-common-test-docker/src/test/java/org/springframework/cloud/dataflow/common/test/docker/compose/execution/DockerTests.java
```diff
@@ -29,60 +29,53 @@
...
         assertThat(docker.listNetworks(), is(lsOutput));
-
         verify(executor).execute("network", "ls");
     }
 
     @Test
-    public void call_docker_network_prune() throws IOException, InterruptedException {
+    void callDockerNetworkPrune() throws Exception {
         String lsOutput = "0.0.0.0:7000->7000/tcp";
         when(executedProcess.getInputStream()).thenReturn(toInputStream(lsOutput));
```
cppwfs, line 64: Nitpick: Can we add this mock to the prepareForTest since it is also used in the `callDockerRmWithForceFlagOnRm` test?

spring-cloud-dataflow-common/spring-cloud-dataflow-common-test-docker/src/test/java/org/springframework/cloud/dataflow/common/test/docker/compose/execution/CommandTests.java
```diff
@@ -20,88 +20,80 @@
...
+    void returnOutputWhenExitCodeOfTheExecutedProcessIsZero() throws Exception {
         String expectedOutput = "test output";
         givenTheUnderlyingProcessHasOutput(expectedOutput);
         String commandOutput = dockerComposeCommand.execute(errorHandler, "rm", "-f");
-
         assertThat(commandOutput, is(expectedOutput));
     }
 
-    @Test public void
-    give_the_output_to_the_specified_consumer_as_it_is_available() throws IOException, InterruptedException {
```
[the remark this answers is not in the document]
onobc: Because this is a test and nobody is going to do anything w/ the annoying checked exception (nobody cares at this point) it is just clutter to list them out IMO. To me it provides no value - only clutter.

spring-cloud-dataflow-common/spring-cloud-dataflow-common-test-docker/src/test/java/org/springframework/cloud/dataflow/common/test/docker/compose/execution/DockerTests.java
onobc: Sure. I should have added some more description around these method names, but these tests were in very bad shape (I mean snake cases test names in Java LOLWUT?) The remaining of the tests in the module are still like that.

spring-cloud-dataflow-common/spring-cloud-dataflow-common-test-docker/src/test/java/org/springframework/cloud/dataflow/common/test/docker/compose/execution/DockerComposeTests.java
onobc: Sure. But my guess is that this is proliferated all through this module/tests. I would like to delete all of this Docker compose stuff and use something off the shelf that probably now exists.

spring-cloud-dataflow-common/spring-cloud-dataflow-common-test-docker/src/test/java/org/springframework/cloud/dataflow/common/test/docker/compose/execution/DockerTests.java
onobc: Actually, every test does some mocking around this method but the `thenReturn` is different in all except 2 of them. If we pull this up to the the prepare method I think it will be more confusing to follow what each is doing.

spring-cloud-dataflow-common/spring-cloud-dataflow-common-test-docker/src/test/java/org/springframework/cloud/dataflow/common/test/docker/compose/execution/DockerComposeTests.java
onobc: The Javadoc should be somewhere in the DockerCompose code. I don't mind adding that simple thing but I would rather not go back through all this DC code and make it nice - I would prefer spending the time removing it ;)

onobc: So this 1.7 is not a minimum version per-se. You can see elsewhere in the code that 1.7 is fine for most commands. It is only for `docker-compose exec` that this is a minimum version. We really don't care as we just use this for our tests and that is an ancient version.

onobc: Renamed methods though - PTAL.

cppwfs:
LGTM Thanks for the great work on this and the discussion!

cppwfs:
Squashed and Merged

PR #5668, opened 2024-02-08
Use AutoConfiguration.imports for all auto-configs

onobc:


spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/ApplicationConfigurationMetadataResolverAutoConfiguration.java
```diff
@@ -29,6 +30,7 @@
  * @author Eric Bottard
  * @author Christian Tzolov
  */
+@AutoConfiguration
 @Configuration
```
cppwfs, line 34: Should this configuration be removed?
`@AutoConfiguration`'s definition  contains `@Configuration(proxyBeanMethods = false)`.
Unless there is a reason to maintain it.   
I have seen this a couple times in the PR.

spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/config/DataFlowClientAutoConfiguration.java
```diff
@@ -141,7 +142,6 @@ public StreamBuilder streamBuilder(DataFlowOperations dataFlowOperations){
 	}
 
 	@ConditionalOnProperty(prefix = DataFlowPropertyKeys.PREFIX + "client.authentication", name = "client-id")
-	@Configuration
```
cppwfs, line 144: Just curious if this removal was necessary?   Seems like this should not have been removed.   Thoughts?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowControllerAutoConfiguration.java
```diff
@@ -223,7 +223,6 @@ public RestControllerAdvice restControllerAdvice() {
 
 
 
-	@Configuration
```
cppwfs, line 226: Again, I may be incorrect.   Probably am...   But should we remove `@Configuration` for these static classes?

cppwfs (commented):
Looks great, just a couple of nitpicks.   Thanks for digging into this!

spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/ApplicationConfigurationMetadataResolverAutoConfiguration.java
onobc: Yep. Not sure why it was left in by OpenRewrite. I'll yank. Good catch.

spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/config/DataFlowClientAutoConfiguration.java
onobc: This is a Spring Framework 6.x (maybe earlier) best practice. Only top-level config needs to be registered - nested configurations are automatically registered. Look for `With nested @Configuration` in the [@Configuration javadocs](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/annotation/Configuration.html) for more details.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowControllerAutoConfiguration.java
onobc: Ditto https://github.com/spring-cloud/spring-cloud-dataflow/pull/5668/files/7919ace73a567e942b9e0f86bbfa2fca8ca78e27#r1483619438

spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/config/DataFlowClientAutoConfiguration.java
cppwfs: But when I read it, it still shows the `@Configuration` added to the inner static class in the sample  and in the doc, it says it means we don't have to add the `@Import` for the static inner class.   
Or is Glenn just being stubborn.   :-)

onobc: Ahh yeh, that is not the proper doc for this. Let me back up my comment or we can simply remove this. It was put in by OpenRewrite.

cppwfs: I was just making sure rewrite didn't take out the wrong `@Configuration`.

onobc: Updated - added back in.

cppwfs:
LGTM Thanks for doing this!

cppwfs:
LGTM!    Thanks for doing this! Squashed and Merged

PR #5664, opened 2024-02-07
Refactor TaskLauncherFunction to be a Consumer.

corneil:
Configure RetryTemplate.

PR #5669, opened 2024-02-09
Adjust rest-client and rest-resource tests (SCDF 3.x migration)

onobc:
> [!NOTE]
> Please do not squash when merge - use "Rebase and merge"

## First commit
The 1st commit addresses the changes initially made to the `spring-cloud-dataflow-rest-resource` module in https://github.com/spring-cloud/spring-cloud-dataflow/pull/5664

* Before the change to using HTTP client5 in the HTTPCC it did not need a live url - it now does.
* In your PR you turned this into a SBT to start a webserver to get the live url (makes sense)
* However, this module is currently ill-suited to launch a SBT as it was not doing this before. 
  * As you found it, it bring in the kitchen sink from skipper but lacks proper db configuration to make that a success (we can add the H2 dep at test scope to deal w/ this
  * The web server will not start as spring web is not on the classpath (if we add starter web at test scope it does start)

See commit for details. NOTE: I chopped out one of the tests - just add it back when you take the changes into your PR. OR we can use this PR to fix this. I don't think either rest-client or rest-resource modules have anythign to do w/ tasklauncher PR - do they? 

cc: @cppwfs @jvalkeal 

## Second commit
The 2nd commit fixes `spring-cloud-dataflow-rest-client`. Really this is the same as the original PR w/ the exception that none  of the changes to pom.xml are required. 

> [!NOTE]
> This is the same code and not much has changed on our side, for whatever reason boot has started initializing the datasource where it was not previously needed. In these cases we either enable H2 or we disable the DatasourceAutoConfiguration. It would be nice to know exactly why but it does make sense that it would attempt to configure a datasource as we probably are asking it to.

### DIFF
--- a/spring-cloud-dataflow-rest-client/src/test/java/org/springframework/cloud/dataflow/rest/client/DataFlowClientAutoConfigurationTests.java
+++ b/spring-cloud-dataflow-rest-client/src/test/java/org/springframework/cloud/dataflow/rest/client/DataFlowClientAutoConfigurationTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016-2018 the original author or authors.
+ * Copyright 2016-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


corneil:
I created separate PR for rest-client #5671

cppwfs:
LGTM

cppwfs:
LGTM Thanks for the great work.   Merged.

PR #5672, opened 2024-02-09
Update SpringDoc OpenAPI to latest 2.x

onobc:
This commit updates to Open API 2.x which is the version that supports Boot 3.

The SpringDocs migration guide was followed:
https://springdoc.org/#migrating-from-springdoc-v1


cppwfs:
LGTM!

PR #5673, opened 2024-02-12
Config client legacy processing

jvalkeal:
This is a same fix for skipper server done in #5670 for dataflow server.

### DIFF
--- a/spring-cloud-skipper/spring-cloud-skipper-server-core/src/main/resources/application.yml
+++ b/spring-cloud-skipper/spring-cloud-skipper-server-core/src/main/resources/application.yml
@@ -41,6 +41,8 @@ server:
 spring:
   main:
     banner-mode: "off"
+  config:
+    use-legacy-processing: true
   data:
     rest:
       base-path: /api


corneil (approved):

spring-cloud-skipper/spring-cloud-skipper-server-core/src/main/resources/application.yml
```diff
@@ -41,6 +41,8 @@ server:
 spring:
   main:
     banner-mode: "off"
+  config:
```
onobc, line 44: This change is fine and we are doing it in Dataflow server already. It is a switch that is [still supported by Boot](https://github.com/spring-projects/spring-boot/blob/107dc5338ba8aff35e2ecb8f6872e23f4ab9e979/spring-boot-project/spring-boot-test/src/test/java/org/springframework/boot/test/context/ConfigDataApplicationContextInitializerWithLegacySwitchTests.java#L39) and we will create an issue to follow and remove post 3.0.0.

onobc (approved):
LGTM - thanks @jvalkeal

PR #5677, opened 2024-02-14
Boot 3.x Task/Batch tables should use default prefixes 

cppwfs:
### What was done
For each database type the following migration was implemented:
1. Prefix the TASK V2 and BATCH V4 tables with `V2_` . 
     This allows user to determine what they wish to do with this data
2. Remove BOOT3_prefix for TASK V2 and BATCH V5 tables 
3. Make sure that these migrations were supported by flyway
4. Make sure that the Migration SQL scripts were added to associated yaml files. 

### What was affected
The following databases were migrated:
1. H2
2. MariaDB
3. Mysql
4. Postgres
5. Oracle
6. SQLServer
7. DB2

### Types of migrations
There were 3 types of migrations that occurred.
1. Default - These types of migrations were typically alter tables and alter sequences.   
The following databases belong to this group:
    a. MariaDB
    b. Mysql
    c. Postgres
    d. Oracle (with some exceptions)
    e. Sql Server (commands different than others, but principles remained).
    
2. In-Memory - For this case the Task V2 and Batch V4 tables/sequences DDL 
was removed and BOOT3_ prefix was removed.  The only in-memory we support is H2.
3. DB2 - Gets its own category
To rename the tables db2 requires all primary and foreign keys to be dropped.
In this case we decided to create the new tables and copy the contents of the 
original tables to the new tables.  Then remove the original. 
This was to avoid errors when recreating  the keys

### How to eat this Elephant
This PR is a bit large so let's discuss how we can handle this review.  
Let's review this by sampling one database from each type of migration.  
This is so that we can make sure the general pattern works for folks.
* Default type, look at **Mariadb**
* In Memory type look at **H2** 
* DB2 Type look at (well... ) **DB2** :-D

After we finish this sampling review, those changes will be applied to the other databases and then a full review 
can be made.


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/db/migration/mariadb/V11__Remove_Task2_Batch4_Support.java
```diff
@@ -0,0 +1,230 @@
...
+package org.springframework.cloud.dataflow.server.db.migration.mariadb;
+
+import java.util.Arrays;
+import java.util.List;
+
+import org.springframework.cloud.dataflow.common.flyway.SqlCommand;
+import org.springframework.cloud.dataflow.server.db.migration.AbstractRemoveBatch4Task2Tables;
+
+/**
+ * Prefixes Task V2 tables and V4 Batch tables with a V2_prefix as well as remove the BOOT3_ prefix for V3 task and v5 batch tables.
```
onobc, line 25: **[TinyNit]** spacing between `V2_` and `prefix`

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/db/migration/mariadb/V11__Remove_Task2_Batch4_Support.java
```diff
@@ -0,0 +1,230 @@
...
+	private final static String DROP_VIEW_AGGREGATE_JOB_INSTANCE =
+		"DROP VIEW AGGREGATE_JOB_INSTANCE";
+
+	private final static String DROP_VIEW_AGGREGATE_TASK_BATCH =
+		"DROP VIEW AGGREGATE_TASK_BATCH";
+
+	private final static String DROP_VIEW_AGGREGATE_STEP_EXECUTION =
+		"DROP VIEW AGGREGATE_STEP_EXECUTION";
+
+	/**
```
onobc, line 52: **[GlobalComment]** I think these should just be regular comments rather than javadoc, otherwise they will be attributed to the variable immediately below it, but nothing else.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/db/migration/oracle/V11__Remove_Task2_Batch4_Support.java
```diff
@@ -0,0 +1,230 @@
...
+			SqlCommand.from(DROP_VIEW_AGGREGATE_JOB_INSTANCE),
+			SqlCommand.from(DROP_VIEW_AGGREGATE_TASK_BATCH),
+			SqlCommand.from(DROP_VIEW_AGGREGATE_STEP_EXECUTION)
+		);
+	}
+
+	@Override
+	public List<SqlCommand> renameTask3Tables() {
+		return Arrays.asList(
+				SqlCommand.from(RENAME_TASK_EXECUTION_V3_TABLE),
```
onobc, line 177: **[NIT]** Indentation is off between the different methods

spring-cloud-dataflow-server-core/src/main/resources/application-init-mariadb.yml
```diff
@@ -15,4 +15,6 @@ spring:
         - classpath*:/schemas/mariadb/V6-dataflow.sql
         - classpath*:/schemas/mariadb/V7-dataflow.sql
         - classpath*:/schemas/mariadb/V9-dataflow.sql
-        - classpath*:/schemas/mariadb/V8-dataflow.sql
\ No newline at end of file
+        - classpath*:/schemas/mariadb/V8-dataflow.sql
```
onobc, line 18: Looks like v9 is before v10 although v9 was before v8 already??

spring-cloud-dataflow-server-core/src/main/resources/org/springframework/cloud/dataflow/server/db/migration/h2/V1__INITIAL_SETUP.sql
```diff
@@ -51,136 +51,7 @@ create table task_definitions (
...
-CREATE SEQUENCE task_execution_metadata_seq START WITH 1 INCREMENT BY 1;
-CREATE SEQUENCE BATCH_STEP_EXECUTION_SEQ;
-CREATE SEQUENCE BATCH_JOB_EXECUTION_SEQ;
-CREATE SEQUENCE BATCH_JOB_SEQ;
-create index STEP_NAME_IDX on BATCH_STEP_EXECUTION (STEP_NAME);
-create index TASK_EXECUTION_ID_IDX on TASK_EXECUTION_PARAMS (TASK_EXECUTION_ID);
-
-CREATE TABLE BOOT3_TASK_EXECUTION  (
+CREATE TABLE TASK_EXECUTION  (
                                TASK_EXECUTION_ID BIGINT NOT NULL PRIMARY KEY ,
```
onobc, line 55: Do you mind removing all of the formatting as it is hard to see what has actually changed. Mostly it is slight column type adjustments and a few new tables.

onobc (changes requested):
Nice work @cppwfs. 

I have added a few comments, mostly of the nit variety. 

These flyway migrations always rings the DRY violation alarm but that is the nature of the beast. I think for the N common ones we can combine into a base class to avoid the duplication as they are identical. 

Not in this PR, and not introduced by this PR, but it would be really nice if we could just play the non-flyway SQL file from Flyway rather than have the same duplicated thing in 2 locations.

spring-cloud-dataflow-server-core/src/main/resources/application-init-mariadb.yml
cppwfs: It was like this before the introduction of this PR.

onobc: Agreed. But was that intentional?

cppwfs: I'll investigate and remediate if necessary

spring-cloud-dataflow-server-core/src/main/resources/org/springframework/cloud/dataflow/server/db/migration/h2/V1__INITIAL_SETUP.sql
cppwfs: I don't think there is file formatting changes.
2 significant changes here:
1. Remove Task V2 and Batch V4  schema declarations
2. Rename all Task V3 and Batch V5 schema declarations Removing the BOOT3_ prefixes.

onobc: Looks good w/ the update.

onobc (approved):
LGTM - thanks @cppwfs for knocking this out.

cppwfs:
Squashed and Merged with Approval.

PR #5678, opened 2024-02-14
Add Autoconfiguration.imports to resource includes in pom.xmls

cppwfs:
### 1st Commit
* We had an issue where some of the auto configurations were not firing.
This is because when a user specifies the resource tag in maven,
the boot plugin expects the user to fill the resources manually vs. Its default behavior
* Also removed the version from the H2 dependency so that it can be managed by the bom
* Removed unnecessary `@Configuration` annotations where a `@Autoconfiguration` annotation is present
### 2nd Commit
- Currently SCDF fails to start with errors around metrics.
- These will need to be re-added when metric migration begins
- Remove debug settings from previous commit


onobc (approved):
LGTM - thanks @cppwfs

PR #5679, opened 2024-02-14
Stub out DataflowPagingProviders so that beans can be created

cppwfs:
This will facilitate Dataflow to start

* Made sure JobExplorer has access to PlatformTransactionManager since it is no longer provided by batch
* WARNING I put a `@Primary` runtimeAppsStatusFJPFB for to resolve mulitple bean error.  This looks right, but I can be wrong!!!

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowControllerAutoConfiguration.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowControllerAutoConfiguration.java
@@ -135,6 +135,7 @@ import org.springframework.cloud.skipper.client.util.HttpClientConfigurer;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
+import org.springframework.context.annotation.Primary;
 import org.springframework.core.env.PropertyResolver;
 import org.springframework.data.jpa.repository.config.EnableJpaAuditing;
 import org.springframework.data.jpa.repository.config.EnableJpaRepositories;
@@ -451,6 +452,8 @@ public class DataFlowControllerAutoConfiguration {
 			return new StreamLogsController(streamDeployer);
 		}
 
+
+		@Primary
 		@Bean
 		@ConditionalOnMissingBean(name = "runtimeAppsStatusFJPFB")
 		public ForkJoinPoolFactoryBean runtimeAppsStatusFJPFB() {
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowControllerAutoConfiguration.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowControllerAutoConfiguration.java
@@ -135,7 +135,6 @@ import org.springframework.cloud.skipper.client.util.HttpClientConfigurer;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
-import org.springframework.context.annotation.Primary;
 import org.springframework.core.env.PropertyResolver;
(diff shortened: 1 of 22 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowControllerAutoConfiguration.java
```diff
@@ -451,6 +452,8 @@ public StreamLogsController streamLogsController(StreamDeployer streamDeployer)
 			return new StreamLogsController(streamDeployer);
 		}
 
+
+		@Primary
```
onobc, line 456: Yeh I don't think `@Primary` would do what we want here. In fact, it would inject the one marked as primary in both injection targets. 

## Cause
So this is due to the removal of the `LocalVariableTableParameterNameDiscoverer` in SF 6.1 ([details here](https://github.com/spring-projects/spring-framework/wiki/Upgrading-to-Spring-Framework-6.x#parameter-name-retention). 

## Solution
2 options:

### Option A
Add the `-parameters` flag when we compile.

### Option B
Add `@Qualifier("<yourBeanName>")` to the injection point. Example: 

```
@Bean
@ConditionalOnMissingBean
public AppRegistryController appRegistryController(
		...
		@Qualifier("appRegistryFJPFB") ForkJoinPool appRegistryFJPFB, 
        ...
```

I am in favor of the former option as the latter option will be a Whac-A-Mole.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/JobExplorerContainer.java
```diff
@@ -9,16 +9,18 @@
...
 import org.springframework.cloud.dataflow.schema.service.SchemaService;
 import org.springframework.cloud.dataflow.server.controller.NoSuchSchemaTargetException;
+import org.springframework.transaction.PlatformTransactionManager;
 import org.springframework.util.StringUtils;
 
 public class JobExplorerContainer {
 	private final Map<String, JobExplorer> container = new HashMap<>();
 
-	public JobExplorerContainer(DataSource dataSource, SchemaService schemaService) {
+	public JobExplorerContainer(DataSource dataSource, SchemaService schemaService, PlatformTransactionManager platformTransactionManager) {
```
onobc, line 18: Who is injecting the `ResourcelessTransactionManager` ?

onobc (commented):
Thanks @cppwfs  - couple of points to address.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/JobExplorerContainer.java
cppwfs: Not sure that I understand.  We are not using ResourceslessTransactionManager.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowControllerAutoConfiguration.java
cppwfs: Thank you for the advice!!!!

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/JobExplorerContainer.java
onobc: Sorry, who creates the tx mgr?

cppwfs: It is the one used by SCDF 
https://github.com/spring-cloud/spring-cloud-dataflow/blob/main/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfiguration.java#L93C1-L100C3

onobc (approved):
LGTM - thanks for the updates @cppwfs

PR #5680, opened 2024-02-16
Update Tests to resolve failures due to updated dependency versions

cppwfs:
This commit does not cover all the cases where updated dependencies broke existing tests. The main ones resolved here are some basics on Hibernate dialect versions, batch 5 updates, and some basic removal of BOOT3/BOOT2 additions in 2.11.x.
Also had to add additional deletes for the DAO tests as seen drop-table-schema-*.sql files

There are other updates that are causing alot of failures.  Here are the issues and a brief description:
* Currently Hibernate 6 does not create sequences out of the box.   More can be read here: https://github.com/hibernate/hibernate-orm/blob/6.0/migration-guide.adoc#implicit-identifier-sequence-and-table-name
One test that shows this error:TabOnTapCompletionProviderTests

* Hibernate Dialect.  Some tests require the hibernate dialects to be updated to hibernate 6. I did resolve the ones I found.   But may have missed others.

* Many tests are failing trying with the following exception: `No bean named 'mvcConversionService' available`
For example: SpringDocAutoConfigurationTests.enabledWithCustomSpringDocSettings

* The TODO excluding wavefront are causing some errors for example: Error creating bean with name 'management.wavefront-org.springframework.boot.actuate.autoconfigure.wavefront.WavefrontProperties': Lookup method resolution failed
An Example can be found here: SpringDocIntegrationTests.disabledByDefault


onobc (approved):
LGTM @cppwfs - going to go ahead and merge and then we can continue the exercise from this point.

PR #5681, opened 2024-02-17
Use the single hibernate_sequence strategy

onobc:
Hibernate 6.0 now creates a sequence per entity hierarchy instead of a single hibernate_sequence. 
This commit sets the `hibernate.id.db_structure_naming_strategy` property to `single` to preserve the previous behavior of using a single hibernate_sequence.

The setting is in both Dataflow and Skipper default properties so that it is in effect system-wide.

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
+++ b/spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
@@ -60,9 +60,10 @@ spring:
   jpa:
     properties:
       hibernate:
+        # continue using a single hibernate_sequence table
+        id.db_structure_naming_strategy: single
+        # use the sequence style identifier generator
         id.new_generator_mappings: true
-        # Statistics generation is required for publishing JPA micrometer metrics.
-        # generate_statistics: true
     hibernate:
       naming:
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
```diff
@@ -60,9 +60,10 @@ spring:
   jpa:
     properties:
       hibernate:
+        # continue using a single hibernate_sequence table
+        id.db_structure_naming_strategy: single
+        # use the sequence style identifier generator
         id.new_generator_mappings: true
-        # Statistics generation is required for publishing JPA micrometer metrics.
```
onobc, line 64: **[Unrelated-Cleanup]** Property not used - lets keep only what we use

spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
```diff
@@ -60,9 +60,10 @@ spring:
   jpa:
     properties:
       hibernate:
+        # continue using a single hibernate_sequence table
+        id.db_structure_naming_strategy: single
+        # use the sequence style identifier generator
```
onobc, line 65: Just add a comment as to why this one is in here as well

cppwfs:
Rebased and Merged.   Thank you Again!

PR #5682, opened 2024-02-17
Add DatabaseAwareLobUserType for @Lob columns

onobc:
Hibernate 6.0 removed support for string values in `@Type` mappings in favor of specifying UserTypes. 
This commit creates a `DatabaseAwareLobUserType` that provides the same functionality as the previous `DatabaseAwareLobType`.

### DIFF
--- a/spring-cloud-dataflow-common/spring-cloud-dataflow-common-persistence/src/main/java/org/springframework/cloud/dataflow/common/persistence/type/DatabaseAwareLobUserType.java
+++ b/spring-cloud-dataflow-common/spring-cloud-dataflow-common-persistence/src/main/java/org/springframework/cloud/dataflow/common/persistence/type/DatabaseAwareLobUserType.java
@@ -1,0 +1,45 @@
+package org.springframework.cloud.dataflow.common.persistence.type;
+
+import java.util.function.BiConsumer;
+
+import org.hibernate.type.descriptor.java.BasicJavaType;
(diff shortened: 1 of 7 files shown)


spring-cloud-dataflow-common/spring-cloud-dataflow-common-persistence/src/main/java/org/springframework/cloud/dataflow/common/persistence/type/DatabaseAwareLobUserType.java
```diff
@@ -0,0 +1,45 @@
+package org.springframework.cloud.dataflow.common.persistence.type;
```
cppwfs, line 1: Ye old Copyright Nag :-D

cppwfs (commented):
Do we have tests for this code?

onobc:
> Do we have tests for this code?

For which code, the `DatabaseAwareLobUserType`? No more tests than we had for the previous custom user type. No more tests than we had for the previously introduced `@JdbcTypeMetadata`. These lower level pieces are tested during the data access operations that we should surely have tests for. 

I am more confident that this will work as we expect it than the `@JdbcTypeCode` changes after reading up more on this topic.

corneil (approved):

cppwfs:
LGTM 
Thanks for the contribution!
Rebased Squashed Merged

PR #5683, opened 2024-02-17
Fix SpringDocAutoConfigurationTests

onobc:
This commit adds a mock "mvcConversionService" bean to the app context for the SpringDocAutoConfigurationTests as SpringDoc 2.x expects this bean to be available.

This commit also updates the use of Mockito verify for varargs in the SpringDocAutoConfigurationTests.


cppwfs:
LGTM.
Rebased/Merged.
Thanks again for looking into this!!!!

PR #5685, opened 2024-02-20
Overhaul metric deps and options

### LINKED ISSUE #5675: Revisit metric exports
This is a short recap when going through various issues:

- `spring-cloud-dataflow-parent/pom.xml` manages `wavefront-spring-boot-bom.version` as `2.3.4`. Bit nasty as boot doesn't manage wavefront but depends something which then pulls specific wavefront sdk version. Should use `3.2.0` and keep that aligned.
- Our server configs currently use `influx`, `prometheus` and `wavefront` and export is disabled on default. Boot changed property format in https://github.com/spring-projects/spring-boot/commit/be3523b1cda41f178df25fdfe8f8f27199259530, we need to change property names. In short `management.metrics.export.prometheus.enabled` vs. `management.prometheus.metrics.export.enabled`.
- However there's `micrometer-metrics/prometheus-rsocket-proxy` which still uses `management.metrics.export.prometheus.rsocket.enabled`, they never migrated to new property structure.
- Thought there's no "release" in `prometheus-rsocket-proxy` claiming boot 3.x support so wondering if we should just remove this dependency.

jvalkeal:
Align wavefront version so that we don't have misaligned versions coming out from other parts of a metric system. Short story is that boot doesn't manage wavefront but there is an explicit dependency to wavefront sdk libs in metric system.

Rename metric options within management for influx, prometheus and wavefront to align changes in boot itself.

For rsocket proxy keep old management.metrics.export.prometheus.rsocket.enabled as that is going to get moved under micrometer spesific namespace when they release boot3 support.

This commit was supposed to get skipper server to start but there is a new issue about missing bean
org.springframework.statemachine.data.jpa.JpaStateMachineRepository which will need to get fixed in a separate commit.

Fixes #5675

### DIFF
--- a/spring-cloud-dataflow-parent/pom.xml
+++ b/spring-cloud-dataflow-parent/pom.xml
@@ -40,7 +40,7 @@
 		<aws-java-sdk-ecr.version>1.12.513</aws-java-sdk-ecr.version>
 		<!-- only used for dataflow managed stream applications, e.g., tasklauncher -->
 		<stream-applications.version>3.2.1</stream-applications.version>
-		<wavefront-spring-boot-bom.version>2.3.4</wavefront-spring-boot-bom.version>
+		<wavefront-spring-boot-bom.version>3.2.0</wavefront-spring-boot-bom.version>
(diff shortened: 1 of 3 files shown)


cppwfs:
LGTM
Squashed Merged

Thank you for the contribution, more tests are showing up green now!

PR #5688, opened 2024-02-20
Disable all tests that are failling because of schema changes.

cppwfs:
The following tests contain a significant number of failing tests because of the Boot2 and Boot3 schema changes.
They have been disabled and tagged as `TODO: Boot3x followup` so that they can be re-enabled
in the future once we update the code to handle the schema changes


corneil:
I suggest adding the following after `mvn -B -s .github/settings.xml clean install` in ci-pr.xml

```
    - name: Test Report
      uses: dorny/test-reporter@v1
      if: ${{ success() || failure() }}
      with:
        name: Unit Tests
        path: '**/surefire-reports/*.xml'
        reporter: java-junit
        list-tests: failed
```

cppwfs:
@corneil I like the idea.   I'll add it as another PR.

onobc (approved):
LGTM - thanks @cppwfs

PR #5693, opened 2024-02-21
Remove the tailing / mark in the mockMVC urls

cppwfs:
These slash marks are not required by Boot3 and cause the tests to fail.

* Simplify TestDependencies class so that it implements WebMvcConfigurer 
instead of extending WebMvcConfigurationSupport while creating the WebMvcConfigurer bean.   
* This PR will reduce the number of test failures and expose other test failures.


onobc (approved):
Thanks @cppwfs  - LGTM.

We need to start a document w/ "What's New / Breaking Changes" and add these things as we go. This could be the 1st entry. "Users, if you are calling into our REST API and you use traliling slashes - it will no longer find the API; please remove any trailing slashes". 

I added this here https://github.com/spring-cloud/spring-cloud-dataflow/wiki/SCDF-3.x-What's-New---Breaking-Changes

cc: @jvalkeal @corneil if you run into any breaking changes please add to the above doc as we go.

PR #5694, opened 2024-02-22
Fix BootClassLoaderFactory and Snakeyaml

onobc:
Two commits - details in the commit messages.

> [!NOTE]
> Please do not SQUASH, just Rebase/Merge

### DIFF
--- a/spring-cloud-dataflow-container-registry/pom.xml
+++ b/spring-cloud-dataflow-container-registry/pom.xml
@@ -56,10 +56,6 @@
 		</dependency>
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-container-registry/pom.xml
```diff
@@ -54,10 +54,6 @@
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter</artifactId>
 		</dependency>
-		<dependency>
```
onobc, line 57: This is not actually used by anything. 
We only use the classic loader.

corneil (approved):

cppwfs:
LGTM Rebased Merged.

PR #5702, opened 2024-02-22
Reuse CriteriaQuery when retrieving result count.

cppwfs:
When migrating to the latest Hibernate we saw the following exception:
```
Caused by: java.lang.IllegalArgumentException: Already registered a copy: SqmBasicValuedSimplePath
```
This is because we were recreating a CriteriaQuery for our count Queries. Hibernate no longer allows us to do that,
but rather allows us to use the existing CriteriaQuery,
but use the convenience method createCountQuery to provide the criteria query for our createQuery.

Also removed 6.1.7 version for hibernate core  that allows us to use the Boot bom. 
This also caused some downstream issues where dataflow was using the old version brought in from skipper

### DIFF
--- a/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/repository/AppRegistrationRepositoryImpl.java
+++ b/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/repository/AppRegistrationRepositoryImpl.java
@@ -28,6 +28,7 @@ import jakarta.persistence.criteria.CriteriaQuery;
 import jakarta.persistence.criteria.Predicate;
 import jakarta.persistence.criteria.Root;
 
+import org.hibernate.query.sqm.tree.select.SqmSelectStatement;
 import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.cloud.dataflow.core.AppRegistration;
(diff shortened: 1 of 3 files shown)


spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/repository/AppRegistrationRepositoryImpl.java
```diff
@@ -91,17 +92,11 @@ public Page<AppRegistration> findAllByTypeAndNameIsLikeAndVersionAndDefaultVersi
...
-	private Long getTotalCount(CriteriaBuilder criteriaBuilder, Predicate[] predicateArray) {
-		CriteriaQuery<Long> criteriaQuery = criteriaBuilder.createQuery(Long.class);
-		Root<AppRegistration> root = criteriaQuery.from(AppRegistration.class);
-
-		criteriaQuery.select(criteriaBuilder.count(root));
-		criteriaQuery.where(predicateArray);
-
-		return entityManager.createQuery(criteriaQuery).getSingleResult();
+	private Long getTotalCount(CriteriaQuery<AppRegistration> criteriaQuery) {
+		return (Long) entityManager.createQuery(((SqmSelectStatement)criteriaQuery).createCountQuery()).getSingleResult();
```
onobc, line 99: Is `criteraQuery` always a `SqmSelectStatement` ?

onobc (approved):
Nice work @cppwfs - thanks for the update. 

I have one question about the casting, but other than that LGTM.

spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/repository/AppRegistrationRepositoryImpl.java
cppwfs: for these two statements, yes.

PR #5707, opened 2024-02-27
Upgrade spring-statemachine 4.0.0

jvalkeal:
This should make skipper to start and run. Simple ticktock stream should work ok.

There's small changes to fix changes in sm api's.


onobc (approved):
Looks simple enough. Thanks @jvalkeal

PR #5709, opened 2024-02-27
Remove JobService and JobExplorer Containers

cppwfs:
Since the next series of the batch migrations are large in scope we will break this up into multiple steps so that they have a certain theme.
The goal of each of the steps is complete the goal while allowing dataflow to start as well as keep the test success percentage at its current level (or above).   
This may not always be possible but should be a driver since we don't want to affect other efforts.

This PR is spread across 3 commits based on a specific focus.   

- 7055ae7e7 Remove JobRepositoryContainer and update Code to use JobRepository bean directly.
- ed8b03da0 Remove JobServiceContainer and JobExplorerContainer
 and allow services to use JobService and JobExplorer directly
- 8bb510d31  Update Project as to restore tests success percentage to original state and to 
update code project so dataflow can start.


spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/JobExecutionControllerTests.java
```diff
@@ -73,8 +70,8 @@
  */
 
 //TODO: Boot3x followup
-@Disabled("TODO: Boot3 followup after boot3/boot2 task changes are complete")
-@RunWith(SpringRunner.class)
+//@Disabled("TODO: Boot3 followup after boot3/boot2 task changes are complete")
+@ExtendWith(SpringExtension.class)
```
onobc, line 74: Not needed as `@SpringBootTest` includes this annotation.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/JobExecutionControllerTests.java
```diff
@@ -73,8 +70,8 @@
  */
 
 //TODO: Boot3x followup
-@Disabled("TODO: Boot3 followup after boot3/boot2 task changes are complete")
-@RunWith(SpringRunner.class)
+//@Disabled("TODO: Boot3 followup after boot3/boot2 task changes are complete")
```
onobc, line 73: Can we remove this comment then now?

onobc (changes requested):
Thanks for the updates @cppwfs - LGTM.

Nit comment to address and then this can merge.

cppwfs:
Updated PR based on code review requests.
Rebased Squashed Merged.

PR #5710, opened 2024-02-27
Adjust metrics replicator for Spring Boot 3

onobc:
This commit adjusts the MetricsReplicationEnvironmentPostProcessor to account for the Spring Boot 3 change in the metrics config props prefix scheme from 
`management.metrics.export.<meter-registry>.<property-path>` to 
`management.<meter-registry>.metrics.export.<property-path>`.

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/MetricsReplicationEnvironmentPostProcessor.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/MetricsReplicationEnvironmentPostProcessor.java
@@ -131,19 +131,18 @@ public class MetricsReplicationEnvironmentPostProcessor implements EnvironmentPo
 	}
 
 	/**
-	 * Checks if the management.metrics.export.<meter-registry>.enabled property is set to ture for the provided
-	 * meterRegistryPropertyClass.
+	 * Checks if the 'management.<meter-registry>.metrics.export.enabled' property is set to true for the specified
+	 * meter registry.
 	 *
-	 * @param meterRegistryPropertyClass Property class that follows Boot's meter-registry properties convention.
-	 * @param environment Spring configuration environment.
-	 * @return Returns true if the provide class contains {@link ConfigurationProperties} prefix of type:
-	 *         management.<meter-registry>.metrics.export and the management.<meter-registry>.metrics.export.enabled
-	 *         property is set to true. Returns false otherwise.
+	 * @param meterRegistryConfigPropsClass the SpringBoot configuration properties for the meter registry
+	 * @param environment the application environment
+	 * @return whether the 'management.<meter-registry>.metrics.export.enabled' property is set to true for the
+	 * specified meter registry class.
 	 */
-	private boolean isMetricsRegistryEnabled(Class<?> meterRegistryPropertyClass, ConfigurableEnvironment environment) {
-		String metricsPrefix = retrievePropertyPrefix(meterRegistryPropertyClass);
+	private boolean isMetricsRegistryEnabled(Class<?> meterRegistryConfigPropsClass, ConfigurableEnvironment environment) {
+		String metricsPrefix = retrievePropertyPrefix(meterRegistryConfigPropsClass);
 		if (!StringUtils.hasText(metricsPrefix)) {
-			logger.warn("Meter registry properties class %s is not a @ConfigurationProperties".formatted(meterRegistryPropertyClass));
+			logger.warn("Meter registry properties class %s is not a @ConfigurationProperties".formatted(meterRegistryConfigPropsClass));
 			return false;
 		}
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/MetricsReplicationEnvironmentPostProcessor.java
```diff
@@ -280,13 +262,16 @@ private void traversePropertyClassFields(Class<?> metricsPropertyClass, Consumer
 	private void traverseClassFieldsRecursively(Class<?> metricsPropertyClass, String metricsPrefix,
 			Consumer<String> metricsReplicationHandler) {
 		for (Field field : metricsPropertyClass.getDeclaredFields()) {
-			if (field.getType().isMemberClass() && Modifier.isStatic(field.getType().getModifiers())) {
+			var isStaticMemberClass = field.getType().isMemberClass() && Modifier.isStatic(field.getType().getModifiers());
+			if (isStaticMemberClass && !field.getType().isEnum()) {
```
onobc, line 266: I found an unrelated bug that has been in this class since its inception... enum fields that are also nested static members (e.g. `WavefrontProperties.apiTokenType` where the `ApiTokenType` is defined in the config props class). It would just die in the recursion and was not visible to end user other than missing expected replicated properties.

cppwfs: Good catch

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/config/MetricsReplicationEnvironmentPostProcessorTests.java
```diff
@@ -154,71 +150,42 @@ void wavefrontPropertiesReplicationWithPlaceholders() {
...
-				"--management.metrics.export.influx.batch-size=20000")) {
+				"--management.influx.metrics.export.enabled=true",
+				"--management.influx.metrics.export.db=myinfluxdb",
+				"--management.influx.metrics.export.uri=http://influxdb:8086",
+				"--management.influx.metrics.export.batch-size=20000")) {
 			for (String commonPropPrefix : COMMON_APPLICATION_PREFIXES) {
-				assertEnvHasProperty(ctx, commonPropPrefix + "management.metrics.export.influx.enabled", "true");
-				assertEnvHasProperty(ctx, commonPropPrefix + "management.metrics.export.influx.db", "myinfluxdb");
-				assertEnvHasProperty(ctx, commonPropPrefix + "management.metrics.export.influx.uri", "http://influxdb:8086");
-				assertEnvHasProperty(ctx, commonPropPrefix + "management.metrics.export.influx.batch-size", "20000");
```
cppwfs, line 220: Just curious as to why we removed batch-size from the assert tests?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/MetricsReplicationEnvironmentPostProcessor.java
```diff
@@ -137,46 +130,35 @@ private void ensurePropIsReplicatedExactlyOnceToCommonProps(String metricsPropNa
...
-		//
-		Matcher matcher = METRIC_PROP_NAME_PATTERN.matcher(metricsPropertyName);
-		if (matcher.matches()) {
-			return Optional.of(matcher.group(1) + matcher.group(3) + matcher.group(2) + matcher.group(4));
-		}
-		return Optional.empty();
-	}
-
 	/**
 	 * Checks if the management.metrics.export.<meter-registry>.enabled property is set to ture for the provided
```
cppwfs, line 134: We can correct the spelling of ture :-)

cppwfs:
Looks good just a nitpick, a question, and a compliment :-D

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/MetricsReplicationEnvironmentPostProcessor.java
onobc: Never

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/config/MetricsReplicationEnvironmentPostProcessorTests.java
onobc: They are still there - they are hidden below

cppwfs:
Great work here!

cppwfs:
LGTM
Squashed Merged

PR #5712, opened 2024-02-28
Scdf 3.0 container rem 2

cppwfs:
This is Phase 2 of 3 to replace *Containers with the task or Job implementations.    
This PR is broken across 2 different commits.  It should have been 3, but I messed up.   
* fce72eb Handles the replacement of JobExecutionDaoContainer and TaskBatchDaoContainer.
* 3899553 Handles the replacement of JdbcTaskExecutionDaoContainer.

You will also note that dataflow is now  relying on BatchAutoConfiguration for the JobRepository .


cppwfs:
Squashed and merged

PR #5714, opened 2024-02-29
Add column missed from conversion from OID to Text

tomgeraghty3:
It seems there is a bug in the latest SCDF version where the `data` column in the `skipper_manifest` was **not** converted from OID into text like most of the other columns. This addition fixes the this


corneil:
We have applied https://github.com/spring-cloud/spring-cloud-dataflow/pull/5721 to branch `v2.11.x`

dhpizza:
> We have applied #5721 to branch `v2.11.x`

Hi @corneil 
does that mean that a frech pull of any 2.11.x docker image of the skipper has the fix or do we need to wait for 2.11.3?

PR #5716, opened 2024-02-29
Remove the remaining Task Containers from SCDF

cppwfs:
This is the 3rd phase of removing the TaskContainers and allowing SCDF to use the Task Repositories And Task DAOs directly.   This PR also begins the migration of DAO's and Repositories from the AggregateTaskConfiguration to TaskConfiguration.   Eventually leading to the remove of AggregateTaskConfiguration.

This PR is spread across 5 commits based on a specific focus.

- 989036fca  Move TaskRepository and SCDFDao beans from AggregateTaskConfiguration to TaskConfiguration
- 2fbcb32b3 Replace TaskRepositoryContainer with TaskRepository
- 93e1b3c34 Replace DataflowTaskExecutionMetadataDaoContainer with DAO from dataflow
- d896f1856 Replace DataflowJobExecutionDaoContainer with dataflow's JobExecutionDao.
- f1aa305cd Replace DataflowTaskExecutionMetadataDaoContainer with DataflowTaskExecutionMetadataDao


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/TaskConfiguration.java
```diff
@@ -211,6 +214,19 @@ public TaskAppDeploymentRequestCreator taskAppDeploymentRequestCreator(
 				metadataResolver, dataflowServerUri);
 	}
 
+	@Bean
+	public TaskRepository taskRepository(DataSource dataSource) {
+		MultiSchemaTaskExecutionDaoFactoryBean taskExecutionDaoFactoryBean = new MultiSchemaTaskExecutionDaoFactoryBean(dataSource, "TASK_");
```
onobc, line 219: So will the MultiSchema* components be removed when the aggregate is removed? Those were invented for the aggregate multi version support.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/JobInstanceControllerTests.java
```diff
@@ -169,7 +171,9 @@ private void createSampleJob(String jobName, int jobExecutionCount)
 		TaskExecution taskExecution = taskExecutionDao.createTaskExecution(jobName, LocalDateTime.now(), new ArrayList<String>(), null);
 
 		for (int i = 0; i < jobExecutionCount; i++) {
-			JobExecution jobExecution = jobRepository.createJobExecution(jobName, new JobParameters());
+			JobParameters jobParameters =
+				new JobParameters(Collections.singletonMap("parm", new JobParameter<>(i, Integer.class)));
```
onobc, line 175: Is `parm` the correct key?

spring-cloud-dataflow-composed-task-runner/src/test/java/org/springframework/cloud/dataflow/composedtaskrunner/TaskLauncherTaskletTests.java
```diff
@@ -524,7 +524,7 @@ TaskRepositoryInitializer taskRepositoryInitializer() {
 
 		@Bean
 		TaskExecutionDao taskExecutionDao(DataSource dataSource) {
-			return new JdbcTaskExecutionDao(dataSource);
+			return = new JdbcTaskExecutionDao(dataSource);
```
onobc, line 527: Is this a Java 17 feature? ;p

onobc (approved):
@cppwfs thanks,
A few minor comments to address but other than that LGTM

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/features/TaskConfiguration.java
cppwfs: Correct.   Working toward their retirement.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/JobInstanceControllerTests.java
cppwfs: Great question.    In this case we used to be able to create multiple job executions with a single instance for this command.   But that loophole was closed in Batch 5.  So now we need to make each JobExecution unique by either changing the name of the Job or adding a unique parameter value.   In our case we want the same job name, so we can add a unique job parameter.

onobc: Oh ok, so this is in effect a "dummy param", correct? If so, maybe we should name the key something besides `parm` - maybe `unused` (or something like that). Wdyt?

Where does this show up as `parm`? In the UI?

cppwfs: Job Parameters can be viewed as a part of the Job Instance in the UI.   I can change it to test-parameter.

onobc: Oh nevermind -  This is a test. Carry on.

cppwfs:
Updated based on code review.  Squashed Merged.   Thanks @onobc !!!!

PR #5717, opened 2024-03-02
Adjust order of DefaultEnvironmentPostProcessor

onobc:
This commit adjusts the change of the order attribute made in commit de5824797fe7d765d4cdeb4f5a8b1aebed2ed9d4. The default EPP in fact needs to run after the ConfigDataEnvironmentPostProcessor and before the ConfigDataMissingEnvironmentPostProcessor.

### DIFF
--- a/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/config/DefaultEnvironmentPostProcessorTests.java
+++ b/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/config/DefaultEnvironmentPostProcessorTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2019 the original author or authors.
+ * Copyright 2015-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,13 +16,12 @@
 
 package org.springframework.cloud.dataflow.server.config;
 
-import org.junit.Test;
+import org.junit.jupiter.api.Test;
 
 import org.springframework.boot.SpringApplication;
 import org.springframework.context.ConfigurableApplicationContext;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotNull;
+import static org.assertj.core.api.Assertions.assertThat;
 
 /**
  * Tests for {@link DefaultEnvironmentPostProcessor}.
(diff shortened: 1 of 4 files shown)


spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/config/DefaultEnvironmentPostProcessorTests.java
```diff
@@ -16,46 +16,54 @@
...
+class DefaultEnvironmentPostProcessorTests {
 
-	private static final String MANAGEMENT_CONTEXT_PATH = "management.contextPath";
-
-	private static final String CONTRIBUTED_PATH = "/bar";
+	private static final String MANAGEMENT_CONTEXT_PATH = "management.context-path";
 
 	@Test
-	public void testDefaultsBeingContributedByServerModule() throws Exception {
+	void defaultsAreContributedByServerModule() {
```
onobc, line 37: **[UNRELATED]** Their are various junit4->junit5 changes in this test.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/config/DefaultEnvironmentPostProcessorTests.java
```diff
@@ -16,46 +16,54 @@
...
 		try (ConfigurableApplicationContext ctx = SpringApplication.run(EmptyDefaultTestApplication.class, "--server.port=0",
 				"--spring.main.allow-bean-definition-overriding=true",
 				"--spring.autoconfigure.exclude=org.springframework.cloud.deployer.spi.cloudfoundry.CloudFoundryDeployerAutoConfiguration,org.springframework.cloud.deployer.spi.kubernetes.KubernetesAutoConfiguration")) {
-			String cp = ctx.getEnvironment().getProperty(MANAGEMENT_CONTEXT_PATH);
-			assertEquals(CONTRIBUTED_PATH, cp);
+
+			// this one comes from <test>/resources/dataflow-server.yml
+			assertThat(ctx.getEnvironment().getProperty(MANAGEMENT_CONTEXT_PATH)).isEqualTo("/foo");
```
onobc, line 43: **[UNRELATED]** Having the value as constant was not helping readability (in fact the opposite). 

I also moved the default to be `/foo` and then the override to be `/bar` as the order in our head is "foo -> bar" rather than "bar -> foo".

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/config/DefaultEnvironmentPostProcessorTests.java
```diff
@@ -16,46 +16,54 @@
...
+
+			// this one comes from dataflow-server-defaults.yml (we use 'spring.flyway.enabled' as the indicator)
```
onobc, line 45: **[UNRELATED]** We are simply picking a well-known simple property that is in the default yml (now and in the future) in order to get the signal that the default properties yml are in fact being contributed.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/config/DefaultEnvironmentPostProcessorTests.java
```diff
@@ -16,46 +16,54 @@
...
+				"--spring.config.name=test",
+				"--spring.flyway.enabled=false",
 				"--spring.main.allow-bean-definition-overriding=true",
 				"--spring.cloud.dataflow.server.profileapplicationlistener.ignore=true",
 				"--spring.autoconfigure.exclude=org.springframework.cloud.deployer.spi.cloudfoundry.CloudFoundryDeployerAutoConfiguration,org.springframework.cloud.deployer.spi.kubernetes.KubernetesAutoConfiguration")) {
-			String cp = ctx.getEnvironment().getProperty(MANAGEMENT_CONTEXT_PATH);
-			assertEquals(cp, "/foo");
-			assertNotNull(ctx.getEnvironment().getProperty("spring.flyway.locations[0]"));
+
+			// this one comes from <test>/resources/test.yml and overrides the entry from <test>/resources/dataflow-server.yml
```
onobc, line 61: **[UNRELATED]** Updated test comments to demystify where these values were coming from and why

corneil (approved):

cppwfs:
LGTM!   Thanks for the contribution it is AWESOME!

PR #5720, opened 2024-03-04
commons compress update

### LINKED ISSUE #5719: [CVE-2024-25710] Security issues in transitive dependency of commons-compress 1.24.0 (SPRING-CLOUD-DATAFLOW 2.11.2)
Hello,

We've found the following security issue in the transitive dependency of commons-compress 1.24.0  in version 2.11.2.
https://www.mend.io/vulnerability-database/CVE-2024-25710
Would appreciate your kind assistance with fixing the issue.

Thanks & BR,
Omri.

corneil:
Fixes #5719

### DIFF
--- a/spring-cloud-skipper/pom.xml
+++ b/spring-cloud-skipper/pom.xml
@@ -41,7 +41,7 @@
 		<jmockit.version>1.24</jmockit.version>
 
 		<testcontainers.version>1.17.6</testcontainers.version>
-		<commons-compress.version>1.24.0</commons-compress.version>
+		<commons-compress.version>1.26.0</commons-compress.version>
 		<asciidoctor.maven.plugin.version>2.2.4</asciidoctor.maven.plugin.version>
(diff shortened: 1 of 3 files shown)


spring-cloud-skipper/pom.xml
```diff
@@ -41,7 +41,7 @@
 		<jmockit.version>1.24</jmockit.version>
 
 		<testcontainers.version>1.17.6</testcontainers.version>
-		<commons-compress.version>1.24.0</commons-compress.version>
+		<commons-compress.version>1.26.0</commons-compress.version>
```
onobc, line 44: Looks like this was already here, but we should not need to override here in skipper as it inherits from one of the above. I am ok leaving it in for this PR (up to you).

corneil: I'm leaving the build in 2.11.x as is.

PR #5722, opened 2024-03-05
Upgrade spring-shell 3.2.2

jvalkeal:
- Exclude spring-cloud-dataflow-common-persistence as it caused jdbc stuff in shell and then failures with datasource autoconfig
- Temporarily in tests use reflection as some methods in spring-shell are not public anymore. Long term todo is to migrate to new spring-shell testing utils
- Add commented shell log file settings which is a way to log hard shell startup errors

### DIFF
--- a/spring-cloud-dataflow-shell-core/src/test/java/org/springframework/cloud/dataflow/shell/ShellCommandRunner.java
+++ b/spring-cloud-dataflow-shell-core/src/test/java/org/springframework/cloud/dataflow/shell/ShellCommandRunner.java
@@ -24,6 +24,7 @@ import org.jline.reader.impl.DefaultParser;
 import org.springframework.shell.Input;
 import org.springframework.shell.Shell;
 import org.springframework.shell.Utils;
+import org.springframework.test.util.ReflectionTestUtils;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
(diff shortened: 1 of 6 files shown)


spring-cloud-dataflow-shell-core/src/test/java/org/springframework/cloud/dataflow/shell/ShellCommandRunner.java
```diff
@@ -62,7 +63,9 @@ public ShellCommandRunner withValidateCommandSuccess() {
 	public Object executeCommand(String command) {
 		Parser parser = new DefaultParser();
 		ParsedLine parsedLine = parser.parse(command, command.length() + 1);
-		Object rawResult = this.shell.evaluate(new ParsedLineInput(parsedLine));
+		// TODO: evaluate is not private method in spring-shell so calling it via
```
onobc, line 66: `s/not/now`

spring-cloud-dataflow-shell/src/main/resources/application.yml
```diff
@@ -6,3 +6,9 @@ spring:
 logging:
   pattern:
     console:
+  file:
```
onobc, line 9: > Add commented shell log file settings which is a way to log hard shell startup errors

**Q1**: What is the behavior w/o this file (is this something new in 3.2.2)?

**Q2**: If we ship this w/ debug logging, how does a user override this?

onobc (changes requested):
Thanks @jvalkeal ,

One tiny comment spelling fix and then a question around the log/file. Other than that this LGTM.

spring-cloud-dataflow-shell/src/main/resources/application.yml
jvalkeal: Hold on, this was meant to left there as commented out, just like in skipper.

onobc (approved):
LGTM - thanks @jvalkeal

PR #5729, opened 2024-03-07
Fixed Skipper Smoke Tests

corneil:
Fixed Skipper Smoke Tests after removing `org.springframework.cloud.common.security.CommonSecurityAutoConfiguration` that IDE identified as not being an AutoConfiguration class.


onobc (changes requested):
Thanks for the update @corneil . 

I think the `.run/**` was accidentally staged.

PR #5731, opened 2024-03-09
Update Job to Batch 5 and remove schema usage

cppwfs:
* Removed SchemaController/Tests/Documentation
* Removed the BatchVersion
* Service calls used by JobInstanceController have been migrated to batch 5.x
*  Migrated SQL Statements in AggregateDataFlowTaskExecutionQueryDao to use Task Execution
     This is a first step to retire the use of the aggregates.
* DefaultTaskExecutionServiceTests - since we are no longer actively passing table prefixes the properties summary tests were adjusted
   * Also Removed BATCH_ and TASK_ tests
* JobExecutionControllerTests and JobExecutionThinControllerTests also had similar changes
* Some tests were either @Disabled or the @Test was commented out because the fixes need more investigation.
   * they were marked with todo
* We also need to revisit the use of locals over timezones.  The stop gap is to use the default locale.  This was marked with todo.


corneil (approved):

cppwfs:
Got a verbal greenlight from @onobc.  
Will merge.

Issue #5735, opened 2024-03-12
Oracle tables for Spring Batch require SEGMENT CREATION IMMEDIATE


corneil:
**Description:**
When using jobRepository to create test data you encounter an exception when using Oracle database.

**Solution**
The solution is to add `SEGMENT CREATION IMMEDIATE` to the create table statements for Oracle.

cppwfs:
@corneil I think we can close this one.   WDYT?

corneil:
Not required.

Issue #5718, opened 2024-03-04
Unable to get list of task executions with a large amount of task execution rows


gyugyu90:
**Description:**
We are using 2.11.2 version of SCDF. After migration, we cannot get list from `Task Executions` tab. It is `GET /tasks/executions` API that causes this issue, as it takes longer than our server's timeout, which is 60 seconds. It seems that time took for getting list takes longer as the number of rows of table AGGREGATE_TASK_EXECUTION and related tables grows.

Why don't you provide with lighter version of API just like `GET /jobs/thinexecutions` from `Job Executions` tab? 
I think it takes much time while fetching task arguments, which is not shown in the grid UI. We have over a million AGGREGATE_TASK_EXECUTION_PARAMS rows, and the application tries to search from it n times.

**Release versions:**
2.11.x

**Custom apps:**
N/A

**Steps to reproduce:**
Make a large number of Task Executions and try to `GET /task/executions`

**Screenshots:**
<img width="1884" alt="스크린샷 2024-02-29 오후 4 26 36" src="https://github.com/spring-cloud/spring-cloud-dataflow/assets/14208790/b306db72-3b07-4901-8818-fe2159ca3643">
<img width="1979" alt="스크린샷 2024-02-29 오후 4 25 27" src="https://github.com/spring-cloud/spring-cloud-dataflow/assets/14208790/450ef226-63af-4643-8f97-9eabcc59ab99">

**Additional context:**
We don't see this issue in our development environment which has around 1k rows of AGGREGATE_TASK_EXECUTION, but our production server has around 100k rows and can't get the list in time.

corneil:
There is already a change in snapshot for fixing this.
It will be in the form of an automatic migration.
If you are in a hurry you can apply the following, but remember to drop the indexes before deploying the next update.
```sql
create index BATCH_STEP_EXECUTION_JOB_EXECUTION_ID_IX on BATCH_STEP_EXECUTION(JOB_EXECUTION_ID);
create index BOOT3_BATCH_STEP_EXECUTION_JOB_EXECUTION_ID_IX on BOOT3_BATCH_STEP_EXECUTION(JOB_EXECUTION_ID);
create index BOOT3_TASK_TASK_BATCH_JOB_EXECUTION_ID_IX on BOOT3_TASK_TASK_BATCH(JOB_EXECUTION_ID);
[...]

PR #5736, opened 2024-03-12
task executor performance

corneil:
`TaskExecutionQueryTests` was created to test performance of the `tasks/execution` API call which does a lot of sub queries at the moment. The subqueries has been reduced.
Added `tasks/thinexecutions` which returns a thin resource that satisfies the UI page list.

This PR applies to #5735 and #5718


corneil:
The Oracle test is commented because it fails in spite of the migration script adding `SEGMENT CREATION IMMEDIATE`

cppwfs:
This PR is associated with #5735

corneil:
The schema migration for Oracle may not be required after all. I will have to do some test to see what breaks wirhout it in normal operation.

cppwfs:
Is this going to be used only for a quick listing of task executions.   i.e. task execution page?

spring-cloud-dataflow-server/src/test/java/org/springframework/cloud/dataflow/server/db/tasks/TaskExecutionQueryTests.java
```diff
@@ -0,0 +1,240 @@
...
+		}
+
+		private void add(Map<String, Map<String, Integer>> columnTypes, String tableName, String columnName, int type) {
+			Map<String, Integer> tableColumnTypes = columnTypes.computeIfAbsent(tableName, k -> new HashMap<>());
+            tableColumnTypes.put(columnName, type);
+		}
+
+
+		@Test
+		void queryWithLargeNumberOfTaskExecutions() throws Exception {
```
cppwfs, line 153: Not sure we should be testing speed in our Unit tests.  I think this is a test that is done to verify the change works, not one that is repeated for each release.

spring-cloud-dataflow-test/pom.xml
```diff
@@ -31,6 +31,16 @@
...
 			<artifactId>junit-jupiter</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework</groupId>
+			<artifactId>spring-jdbc</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.apache.commons</groupId>
+			<artifactId>commons-csv</artifactId>
+			<version>1.10.0</version>
```
cppwfs, line 41: If we decide to keep the test, we need to set the scope to test.

.github/workflows/ci-pr.yml
```diff
@@ -27,10 +27,6 @@ jobs:
       with:
         maven-version: 3.8.8
         maven-mirror: 'https://dlcdn.apache.org/maven/maven-3/'
-    - name: Setup Testcontainers Cloud Client
```
cppwfs, line 30: Just curious why we are removing this configuration in this PR.

spring-cloud-dataflow-server/src/test/java/org/springframework/cloud/dataflow/server/db/tasks/TaskExecutionQueryTests.java
corneil: The tests tasks a few seconds for each database and verifies a lot more

.github/workflows/ci-pr.yml
corneil: CI PR doesn't have access to secrets.

cppwfs: Understood.  We probably want to enumerate this in our commit message since it didn't fit the rest of the PR.

onobc: This should just be in a separate commit (it can be in this PR).

spring-cloud-dataflow-server/src/test/java/org/springframework/cloud/dataflow/server/db/tasks/TaskExecutionQueryTests.java
onobc: I think all of these performance tests should be repurposed outside of the CI runs.

spring-cloud-dataflow-test/pom.xml
onobc: Otherwise we increase the risk of potential hitch hiker CVEs.

spring-cloud-dataflow-server/src/test/java/org/springframework/cloud/dataflow/server/db/tasks/TaskExecutionQueryTests.java
corneil: Change to IT and created ci-performance.yml

.github/workflows/ci-pr.yml
corneil: Changed update to ci-pr.yml to separate commit

corneil:
Replaced

PR #5739, opened 2024-03-14
Give SCDF ability to compile and run tests to completion

cppwfs:
* Disable tests that need more investigation
* Refresh the TODO and Disable messages to reflect an accurate message of what is required
* These changes do not include those for skipper
* To compile to completion comment out spring-cloud-skipper module then execute mvn clean install

### DIFF
--- a/spring-cloud-dataflow-aggregate-task/src/main/java/org/springframework/cloud/dataflow/aggregate/task/impl/AggregateDataFlowTaskExecutionQueryDao.java
+++ b/spring-cloud-dataflow-aggregate-task/src/main/java/org/springframework/cloud/dataflow/aggregate/task/impl/AggregateDataFlowTaskExecutionQueryDao.java
@@ -17,6 +17,7 @@ package org.springframework.cloud.dataflow.aggregate.task.impl;
 
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.sql.Timestamp;
 import java.sql.Types;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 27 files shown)


spring-cloud-dataflow-aggregate-task/src/main/java/org/springframework/cloud/dataflow/aggregate/task/impl/AggregateDataFlowTaskExecutionQueryDao.java
```diff
@@ -484,11 +485,12 @@ public TaskExecution mapRow(ResultSet rs, int rowNum) throws SQLException {
 			if (rs.wasNull()) {
 				parentExecutionId = null;
 			}
+			Timestamp endTimestamp = rs.getTimestamp("END_TIME");
 			return new TaskExecution(id,
 					getNullableExitCode(rs),
 					rs.getString("TASK_NAME"),
 					rs.getTimestamp("START_TIME").toLocalDateTime(),
-					rs.getTimestamp("END_TIME").toLocalDateTime(),
+					(endTimestamp != null) ? rs.getTimestamp("END_TIME").toLocalDateTime() : null,
```
onobc, line 493: Why not use `endTimestamp` instead of `rs.getTimestamp("END_TIME")` as we already pulled that into `endTimestamp`?

onobc (approved):
LGTM @cppwfs

spring-cloud-dataflow-aggregate-task/src/main/java/org/springframework/cloud/dataflow/aggregate/task/impl/AggregateDataFlowTaskExecutionQueryDao.java
cppwfs: DOH!   Good catch

cppwfs:
Squashed, Merged.   Thanks @onobc for the Review!

PR #5740, opened 2024-03-18
improve simple job service tests

corneil:
Improved Failing `SimpleJobServiceMariadbTests` and `SimpleJobServicePostgresTests` that was running out of connections. Dirties Context was remove connection pool without closing connections.


cppwfs (approved):
LGTM

PR #5746, opened 2024-03-21
Update Data Flow to remove Schema Version Modules

cppwfs:
The goal of this PR was to give SCDF a BOOT3 only view of the world.  

There are 4 commits that may help in reviewing the changes.

1. Update DAOs not to use the Aggregate views
2. Update Tests not to use the schema versions
3. Remove AggregateJobQueryDAO
4. Update Explorer and configurations to no longer reflect the use of aggregates
5. Remove spring-cloud-dataflow-schema-core modules and where it used.

You will also see
1. Disabled removed from some tests.  We should be down to 7 tests that are disabled because we need to add query features
2. Updated SimpleJobServiceMariadbTests and SimpleJobServicePostgrestTests, to remove @DirtiesContext and increase connection pool size as pushed in the 2.11.x branch: https://github.com/spring-cloud/spring-cloud-dataflow/pull/5740


spring-cloud-dataflow-registry/src/test/java/org/springframework/cloud/dataflow/registry/service/DefaultAppRegistryServiceTests.java
```diff
@@ -552,20 +549,6 @@ public void testDeleteAll() throws URISyntaxException {
...
-		// then
-		List<AppRegistration> boot2 = result.stream().filter(r -> r.getBootVersion().equals(AppBootSchemaVersion.BOOT2)).collect(Collectors.toList());
-		List<AppRegistration> boot3 = result.stream().filter(r -> r.getBootVersion().equals(AppBootSchemaVersion.BOOT3)).collect(Collectors.toList());
-		assertEquals(1L, boot2.size());
-		assertEquals(1L, boot3.size());
-		assertEquals("2.0.1", boot2.get(0).getVersion());
-		assertEquals("3.0.0", boot3.get(0).getVersion());
-	}
 	@Test
 	public void testMultipleBootVersionsExpectError() {
```
onobc, line 553: I believe this test either is no longer valid OR it needs to be renamed. The corresponding test yml still references the now unsupported `bootVersion` property.

onobc (changes requested):
Looking good @cppwfs !

I am assuming the plan is to clean up the schema in a subsequent PR as this one is missing the schema migrations to remove the `boot_version` column from `app_registration`as well as I see joins that are still locking onto `schema_target`. 

One minor comment to address but consider this approved regardless of the outcome of that test. 

Thanks

spring-cloud-dataflow-aggregate-task/src/main/java/org/springframework/cloud/dataflow/composite/task/impl/DefaultCompositeTaskExplorer.java
```diff
@@ -45,8 +45,8 @@
  *
  * @author Corneil du Plessis
  */
-public class DefaultAggregateTaskExplorer implements AggregateTaskExplorer {
-	private final static Logger logger = LoggerFactory.getLogger(DefaultAggregateTaskExplorer.class);
+public class DefaultCompositeTaskExplorer implements CompositeTaskExplorer {
```
onobc, line 48: Do we intend on keeping this long term? If so, what is it providing?

cppwfs:
Squashed Merged.

Issue #5406, opened 2023-07-19
Add ttl support for task with status `COMPLETE`


seonwoo960000:
**Problem description:**
We use cloud functions to call spring cloud dataflow tasks. Because we are executing it frequently, there are a lot of completed pods that pile up. It's bit hard to monitor which pods are being created and completed using `k9s` if there is some operations as above. 
![image](https://github.com/spring-cloud/spring-cloud-dataflow/assets/69591622/acebc7ba-d8d1-4fea-ba9a-034c96f5947e)

**Solution description:**
I read in the documentation that we can add ttl for scheduled tasks using below property. 
```
deployer.<application>.kubernetes.cron.ttlSecondsAfterFinished=86400
```
However doesn't seem to work when we pass above property when launching a single task. Will it be possible to support `ttl` features when launching single task as well? 

**Description of alternatives:**
Possible solution I guess is to call API that cleans up the completed tasks periodically, but I still think that using ttl would be useful  because we wouldn't have to do API calls to remove completed tasks. 

**Additional context:**
-

seonwoo960000:
@corneil I've tries that but it `Completed` pods were not being removed. I solved this issue by scheduling a cloud function to call tasks `cleanup` API periodically.

dhpizza:
As [documentation ](https://docs.spring.io/spring-cloud-dataflow/docs/current/reference/htmlsingle/#_ttlsecondsafterfinished) says, this  `deployer.<application>.kubernetes.cron.ttlSecondsAfterFinished` works on **jobs** only i.e. on scheduled tasks only since it uses the k8s jobs ttl controller.

corneil:
@dhpizza You can configure `spring.cloud.deployer.kubernetes.ttlSecondsAfterFinished` with task deployment or configure it on the Data Flow Server deployment.

PR #5748, opened 2024-03-25
update k8s docs

corneil:
References: #5406

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/configuration-kubernetes.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/configuration-kubernetes.adoc
@@ -124,12 +124,36 @@ These properties are also used when configuring the <<configuration-kubernetes-t
 |The CPU limit, maximum needed value to allocate a pod
 |<none>
 
+|limits.ephemeral-storage
+|The ephemeral-storage limit, maximum needed value to allocate a pod.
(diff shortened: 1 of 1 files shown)


cppwfs (approved):
LGTM

spring-cloud-dataflow-docs/src/main/asciidoc/configuration-kubernetes.adoc
```diff
@@ -839,6 +863,15 @@ deployer.<application>.kubernetes.cron.ttlSecondsAfterFinished=86400
 ----
 ====
 
+The following example shows how you can individually configure application jobs:
+
+====
+[source,options=nowrap]
+----
+deployer.<application>.kubernetes.ttlSecondsAfterFinished=86400
```
onobc, line 871: What is this related to?

corneil: There was an addition done a while ago for cron.ttlSecondsAfterFinished and this was added but not documented. #5406

onobc: Got it. Even though it is small I would probably put this in its own commit in this PR for later find/perusal. I think it is not significant enough to change now though.

PR #5749, opened 2024-03-25
Provide for explicit thin results in queryForPageableResults

corneil:


spring-cloud-dataflow-aggregate-task/src/main/java/org/springframework/cloud/dataflow/aggregate/task/impl/AggregateDataFlowTaskExecutionQueryDao.java
```diff
@@ -490,7 +490,7 @@ public List<String> getTaskNames() {
 	@Override
 	public Page<AggregateTaskExecution> findAll(Pageable pageable) {
 		return queryForPageableResults(pageable, SELECT_CLAUSE, FROM_CLAUSE, null,
-				new MapSqlParameterSource(), getTaskExecutionCount());
+				new MapSqlParameterSource(), getTaskExecutionCount(), true);
```
cppwfs, line 493: I think we need to add a `findAll(Pageable pageable, boolean thinResults) ` that is used by our service that will set it to true.   And the `findAll(Pageable pageable) should set this thinResults to `false`.  That way the original behavior of the findAll is not compromised.   Thoughts?

corneil: Agreed

PR #5763, opened 2024-04-04
Updated representation of start time and date given timezone.

corneil:
I believe for Spring Batch to choose not to use ZonedDateTime but rather LocalDateTime is a mistake.
With the possibility that execution could originate on different machine with different local timezones all writing to the same database means that auditability is a potential minefield.


cppwfs (approved):
Looks good to me.  
Yes let's talk to Mahmoud.   This is something we have discussed in the far past.

cppwfs:
Rebased and Merged.   @corneil Thanks for the PR!

PR #5766, opened 2024-04-08
SCDF does not need to support JSR-352 as of Batch 5.

cppwfs:
Remove JobName search code for JSR-352, it is not used by Batch after 4.x

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
@@ -228,7 +228,8 @@ public class SimpleJobService implements JobService, DisposableBean {
 		Collection<JobExecution> result = jobExecutionDao.getRunningJobExecutions();
 		for (JobExecution jobExecution : result) {
 			try {
-				stopJobExecution(jobExecution);
+				jobExecution.getStepExecutions().forEach(StepExecution::setTerminateOnly);
+				jobExecution.setStatus( BatchStatus.STOPPING);
 				jobRepository.update(jobExecution);
 			} catch (Exception e) {
 				throw new IllegalArgumentException("The following JobExecutionId was not found: " + jobExecution.getId(), e);
@@ -236,18 +237,6 @@ public class SimpleJobService implements JobService, DisposableBean {
 		}
 
 		return result.size();
-	}
-
-	/**
-	 * Signal the {@link JobExecution} to stop. Iterates through the associated
-	 * {@link StepExecution}s, calling {@link StepExecution#setTerminateOnly()}.
-	 *
-	 */
-	public void stopJobExecution(JobExecution jobExecution) {
-		for (StepExecution stepExecution : jobExecution.getStepExecutions()) {
-			stepExecution.setTerminateOnly();
-		}
-		jobExecution.setStatus( BatchStatus.STOPPING);
 	}
 
 	@Override
@@ -259,8 +248,8 @@ public class SimpleJobService implements JobService, DisposableBean {
 
 		logger.info("Stopping job execution: " + jobExecution);
 
-			jobExecution.setStatus(BatchStatus.STOPPED);
-			jobRepository.update(jobExecution);
+		jobExecution.setStatus(BatchStatus.STOPPED);
+		jobRepository.update(jobExecution);
 		return jobExecution;
 
 	}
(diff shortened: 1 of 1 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
```diff
@@ -263,6 +238,18 @@ public int stopAll() {
...
 
+	/**
+	 * Signal the {@link JobExecution} to stop. Iterates through the associated
+	 * {@link StepExecution}s, calling {@link StepExecution#setTerminateOnly()}.
+	 *
+	 */
+	public void stopJobExecution(JobExecution jobExecution) {
+		for (StepExecution stepExecution : jobExecution.getStepExecutions()) {
+			stepExecution.setTerminateOnly();
+		}
```
onobc, line 249:
```suggestion
		jobExecution.getStepExecutions().forEach(StepExecution::setTerminateOnly);
```
This can be simplified a bit w/ the above. Wdyt?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
```diff
@@ -245,16 +226,10 @@ public int countJobs() {
...
-
 		for (JobExecution jobExecution : result) {
 			try {
-				if (jobNames.contains(jobExecution.getJobInstance().getJobName())) {
-					jobOperator.stop(jobExecution.getId());
-
-				} else {
-					throw new JobStopException(jobExecution.getId());
-				}
+				stopJobExecution(jobExecution);
```
onobc, line 231: Since nobody else is using `stopJobExecution` I would suggest one of the following:

1. Inline the code here and remove the method
2. Make the method private

I am in favor of the former.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
```diff
@@ -263,6 +238,18 @@ public int stopAll() {
...
+	/**
+	 * Signal the {@link JobExecution} to stop. Iterates through the associated
+	 * {@link StepExecution}s, calling {@link StepExecution#setTerminateOnly()}.
+	 *
+	 */
+	public void stopJobExecution(JobExecution jobExecution) {
+		for (StepExecution stepExecution : jobExecution.getStepExecutions()) {
+			stepExecution.setTerminateOnly();
+		}
+		jobExecution.setStatus( BatchStatus.STOPPING);
```
onobc, line 250: **[TinyNit]** Spacing and parens

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
```diff
@@ -484,8 +456,7 @@ private List<JobExecution> getJobExecutions(String jobName, BatchStatus status,
 	}
 
 	private void checkJobExists(String jobName) throws NoSuchJobException {
-		if (getJobNames().stream().anyMatch(e -> e.contains(jobName)) ||
-				jobInstanceDao.countJobInstances(jobName) > 0) {
+		if (jobInstanceDao.countJobInstances(jobName) > 0) {
```
onobc, line 459: I know it was like this before but I think it would be more readable if we inverted and if the count is`<= 0` throw the ex since its a void return the other branch falls off.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
```diff
@@ -272,15 +259,8 @@ public JobExecution stop(Long jobExecutionId) throws NoSuchJobExecutionException
 
 		logger.info("Stopping job execution: " + jobExecution);
 
-		Collection<String> jobNames = getJobNames();
-
-		if (jobNames.contains(jobExecution.getJobInstance().getJobName())) {
-			jobOperator.stop(jobExecutionId);
-			jobExecution = getJobExecution(jobExecutionId);
-		} else {
 			jobExecution.setStatus(BatchStatus.STOPPED);
```
onobc, line 262: **[TinyNit]** Indentations

onobc (changes requested):
A few minor comments to address but other than that LGTM. Thanks @cppwfs .

Any tests need to be removed or updated?

cppwfs:
Good question on the tests.   
No tests failed for the removal of this feature.   i.e. we didn't test for it originally.   (at least none that I could find).

onobc (approved):
Thanks for the updates @cppwfs - LGTM!

cppwfs:
Rebased Squashed Merged
Thank you @onobc for the review!

PR #5781, opened 2024-04-17
Added taskExecutionStatus to TaskExecutionThinResource.

corneil:
The synthesized method did not make it into the TaskExecutionThinResource.
There will be an issue with CTR tasks. We need to determine update the new thin query to collect the CTR status.


cppwfs (approved):
LGTM

cppwfs:
Is this for https://github.com/spring-cloud/spring-cloud-dataflow/pull/5781?

PR #5788, opened 2024-04-26
Update Spring Framework to 5.3.34

corneil:
Update Spring Security to 5.7.12
Update Spring Shell to 2.1.15

Only known snake-yaml and spring-web remain.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -50,11 +50,10 @@
 		</developer>
 	</developers>
 	<modules>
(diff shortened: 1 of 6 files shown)


corneil:
Placing spring-cloud-dataflow-parent first in project module list does somehow ensure it is always resolved before spring-cloud-dataflow-build and properly overrides the spring-boot-parent in dependency resolution

PR #5793, opened 2024-05-03
fix documentation for boot 3.x

corneil:
Remove documentation for boot version and schemaTarget.
Update for newer version of rest docs.


cppwfs (changes requested):
Just a small nitpick on adding the comment.

PR #5801, opened 2024-05-09
Disabling DB2 tests by default.

corneil:
Add job to integration test to explicitly test with DB2.


onobc (approved):
Nice work @corneil - LGTM

Issue #5780, opened 2024-04-17
Vulnerabilities affecting Spring Cloud Dataflow dependencies 


shalomyasap:
Running a vulnerabilities scan through the Spring Cloud Dataflow server returns some CVEs affecting the latest release:

1. CVE-2024-23672 - tomcat-embed-websocket-9.0.83.jar 
2. CVE-2024-24549 - tomcat-embed-core-9.0.83.jar 
3. CVE-2024-22257 - spring-security-core-5.7.6.jar 
4. CVE-2024-29025 - netty-codec-http-4.1.101.Final.jar 
5. CVE-2023-52428 - nimbus-jose-jwt-9.22.jar 
6. CVE-2024-31033 - jjwt-impl-0.11.2.jar 
7. CVE-2024-22262 - spring-web-5.3.31.jar 
8. CVE-2016-1000027 - spring-web-5.3.31.jar 

Could you confirm whether the App is affected by these vulnerabilities and if so, are there plans to update the related dependencies and release it soon?

This reference to latest release v2.11.2

Many thanks,
Shalom

corneil:
2.11.3-SNAPSHOT is updating to Spring Framework 5.3.33
The only one that will remain is [CVE-2016-1000027](https://github.com/advisories/GHSA-4wrc-f8pq-fpqp) with mitigation described [here](https://github.com/spring-cloud/spring-cloud-dataflow/security/advisories/GHSA-5685-fp7w-h5cr).

shalomyasap:
Thank you for the update. 
When is version 2.11.3-SNAPSHOT expected to be released?

PR #5803, opened 2024-05-09
Fix Tomcat embedded dep. mgmt.

onobc:
Update the dep. mgmt. entries for Tomcat embedded as the previous commit used the wrong groupId (missing `.embed` suffix).

See #5780


corneil (approved):
Great catch!

PR #5809, opened 2024-05-13
[DO_NOT_MERGE] Update scripts so that they are compatible with bash and zsh.

corneil:


cppwfs:

PR #5811, opened 2024-05-14
Add trailing slash to dashboard

jvalkeal:
- Boot 3.x changed handling of trailing slash meaning it's now adviced to define both /dashboard and /dashboard/ in UiController

Essentially to fix latter:
```
jvalkealahti@cypher:/tmp/xxx$ http --headers http://localhost:9393/dashboard
HTTP/1.1 302 
Connection: keep-alive
Content-Language: en-GB
Content-Length: 0
Date: Mon, 13 May 2024 12:36:17 GMT
Keep-Alive: timeout=60
Location: http://localhost:9393/dashboard/index.html

jvalkealahti@cypher:/tmp/xxx$ http --headers http://localhost:9393/dashboard/
HTTP/1.1 500 
Connection: close
Content-Type: application/hal+json
Date: Mon, 13 May 2024 12:36:21 GMT
Transfer-Encoding: chunked
Vary: Origin
Vary: Access-Control-Request-Method
Vary: Access-Control-Request-Headers
```

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/UiController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/UiController.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2017 the original author or authors.
+ * Copyright 2015-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -27,7 +27,7 @@ import org.springframework.web.bind.annotation.RequestMapping;
  * @author Gunnar Hillert
  */
 @Controller
-@RequestMapping(UiController.WEB_UI_INDEX_PAGE_ROUTE)
+@RequestMapping({ UiController.WEB_UI_INDEX_PAGE_ROUTE, UiController.WEB_UI_INDEX_PAGE_ROUTE + "/" })
 public class UiController {
 
 	public static final String WEB_UI_INDEX_PAGE_ROUTE = "/dashboard";


corneil (approved):
LGTM

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/UiController.java
```diff
@@ -27,7 +27,7 @@
  * @author Gunnar Hillert
  */
 @Controller
-@RequestMapping(UiController.WEB_UI_INDEX_PAGE_ROUTE)
+@RequestMapping({ UiController.WEB_UI_INDEX_PAGE_ROUTE, UiController.WEB_UI_INDEX_PAGE_ROUTE + "/" })
```
onobc, line 30: Because this approach only fixes this controller, until we adjust globally ([recommended by the release notes](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.0.0-M4-Release-Notes#spring-mvc-and-webflux-url-matching-changes)) I would rather set the default as it was before using something like:

```java
@Configuration
public class WebConfiguration implements WebMvcConfigurer {

    @Override
    public void configurePathMatch(PathMatchConfigurer configurer) {
      configurer.setUseTrailingSlashMatch(true);
    }
  
}
```
Wdyt?

onobc (changes requested):
Thanks @jvalkeal - please see my comment.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/UiController.java
jvalkeal: @onobc sorry I missed your question. I'd rather not to set it globally because `setUseTrailingSlashMatch` is already deprecated. We anyway need to handle this in a recommended way going forward.

onobc: > @onobc sorry I missed your question

No worries @jvalkeal. 

> I'd rather not to set it globally because setUseTrailingSlashMatch is already deprecated. We anyway need to handle this in a recommended way going forward.

If we don't fix it globally then maybe we should not fix the dashboard either. Since this is a major update we can just list this as a breaking change.

jvalkeal: We'd have working redirect without slash but not with slash which would not make sense not to fix(as this may be only case):

```
$ http http://localhost:9393/dashboard
HTTP/1.1 302 
Connection: keep-alive
Content-Language: en-GB
Content-Length: 0
Date: Thu, 13 Jun 2024 13:35:46 GMT
Keep-Alive: timeout=60
Location: http://localhost:9393/dashboard/index.html
```
vs.
```
$ http http://localhost:9393/dashboard/
HTTP/1.1 500 
Connection: close
Content-Type: application/hal+json
Date: Thu, 13 Jun 2024 13:35:49 GMT
Transfer-Encoding: chunked
Vary: Origin
Vary: Access-Control-Request-Method
Vary: Access-Control-Request-Headers

{
    "_embedded": {
        "errors": [
            {
                "_links": {
                    "self": {
                        "href": "/"
                    }
                },
                "logref": "NoResourceFoundException",
                "message": "No static resource dashboard."
            }
        ]
    }
}
```

onobc: Sure, this fixes this "dashboard" but we may get reports from other endpoints (REST API users) etc..

Those cases are probably less likely to be an issue though. 

This is probably fine to fix in the "dashboard" only and if we get reports on other endpoints we can decide what to do at that point. 

Thanks for the discussion @jvalkeal

onobc (approved):

PR #5813, opened 2024-05-15
Fix JdbcAggregateJobQueryDao 

corneil:
Current users who utilize the thinExecutionController will receive a status of `COMPLETE` when a child task or the CTR fails.    The state should be `FAILED`.   The following fix will resolve this issue as enumerated below:
Fix `JdbcAggregateJobQueryDao` to read `AGGREGATE_JOB_EXECUTION.EXIT_CODE` instead of `EXIT_MESSAGE`.


cppwfs (changes requested):

cppwfs:
LGTM
Polish, squash, merge.

Thanks @corneil

PR #5815, opened 2024-05-16
Updated pro deployment for local

corneil:
The scripts for local k8s PRO deployment had 1.6.1-SNAPSHOT instead of 1.6.2-SNAPSHOT

### DIFF
--- a/src/deploy/k8s/deploy-scdf.sh
+++ b/src/deploy/k8s/deploy-scdf.sh
@@ -14,7 +14,7 @@ if [ ! -d "$K8S" ]; then
 fi
 PARENT=$(realpath "$SCDIR/../../..")
 if [ "$DATAFLOW_PRO_VERSION" = "" ]; then
-  DATAFLOW_PRO_VERSION=1.6.1-SNAPSHOT
+  export DATAFLOW_PRO_VERSION=1.6.2-SNAPSHOT
 fi
 if [ "$DATAFLOW_VERSION" = "" ]; then
   export DATAFLOW_VERSION=2.11.3-SNAPSHOT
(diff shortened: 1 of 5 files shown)


src/deploy/k8s/deploy-scdf.sh
```diff
@@ -137,12 +137,11 @@ if [ "$K8S_DRIVER" != "tmc" ] && [ "$K8S_DRIVER" != "gke" ]; then
 
     if [ "$USE_PRO" = "true" ]; then
         sh "$SCDIR/load-image.sh" "dev.registry.tanzu.vmware.com/p-scdf-for-kubernetes/scdf-pro-server:$DATAFLOW_PRO_VERSION" true
-#        if [[ "$DATAFLOW_PRO_VERSION" == *"1.6"* ]]; then
-#            sh "$SCDIR/load-image.sh" "dev.registry.tanzu.vmware.com/p-scdf-for-kubernetes/scdf-pro-skipper:$DATAFLOW_PRO_VERSION" true
-#
-#        else
+        if [[ "$DATAFLOW_PRO_VERSION" == *"1.6"* ]]; then
+            sh "$SCDIR/load-image.sh" "dev.registry.tanzu.vmware.com/p-scdf-for-kubernetes/scdf-pro-skipper:$DATAFLOW_PRO_VERSION" true
```
cppwfs, line 141: So this change is so that we use the PRO instead of the Skipper OSS version if the dataflow is 1.6?  Is there a way to check for 1.6 or greater?   So we don't have to update it in the future.  Thoughts?

cppwfs (changes requested):

Issue #5779, opened 2024-04-16
Investigate the Maven Develocity plugin


onobc:
cc: @corneil

onobc:
@corneil can you add some details here? Thanks

corneil:
We will be configuring build to use our Gradle Enterprise instance and a profile to enable/disable build scan etc.

PR #5821, opened 2024-05-21
Add develocity configuration.

corneil:
Issue #5779

We will add profile for doing buildscan.

### DIFF
--- a/.mvn/develocity.xml
+++ b/.mvn/develocity.xml
@@ -1,0 +1,30 @@
+<develocity
+		xmlns="https://www.gradle.com/develocity-maven" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+		xsi:schemaLocation="https://www.gradle.com/develocity-maven https://www.gradle.com/schema/develocity-maven.xsd">
+	<server>
+		<url>https://ge.spring.io</url>
+	</server>
+	<buildScan>
+		<backgroundBuildScanUpload>false</backgroundBuildScanUpload>
+		<captureGoalInputFiles>true</captureGoalInputFiles>
+		<publishIfAuthenticated>true</publishIfAuthenticated>
+		<obfuscation>
+			<ipAddresses>#{{'0.0.0.0'}}</ipAddresses>
+		</obfuscation>
+	</buildScan>
+	<buildCache>
+		<local>
+			<enabled>true</enabled>
+		</local>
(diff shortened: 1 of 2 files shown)


.mvn/develocity.xml
```diff
@@ -0,0 +1,30 @@
+<develocity
+		xmlns="https://www.gradle.com/develocity-maven" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+		xsi:schemaLocation="https://www.gradle.com/develocity-maven https://www.gradle.com/schema/develocity-maven.xsd">
+	<server>
+		<url>https://ge.spring.io</url>
+	</server>
+	<buildScan>
+		<backgroundBuildScanUpload>false</backgroundBuildScanUpload>
```
cppwfs, line 8: Is this something we want to run on a schedule or every time?
Have we tested how much time this will add to our CI builds?

.mvn/develocity.xml
```diff
@@ -0,0 +1,30 @@
...
+		<captureGoalInputFiles>true</captureGoalInputFiles>
```
cppwfs, line 9: Do we want to set the  `projectId`  i.e.  `<projectId>SC-DataFlow</projectId>`

.mvn/develocity.xml
```diff
@@ -0,0 +1,30 @@
...
+		<publishIfAuthenticated>true</publishIfAuthenticated>
+		<obfuscation>
+			<ipAddresses>#{{'0.0.0.0'}}</ipAddresses>
+		</obfuscation>
+	</buildScan>
```
cppwfs, line 14: Do we want to have this on our CI Builds or also on releases?   If so do we want to use tags?
```
    <tags>
      <tag>my tag</tag>
    </tags>
```

cppwfs (changes requested):
Do we have an issue for the addition of this feature?   
That way we can explain why it is being added. 

Thanks!

Correction... I see the issue.  Apologies for the noise.

onobc:
> Issue #5779
> 
> We will add profile for doing buildscan.

This may be helpful to the reviewer: https://docs.gradle.com/develocity/maven-extension/current/

cppwfs:
We probably want to have a chat on how we want to configure it.   I think that is the gist of my questions, "how should this be configured?"

PR #5822, opened 2024-05-21
Add spring-cloud-dataflow-server to version-info.

corneil:
This adds an entry to version-info that the UI can display to ensure OSS and PRO UI shows product name and version differently.

In Pro the entry has the following:

```yaml
    spring-cloud-dataflow-server:
        name: Spring Cloud Dataflow Pro Server

```            
in OSS:

```yaml
    spring-cloud-dataflow-server:
        name: Spring Cloud Dataflow Server

```

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
+++ b/spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
@@ -94,6 +94,9 @@ spring:
         dependency-fetch:
           enabled: false
         dependencies:
+          spring-cloud-dataflow-server:
+            name: Spring Cloud Dataflow
+            version: "@project.version@"
           spring-cloud-dataflow-dashboard:
             name: Spring Cloud Dataflow UI
             version: "@spring-cloud-dataflow-ui.version@"


spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
```diff
@@ -228,7 +231,7 @@ spring:
             - GET    /tasks/info/*                   => hasRole('ROLE_VIEW')
 
             - GET    /tasks/thinexecutions           => hasRole('ROLE_VIEW')
-
+            - GET    /tasks/thinexecutions/*         => hasRole('ROLE_VIEW')
```
cppwfs, line 234: This looks like a bug fix.  This probably needs to be in a separate PR with an explanation as to why it is needed.

onobc: Yes please

spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
```diff
@@ -94,6 +94,9 @@ spring:
         dependency-fetch:
           enabled: false
         dependencies:
+          spring-cloud-dataflow-server:
+            name: Spring Cloud Dataflow
+            version: "@project.version@"
```
onobc, line 99: Should this be `@dataflow.version@` ?

spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
```diff
@@ -94,6 +94,9 @@ spring:
...
           spring-cloud-dataflow-dashboard:
```
onobc, line 100: Where these entries in PRO required since they are the same here (OSS)? https://github.com/pivotal/scdf-pro/blob/1483cf2b965dbc4e0d7be3da3a8ada2aff6473d3/scdf-pro-server/src/main/resources/application.yml#L19-L30

onobc (changes requested):
@corneil I have a couple of comments/questions. Other than that, please remove the [bug fix](https://github.com/spring-cloud/spring-cloud-dataflow/pull/5822/files#r1608390980) and create separate PR (or at the **very least** a separate commit describing what the change is).

spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
corneil: It will only build `spring-cloud-dataflow-server-core` in the scope of `spring-cloud-dataflow` and the project.version will be the same as `dataflow.version`

spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
corneil: Yes. The properties from the 2 files won't be merged.

spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
onobc: As long as you have tested both OSS and PRO - this is fine.

onobc (approved):
LGTM - thanks @corneil

cppwfs (approved):

Issue #5819, opened 2024-05-21
Failure to get Application properties for images with OCI manifest


vicSenior:
###  **Description:**
My purpose is to create SCDF Streams of Applications that make use of customized Docker images.
All custom Docker images have been customized with application properties defined in the json `org.springframework.cloud.dataflow.spring-configuration-metadata.json`, declared with LABEL in DockerFile.

So, suppose that a Stream of Applications was declared as `http | my-custom-app | log` pipe.
In detail:
- `http` point to official SCDF image `docker:springcloudstream/http-source-kafka:3.2.1`
- `my-custom-app` point to image stored in Private Registry `docker:my-quay-registry.local/poc-scdf/my-custom-app:0.0.1`
- `log` point to official SCDF image `docker:springcloudstream/log-sink-kafka:3.2.1`

The problem arises when SCDF tries to collect `.config` information from `my-custom-app` manifest, because `my-custom-app` has been published (using Azure DevOps automation) in Quay private Registry using OCI Standard (so mediaType manifest is `vnd.oci.image.manifest.v1+json`).

In fact, looking at the source code, I understood that Rest http GET calls are performed using (always) the header `Accept:application/vnd.docker.container.image.v1+json`.
That call, performed for an image published with Standard OCI, returns a manifest with `SchemaVersion=1` (that is devoid of the `.config` object).

### **Follow the smoke trail:**
The exception `ContainerRegistryException` reported by the SCDF `server` component is the following:
```
2024-05-17 16:22:30.068  WARN 1 --- [io-8080-exec-10] ApplicationConfigurationMetadataResolver : Failed to retrieve properties for resource Docker Resource [docker:my-quay-registry.local/poc-scdf/my-custom-app:0.0.1] because of ContainerRegistryException: Image [my-quay-registry.local/poc-scdf/my-custom-app:0.0.1] has incorrect or missing manifest config element: {tag=0.0.1, name=poc-scdf/my-custom-app, architecture=amd64, schemaVersion=1, history=[{v1Compatibility={"created": "2024-05-16T13:20:41.393484176Z", "architecture": "amd64", "os": "linux", "config": {"User": "default", "Env": ["container=oci", "GECOS=JBoss user", "HOME=/home/jboss", ...
``` 
which is raised on line 55 of the `DefaultContainerImageMetadataResolver.java` file of method `getImageLabels(String imageName)`:
```java
// DefaultContainerImageMetadataResolver.java
[...]

PR #5823, opened 2024-05-22
Try to resolve manifest with OCI manifest type when config is absent

corneil:
Issue #5819

### DIFF
--- a/spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/container/DefaultContainerImageMetadataResolver.java
+++ b/spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/container/DefaultContainerImageMetadataResolver.java
@@ -20,6 +20,7 @@ import java.util.Collections;
 import java.util.Map;
 
 import org.springframework.cloud.dataflow.container.registry.ContainerRegistryException;
+import org.springframework.cloud.dataflow.container.registry.ContainerRegistryProperties;
 import org.springframework.cloud.dataflow.container.registry.ContainerRegistryRequest;
 import org.springframework.cloud.dataflow.container.registry.ContainerRegistryService;
 import org.springframework.util.StringUtils;
@@ -30,6 +31,7 @@ import org.springframework.util.StringUtils;
  *
  * @author Christian Tzolov
  * @author Ilayaperumal Gopinathan
+ * @author Corneil du Plessis
  */
 public class DefaultContainerImageMetadataResolver implements ContainerImageMetadataResolver {
 
@@ -39,6 +41,7 @@ public class DefaultContainerImageMetadataResolver implements ContainerImageMeta
 		this.containerRegistryService = containerRegistryService;
 	}
 
+	@SuppressWarnings("unchecked")
 	@Override
 	public Map<String, String> getImageLabels(String imageName) {
 
@@ -48,12 +51,23 @@ public class DefaultContainerImageMetadataResolver implements ContainerImageMeta
 
 		ContainerRegistryRequest registryRequest = this.containerRegistryService.getRegistryRequest(imageName);
 
-		Map manifest = this.containerRegistryService.getImageManifest(registryRequest, Map.class);
+		Map<String, Object> manifest = this.containerRegistryService.getImageManifest(registryRequest, Map.class);
 
-		if (manifest != null && !isNotNullMap(manifest.get("config"))) {
-			throw new ContainerRegistryException(
-					String.format("Image [%s] has incorrect or missing manifest config element: %s",
-							imageName, manifest.toString()));
+		if (manifest != null && manifest.get("config") == null) {
+			// when both Docker and OCI images are stored in repository the response for OCI image when using Docker manifest type will not contain config.
+			// In the case where we don't receive a config and schemaVersion is less than 2 we try OCI manifest type.
+			String manifestMediaType = registryRequest.getRegistryConf().getManifestMediaType();
+			if (asInt(manifest.get("schemaVersion")) < 2
+					&& !manifestMediaType.equals(ContainerRegistryProperties.OCI_IMAGE_MANIFEST_MEDIA_TYPE)) {
+				registryRequest.getRegistryConf()
+					.setManifestMediaType(ContainerRegistryProperties.OCI_IMAGE_MANIFEST_MEDIA_TYPE);
+				manifest = this.containerRegistryService.getImageManifest(registryRequest, Map.class);
+			}
+			if (manifest.get("config") == null) {
+				String message = String.format("Image [%s] has incorrect or missing manifest config element: %s",
+						imageName, manifest);
+				throw new ContainerRegistryException(message);
+			}
 		}
 		if (manifest != null) {
 			String configDigest = ((Map<String, String>) manifest.get("config")).get("digest");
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/container/DefaultContainerImageMetadataResolver.java
```diff
@@ -85,12 +99,24 @@ public Map<String, String> getImageLabels(String imageName) {
...
 		}
 		else {
-			throw new ContainerRegistryException(
-					String.format("Image [%s] is missing manifest", imageName));
+			throw new ContainerRegistryException(String.format("Image [%s] is missing manifest", imageName));
+		}
+	}
+
+	private static int asInt(Object value) {
+		if (value instanceof Number) {
```
cppwfs, line 107: Probably want to check for a null.

spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/container/DefaultContainerImageMetadataResolver.java
```diff
@@ -48,12 +51,23 @@ public Map<String, String> getImageLabels(String imageName) {
...
-			throw new ContainerRegistryException(
-					String.format("Image [%s] has incorrect or missing manifest config element: %s",
-							imageName, manifest.toString()));
+		Map<String, Object> manifest = this.containerRegistryService.getImageManifest(registryRequest, Map.class);
+
+		if (manifest != null && manifest.get("config") == null) {
+			// when both Docker and OCI images are stored in repository the response for OCI image when using Docker manifest type will not contain config.
+			// In the case where we don't receive a config and schemaVersion is less than 2 we try OCI manifest type.
+			String manifestMediaType = registryRequest.getRegistryConf().getManifestMediaType();
+			if (asInt(manifest.get("schemaVersion")) < 2
```
cppwfs, line 60: Schema version maybe removed in the future so we need to handle a null condition.  https://github.com/opencontainers/image-spec/blob/main/manifest.md

spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/container/DefaultContainerImageMetadataResolver.java
```diff
@@ -48,12 +51,23 @@ public Map<String, String> getImageLabels(String imageName) {
...
+					&& !manifestMediaType.equals(ContainerRegistryProperties.OCI_IMAGE_MANIFEST_MEDIA_TYPE)) {
```
cppwfs, line 61: Is there another media type for schemaVersion == 1?

spring-cloud-dataflow-configuration-metadata/src/test/java/org/springframework/cloud/dataflow/container/registry/DefaultContainerImageMetadataResolverTest.java
```diff
@@ -210,6 +211,25 @@ public void getImageLabelsWithInvalidLabels() throws JsonProcessingException {
 		assertThat(labels).isEmpty();
 	}
 
+	@Test
+	public void getImageLabelsWithMixedOCIResponses() throws JsonProcessingException {
```
cppwfs, line 215: Probably want to have a negative test in case the config is missing.

spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/container/DefaultContainerImageMetadataResolver.java
corneil: `value instanceof Number` is never true when `value` is `null`

spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/container/DefaultContainerImageMetadataResolver.java
corneil: `asInt` will return 0 when `schemaVersion` is not present.

spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/container/DefaultContainerImageMetadataResolver.java
corneil: There are only 2 media types. The default is Docker manifest. If it doesn't find the config then switching to accepting the OCI manifest should retrieve the info.

spring-cloud-dataflow-configuration-metadata/src/test/java/org/springframework/cloud/dataflow/container/registry/DefaultContainerImageMetadataResolverTest.java
corneil: An exception is thrown if the config is missing.

spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/container/DefaultContainerImageMetadataResolver.java
cppwfs: For now this field is required.    But it says that it could be removed in the future.   So for now a null is invalid, but in the future a null may indicate something different.   So would this indicate an exception be thrown.

spring-cloud-dataflow-configuration-metadata/src/test/java/org/springframework/cloud/dataflow/container/registry/DefaultContainerImageMetadataResolverTest.java
cppwfs: Wouldn't we want a test for that?

spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/container/DefaultContainerImageMetadataResolver.java
cppwfs: Gotcha !   Thanks!

spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/container/DefaultContainerImageMetadataResolver.java
corneil: If there is no schemaVersion or schemaVersion is 1 we want to try the other manifest type query.

spring-cloud-dataflow-configuration-metadata/src/main/java/org/springframework/cloud/dataflow/configuration/metadata/container/DefaultContainerImageMetadataResolver.java
corneil: If we cannot read the manifest we try something else. If that fails we throw an exception.

spring-cloud-dataflow-configuration-metadata/src/test/java/org/springframework/cloud/dataflow/container/registry/DefaultContainerImageMetadataResolverTest.java
corneil: If happens when people push OCI and Docker containers to the same registry. With Quay it is the case. Data flow configuration provides for specifying a manifest type to accept but when the registry has both type it fails on containers of the other type. So users should not configure a manifest type and we will use docker manifest type first and if that fails we use OCI manifest type to obtain the sha. Then the labels are retrieved using the sha.

vicSenior:
Guys, this alternative algorithm could be interesting. 

https://github.com/spring-cloud/spring-cloud-dataflow/issues/5819#issuecomment-2133862495 

It's a strategy that allows SCDF to overcome future problems if new Manifest mediaType standards were to be defined ( ... and it doesn't use the `schemaVersion` 😄  )

cppwfs (approved):

cppwfs:
Merged and cherry-picked to main-3.

Thanks for your work on this and the discussion with me on it.

PR #5826, opened 2024-05-23
Update scope of spring-boot-starter-test

corneil:
Update scope of spring-boot-starter-test for spring-cloud-dataflow-aggregate-task.


onobc (approved):
LGTM - thanks @corneil

PR #5831, opened 2024-05-30
the document plugin needs to use the current spring javadoc location

cppwfs:
It was set to use a configurable spring version that was not set in the main-3.   When ever possible we want to use the BOM or current when determining versions.

### DIFF
--- a/spring-cloud-dataflow-build/pom.xml
+++ b/spring-cloud-dataflow-build/pom.xml
@@ -90,6 +90,8 @@
 		<maven-dependency-plugin-for-guides.phase>generate-resources</maven-dependency-plugin-for-guides.phase>
 		<groups />
 		<excludedGroups>slow,docker</excludedGroups>
+		<!--Specific Spring framework javadoc version for attach-javadocs plugin-->
+		<spring-javadoc.version>6.1.3</spring-javadoc.version>
 	</properties>
 	<dependencyManagement>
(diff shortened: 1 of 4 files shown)


corneil (changes requested):
`spring-boot-dependencies` has a property named `spring-framework.version`

cppwfs:
@corneil I don't think that will work.   That is for overriding the dependency in the bom.

spring-cloud-dataflow-build/pom.xml
```diff
@@ -90,6 +90,8 @@
 		<maven-dependency-plugin-for-guides.phase>generate-resources</maven-dependency-plugin-for-guides.phase>
 		<groups />
 		<excludedGroups>slow,docker</excludedGroups>
+		<!--Specific Spring framework javadoc version for attach-javadocs plugin-->
+		<spring-javadoc.version>6.1.3</spring-javadoc.version>
```
onobc, line 94: I think `javadoc-spring.version` might be clearer, plus if we end up needing to add another javadoc version property, the prefix keeps them "close" together. Wdyt?

onobc (approved):
LGTM - thanks @cppwfs

corneil (approved):
Let's fly.

cppwfs:
Squashed Merged

PR #5838, opened 2024-06-10
Migrate CTR to Boot 3.3 and Batch 5

cppwfs:
* Remove the Batch Configurer and replace with a Configuration
* BeanPostProcessor has been added so that CTR can use its jobRepository vs. the one provided by BatchAutoConfiguration
* BeanPostProcessor has been added so that CTR can use its own JobLauncher.  
    * This is because CTR needs a  ThreadPoolTaskExecutor vs the SyncTaskExecutor used by default.
* Update Tests so that they will work with Boot3
* Removed EnableBatchAutoConfiguration no longer needed if using BatchAutoConfiguration
* Removed schema code since this release focusses on Boot3


cppwfs:
Closing this PR to create a new one where we have 2 commits.  One with the base conversion, the other containing the elements that allow us to override the JobLauncher and JobRepository.

PR #5839, opened 2024-06-10
Migrate CTR to Boot 3.x and Batch 5.x

cppwfs:
* Commit: [cecbb1d](https://github.com/spring-cloud/spring-cloud-dataflow/commit/cecbb1d8a4223bd8626e83a508aa26eab7a5e717 ) Has the base changes to migrate to  Boot3/Batch5 
* Commit: [46ad2e1](https://github.com/spring-cloud/spring-cloud-dataflow/commit/46ad2e14dbc02250ae0ee926e186174e1d0e80c5) Contains Bean Post Processor to give the ability to override the JobRepository.


cppwfs:
I think we should proceed with this PR (after comments are resolved), but I want to revisit the JobLauncher here.   I think we are still using the SyncTaskExecutor, which will work for now and will let us proceed to let us finish the build in this branch.   
I will create a story on trello to revisit this.

cppwfs:
Update: 
We don't need to override the JobLauncher.   We can use the existing behavior.  The executor I was worried about is not used by the JobLauncher.

spring-cloud-dataflow-composed-task-runner/src/test/java/org/springframework/cloud/dataflow/composedtaskrunner/properties/ComposedTaskPropertiesTests.java
```diff
@@ -127,12 +127,12 @@ public void testComposedTaskAppArguments() {
 	@Test
 	public void testAssignmentOfOauth2ClientCredentialsClientAuthenticationMethod(){
 		this.contextRunner
-				.withSystemProperties("OAUTH2_CLIENT_CREDENTIALS_CLIENT_AUTHENTICATION_METHOD=POST")
+				.withSystemProperties("OAUTH2_CLIENT_CREDENTIALS_CLIENT_AUTHENTICATION_METHOD=client_secret_post")
```
klopfdreh, line 130: Just a small question - why don't you use
```java
.withSystemProperties("OAUTH2_CLIENT_CREDENTIALS_CLIENT_AUTHENTICATION_METHOD="+ClientAuthenticationMethod.CLIENT_SECRET_POST.value())
```
so that the value of the property is also derived from the enum?

cppwfs: @klopfdreh Good point.  There will be a PR coming up soon updating the security in this branch!   Stay tuned!

onobc: https://github.com/spring-cloud/spring-cloud-dataflow/issues/5883

onobc:
**TODO** Close once we decide if it is to port forward to `3.0.x`

PR #5842, opened 2024-06-12
Remove thinTaskExecutionList from TaskTemplate

cppwfs:
ThinTaskExecutions are not a part of SCDF-3.0.x.
During the migration some of the ThinTaskExecution URL calls were left in the rest client and this caused some test failures in IT. This PR removes the API access from the client for ThinTaskExecutions.


onobc (approved):
LGTM - ty @cppwfs

PR #5845, opened 2024-06-19
Update Jackson to 2.17.1

onobc:


corneil (approved):
LGTM

spring-cloud-dataflow-parent/pom.xml
```diff
@@ -45,6 +45,7 @@
 		<snappy-java.version>1.1.10.5</snappy-java.version>
 		<commons-compress.version>1.26.1</commons-compress.version>
 		<commons-io.version>2.15.1</commons-io.version>
+		<jackson-bom.version>2.17.1</jackson-bom.version>
```
cppwfs, line 48: LGTM but this is a pretty big bump in version from the bom.

onobc: Yep, however, the CVE is only fixed in `2.15.x` (we are coming from `2.13.x`). 
So if we decide to fix this CVE we will have to jump to `2.15.x` and at that point we might as well jump to `2.17.1` (latest).
I scanned the release notes and did not see anything concerning. My thinking is if the tests all pass then we should be good to go.

**[MyBad]** I should have added this in the PR description as it was definitely something I researched and thought about before doing.

cppwfs (approved):
LGTM.  Thank you @onobc !!!!

cppwfs:
LGTM!!!
Rebased Merged!
Thanks @onobc for the great work!

PR #5850, opened 2024-06-26
Serialize Job Parameters to the command Line Arguments on Job Restarts per Batch Standards

cppwfs:
### What this PR is trying to do
Currently Spring Batch allows users to specify 2 formats for accepting Job parameters via the command line.  This PR allows SCDF to use both formats, where before it was only using the default.
* Default: `parameterValue,parameterType,identificationFlag`
* JSON

### This PR is initially broken up into 4 commits

1. Updates DefaultJobService to accept dataflow global configuration for which format should be use as its default.
     a. Update the controller so that user can specify a different format for a specific launch instead of using dataflow's configured default
     b. Update the model so if the user clicks the link for a restart it will use dataflow's configured default. 
     c. Notice some of the old tests had to be updated because the encoder provided by batch provides the isIdentifying field.  
2. Update the restful docs so can demonstrate the use  of `useJsonJobParameters`
3. Update shell and Data flow client API to allow users to specify a `useJsonJobParameters` if necessary.
    a. There are no tests for shell nor Data flow client yet.  This is because @EnableDataflowServer needs to be fixed first.
 4. Polish the PR to cleanup some things I missed or noticed during review.

### DIFF
--- a/spring-cloud-dataflow-rest-client/src/test/java/org/springframework/cloud/dataflow/rest/client/DataflowTemplateTests.java
+++ b/spring-cloud-dataflow-rest-client/src/test/java/org/springframework/cloud/dataflow/rest/client/DataflowTemplateTests.java
@@ -24,8 +24,6 @@ import java.util.Optional;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
-import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
@@ -41,10 +39,8 @@ import org.springframework.batch.item.ExecutionContext;
 import org.springframework.cloud.dataflow.rest.Version;
 import org.springframework.cloud.dataflow.rest.job.StepExecutionHistory;
 import org.springframework.cloud.dataflow.rest.resource.RootResource;
-import org.springframework.cloud.dataflow.rest.support.jackson.Jackson2DataflowModule;
 import org.springframework.hateoas.Link;
 import org.springframework.hateoas.LinkRelation;
-import org.springframework.hateoas.mediatype.hal.Jackson2HalModule;
 import org.springframework.http.converter.HttpMessageConverter;
 import org.springframework.http.converter.json.MappingJackson2HttpMessageConverter;
 import org.springframework.web.client.ResourceAccessException;
@@ -71,10 +67,7 @@ public class DataflowTemplateTests {
 	@Before
 	public void setup() {
 		mapper = new ObjectMapper();
-		mapper.registerModule(new Jdk8Module());
-		mapper.registerModule(new Jackson2HalModule());
-		mapper.registerModule(new JavaTimeModule());
-		mapper.registerModule(new Jackson2DataflowModule());
+		DataFlowTemplate.prepareObjectMapper(mapper);
 		System.setProperty("sun.net.client.defaultConnectTimeout", String.valueOf(100));
 	}
 
@@ -104,9 +97,7 @@ public class DataflowTemplateTests {
 
 	@Test
 	public void testThatObjectMapperGetsPrepared() {
-		final ObjectMapper objectMapper = new ObjectMapper();
-		DataFlowTemplate.prepareObjectMapper(objectMapper);
-		assertCorrectMixins(objectMapper);
+		assertCorrectMixins(this.mapper);
 	}
 
 	@Test
@@ -116,7 +107,6 @@ public class DataflowTemplateTests {
 		jobParametersBuilder.addString("bar", "bar");
 
 		JobParameters jobParameters = jobParametersBuilder.toJobParameters();
-		DataFlowTemplate.prepareObjectMapper(this.mapper);
 		assertCorrectMixins(this.mapper);
 		String jobParametersSerialized = this.mapper.writeValueAsString(jobParameters);
 		jobParameters = this.mapper.readValue(jobParametersSerialized, JobParameters.class);
--- a/spring-cloud-dataflow-rest-client/src/test/java/org/springframework/cloud/dataflow/rest/client/DataflowTemplateTests.java
+++ b/spring-cloud-dataflow-rest-client/src/test/java/org/springframework/cloud/dataflow/rest/client/DataflowTemplateTests.java
@@ -22,6 +22,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Optional;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
@@ -34,6 +35,7 @@ import org.springframework.batch.core.JobExecution;
 import org.springframework.batch.core.JobInstance;
 import org.springframework.batch.core.JobParameter;
 import org.springframework.batch.core.JobParameters;
+import org.springframework.batch.core.JobParametersBuilder;
 import org.springframework.batch.core.StepExecution;
 import org.springframework.batch.item.ExecutionContext;
 import org.springframework.cloud.dataflow.rest.Version;
@@ -105,6 +107,23 @@ public class DataflowTemplateTests {
 		final ObjectMapper objectMapper = new ObjectMapper();
 		DataFlowTemplate.prepareObjectMapper(objectMapper);
 		assertCorrectMixins(objectMapper);
+	}
+
+	@Test
+	public void testJobParameters() throws JsonProcessingException {
+		final ObjectMapper objectMapper = new ObjectMapper();
+		JobParametersBuilder jobParametersBuilder = new JobParametersBuilder();
+		jobParametersBuilder.addString("foo", "foo");
+		jobParametersBuilder.addString("bar", "bar");
+
+		JobParameters jobParameters = jobParametersBuilder.toJobParameters();
+		DataFlowTemplate.prepareObjectMapper(objectMapper);
+		assertCorrectMixins(objectMapper);
+		String jobParametersSerialized = objectMapper.writeValueAsString(jobParameters);
+		jobParameters = objectMapper.readValue(jobParametersSerialized, JobParameters.class);
+		assertEquals(jobParameters.getParameter("foo").getValue(), "foo");
+		assertEquals(jobParameters.getParameter("bar").getValue(), "bar");
+		assertEquals(jobParameters.getParameters().size(), 2);
 	}
 
 	@Test
--- a/spring-cloud-dataflow-rest-client/src/test/java/org/springframework/cloud/dataflow/rest/client/DataflowTemplateTests.java
+++ b/spring-cloud-dataflow-rest-client/src/test/java/org/springframework/cloud/dataflow/rest/client/DataflowTemplateTests.java
@@ -111,16 +111,15 @@ public class DataflowTemplateTests {
 
 	@Test
 	public void testJobParameters() throws JsonProcessingException {
-		final ObjectMapper objectMapper = new ObjectMapper();
 		JobParametersBuilder jobParametersBuilder = new JobParametersBuilder();
 		jobParametersBuilder.addString("foo", "foo");
 		jobParametersBuilder.addString("bar", "bar");
 
 		JobParameters jobParameters = jobParametersBuilder.toJobParameters();
-		DataFlowTemplate.prepareObjectMapper(objectMapper);
-		assertCorrectMixins(objectMapper);
-		String jobParametersSerialized = objectMapper.writeValueAsString(jobParameters);
-		jobParameters = objectMapper.readValue(jobParametersSerialized, JobParameters.class);
+		DataFlowTemplate.prepareObjectMapper(this.mapper);
+		assertCorrectMixins(this.mapper);
+		String jobParametersSerialized = this.mapper.writeValueAsString(jobParameters);
+		jobParameters = this.mapper.readValue(jobParametersSerialized, JobParameters.class);
 		assertEquals(jobParameters.getParameter("foo").getValue(), "foo");
 		assertEquals(jobParameters.getParameter("bar").getValue(), "bar");
 		assertEquals(jobParameters.getParameters().size(), 2);
--- a/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/support/jackson/JobParametersJacksonMixIn.java
+++ b/spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/support/jackson/JobParametersJacksonMixIn.java
@@ -16,9 +16,12 @@
 
 package org.springframework.cloud.dataflow.rest.support.jackson;
 
+import java.util.Map;
+
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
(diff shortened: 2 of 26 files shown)


spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/support/jackson/JobParametersJacksonMixIn.java
```diff
@@ -27,9 +30,12 @@
...
  */
-@JsonIgnoreProperties("empty")
+@JsonIgnoreProperties({"empty", "identifyingParameters"})
 public abstract class JobParametersJacksonMixIn {
 
 	@JsonProperty
 	abstract boolean isEmpty();
+
+	@JsonProperty
+	abstract Map<String, JobParameter<?>> getIdentifyingParameters();
```
corneil, line 40: I suggest a test to valid use of `@JsonIgnoreProperties`

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/job/support/StepExecutionProgressInfo.java
```diff
@@ -64,7 +64,7 @@ public StepExecutionProgressInfo(StepExecution stepExecution, StepExecutionHisto
 		if (startTime == null) {
 			startTime = LocalDateTime.now();
 		}
-		duration = Duration.between(startTime, endTime).get(ChronoUnit.MILLIS);
+		duration = Duration.between(startTime, endTime).get(ChronoUnit.NANOS);
 		percentageComplete = calculatePercentageComplete();
```
corneil, line 68: I notice the `calculatePercentageComplete` is called after duration is set. It is a bit weird that the only call to `getDurationBasedEstimate` accepts the duration as parameter and in other uses references the field.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/job/support/StepExecutionProgressInfo.java
```diff
@@ -64,7 +64,7 @@ public StepExecutionProgressInfo(StepExecution stepExecution, StepExecutionHisto
 		if (startTime == null) {
 			startTime = LocalDateTime.now();
 		}
-		duration = Duration.between(startTime, endTime).get(ChronoUnit.MILLIS);
+		duration = Duration.between(startTime, endTime).get(ChronoUnit.NANOS);
```
onobc, line 67: Why the switch from millis to nanos?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/TaskJobService.java
```diff
@@ -99,15 +99,32 @@ public interface TaskJobService {
 	JobInstanceExecutions getJobInstance(long id) throws NoSuchJobInstanceException, NoSuchJobException;
 
 	/**
-	 * Restarts a {@link JobExecution} IF the respective {@link JobExecution} is actually
+	 * Restarts a {@link JobExecution} if the respective {@link JobExecution} is actually
 	 * deemed restartable. Otherwise a {@link JobNotRestartableException} is being thrown.
+	 * The system will use {@link TaskConfigurationProperties#isUseJsonJobParameters()} to
```
onobc, line 104: I ❤️ the detailed info but I think this new sentence is actually related to the default impl and not really the API (interface). Wdyt?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/ScdfJobParametersConverter.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+package org.springframework.cloud.dataflow.server.service.impl;
+
+import org.springframework.batch.core.JobParameter;
+
+/**
+ * Provides methods to serialize a Spring Batch {@link JobParameter} to the proper format.
+ */
+public interface ScdfJobParametersConverter {
```
onobc, line 24: It looks like this thin layer of indirection is here only because the `DefaultJobParametersConverter` `encode` method is `protected`. Is that correct?

cppwfs: That is correct.   We need access so that SCDF can utilize the the serialized String.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/job/support/StepExecutionProgressInfo.java
cppwfs: I can not remember why this was switched to NANOS.   Will revert.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskJobService.java
```diff
@@ -269,13 +278,24 @@ public void restartJobExecution(long jobExecutionId) throws NoSuchJobExecutionEx
 	 *
 	 * @param taskExecutionArgs original set of task execution arguments
 	 * @param jobParameters     for the job to be restarted.
+	 * @param useJsonJobParameters determine what converter to use to serialize the job parameter to the command line arguments.
 	 * @return deduped list of arguments that contains the original arguments and any
 	 * identifying job parameters not in the original task execution arguments.
 	 */
-	private List<String> restartExecutionArgs(List<String> taskExecutionArgs, JobParameters jobParameters) {
+	private List<String> restartExecutionArgs(List<String> taskExecutionArgs, JobParameters jobParameters,
+		Boolean useJsonJobParameters) {
```
onobc, line 286: This method can be simplified a bit via stream api - something like this:
```java
		if (useJsonJobParameters == null) {
			useJsonJobParameters = taskConfigurationProperties.isUseJsonJobParameters();
		}
		var jobParamsConverter = useJsonJobParameters ? new ScdfJsonJobParametersConverter()
				: new ScdfDefaultJobParametersConverter();
		List<String> result = new ArrayList<>(taskExecutionArgs);
		jobParameters.getParameters().entrySet().stream()
				.filter((e) -> !e.getKey().startsWith("-"))
				.filter((e) -> !taskExecutionArgs.stream().anyMatch((arg) -> arg.startsWith(e.getKey())))
				.map((e) -> e.getKey() + "=" + jobParamsConverter.deserializeJobParameter(e.getValue()))
				.forEach(result::add);
		return result;
```

Wdyt?

cppwfs: LGTM.   Tweaked it to use `noneMatch` instead of `!taskExecutionArgs.stream().anyMatch` 
Will merge change when review is complete.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/service/impl/DefaultTaskJobServiceTests.java
```diff
@@ -160,8 +160,19 @@ public void testRestart() throws Exception {
...
+	}
+
+	@Test
+	public void testRestartWithJsonParameters() throws Exception {
+		createBaseLaunchers();
+		initializeJobs(true);
 
-		assertThat(appDeploymentRequest.getCommandlineArguments()).contains("identifying.param=testparam,java.lang.String");
+		this.taskJobService.restartJobExecution(jobInstanceCount, true);
+		final ArgumentCaptor<AppDeploymentRequest> argument = ArgumentCaptor.forClass(AppDeploymentRequest.class);
```
onobc, line 172: **[TinyNit]** Do you mind dropping these inline `final`s?

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/JobExecutionControllerTests.java
```diff
@@ -188,7 +188,17 @@ public void testGetExecution() throws Exception {
 	@Test
 	public void testGetExecutionWithJobProperties() throws Exception {
 		MvcResult result = mockMvc.perform(get("/jobs/executions/10").accept(MediaType.APPLICATION_JSON))
-			.andDo(print())
+			.andExpect(status().isOk())
+			.andExpect(jsonPath("$.executionId", is(10)))
+			.andExpect(jsonPath("$.jobExecution.jobParameters.parameters", Matchers.hasKey(("javaUtilDate"))))
+			.andExpect(jsonPath("$.jobExecution.stepExecutions", hasSize(1))).andReturn();
+		assertThat(result.getResponse().getContentAsString()).contains("\"identifying\":true");
```
onobc, line 195: **[TIP]** You can combine this into single stmt using `contains(CharSequence... values)`

spring-cloud-dataflow-shell-core/src/test/java/org/springframework/cloud/dataflow/shell/command/JobCommandTests.java
```diff
@@ -174,22 +175,33 @@ public void testViewExecution() {
...
+
+		assertTrue("the table did not contain the correct job parameters for job parameter value bar",
+			checkModelColumn(paramRowOne, table, "bar(java.lang.String) "));
+
+		assertTrue("the table did not contain the correct job parameters for job parameter value baz",
+			checkModelColumn(paramRowOne, table, "baz(java.lang.Long) "));
+
+	}
+
+	private boolean checkModelColumn(int rowNumber, Table table, String value) {
```
onobc, line 191: Nice refactor!

spring-cloud-dataflow-shell-core/src/test/java/org/springframework/cloud/dataflow/shell/AbstractShellIntegrationTest.java
```diff
@@ -119,7 +119,15 @@ public static void startUp() {
...
+					"--spring.autoconfigure.exclude=org.springframework.boot.autoconfigure.security.servlet.SecurityAutoConfiguration," +
+						"org.springframework.boot.autoconfigure.security.servlet.SecurityFilterAutoConfiguration," +
+						"org.springframework.boot.actuate.autoconfigure.security.servlet.ManagementWebSecurityAutoConfiguration," +
+						"org.springframework.boot.autoconfigure.session.SessionAutoConfiguration," +
+						"org.springframework.cloud.deployer.spi.cloudfoundry.CloudFoundryDeployerAutoConfiguration," +
+						"org.springframework.cloud.deployer.spi.kubernetes.KubernetesAutoConfiguration," +
+						"org.springframework.boot.autoconfigure.security.servlet.UserDetailsServiceAutoConfiguration," +
```
onobc, line 128: It looks like this is mostly formatting w/ the exception of these 3 new additions at the tail. Curious as to why its needed here?

spring-cloud-dataflow-rest-client/src/test/java/org/springframework/cloud/dataflow/rest/client/DataflowTemplateTests.java
```diff
@@ -107,6 +109,23 @@ public void testThatObjectMapperGetsPrepared() {
 		assertCorrectMixins(objectMapper);
 	}
 
+	@Test
+	public void testJobParameters() throws JsonProcessingException {
+		final ObjectMapper objectMapper = new ObjectMapper();
```
onobc, line 114: I think you can remove the constuction (and prepare) of the OM here and just use `this.mapper`.

onobc (changes requested):
Nice work @cppwfs .  A few minor comments/questions - but other than that LGTM.

spring-cloud-dataflow-shell-core/src/test/java/org/springframework/cloud/dataflow/shell/AbstractShellIntegrationTest.java
cppwfs: They are not needed.  It was some transition code that was added then was not removed.   Will remove.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/service/impl/TaskConfigurationProperties.java
```diff
@@ -68,6 +68,14 @@ public class TaskConfigurationProperties {
 	 */
 	private boolean useKubernetesSecretsForDbCredentials;
 
+	/**
```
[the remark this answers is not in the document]
cppwfs: Nice change.  Updated with small polish.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/JobExecutionController.java
```diff
@@ -148,8 +148,10 @@ public ResponseEntity<Void> stopJobExecution(
 	@RequestMapping(value = {"/{executionId}"}, method = RequestMethod.PUT, params = "restart=true")
 	@ResponseStatus(HttpStatus.OK)
 	public ResponseEntity<Void> restartJobExecution(
-			@PathVariable("executionId") long jobExecutionId) throws NoSuchJobExecutionException {
-		taskJobService.restartJobExecution(jobExecutionId);
+			@PathVariable("executionId") long jobExecutionId,
+			@RequestParam(value = "useJsonJobParameters", required = false) Boolean useJsonJobParameters
+	) throws NoSuchJobExecutionException {
+		taskJobService.restartJobExecution(jobExecutionId, useJsonJobParameters);
```
onobc: Sorry, just noticed this on the re-review but should we log at warn (or at the very least debug) the reason of the NSJEE or is this just a standard expected case here? If the latter, maybe a little comment as to why we don't need to log the exception details. Wdyt?

spring-cloud-dataflow-rest-client/src/test/java/org/springframework/cloud/dataflow/rest/client/DataflowTemplateTests.java
```diff
@@ -111,16 +111,15 @@ public void testThatObjectMapperGetsPrepared() {
...
 		JobParametersBuilder jobParametersBuilder = new JobParametersBuilder();
 		jobParametersBuilder.addString("foo", "foo");
 		jobParametersBuilder.addString("bar", "bar");
 
 		JobParameters jobParameters = jobParametersBuilder.toJobParameters();
-		DataFlowTemplate.prepareObjectMapper(objectMapper);
-		assertCorrectMixins(objectMapper);
-		String jobParametersSerialized = objectMapper.writeValueAsString(jobParameters);
-		jobParameters = objectMapper.readValue(jobParametersSerialized, JobParameters.class);
+		DataFlowTemplate.prepareObjectMapper(this.mapper);
```
onobc, line 119: I believe prepareObjectMapper does the exact things to `this.mapper` that is done in `@BeforeEach`. Maybe `this.mapper` in @BeforeEach` could use `prepareObjectMapper`? I could me missing something though.

onobc (approved):
LGTM @cppwfs . Thank you for the updates. I have a couple of comments of the nit variety. Do what you want w/ them. I am approving now regardless.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/JobExecutionController.java
cppwfs: Adding a warning log entry.

cppwfs:
Squashed Rebased Merged.  Thanks for the reviews @onobc @corneil

PR #5870, opened 2024-07-16
Remove Dependency Mgmt for H2 and rely on Boot Bom

cppwfs:
This change is only for the Single Step Batch Job and Composed Task Runner.


onobc (approved):
Thanks for this improvement and fix @cppwfs

PR #5875, opened 2024-07-30
Fix mariadb mount to be /var/lib/mysql

### LINKED ISSUE #5877: MariaDB container volume mounts are incorrect
**Description:**
The volume mount for mariadb deployment was created with `/var/lib/mariadb` and results in a blank database after a restart.

**Release versions:**
2.8.x

**Fix**
The correct mount path is `/var/lib/mysql`

corneil:
Fixes #5877


onobc:
Why? Is there an associated issue?

corneil:
Added reproduction steps to https://github.com/spring-cloud/spring-cloud-dataflow/issues/5877#issuecomment-2260317394

PR #5888, opened 2024-08-05
Update prometheus-rsocket-proxy to 2.0.0-M1

onobc:


spring-cloud-dataflow-docs/src/main/asciidoc/spring-boot-3x.adoc
```diff
@@ -20,8 +20,6 @@ The naming of the metrics registry-specific properties differ as follows:
 * `2.x`: `management.metrics.export.prometheus.enabled=true`
```
cppwfs, line 20: Do we want to keep the spring-boot-3x.adoc? It contains both boot 3 and boot 2 instructions on prometheus?
We really don't need the boot2 docs anymore do we?

cppwfs (commented):
Just a little nitpick.

spring-cloud-dataflow-docs/src/main/asciidoc/spring-boot-3x.adoc
onobc: Nope, it will be deleted, I just did not delete it as part of "Update prometheus-rsocket-proxy...`

corneil (approved):
LGTM

cppwfs (approved):
LGTM

PR #5891, opened 2024-08-07
Fix problem with large Oracle container causing disk space issue

corneil:
Fix issue with large Oracle container causing disk space issue by running DB2 and Oracle specific tests in separate parallel jobs.

Found the following in ci and build-snapshot-worker logs.

```
##[warning]You are running out of disk space. The runner will stop working when the machine runs out of disk space. Free space left: 20 MB
```


cppwfs:
Can you take a peak at why the PR-CI build did not run for this PR?  Thanks!

corneil:
> Can you take a peak at why the PR-CI build did not run for this PR? Thanks!

I added [skip-ci] because it wasn't going to exercise the code in question.

PR #5896, opened 2024-08-14
Update apps version refs to 5.0.0 / 3.0.0

onobc:
This commit is the 1st step in the effort to update the default version of stream apps from 3.2.1 to 5.0.0.

* All references to `dataflow.spring.io/*-*-latest` are replaced w/ `dataflow.spring.io/*-*-(5-0-x|3-0-x)`
* All references to stream apps version `3.2.1` are replaced w/ `5.0.0` (unless it is a version diff or version upgrade test)
* All references to task apps version `2.0.2` are replaced w/ `3.0.0` (unless it is a version diff or version upgrade test)

> [!NOTE]
> @corneil we are still registering task apps version 2 and 3 in many of the scripts under `/src/` and I am pretty sure we should remove the 2.x ones but I wanted to understand the implications before we do. Will this break the ATs?  

> [!NOTE]
> The following PRs were merged to enable this change:
> * https://github.com/spring-cloud/spring-cloud-dataflow-ui/pull/2010
> * https://github.com/spring-io/dataflow.spring.io/pull/524

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/appendix-create-containers.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/appendix-create-containers.adoc
@@ -85,7 +85,7 @@ Downloads all applications needed by `create-containers.sh` from Maven repositor
 
 Usage: `download-apps.sh [version] [broker] [filter]`
 
-* `version` is the stream applications version like `3.2.1` or default is `3.2.2-SNAPSHOT`
+* `version` is the stream applications version like `5.0.0` or default is `5.0.1-SNAPSHOT`
 * `broker` is one of rabbitmq, rabbit or kafka
 * `filter` is a name of an application or a partial name that will be matched.
 
@@ -96,9 +96,9 @@ This script requires link:https://github.com/GoogleContainerTools/jib/tree/maste
 
 Usage: `create-containers.sh [version] [broker] [jre-version] [filter]`
 
-* `version` is the stream-applications version like `3.2.1` or default is `3.2.2-SNAPSHOT`
+* `version` is the stream-applications version like `5.0.0` or default is `5.0.1-SNAPSHOT`
 * `broker` is one of rabbitmq, rabbit or kafka
-* `jre-version` should be one of 11, 17
+* `jre-version` must be 17
 * `filter` is a name of an application or a partial name that will be matched.
 
 If the file is not present required to create the container the script will skip the one.
@@ -110,9 +110,9 @@ This script requires link:https://buildpacks.io/docs/tools/pack[packeto pack]
 
 Usage: `pack-containers.sh [version] [broker] [jre-version] [filter]`
 
-* `version` is the stream-applications version like `3.2.1` or default is `3.2.2-SNAPSHOT`
+* `version` is the stream-applications version like `5.0.0` or default is `5.0.1-SNAPSHOT`
 * `broker` is one of rabbitmq, rabbit or kafka
-* `jre-version` should be one of 11, 17
+* `jre-version` must be 17
 * `filter` is a name of an application or a partial name that will be matched.
 
 If the required file is not present to create the container the script will skip that one.
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
@@ -340,20 +340,20 @@ The following table includes the `dataflow.spring.io` links to the stream applic
 |Artifact Type |Stable Release |SNAPSHOT Release
 
 |RabbitMQ + Maven
-|https://dataflow.spring.io/rabbitmq-maven-latest
-|https://dataflow.spring.io/rabbitmq-maven-latest-snapshot
+|https://dataflow.spring.io/rabbitmq-maven-5-0-x
+|https://dataflow.spring.io/rabbitmq-maven-5-0-x-snapshot
 
 |RabbitMQ + Docker
-|https://dataflow.spring.io/rabbitmq-docker-latest
-|https://dataflow.spring.io/rabbitmq-docker-latest-snapshot
+|https://dataflow.spring.io/rabbitmq-docker-5-0-x
+|https://dataflow.spring.io/rabbitmq-docker-5-0-x-snapshot
 
 |Apache Kafka + Maven
-|https://dataflow.spring.io/kafka-maven-latest
-|https://dataflow.spring.io/kafka-maven-latest-snapshot
+|https://dataflow.spring.io/kafka-maven-5-0-x
+|https://dataflow.spring.io/kafka-maven-5-0-x-snapshot
 
 |Apache Kafka + Docker
-|https://dataflow.spring.io/kafka-docker-latest
-|https://dataflow.spring.io/kafka-docker-latest-snapshot
+|https://dataflow.spring.io/kafka-docker-5-0-x
+|https://dataflow.spring.io/kafka-docker-5-0-x-snapshot
 |======================
 
 NOTE: By default, the out-of-the-box app's actuator endpoints are secured. You can disable security by deploying streams by setting the following property: `[small]#app.*.spring.autoconfigure.exclude=org.springframework.boot.autoconfigure.security.servlet.SecurityAutoConfiguration#`
@@ -370,12 +370,12 @@ The following table includes the `dataflow.spring.io` links to the task applicat
 |Artifact Type |Stable Release |SNAPSHOT Release
 
 |Maven
-|https://dataflow.spring.io/task-maven-latest
-|https://dataflow.spring.io/task-maven-latest-snapshot
+|https://dataflow.spring.io/task-maven-3-0-x
+|https://dataflow.spring.io/task-maven-3-0-x-snapshot
 
 |Docker
-|https://dataflow.spring.io/task-docker-latest
-|https://dataflow.spring.io/task-docker-latest-snapshot
+|https://dataflow.spring.io/task-docker-3-0-x
+|https://dataflow.spring.io/task-docker-3-0-x-snapshot
 |======================
 
 For more information about the available out-of-the-box stream applications see the https://cloud.spring.io/spring-cloud-task-app-starters/[Spring Cloud Stream Applications] project page.
@@ -387,7 +387,7 @@ As an example, if you would like to register all out-of-the-box stream applicati
 ====
 [source,bash,subs=attributes]
 ----
-$ dataflow:>app import --uri https://dataflow.spring.io/kafka-maven-latest
+$ dataflow:>app import --uri https://dataflow.spring.io/kafka-maven-5-0-x
 ----
 ====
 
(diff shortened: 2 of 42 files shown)


corneil (approved):
LGTM

spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/service/DefaultAppRegistryService.java
```diff
@@ -413,14 +424,17 @@ private AppRegistration createAppRegistrations(Map<String, AppRegistration> regi
...
+			if (extra.equals("metadata")) {
 				// metadata app uri
 				try {
 					ar.setMetadataUri(new URI(lineSplit[1]));
 					warnOnMalformedURI(lineSplit[0], ar.getMetadataUri());
 				} catch (Exception e) {
 					throw new IllegalArgumentException(e);
 				}
-
+			} else if (!"bootVersion".equals(extra)) {
```
cppwfs, line 435: I don't think we need to check for `bootVersion` here, because the if block on lines 405-412, should return previous if, `"bootVersion".equals(extra)".   Thus extra should be something else.

spring-cloud-dataflow-registry/src/test/java/org/springframework/cloud/dataflow/registry/support/AppResourceCommonTests.java
```diff
@@ -130,73 +130,73 @@ public void testInvalidUrlResourceURI() throws Exception {
...
+		assertThat(version).isEqualTo("4.0.1-SNAPSHOT");
 		theRest = appResourceCommon.getResourceWithoutVersion(urlResource);
 		assertThat(theRest).isEqualTo("https://repo.maven.apache.org/maven2/org/springframework/cloud/stream/app/file-sink-rabbit/file-sink-rabbit");
 
 		//Standard maven style naming scheme
-		urlResource = new UrlResource("https://repo.maven.apache.org/maven2/org/springframework/cloud/stream/app/file-sink-rabbit/file-sink-rabbit-3.2.1-SNAPSHOT.jar");
+		urlResource = new UrlResource("https://repo.maven.apache.org/maven2/org/springframework/cloud/stream/app/file-sink-rabbit/file-sink-rabbit-4.0.1-SNAPSHOT.jar");
 		version = appResourceCommon.getUrlResourceVersion(urlResource);
-		assertThat(version).isEqualTo("3.2.1-SNAPSHOT");
+		assertThat(version).isEqualTo("4.0.1-SNAPSHOT");
```
cppwfs, line 164: Just curious as to why `file-sink-rabbit` is `4.0.1-SNAPSHOT` and not `5.0.1-SNAPSHOT`?

spring-cloud-dataflow-server/src/test/java/org/springframework/cloud/dataflow/integration/test/db/AbstractPostgresDatabaseTests.java
```diff
@@ -93,7 +93,7 @@ public void testMigration_210_211() throws URISyntaxException {
 		ObjectMapper objectMapper = new ObjectMapper();
 		objectMapper.registerModule(new Jackson2DataflowModule());
 		final DataFlowTemplate dataFlowTemplate = new DataFlowTemplate(new URI(dataflowCluster.getDataflowUrl()), objectMapper);
-		dataFlowTemplate.appRegistryOperations().register("time", ApplicationType.source, "docker:springcloudstream/time-source-rabbit:3.2.1", null, false);
+		dataFlowTemplate.appRegistryOperations().register("time", ApplicationType.source, "docker:springcloudstream/time-source-rabbit:5.0.0", null, false);
 		dataFlowTemplate.appRegistryOperations().register("log", ApplicationType.sink, "docker:springcloudstream/log-sink-rabbit:3.2.1", null, false);
```
cppwfs, line 97: This entry looks like it was skipped.

cppwfs (changes requested):
Looks good.
Just a couple of questions and nitpicks.

cppwfs:
Did some scanning around and we have some tests that are mocked up as well as documentation that refer versions `1.2.0.RELEASE`.   Wondering if we handle this on a different PR.

onobc:
> Did some scanning around and we have some tests that are mocked up as well as documentation that refer versions `1.2.0.RELEASE`. Wondering if we handle this on a different PR.

We should update any doc references. 

Any test references should be updated to be something less ancient than `1.2.0.RELEASE`.  Also, depending on the test, we may not want to update it if its testing upgrade etc.. Let's go ahead and point those out here and get them updated in this PR if I missed some (which is very possible).

spring-cloud-dataflow-docs/src/main/asciidoc/appendix-create-containers.adoc
```diff
@@ -110,9 +110,9 @@ This script requires link:https://buildpacks.io/docs/tools/pack[packeto pack]
 
 Usage: `pack-containers.sh [version] [broker] [jre-version] [filter]`
 
-* `version` is the stream-applications version like `3.2.1` or default is `3.2.2-SNAPSHOT`
+* `version` is the stream-applications version like `5.0.0` or default is `5.0.1-SNAPSHOT`
 * `broker` is one of rabbitmq, rabbit or kafka
-* `jre-version` should be one of 11, 17
```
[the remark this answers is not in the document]
onobc: Rather, why was it not included? Waiting on [packeto support](https://github.com/spring-cloud/spring-cloud-dataflow/pull/5898/files#diff-180b06aba3688b8e993503b50a581ffe67f9925f1bf3e6cc7ef150ed0d96bb57R58) I believe.

spring-cloud-dataflow-docs/src/main/asciidoc/streams.adoc
```diff
@@ -370,12 +370,12 @@ The following table includes the `dataflow.spring.io` links to the task applicat
 |Artifact Type |Stable Release |SNAPSHOT Release
 
 |Maven
-|https://dataflow.spring.io/task-maven-latest
-|https://dataflow.spring.io/task-maven-latest-snapshot
+|https://dataflow.spring.io/task-maven-3-0-x
```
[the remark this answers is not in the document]
onobc: Further digging reveals this is ok because the default SCDF configuration includes the anonymously accessible `repo.spring.io/milestone` where the task samples are available.

spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/service/DefaultAppRegistryService.java
onobc: I will double-check. This was pulled over from the impl in `2.11.x` and simplified to just ignore the bootVersion line in the descriptor as the stream apps descriptors are already published w/ the bootVersion and to remove it would require stream apps release. I will see if we can simplify this further as suggested. Thanks.

spring-cloud-dataflow-registry/src/test/java/org/springframework/cloud/dataflow/registry/support/AppResourceCommonTests.java
onobc: Good catch. The first pass at this was using `4.0.0` apps but we want to use the `5.0.0` apps. I must have missed this search/replace when I went from 4.0->5.0. Adjusting...

spring-cloud-dataflow-server/src/test/java/org/springframework/cloud/dataflow/integration/test/db/AbstractPostgresDatabaseTests.java
onobc: Yep - I misread this test because it was "upgrade" I think I got confused w/ the version numbers and thought it was upgrading the apps - but clearly that is not what its doing. Good catch!

spring-cloud-dataflow-docs/src/main/asciidoc/appendix-create-containers.adoc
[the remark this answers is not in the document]
cppwfs: Just curious if it is the same reason for SCDF.  Had a discussion with @corneil on a solution.    But we'll handle that in another PR.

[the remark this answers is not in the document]
onobc: Not sure. I know this is just the steps on how users can create their own containers if they need to add JDBC drivers etc.. agreed we can look into this in another PR as it is tangential to my changes and in fact I could have waited myself before changing the Java version pieces that I did (hindsight :) )

cppwfs:
Here are the doc links where docs can be updated from 1.2.0.RELEASE to the 5.0.0:
* https://github.com/spring-cloud/spring-cloud-dataflow/blob/main/spring-cloud-skipper/spring-cloud-skipper/src/main/java/org/springframework/cloud/skipper/domain/SpringCloudDeployerApplicationManifest.java#L35-L39
* https://github.com/spring-cloud/spring-cloud-dataflow/blob/main/spring-cloud-dataflow-docs/src/main/asciidoc/dashboard.adoc?plain=1#L70-L71
* https://github.com/spring-cloud/spring-cloud-dataflow/blob/main/spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/support/AppResourceCommon.java#L104
* https://github.com/spring-cloud/spring-cloud-dataflow/blob/main/spring-cloud-skipper/README.adoc?plain=1 (throughout the file)

I excluded the tests that need to be updated.   Wondering if they are best handled as a part of the JUnit 4-5 migration?

onobc:
> I excluded the tests that need to be updated. Wondering if they are best handled as a part of the JUnit 4-5 migration?

Let me take a peek. May just throw them in here as this one is already out of control ;P

cppwfs (approved):
LGTM.   Thanks for the hard work!!!!

onobc:
> LGTM. Thanks for the hard work!!!!

Thank you for the excellent review and back-forth - the PR was too big and I apologize for that.

spring-cloud-dataflow-registry/src/main/java/org/springframework/cloud/dataflow/registry/service/DefaultAppRegistryService.java
onobc: I believe this check is enforcing "the only 3 part keys in the descriptor file in here can be 'metadata' or 'bootVersion'". 

Keys in that file must be one of these:
- `<app-type>.<app-name>` 
- `<app-type>.<app-name>.(metadata|bootVersion)`

Without that check in here it would allow an entry such as `<app-type>.<app-name>.imposter`. I will verify this though before merging.

onobc: Actually, you are right @cppwfs . The check does not nothing, in fact there is nothing that verifies the 3rd token is one of `metadata` or `bootVersion` because line 397 will fail on `getResourceVersion` if you put a key anything other than `metadata` in the extra field. 

E.g. this entry fails - but not in the if block in question (on line 397)...

```java
source.debezium=maven://org.springframework.cloud.stream.app:debezium-source-rabbit:5.0.0
source.debezium.fooVersion=3
source.debezium.metadata=maven://org.springframework.cloud.stream.app:debezium-source-rabbit:jar:metadata:5.0.0
```
Because `fooVersion` is indeed `!bootVersion` so it attempts to get version info that does not exist on the value of the entry (`3`).

Not going to make any changes for this behavior in this PR. Updated the code w/ your original request though. Thanks.

PR #5898, opened 2024-08-15
Update DataFlowIT.multipleComposedTaskWithArguments so that it will pass

cppwfs:
This PR has 3 commits:

1. Update scripts and docker-compose files.
     a. Make sure to remove the `version` from the docker-compose.yml files.  It is marked as obsolete and causes different versions of `docker compose` to fail.
     b.  Update docker-compose files to use the latest version of SCDF 3.x instead of 2.11.x
     c.  Update build image script so that it uses java 17 when creating containers
2. Update the Code so that we have SCDF 3.x enumerated.   Remove the `Disabled flags and TODOs`
    a. Note that some code in the is already in PR #5850 (`JobParametersJacksonMixIn.java` & `DefaultTaskJobService.java`).  This was added so we could provide a successful test.  And these changes can be removed on merge once that PR has been merged.  
3. Start documenting how SCDF's workflows work.   This document currently only contains information on how images are created.  (But its a start).

### DIFF
--- a/.github/workflows/build-image.sh
+++ b/.github/workflows/build-image.sh
@@ -13,8 +13,8 @@ if [ "$TAG" == "" ]; then
     exit 1
 fi
 if [ "$DEFAULT_JDK" = "" ]; then
-    echo "DEFAULT_JDK not found using 11"
-    DEFAULT_JDK=11
+    echo "DEFAULT_JDK not found using 17"
+    DEFAULT_JDK=17
 else
     echo "DEFAULT_JDK=$DEFAULT_JDK"
 fi
@@ -55,7 +55,8 @@ if [ ! -f "$JAR" ]; then
         exit $RC
     fi
 fi
-for v in 8 11 17; do
+# TODO add Java 21 when packeto supports it
+for v in 17; do
     echo "Creating: $REPO:$TAG-jdk$v"
     pack build --builder gcr.io/paketo-buildpacks/builder:base \
         --path "$JAR" \
@@ -93,9 +94,5 @@ for v in 8 11 17; do
         fi
     fi
 done
-#if [ "$PUSH" == "true" ]; then
-#    echo "Pruning Docker"
-#    docker system prune -f
-#    docker system prune --volumes -f
-#fi
+
 
--- a/.github/workflows/workflow.adoc
+++ b/.github/workflows/workflow.adoc
@@ -1,6 +1,8 @@
-= Workflow Reference Readme
+= Workflow Reference
 
-This README serves as a guide to the GitHub Action workflows included in this repository. It outlines the purpose and functionality of each workflow, detailing their role in the CI and release processes. Additionally, this document provides an overview of the scripts and actions utilized in these workflows, offering insights into how they work together in SCDF's CI/CD pipeline.
+This README serves as a guide to the GitHub Action workflows included in this repository.
+It outlines the purpose and functionality of each workflow, detailing their role in the CI and release processes.
+Additionally, this document provides an overview of the scripts and actions utilized in these workflows, offering insights into how they work together in SCDF's CI/CD pipeline.
 
 This document is a work in progress, and as various workflows are updated, the documentation will be revised to reflect both existing and new behaviors.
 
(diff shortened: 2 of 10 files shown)


.github/workflows/build-image.sh
```diff
@@ -93,9 +94,5 @@ for v in 8 11 17; do
         fi
     fi
 done
-#if [ "$PUSH" == "true" ]; then
```
onobc, line 96: Was this supposed to be removed?

.github/workflows/workflow.adoc
```diff
@@ -0,0 +1,66 @@
+= Workflow Reference Readme
+
+This README serves as a guide to the GitHub Action workflows included in this repository. It outlines the purpose and functionality of each workflow, detailing their role in the CI and release processes. Additionally, this document provides an overview of the scripts and actions utilized in these workflows, offering insights into how they work together in SCDF's CI/CD pipeline.
```
onobc, line 3: Nit, but do you mind doing [one sentence per line](https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line)?

.github/workflows/workflow.adoc
```diff
@@ -0,0 +1,66 @@
+= Workflow Reference Readme
```
onobc, line 1: I think the "Readme" in the title is redundant.

spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/support/jackson/JobParameterJacksonDeserializer.java
```diff
@@ -50,19 +50,11 @@ public JobParameter deserialize(JsonParser jsonParser, DeserializationContext de
...
-			}
-			else if ("LONG".equalsIgnoreCase(type)) {
-				jobParameter = new JobParameter(Long.valueOf(value), Long.class, identifying);
-			}
-			else {
-				throw new IllegalStateException("Unsupported JobParameter type: " + type);
+		if (!type.isEmpty()) {
+			try {
+				jobParameter = new JobParameter(value, Class.forName(type), identifying);
+			} catch (ClassNotFoundException e) {
```
onobc, line 56: Would there ever be more helpful in the CNFEX that we would want to log?

spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/support/jackson/JobParametersJacksonMixIn.java
```diff
@@ -27,7 +27,7 @@
  * @author Gunnar Hillert
  * @since 1.0
  */
-@JsonIgnoreProperties("empty")
+@JsonIgnoreProperties({"empty", "identifyingParameters"})
```
onobc, line 30: Why do we want to ignore `identifyingParameters`?

onobc (changes requested):
Thanks for the PR @cppwfs . I have a few comments/questions but other than that LGTM.

.github/workflows/build-image.sh
corneil: It was an attempt to save space before the DB2 and Oracle integration tests were moved to separate jobs.

spring-cloud-dataflow-rest-resource/src/main/java/org/springframework/cloud/dataflow/rest/support/jackson/JobParametersJacksonMixIn.java
cppwfs: This attribute returns an unmodifiableMap and thus when it deseriallizes it fails.  This change is a shim awaiting the merge of https://github.com/spring-cloud/spring-cloud-dataflow/pull/5850.

.github/workflows/workflow.adoc
cppwfs: Whoops!   Thank you!

onobc (approved):
Thanks for the updates @cppwfs - LGTM!

cppwfs:
Squashed rebased Merged.   Thanks for the reviews!

PR #5902, opened 2024-08-21
Remove metrics auto-config exclusions

onobc:
This commit removes temporary auto-config exclusions that were in place to allow compiling against Boot 3.x prior to migrating to the newer Observation API.


corneil (approved):
LGTM

cppwfs (approved):

onobc:
Thanks for the review @cppwfs @corneil

PR #5903, opened 2024-08-22
SCDF should be able to migrate a schema from 2.10 to 3.0

cppwfs:
Currently the migration fails with a validation error. The cause of the validation error was that some commits from 2.11.x that contained flyway migration scripts were not ported to the main-3 branch.

The following commits that contained flyway migrations were migrated to main-3 from the main branch.
* 62ea6c5
* 6f97589

The goal of this PR is to resolve the validation error so DB migrations will work properly.

A subsequent PR will be submitted will add the feature code for commit 6f97589.


onobc (approved):
LGTM - thanks for the forward port @cppwfs

cppwfs:
Rebased Merged.

PR #5905, opened 2024-08-23
[CI] Remove actions that don't add value. 

corneil:
Remove actions that don't add value. 
Ensure documentation tests are executed.

### DIFF
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -16,7 +16,7 @@ jobs:
       shell: bash
       timeout-minutes: 75
       run: |
-        ./mvnw -B -s .github/settings.xml -Pasciidoctordocs,restdocs clean install
+        ./mvnw -B -s .github/settings.xml -Pdocs clean install
   scan:
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -8,13 +8,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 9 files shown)


.github/workflows/ci-pr.yml
```diff
@@ -23,14 +16,7 @@ jobs:
...
       run: |
-        ./mvnw -B -s .github/settings.xml clean install
-    - name: Capture Test Results
-      if: failure()
-      uses: actions/upload-artifact@v3
-      with:
-        name: Unit Tests
-        path: '**/surefire-reports/*.xml'
-        retention-days: 3
+        ./mvnw -B -s .github/settings.xml -Pasciidoctordocs,restdocs clean install
```
onobc, line 19: This seems very specific to `asciidoctordocs,restdocs` -- will all the other tests execute? I did not know we even had these profiles. Where are they defined?

corneil: They have been in place to ensure the rest docs execute even when skipping tests but not during a normal build.

onobc: > will all the other tests execute?

corneil: I will simplify this to a single profile.

corneil: > > will all the other tests execute?

If you build with `./mvnw clean install -DskipTests -Pdocs` the docs will be generated.
If you build with `./mvnw clean install -DskipTests` the docs will be not generated and not tests executed.
If you build with `./mvnw clean install` the docs will be not generated and tests executed.

onobc (approved):
Thanks for the updates @corneil . The profiles are much simpler now. LGTM

Issue #5909, opened 2024-08-27
User should be able to set app version when creating schedule.


cppwfs:
**Description:**
Spring Cloud Data Flow supports the ability to allow users to set the version of the app used when launching a task definition.
This is done by using the following property `version.<app-name>=<version>`.   
However when a user schedules a task definition, the format `version.<app-name>=<version>` does not set the version of the application.   
Schedules work on a task definition, so the format to set the app version is `version.<task-definition-name>=<version>`.  While this works it is inconsistent with the usage when launching a task.   
So SCDF needs to support the format `version.<app-name>=<version>`.

**Release versions:**
Spring Cloud Dataflow 2.11.0 - 2.11.4

**Steps to reproduce:**
1. Import Task apps 
2. Register the 2.0.1 version of timestamp-task
3. Create task definition that contains just the timestamp app.
4. Create schedule for the task definition, but in the properties set the timestamp to user version 2.0.1.

The shell script to reproduce the scenario for Kubernetes is located below:
```
app register timestamp --type task --uri docker:springcloudtask/timestamp-task:2.0.1
task create ts1 --definition "timestamp"
task schedule create --name w1 --definitionName ts1 --expression '*/1 * * * *' --properties 'version.timestamp=2.0.1'
```

**Additionally**
Update reference documentation to show how to set the version when launching and scheduling a task definition.

cppwfs:
Merged

PR #5919, opened 2024-08-29
App version should be set on Schedules using the version.app property

cppwfs:
SCDF allows users to set the version of  an application when creating a schedule using the following deployer property `version.<taskdefinition>` property. 
However this does not follow the pattern used in task launches which is `version.<app-name>`.  For example:
```
task schedule create --name w1 --definitionName ts1 --expression '*/1 * * * *' --properties 'version.timestamp=3.0.0'
```

This PR resolves this discrepancy.   So it supports both usages (task-definition/app-name).
Documentation has also been updated to show users how to set this.

Commit 1: The updated code and newly added tests
Commit2: Documentation.

See #5909

### DIFF
--- a/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/service/impl/DefaultSchedulerServiceTests.java
+++ b/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/service/impl/DefaultSchedulerServiceTests.java
@@ -31,6 +31,7 @@ import org.junit.jupiter.api.Test;
 import org.mockito.ArgumentCaptor;
 import org.mockito.Mockito;
 
+import org.mockito.stubbing.Answer;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.context.PropertyPlaceholderAutoConfiguration;
 import org.springframework.boot.autoconfigure.jdbc.DataSourceProperties;
@@ -60,7 +61,6 @@ import org.springframework.cloud.dataflow.server.service.SchedulerService;
 import org.springframework.cloud.dataflow.server.service.SchedulerServiceProperties;
 import org.springframework.cloud.dataflow.server.service.TaskExecutionInfoService;
 import org.springframework.cloud.deployer.resource.docker.DockerResource;
-import org.springframework.cloud.deployer.spi.core.AppDefinition;
 import org.springframework.cloud.deployer.spi.scheduler.CreateScheduleException;
 import org.springframework.cloud.deployer.spi.scheduler.ScheduleInfo;
 import org.springframework.cloud.deployer.spi.scheduler.ScheduleRequest;
@@ -68,6 +68,7 @@ import org.springframework.cloud.deployer.spi.scheduler.Scheduler;
 import org.springframework.cloud.deployer.spi.task.TaskLauncher;
 import org.springframework.cloud.task.listener.TaskException;
 import org.springframework.core.env.PropertyResolver;
+import org.springframework.core.io.FileSystemResource;
 import org.springframework.core.io.Resource;
 import org.springframework.core.io.ResourceLoader;
 import org.springframework.data.domain.Page;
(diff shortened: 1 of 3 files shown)


corneil (approved):
LGTM

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/service/impl/DefaultSchedulerServiceTests.java
```diff
@@ -431,18 +436,51 @@ public void testScheduleWithoutCommandLineArguments() {
...
+	public void testVersionWithResource() {
+		String validVersionNumber = "3.0.0";
+		ScheduleRequest request = scheduleRequest(validVersionNumber);
+		assertThat(request.getResource().toString()).contains("file:src/test/resources/apps/foo-task");
+	}
+
+	@Test
+	public void testVersionWithResourceInvalidVersion() {
+		String invalidVersionNumber = "2.0.0";
+		assertThatExceptionOfType(IllegalArgumentException.class)
```
onobc, line 451: Take a peek at `assertThatIllegalArgumentException().isThrownBy(() -> ...)`

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/service/impl/DefaultSchedulerServiceTests.java
```diff
@@ -431,18 +436,51 @@ public void testScheduleWithoutCommandLineArguments() {
...
+			.isThrownBy(() -> {
+				scheduleRequest(invalidVersionNumber);
+			}).hasToString("java.lang.IllegalArgumentException: Unknown task app: demo");
```
onobc, line 454: There are a handful of ex assertions available, `havingRootCause().withInstanceOf(IAEX.class).withMessagContaining`.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/service/impl/DefaultSchedulerServiceTests.java
```diff
@@ -431,18 +436,51 @@ public void testScheduleWithoutCommandLineArguments() {
...
+	}
+
+	private ScheduleRequest scheduleRequest(String appVersionToTest) {
+		String definition = "demo";
+		Map<String, String> resourceTestProps = new HashMap<>(testProperties);
```
onobc, line 459: I am enjoying the newer `Map.of(` API. Combined w/ `var` it makes the code a bit less terse.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/service/impl/DefaultSchedulerServiceTests.java
```diff
@@ -470,10 +508,20 @@ private ScheduleRequest getScheduleRequest(List<String> commandLineArguments, St
 		TaskDefinition taskDefinition = new TaskDefinition(BASE_DEFINITION_NAME, definition);
 
 		when(mockTaskDefinitionRepository.findById(BASE_DEFINITION_NAME)).thenReturn(Optional.of(taskDefinition));
-		when(mockAppRegistryService.getAppResource(any())).thenReturn(new DockerResource(resourceToReturn));
-		when(mockAppRegistryService.find(taskDefinition.getRegisteredAppName(), ApplicationType.task))
+		doAnswer((Answer<Resource>) invocation -> {
```
onobc, line 511: Another way to do this w/o `doAnswer` is via the `argThat` - something like:
```java
		when(mockAppRegistryService.getAppResource(argThat((appReg) -> DEMO_APP_NAME.equals(appReg.getName()))))
				.thenReturn(new FileSystemResource("file:src/test/resources/apps/foo-task"));

		when(mockAppRegistryService.getAppResource(argThat((appReg) -> !DEMO_APP_NAME.equals(appReg.getName()))))
				.thenReturn(new DockerResource(resourceToReturn));
```

Both work - I try using `argThat` until it breaks down and then I jump into `doAnswer`. Not sure what is best in this case though (about the same IMO ).

onobc (approved):
LGTM @cppwfs . I made a few nit comments/suggestions but do what you want w/ them as they are not critical. Thanks.

cppwfs:
Squashed and Merged

PR #5921, opened 2024-08-30
Improve Carvel Deployment Support.

corneil:
The change improve the references to imagePullSecrets and serviceAccountNames to ensure deployment works correctly in the TPE cluster that seems to have stricter security than standard GKE or minikube that we have used for testing thus far.

References [scdf-pro#209](https://github.com/pivotal/scdf-pro/issues/209)


onobc (approved):
LGTM - thanks @corneil

jvalkeal (approved):
Looks good. I like idea having dedicated options instead of letting user to change too much.

PR #5925, opened 2024-09-04
Update tracing props

onobc:
This PR updates the tracing properties in the Zipkin and Wavefront docker compose files to enable trace reporting to each.

Also the 2nd commit is unrelated and was a missed property from when we updated the Prometheus RSocket proxy to `2.0.0-M1`.


spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
```diff
@@ -14,10 +14,6 @@ management:
           autotime:
             enabled: true # true is default to Boot 2.3.2 at least.
             percentiles-histogram: true
-    export:
```
ilayaperumalg, line 17: Could you help me understand why these properties are being removed and the addition of `metrics.export.rsocket.enabled`?

src/docker-compose/docker-compose-wavefront.yml
```diff
@@ -14,6 +14,9 @@ services:
       - MANAGEMENT_WAVEFRONT_METRICS_EXPORT_APITOKEN=${WAVEFRONT_KEY:?WAVEFRONT_KEY is not set!}
       - MANAGEMENT_WAVEFRONT_METRICS_EXPORT_URI=${WAVEFRONT_URI:-https://vmware.wavefront.com}
       - MANAGEMENT_WAVEFRONT_METRICS_EXPORT_SOURCE=${WAVEFRONT_SOURCE:-scdf-docker-compose}
+      - SPRING_CLOUD_DATAFLOW_APPLICATIONPROPERTIES_STREAM_MANAGEMENT_TRACING_ENABLED=true
+      - SPRING_CLOUD_DATAFLOW_APPLICATIONPROPERTIES_STREAM_MANAGEMENT_TRACING_SAMPLING_PROBABILITY=1.0
```
ilayaperumalg, line 18: The reasoning to enable tracing for all the requests (1.0) here is to help the users get a better understanding on the tracing capability?

src/docker-compose/docker-compose-wavefront.yml
```diff
@@ -14,6 +14,9 @@ services:
       - MANAGEMENT_WAVEFRONT_METRICS_EXPORT_APITOKEN=${WAVEFRONT_KEY:?WAVEFRONT_KEY is not set!}
       - MANAGEMENT_WAVEFRONT_METRICS_EXPORT_URI=${WAVEFRONT_URI:-https://vmware.wavefront.com}
       - MANAGEMENT_WAVEFRONT_METRICS_EXPORT_SOURCE=${WAVEFRONT_SOURCE:-scdf-docker-compose}
+      - SPRING_CLOUD_DATAFLOW_APPLICATIONPROPERTIES_STREAM_MANAGEMENT_TRACING_ENABLED=true
```
ilayaperumalg, line 17: There is some discussion about the usage of this property [here](https://github.com/spring-projects/spring-boot/issues/38625). While I see the auto configuration is enabled by default, the usage of this property is still not recommended and this property is even removed in the later SB 3.x versions.

ilayaperumalg (commented):
I am also curious to know why we have docker-compose files maintained at `src/templates` and `src`

spring-cloud-dataflow-server-core/src/main/resources/META-INF/dataflow-server-defaults.yml
cppwfs: It was a change in SB 3.x that facilitated the need for this change: https://github.com/spring-projects/spring-boot/commit/be3523b1cda41f178df25fdfe8f8f27199259530

onobc: See https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.0-Migration-Guide#actuator-metrics-export-properties

onobc: The RSocket property was a miss when we bumped to prometheus-rsocket-proxy 2.0.0-M1 (see 2nd half of PR description).

src/docker-compose/docker-compose-wavefront.yml
onobc: The property is not intended to enable/disable tracing, just the exporting to the reporters. 
See [here](https://github.com/spring-projects/spring-boot/blob/f024c193e4ae5c4263077d8fada4959639231248/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/OnEnabledTracingCondition.java#L36-L37). THere is a global and a reporter specific variant. 

This one is actually redundant since we are setting 
```
SPRING_CLOUD_DATAFLOW_APPLICATIONPROPERTIES_STREAM_MANAGEMENT_WAVEFRONT_TRACING_EXPORT_ENABLED=true
```
below. 

I will remove it from the stream apps though and here as well in a subsequent PR as it does not add value.

Thanks for the good eye!

src/docker-compose/docker-compose-wavefront.yml
onobc: Otherwise it is 10% sampled and they say "where are my traces".  We can adjust or they can adjust when they install.

cppwfs (approved):
After applying the fix you mentioned below: LGTM

onobc:
> I am curious to know why we have docker-compose files maintained at `src/templates` and `src`

Good question. Maybe let's create an issue to follow up on that.

onobc:
> After applying the fix you mentioned below: LGTM

Thanks @cppwfs . The comment alluded to improvement but not a "fix". Also not planning on doing that in this PR as it will require same change on stream apps. You ok w/ that?

src/docker-compose/docker-compose-wavefront.yml
onobc: Created https://github.com/spring-cloud/spring-cloud-dataflow/issues/5928 to follow up

src/docker-compose/docker-compose-wavefront.yml
onobc: Created https://github.com/spring-cloud/spring-cloud-dataflow/issues/5928 to follow up

corneil (commented):
LGTM

PR #5962, opened 2024-10-04
Update maven-surefire-report-plugin to 3.5.0

cppwfs:
Replace the hardcoded version with a property in each dependency, Set the property in the spring-cloud-dataflow parent

You will also notice I removed a unused import in a test.   Can remove if it needs to be put in another PR


onobc (approved):
Thanks for the improvement @cppwfs !

ilayaperumalg:
LGTM

PR #5964, opened 2024-10-04
Use Duration.toMillis instead of CronoUnit

cppwfs:
Duration using ChronoUnit only supports nanos and seconds on conversion. In this case use the toMillis() function


onobc (approved):

cppwfs:
Rebased, Squashed, Merged Thanks @onobc for the review!!!

PR #5965, opened 2024-10-04
Enable JobExecutionDeserializationTests

cppwfs:
* Remove JobConfigurationName from test data.  Could not find where it was used in 2.11.x nor in main-3
* Change from old JobParameterType to java class name

### DIFF
--- a/spring-cloud-dataflow-rest-client/pom.xml
+++ b/spring-cloud-dataflow-rest-client/pom.xml
@@ -34,6 +34,10 @@
 			<artifactId>jackson-databind</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>com.fasterxml.jackson.datatype</groupId>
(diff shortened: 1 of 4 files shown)


onobc (approved):

spring-cloud-dataflow-rest-client/pom.xml
```diff
@@ -33,6 +33,10 @@
 			<groupId>com.fasterxml.jackson.core</groupId>
 			<artifactId>jackson-databind</artifactId>
 		</dependency>
+		<dependency>
```
ilayaperumalg, line 36: If this dependency is just for test only, can we just restrict it for tests alone?

cppwfs:
Rebased, Squashed, Merged.  Thank you @onobc and @ilayaperumalg for the reviews!

PR #5968, opened 2024-10-07
Enable tests that were disabled because of TaskRepository issue

cppwfs:
Tests suites were disabled with the following message `"taskRepository not found"`.
This message was incorrect.   They were failing because multiple taskRepository beans were declared.
By excluding SimpleTaskAutoConfiguration from the @SpringBootApplication the issue no longer occured in these tests.

Another issue arose where 2 RestTemplates were declared and only one was required for actuatorOperations.  Since this is not needed for these tests, this was mocked out.

Once past this, individual tests were able to run.   I resolved issues for failed tests or Disabled those that were either waiting for another PR to be merged or needed more investigation.

### DIFF
--- a/spring-cloud-dataflow-shell-core/src/test/java/org/springframework/cloud/dataflow/shell/ShellCommandsTests.java
+++ b/spring-cloud-dataflow-shell-core/src/test/java/org/springframework/cloud/dataflow/shell/ShellCommandsTests.java
@@ -57,7 +57,6 @@ import static org.assertj.core.api.Assertions.assertThat;
  * @author Chris Bono
  * @author Corneil du Plessis
  */
-@Disabled("taskRepository not found")
 class ShellCommandsTests extends AbstractShellIntegrationTest {
 
 	@AfterEach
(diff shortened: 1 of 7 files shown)


onobc (commented):

onobc (commented):

onobc (approved):

spring-cloud-dataflow-shell-core/src/test/java/org/springframework/cloud/dataflow/shell/ShellCommandsTests.java
```diff
@@ -77,6 +76,8 @@ void singleFileCommand() {
 		assertAppExists("timestamp", ApplicationType.task);
 	}
 
+	//TODO: Boot3x followup
```
ilayaperumalg, line 79: Thank you for updating both the TODO as well as description. Let's be consistent with approach.

ilayaperumalg:
LGTM

cppwfs:
Rebased Squashed Merged.   Thank you @ilayaperumalg @onobc for the reviews!

PR #5970, opened 2024-10-09
Mark failing tests caused by main-3 migration as Boot3x

cppwfs:
The goal is to isolate `Disabled` to the test level and provide documentation as to why it is disabled.   And if possible resolve the problem that facilitated the need for `Disabled` 
* Re-Enable  tests that are now functioning or needed small repairs/tweaks
* ReEnable LocalConfigurationTests
* Remove ROW_COUNT test from AbstractSmokeTests. The ROW_COUNT test was to make sure that this filter was applied for specific database types that uses the aggregate views. In this case we no longer need to test for this as SCDF no longer uses views
* Make sure all Disabled Tests have a message even if they are not a main-3 Disabled tests


ilayaperumalg:
Great work @cppwfs for investigating and adding detailed description! LGTM, merging.

PR #5971, opened 2024-10-10
Allow users to disable task features in dataflow

cppwfs:
Currently if the user sets task-enabled to false SCDF will fail to start.

* Rename DataflowTaskConfiguration to DataflowTaskExplorerConfiguration so that it expresses its usage. 
* Rename aggregateTaskExplorer to dataflowTaskExplorer 
* Set the `ConditionalOnTasksEnabled` annotation so that the configuration won't fire if `task-enabled` is set to false. 
* Remove the `Disabled` annotation from the test so that we can verify that the `task-enabled` works properly.


ilayaperumalg (approved):
LGTM

corneil (approved):
Elegant solution

PR #5976, opened 2024-10-14
JdbcSearchableJobExecutionDao needs to support LocalDateTime for DB2

cppwfs:
* Code now handles the NPE that is thrown by DB2 when an a LocalDateTime field contains null when retrieving from the database.
* It will set the value to NULL.

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/JdbcSearchableJobExecutionDao.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/JdbcSearchableJobExecutionDao.java
@@ -38,6 +38,7 @@ import org.springframework.batch.core.JobParameter;
 import org.springframework.batch.core.JobParameters;
 import org.springframework.batch.core.repository.dao.JdbcJobExecutionDao;
 import org.springframework.batch.item.database.Order;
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/JdbcSearchableJobExecutionDao.java
```diff
@@ -639,16 +643,42 @@ JobExecution createJobExecutionFromResultSet(ResultSet rs, int rowNum) throws SQ
...
 		return jobExecution;
 	}
 
+	private  LocalDateTime getLocalDateTime(int columnIndex, ResultSet rs) throws SQLException {
+		LocalDateTime result = null;
+		if (databaseType == DatabaseType.DB2) {
+			try {
+				result = rs.getObject(columnIndex, LocalDateTime.class);
+			} catch (NullPointerException npe) {
+				if (!npe.getMessage().contains("java.sql.Timestamp.toLocalDateTime()\" because \"<local4>\" is null")) {
```
ilayaperumalg, line 662: Since we bypass an NPE which is a bug, let's make a note as TODO when this eventually gets fixed at DB2. Also, it would probably be a better to log a message at the else block stating the actual exception we are bypassing.

cppwfs:
Updated based on code reviews.  Thanks for the reviews!
Rebased, squashed, Merged.

PR #5977, opened 2024-10-14
fix dataflow task explorer bean

corneil:
Added Bean creation because it was previously done in module now removed.


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowTaskConfiguration.java
```diff
@@ -156,4 +158,10 @@ public TaskDeploymentReader taskDeploymentReader(TaskDeploymentRepository reposi
 	public JdbcTaskBatchDao taskBatchDao(DataSource dataSource) {
 		return new JdbcTaskBatchDao(dataSource);
 	}
+
+	@Bean
+	@ConditionalOnMissingBean
```
cppwfs, line 163: The solution in this PR requires that we create a `DataflowTaskExecutionQueryDao` when Tasks are disabled, which is not the desired goal. 

Please look at PR https://github.com/spring-cloud/spring-cloud-dataflow/pull/5971 .  This PR resolves the problem without creating a `DataflowTaskExecutionQueryDao` when tasks are disabled.

cppwfs (changes requested):

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowTaskConfiguration.java
corneil: 5971 fixes this issue in a more elegant way.

corneil:
#5971 fixes this issue in a more elegant way.

PR #5978, opened 2024-10-14
Fix jobExecutionListByName by removing Schema Target.

corneil:


cppwfs (approved):
LGTM !   Thank you for looking into this!

PR #5981, opened 2024-10-14
Remove unused componentscan from DataFlowServerConfiguration

cppwfs:


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/config/DataFlowServerConfiguration.java
```diff
@@ -69,7 +68,6 @@
 		DataFlowTaskConfiguration.class
 })
 @EnableConfigurationProperties({ BatchProperties.class, CommonApplicationProperties.class })
-@ComponentScan(basePackages = {"org.springframework.cloud.dataflow.schema.service", "org.springframework.cloud.dataflow.aggregate.task"})
```
ilayaperumalg, line 72: I believe this was needed when we enabled DataFlowServer configuration as AutoConfiguration which then be used by any other custom DataFlow server implementations. We moved away from supporting customised ones. But, we can always revisit when needed. 

LGTM

corneil: I believe if someone wants to customize they can add those to their application configuration.

corneil (approved):

cppwfs:
Rebased, Merged.   Thank you for the reviews!!!!

PR #5984, opened 2024-10-15
The authorization header was not dropped by redirect strategy as expected.

corneil:
Edit:
Discovered that `RedirectStrategy` was change from Apache HTTP Client 4 -> 5. The change meant header changes were not possible.
I copied to `RedirectExec` interceptor / chain handler and modified it to use the modified request instead of the original request to create a request builder.

### DIFF
--- a/spring-cloud-dataflow-configuration-metadata/src/test/java/org/springframework/cloud/dataflow/container/registry/authorization/support/S3SignedRedirectRequestController.java
+++ b/spring-cloud-dataflow-configuration-metadata/src/test/java/org/springframework/cloud/dataflow/container/registry/authorization/support/S3SignedRedirectRequestController.java
@@ -39,7 +39,6 @@ import org.springframework.web.bind.annotation.RestController;
  */
 @RestController
 public class S3SignedRedirectRequestController {
-	private final static Logger logger = LoggerFactory.getLogger(S3SignedRedirectRequestController.class);
 
 	@RequestMapping("/service/token")
 	public ResponseEntity<Map<String, String>> getToken() {
@@ -57,7 +56,6 @@ public class S3SignedRedirectRequestController {
 	@RequestMapping("/v2/test/s3-redirect-image/blobs/signed_redirect_digest")
 	public ResponseEntity<Map<String, String>> getBlobRedirect(@RequestHeader("Authorization") String token) {
 		if (!"bearer my_token_999".equals(token.trim().toLowerCase())) {
-			logger.info("getBlobRedirect=BAD_REQUEST");
 			return new ResponseEntity<>(HttpStatus.BAD_REQUEST);
 		}
 		HttpHeaders redirectHeaders = new HttpHeaders();
@@ -68,15 +66,12 @@ public class S3SignedRedirectRequestController {
 				"&X-Amz-Expires=1200" +
 				"&X-Amz-SignedHeaders=host" +
 				"&X-Amz-Signature=test");
-		logger.info("getBlobRedirect:{}", redirectHeaders);
 		return new ResponseEntity<>(redirectHeaders, HttpStatus.TEMPORARY_REDIRECT);
 	}
 
 	@RequestMapping("/test/docker/registry/v2/blobs/test/data")
 	public ResponseEntity<Resource> getSignedBlob(@RequestHeader Map<String, String> headers) {
-		logger.info("getSignedBlob:{}", headers);
 		if (headers.containsKey("authorization")) {
-			logger.info("getSignedBlob=BAD_REQUEST");
 			return new ResponseEntity<>(HttpStatus.BAD_REQUEST);
 		}
 		return buildFromString("{\"config\": {\"Labels\": {\"foo\": \"bar\"} } }");
--- a/spring-cloud-dataflow-configuration-metadata/src/test/java/org/springframework/cloud/dataflow/container/registry/authorization/support/S3SignedRedirectRequestController.java
+++ b/spring-cloud-dataflow-configuration-metadata/src/test/java/org/springframework/cloud/dataflow/container/registry/authorization/support/S3SignedRedirectRequestController.java
@@ -19,6 +19,9 @@ package org.springframework.cloud.dataflow.container.registry.authorization.supp
 import java.util.Collections;
 import java.util.Map;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import org.springframework.cloud.dataflow.container.registry.ContainerRegistryProperties;
 import org.springframework.core.io.ByteArrayResource;
 import org.springframework.core.io.Resource;
@@ -36,6 +39,7 @@ import org.springframework.web.bind.annotation.RestController;
  */
 @RestController
 public class S3SignedRedirectRequestController {
+	private final static Logger logger = LoggerFactory.getLogger(S3SignedRedirectRequestController.class);
 
 	@RequestMapping("/service/token")
 	public ResponseEntity<Map<String, String>> getToken() {
@@ -53,6 +57,7 @@ public class S3SignedRedirectRequestController {
 	@RequestMapping("/v2/test/s3-redirect-image/blobs/signed_redirect_digest")
 	public ResponseEntity<Map<String, String>> getBlobRedirect(@RequestHeader("Authorization") String token) {
 		if (!"bearer my_token_999".equals(token.trim().toLowerCase())) {
+			logger.info("getBlobRedirect=BAD_REQUEST");
 			return new ResponseEntity<>(HttpStatus.BAD_REQUEST);
 		}
 		HttpHeaders redirectHeaders = new HttpHeaders();
@@ -63,13 +68,15 @@ public class S3SignedRedirectRequestController {
 				"&X-Amz-Expires=1200" +
 				"&X-Amz-SignedHeaders=host" +
 				"&X-Amz-Signature=test");
-
+		logger.info("getBlobRedirect:{}", redirectHeaders);
 		return new ResponseEntity<>(redirectHeaders, HttpStatus.TEMPORARY_REDIRECT);
 	}
 
 	@RequestMapping("/test/docker/registry/v2/blobs/test/data")
 	public ResponseEntity<Resource> getSignedBlob(@RequestHeader Map<String, String> headers) {
-		if (!headers.containsKey("authorization")) {
+		logger.info("getSignedBlob:{}", headers);
+		if (headers.containsKey("authorization")) {
+			logger.info("getSignedBlob=BAD_REQUEST");
 			return new ResponseEntity<>(HttpStatus.BAD_REQUEST);
 		}
 		return buildFromString("{\"config\": {\"Labels\": {\"foo\": \"bar\"} } }");
--- a/spring-cloud-dataflow-configuration-metadata/src/test/java/org/springframework/cloud/dataflow/container/registry/authorization/support/S3SignedRedirectRequestController.java
+++ b/spring-cloud-dataflow-configuration-metadata/src/test/java/org/springframework/cloud/dataflow/container/registry/authorization/support/S3SignedRedirectRequestController.java
@@ -32,6 +32,7 @@ import org.springframework.web.bind.annotation.RestController;
 
 /**
  * @author Adam J. Weigold
+ * @author Corneil du Plessis
  */
 @RestController
 public class S3SignedRedirectRequestController {
@@ -68,7 +69,7 @@ public class S3SignedRedirectRequestController {
 
 	@RequestMapping("/test/docker/registry/v2/blobs/test/data")
 	public ResponseEntity<Resource> getSignedBlob(@RequestHeader Map<String, String> headers) {
-		if (headers.containsKey("authorization")) {
+		if (!headers.containsKey("authorization")) {
 			return new ResponseEntity<>(HttpStatus.BAD_REQUEST);
 		}
 		return buildFromString("{\"config\": {\"Labels\": {\"foo\": \"bar\"} } }");
--- a/spring-cloud-dataflow-container-registry/src/main/java/org/springframework/cloud/dataflow/container/registry/ContainerImageRestTemplateFactory.java
+++ b/spring-cloud-dataflow-container-registry/src/main/java/org/springframework/cloud/dataflow/container/registry/ContainerImageRestTemplateFactory.java
@@ -21,6 +21,7 @@ import java.security.NoSuchAlgorithmException;
 import java.security.cert.X509Certificate;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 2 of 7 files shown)


spring-cloud-dataflow-configuration-metadata/src/test/java/org/springframework/cloud/dataflow/container/registry/authorization/support/S3SignedRedirectRequestController.java
```diff
@@ -68,7 +69,7 @@ public ResponseEntity<Map<String, String>> getBlobRedirect(@RequestHeader("Autho
 
 	@RequestMapping("/test/docker/registry/v2/blobs/test/data")
 	public ResponseEntity<Resource> getSignedBlob(@RequestHeader Map<String, String> headers) {
-		if (headers.containsKey("authorization")) {
+		if (!headers.containsKey("authorization")) {
```
cppwfs, line 72: I'm curious why it worked in 2.11.x  when the test said it should not contain `authorized` but now requires  authorized?

cppwfs (changes requested):

spring-cloud-dataflow-configuration-metadata/src/test/java/org/springframework/cloud/dataflow/container/registry/authorization/support/S3SignedRedirectRequestController.java
corneil: The question is then how does the authorization header not get dropped?

corneil: `org.springframework.cloud.dataflow.container.registry.authorization.DropAuthorizationHeaderRequestRedirectStrategy` doesn't reach the code to check if a header should be dropped.

corneil: Comparing [org.apache.hc.client5.http.protocol.RedirectStrategyorg.apache.http.client.RedirectStrategy](https://hc.apache.org/httpcomponents-client-5.4.x/current/httpclient5/apidocs/org/apache/hc/client5/http/protocol/RedirectStrategy.html)
to [org.apache.http.client.RedirectStrategy](https://hc.apache.org/httpcomponents-client-4.5.x/current/httpclient/apidocs/org/apache/http/client/RedirectStrategy.html)
shows the return type change from `HttpUriRequest` to `URI`
So we will have to figure out where we can actually deal with the headers and if we need to actually drop the `Authorization` header.

rstoyanchev (commented):
I can't comment the solution, but it's worth reaching out for comment through one of the channels on https://hc.apache.org/get-involved.html, mailing list or issue tracker. There may be related issues and discussions there as well.

onobc:
This is on the mailing list [here](https://lists.apache.org/thread/rwf87n9z6kgspjombc1hxl9gq4dvzx9h) w/ 3 options.

spring-cloud-dataflow-configuration-metadata/src/test/java/org/springframework/cloud/dataflow/container/registry/authorization/support/S3SignedRedirectRequestController.java
```diff
@@ -62,13 +68,15 @@ public ResponseEntity<Map<String, String>> getBlobRedirect(@RequestHeader("Autho
...
 				"&X-Amz-SignedHeaders=host" +
 				"&X-Amz-Signature=test");
-
+		logger.info("getBlobRedirect:{}", redirectHeaders);
 		return new ResponseEntity<>(redirectHeaders, HttpStatus.TEMPORARY_REDIRECT);
 	}
 
 	@RequestMapping("/test/docker/registry/v2/blobs/test/data")
 	public ResponseEntity<Resource> getSignedBlob(@RequestHeader Map<String, String> headers) {
+		logger.info("getSignedBlob:{}", headers);
```
onobc, line 77: I don't think we should log the `headers` as they may contain sensitive info.

spring-cloud-dataflow-configuration-metadata/src/test/java/org/springframework/cloud/dataflow/container/registry/authorization/support/S3SignedRedirectRequestController.java
```diff
@@ -62,13 +68,15 @@ public ResponseEntity<Map<String, String>> getBlobRedirect(@RequestHeader("Autho
 				"&X-Amz-Expires=1200" +
 				"&X-Amz-SignedHeaders=host" +
 				"&X-Amz-Signature=test");
-
+		logger.info("getBlobRedirect:{}", redirectHeaders);
```
onobc, line 71: Please add some human-readable context to these log message as the method name is already in the log prefix.

corneil:
> This is on the mailing list [here](https://lists.apache.org/thread/rwf87n9z6kgspjombc1hxl9gq4dvzx9h) w/ 3 options.

I will look into a chain handler that strips the header if I can consistently figure out that we are handling a redirect.

spring-cloud-dataflow-configuration-metadata/src/test/java/org/springframework/cloud/dataflow/container/registry/authorization/support/S3SignedRedirectRequestController.java
```diff
@@ -52,6 +57,7 @@ public ResponseEntity<Resource> getManifests(@RequestHeader("Authorization") Str
 	@RequestMapping("/v2/test/s3-redirect-image/blobs/signed_redirect_digest")
 	public ResponseEntity<Map<String, String>> getBlobRedirect(@RequestHeader("Authorization") String token) {
 		if (!"bearer my_token_999".equals(token.trim().toLowerCase())) {
+			logger.info("getBlobRedirect=BAD_REQUEST");
```
onobc, line 60: Please add some human readable context to the log message as the method name is already in the log prefix. Maybe something like `Token did not contain expected bearer signal therefore responding BAD_REQUEST`.

spring-cloud-dataflow-configuration-metadata/src/test/java/org/springframework/cloud/dataflow/container/registry/authorization/support/S3SignedRedirectRequestController.java
```diff
@@ -62,13 +68,15 @@ public ResponseEntity<Map<String, String>> getBlobRedirect(@RequestHeader("Autho
...
-
+		logger.info("getBlobRedirect:{}", redirectHeaders);
 		return new ResponseEntity<>(redirectHeaders, HttpStatus.TEMPORARY_REDIRECT);
 	}
 
 	@RequestMapping("/test/docker/registry/v2/blobs/test/data")
 	public ResponseEntity<Resource> getSignedBlob(@RequestHeader Map<String, String> headers) {
+		logger.info("getSignedBlob:{}", headers);
 		if (headers.containsKey("authorization")) {
+			logger.info("getSignedBlob=BAD_REQUEST");
```
onobc, line 79: Same motivation as https://github.com/spring-cloud/spring-cloud-dataflow/pull/5984/files#r1803857559, maybe something like `Request contained unexpected authorization header therefore responding BAD_REQUEST`.

spring-cloud-dataflow-container-registry/src/main/java/org/springframework/cloud/dataflow/container/registry/authorization/DropAuthorizationHeaderRequestRedirectStrategy.java
```diff
@@ -92,109 +91,75 @@ public URI getLocationURI(final HttpRequest request, final HttpResponse response
...
+					if (isHeadMethod(method)) {
+						return new HttpHead(httpUriRequest).getUri();
+					}
+					else {
+						return new HttpGet(httpUriRequest).getUri();
+					}
+				}
 			}
 		}
-		} catch (URISyntaxException e) {
```
onobc, line 128: Nice cleanup!

spring-cloud-dataflow-configuration-metadata/src/test/java/org/springframework/cloud/dataflow/container/registry/authorization/support/S3SignedRedirectRequestController.java
```diff
@@ -32,9 +35,11 @@
 
 /**
  * @author Adam J. Weigold
+ * @author Corneil du Plessis
  */
 @RestController
 public class S3SignedRedirectRequestController {
```
onobc, line 41: **[UNRELATED-TO-YOUR-CHANGES]** These 2 controllers are terribly named and there is no comment telling us that these are test controllers used to mimic what S3 signed redirects look like.

spring-cloud-dataflow-container-registry/src/test/java/org/springframework/cloud/dataflow/container/registry/authorization/support/S3SignedRedirectRequestController.java
```diff
@@ -68,7 +69,7 @@ public ResponseEntity<Map<String, String>> getBlobRedirect(@RequestHeader("Autho
 
 	@RequestMapping("/test/docker/registry/v2/blobs/test/data")
 	public ResponseEntity<Resource> getSignedBlob(@RequestHeader Map<String, String> headers) {
-		if (headers.containsKey("authorization")) {
+		if (!headers.containsKey("authorization")) {
```
onobc, line 72: This needs to be removed as it is expected that the auth header has been removed. Same as the other S3 test controller thing.

spring-cloud-dataflow-container-registry/src/main/java/org/springframework/cloud/dataflow/container/registry/authorization/DropAuthorizationHeaderRequestRedirectStrategy.java
```diff
@@ -92,109 +91,75 @@ public URI getLocationURI(final HttpRequest request, final HttpResponse response
...
+						return new HttpHead(httpUriRequest).getUri();
+					}
+					else {
+						return new HttpGet(httpUriRequest).getUri();
+					}
+				}
+				catch (URISyntaxException e) {
+					throw new HttpException("Unable to get location URI", e);
+				}
+			}
```
onobc, line 109: These lines are repeated in all 3 cases. Let's factor this out for DRY.

spring-cloud-dataflow-container-registry/src/main/java/org/springframework/cloud/dataflow/container/registry/ContainerImageRestTemplateFactory.java
```diff
@@ -197,30 +203,34 @@ private HttpClientBuilder httpClientBuilder(SSLContext sslContext) {
...
-				new HttpComponentsClientHttpRequestFactory(
-						clientBuilder
-								.setRedirectStrategy(new DropAuthorizationHeaderRequestRedirectStrategy(extra))
-								// Azure redirects may contain double slashes and on default those are normilised
-								.setDefaultRequestConfig(RequestConfig.custom().build())
-								.build());
+		DropAuthorizationHeaderRequestRedirectStrategy redirectStrategy = new DropAuthorizationHeaderRequestRedirectStrategy(extra);
+		HttpComponentsClientHttpRequestFactory customRequestFactory = new HttpComponentsClientHttpRequestFactory(
+				clientBuilder.setRedirectStrategy(redirectStrategy)
+					.replaceExecInterceptor(ChainElement.REDIRECT.name(), new SpecialRedirectExec(routePlanner, redirectStrategy))
```
onobc, line 223: Based on your reply to the mailing list comment, it sounds like you are going to pursue option 2) which is to append a custom request interceptor immediately after RedirectExec and strip undesired headers from requests. 

I appreciate the scrappiness of figuring out how to get this to work in this case but I would really really like to **NOT** own this copy of the `SpecialRedirectExec` as we do not have tests for it, it is immediately out of date as soon as we commit it to this repo, and so on so forth the cost of copy/pasta/patch approach. I did find this test example in httpclient around adding interceptors (may or may not be helpful). 

https://github.com/apache/httpcomponents-client/blob/master/httpclient5/src/test/java/org/apache/hc/client5/http/examples/ClientInterceptors.java



 Add a custom request interceptor immediately after RedirectExec and
strip undesired headers from requests

onobc (changes requested):
Thanks for tackling this @corneil . It is definitely a hairy section of the codebase and http client 4->5 has changed quite a bit. I do have some concerns/suggestions.

spring-cloud-dataflow-configuration-metadata/src/test/java/org/springframework/cloud/dataflow/container/registry/authorization/support/S3SignedRedirectRequestController.java
corneil: Done

spring-cloud-dataflow-container-registry/src/test/java/org/springframework/cloud/dataflow/container/registry/authorization/support/S3SignedRedirectRequestController.java
corneil: Done

spring-cloud-dataflow-container-registry/src/main/java/org/springframework/cloud/dataflow/container/registry/authorization/DropAuthorizationHeaderRequestRedirectStrategy.java
corneil: Done

spring-cloud-dataflow-container-registry/src/main/java/org/springframework/cloud/dataflow/container/registry/ContainerImageRestTemplateFactory.java
corneil: After a log of back and forth with extra exec interceptor and request interceptor the only conclusion I have is that we need a modified version of `RedirectExec` that makes it possible to remove the header in the `RedirectStrategy`
There is not other way to get involved in the redirect process that doing exactly what `RedirectExec` does.

spring-cloud-dataflow-configuration-metadata/src/test/java/org/springframework/cloud/dataflow/container/registry/authorization/support/S3SignedRedirectRequestController.java
corneil: The initial change I made was invalid and reversed.
The `Authorization` header should be dropped for most of the redirects on container queries.

corneil:
Using a client that supports request changes on redirect is a better option than a custom redirect handler.

PR #5988, opened 2024-10-16
Modifying view queries to improve performance in MySQL / MariaDB.

peter-cheon:
When using the SCDF UI on an SCDF server based on MySQL / MariaDB, executing job executions becomes slow during pagination if there are thousands of job history records.

This occurs because job executions and other histories are basically exposed in descending order. Creating a view based on descending order can lead to performance improvements.

For reference, when an internal SCDF server had about 8,000 job history records, creating a descending order-based view reduced the pagination movement speed for the 8,000-item list from 30 seconds to 5 seconds.

This has only been verified in MySQL / MariaDB, so modifications for other databases have not been made.


cppwfs:

PR #5990, opened 2024-10-16
Provide method for stopping Batch 5 Jobs upon user request

### LINKED ISSUE #5986: SimpleJobService stop/stopAll need to support JobOperator
Currently SimpleJobService sets the JobExecution to `Stopping` or `Stopped`.   This signals that the batch job to stop on the next step, which is normal.   However, it does not support stopping a `Stoppable` tasklet, which is a feature of  the `SimpleJobOperator.stop/stopAll` methods .   
SCDF needs to create a SCDFSimpleJobOperator that overrides the `stop` and `stopAll` methods so that it implements this feature.   The reason SCDF can't use the existing `SimpleJobOperator.stop/stopAll` methods is that batch uses a MapJobRegistry to retrieve step information, which will not be in a consistent state within SCDF.

cppwfs:
In the previous release of SCDF we used the JsrJobOperator to stop job executions. The 2 stages of stopping jobs is as follows:

1.  Sets the Batch Status of the job execution to STOPPING.  This signals to Spring Batch to stop execution at the next step. 
2. If the Job is a StepLocator it will go through each of the StoppableTasklets and stop them.

Since Dataflow never deals with a Job directly, it never executed the 2nd stage.  Meaning the Job is in the application that SCDF launched, not SCDF itself. So when using Batch 4.x it was just a quick check of the JobRegistry to retrieve the job, which was always empty since SCDF never deals with Jobs directly. But with Batch 5.x they loaded the Job Registry and attempted to retrieve the Job in a different way using the SimpleJobOperator. This caused a NPE when it tried to retrieve and determine if the Job was a StepLocator. In the updated solution, SCDF doesn't use the SimpleJobOperator since SCDF doesn't have access to the StepLocator for the Job, nor does it use the JobRegistry and even if it did, it would always be empty.

Resolves #5986

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -70,7 +70,7 @@
 		<module>spring-cloud-dataflow-shell</module>
 		<module>spring-cloud-dataflow-shell-core</module>
 		<module>spring-cloud-dataflow-completion</module>
-		<module>spring-cloud-skipper</module>
+<!--		<module>spring-cloud-skipper</module>-->
 		<module>spring-cloud-starter-dataflow-server</module>
 		<module>spring-cloud-starter-dataflow-ui</module>
 		<module>spring-cloud-dataflow-server</module>
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
@@ -225,9 +225,7 @@ public class SimpleJobService implements JobService, DisposableBean {
 		Collection<JobExecution> result = jobExecutionDao.getRunningJobExecutions();
 		for (JobExecution jobExecution : result) {
 			try {
-				jobExecution.getStepExecutions().forEach(StepExecution::setTerminateOnly);
-				jobExecution.setStatus( BatchStatus.STOPPING);
-				jobRepository.update(jobExecution);
+				stop(jobExecution.getId());
 			} catch (Exception e) {
 				throw new IllegalArgumentException("The following JobExecutionId was not found: " + jobExecution.getId(), e);
 			}
@@ -246,7 +244,7 @@ public class SimpleJobService implements JobService, DisposableBean {
 		// 'STOPPING'. It is assumed that
 		// the step implementation will check this status at chunk boundaries.
 		logger.info("Stopping job execution: " + jobExecution);
-
+		jobExecution.getStepExecutions().forEach(StepExecution::setTerminateOnly);
 		jobExecution.setStatus(BatchStatus.STOPPING);
 		jobRepository.update(jobExecution);
 		return jobExecution;
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
@@ -242,10 +242,12 @@ public class SimpleJobService implements JobService, DisposableBean {
 		if (!jobExecution.isRunning()) {
 			throw new JobExecutionNotRunningException("JobExecution is not running and therefore cannot be stopped");
 		}
-
+		// Indicate the execution should be stopped by setting it's status to
+		// 'STOPPING'. It is assumed that
+		// the step implementation will check this status at chunk boundaries.
 		logger.info("Stopping job execution: " + jobExecution);
 
-		jobExecution.setStatus(BatchStatus.STOPPED);
+		jobExecution.setStatus(BatchStatus.STOPPING);
 		jobRepository.update(jobExecution);
 		return jobExecution;
 
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
@@ -225,7 +225,7 @@ public class SimpleJobService implements JobService, DisposableBean {
 		Collection<JobExecution> result = jobExecutionDao.getRunningJobExecutions();
 		for (JobExecution jobExecution : result) {
 			try {
-				stop(jobExecution.getId());
+				stopJobExecution(jobExecution);
 			} catch (Exception e) {
 				throw new IllegalArgumentException("The following JobExecutionId was not found: " + jobExecution.getId(), e);
 			}
@@ -236,14 +236,17 @@ public class SimpleJobService implements JobService, DisposableBean {
 
 	@Override
 	public JobExecution stop(Long jobExecutionId) throws NoSuchJobExecutionException, JobExecutionNotRunningException {
-		JobExecution jobExecution = getJobExecution(jobExecutionId);
+		return stopJobExecution(getJobExecution(jobExecutionId));
+	}
+
+	private JobExecution stopJobExecution(JobExecution jobExecution) throws JobExecutionNotRunningException{
 		if (!jobExecution.isRunning()) {
 			throw new JobExecutionNotRunningException("JobExecution is not running and therefore cannot be stopped");
 		}
 		// Indicate the execution should be stopped by setting it's status to
 		// 'STOPPING'. It is assumed that
 		// the step implementation will check this status at chunk boundaries.
-		logger.info("Stopping job execution: " + jobExecution);
+		logger.info("Stopping job execution: {}", jobExecution);
 		jobExecution.getStepExecutions().forEach(StepExecution::setTerminateOnly);
 		jobExecution.setStatus(BatchStatus.STOPPING);
 		jobRepository.update(jobExecution);
--- a/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/batch/AbstractSimpleJobServiceTests.java
+++ b/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/batch/AbstractSimpleJobServiceTests.java
@@ -28,6 +28,8 @@ import java.util.List;
 import javax.sql.DataSource;
 
 import org.junit.jupiter.api.Test;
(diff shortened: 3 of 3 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
```diff
@@ -242,10 +242,12 @@ public JobExecution stop(Long jobExecutionId) throws NoSuchJobExecutionException
 		if (!jobExecution.isRunning()) {
 			throw new JobExecutionNotRunningException("JobExecution is not running and therefore cannot be stopped");
 		}
-
+		// Indicate the execution should be stopped by setting it's status to
```
onobc, line 245: This is the replacement for [step 1) from the PR description](https://github.com/spring-projects/spring-batch/blob/fb0ce98a890eae1908bdf25d69eb4ea51b89973e/spring-batch-core/src/main/java/org/springframework/batch/core/jsr/launch/JsrJobOperator.java#L733-L742), correct?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
```diff
@@ -242,10 +242,12 @@ public JobExecution stop(Long jobExecutionId) throws NoSuchJobExecutionException
...
+		// 'STOPPING'. It is assumed that
+		// the step implementation will check this status at chunk boundaries.
 		logger.info("Stopping job execution: " + jobExecution);
 
-		jobExecution.setStatus(BatchStatus.STOPPED);
+		jobExecution.setStatus(BatchStatus.STOPPING);
 		jobRepository.update(jobExecution);
```
onobc, line 251: > [!NOTE]
> This comment is around this general area (the stop method) and not necessarily this line....

> So when using Batch 4.x it was just a quick check of the JobRegistry to retrieve the job, which was always empty since SCDF never deals with Jobs directly. But with Batch 5.x they loaded the Job Registry and attempted to retrieve the Job in a different way using the SimpleJobOperator. 

It looks like `JsrJobOperator` [was using ](https://github.com/spring-projects/spring-batch/blob/fb0ce98a890eae1908bdf25d69eb4ea51b89973e/spring-batch-core/src/main/java/org/springframework/batch/core/jsr/launch/JsrJobOperator.java#L744)
```java
Job job = jobRegistry.getJob(jobExecution.getId());
```
and `SimpleJobOperator` [is using](https://github.com/spring-projects/spring-batch/blob/ffe3687c94412c3a9c156c1cc85e78bf974349b1/spring-batch-core/src/main/java/org/springframework/batch/core/launch/support/SimpleJobOperator.java#L340)
```java
Job job = jobRegistry.getJob(jobExecution.getJobInstance().getJobName());
```
so I'm guessing it is the job instance coming back null off the execution which is now causing the NPE. Correct?

onobc (changes requested):
Thanks for the update @cppwfs . Makes sense. I have one suggestion/comment to address but other than that 👍🏻

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
```diff
@@ -242,10 +242,12 @@ public JobExecution stop(Long jobExecutionId) throws NoSuchJobExecutionException
...
 			throw new JobExecutionNotRunningException("JobExecution is not running and therefore cannot be stopped");
 		}
-
+		// Indicate the execution should be stopped by setting it's status to
+		// 'STOPPING'. It is assumed that
+		// the step implementation will check this status at chunk boundaries.
 		logger.info("Stopping job execution: " + jobExecution);
 
-		jobExecution.setStatus(BatchStatus.STOPPED);
+		jobExecution.setStatus(BatchStatus.STOPPING);
```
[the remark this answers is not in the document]
cppwfs: I agree.   In that focussing the code into a single method instead of having it duplicated is always a good thing.

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
```diff
@@ -225,9 +225,7 @@ public int stopAll() {
 		Collection<JobExecution> result = jobExecutionDao.getRunningJobExecutions();
 		for (JobExecution jobExecution : result) {
 			try {
-				jobExecution.getStepExecutions().forEach(StepExecution::setTerminateOnly);
-				jobExecution.setStatus( BatchStatus.STOPPING);
-				jobRepository.update(jobExecution);
+				stop(jobExecution.getId());
```
onobc, line 228: Because we already have loaded the execution from the DB it would be an improvement to create a private `stop(JobExecution)` and then have both this and the one in the singular `stop(Long)` call into it.

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/batch/AbstractSimpleJobServiceTests.java
```diff
@@ -192,6 +194,18 @@ void exceptionsShouldBeThrownIfRequestForNonExistingJobInstance() {
...
+	void stoppingAllJobExecutionsShouldLeaveJobExecutionsWithStatusOfStopping() throws Exception {
+		JobExecution jobExecutionOne = createJobExecution(BASE_JOB_INST_NAME,true);
+		JobExecution jobExecutionTwo = createJobExecution(BASE_JOB_INST_NAME+"_TWO",true);
+		assertJobHasStopped(jobExecutionOne);
+		assertJobHasStopped(jobExecutionTwo);
+	}
+
+	private void assertJobHasStopped(JobExecution jobExecution) throws NoSuchJobExecutionException, JobExecutionNotRunningException {
 		jobExecution = jobService.getJobExecution(jobExecution.getId());
 		assertThat(jobExecution.isRunning()).isTrue();
```
onobc, line 210: Should this be `isFalse()` ?

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/batch/AbstractSimpleJobServiceTests.java
```diff
@@ -192,12 +194,24 @@ void exceptionsShouldBeThrownIfRequestForNonExistingJobInstance() {
...
 		assertThat(jobExecution.getStatus()).isNotEqualTo(BatchStatus.STOPPING);
 		jobService.stop(jobExecution.getId());
 		jobExecution = jobService.getJobExecution(jobExecution.getId());
-		assertThat(jobExecution.getStatus()).isEqualTo(BatchStatus.STOPPED);
+		assertThat(jobExecution.getStatus()).isEqualTo(BatchStatus.STOPPING);
```
corneil, line 214: should we wait for STOPPED as well?

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/batch/AbstractSimpleJobServiceTests.java
cppwfs: In this case it should be running because the JobExecution is created in a running state.   This test verifies the pre-condition that the test running before we call the stop on line 212.

onobc: Ahh yeh, I did not jump into `assertJobHasStopped` as the name implies that it is only asserting that the job has stopped. 

I think we can simplify this a bit by moving the post-creation verify steps
```java
jobExecution = jobService.getJobExecution(jobExecution.getId());
assertThat(jobExecution.isRunning()).isTrue();
assertThat(jobExecution.getStatus()).isNotEqualTo(BatchStatus.STOPPING);
```
into the `createJobExecution` method.

Then we can move the action (`stop`) into the test followed by a simplified assert method. That way it would read more like `AAA` (arrange / act / assert).

Something like this:
```java
@Test
void stoppingJobExecutionShouldLeaveJobExecutionWithStatusOfStopping() throws Exception {
	JobExecution jobExecution = createJobExecution(BASE_JOB_INST_NAME,true);
	jobService.stop(jobExecution.getId());
	assertJobHasStopped(jobExecution);
}

private void assertJobHasStopped(JobExecution jobExecution) throws NoSuchJobExecutionException, JobExecutionNotRunningException {
	jobExecution = jobService.getJobExecution(jobExecution.getId());
	assertThat(jobExecution.getStatus()).isEqualTo(BatchStatus.STOPPING);
	assertThat(jobExecution.isRunning()).isTrue();
}
```

Wdyt?

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/batch/AbstractSimpleJobServiceTests.java
cppwfs: `STOPPED` only occurs once the batch job has completed.  In this test we are just testing that SCDF sets the state to `STOPPING`.   We are not running real jobs here, just creating / modifying the state of the `JobExecution`.

onobc (approved):
Thanks for the updates @cppwfs . Other than the commented out skipper in pom.xml - LGTM!

pom.xml
```diff
@@ -70,7 +70,7 @@
 		<module>spring-cloud-dataflow-shell</module>
 		<module>spring-cloud-dataflow-shell-core</module>
 		<module>spring-cloud-dataflow-completion</module>
-		<module>spring-cloud-skipper</module>
+<!--		<module>spring-cloud-skipper</module>-->
```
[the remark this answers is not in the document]
cppwfs: 1 word.....  "DOHHH"

cppwfs:
Thank you @onobc  for the reviews that refined this PR.    
Thanks @corneil for the great question!
Polished, Squashed, Merged.

PR #5999, opened 2024-10-18
Re-enable for CommandTest

cppwfs:
This test still fails sporadically


onobc (approved):
Thanks for re-disabling it if its still sporadically failing.

PR #6007, opened 2024-10-23
[CI] Remove Maven profile full in favour of using docs for all document generation.

corneil:
In `spring-cloud-skipper-server-core` I've changed to the surefire in `docs` profile to use the `package` phase which means `test` phase executes normal unit tests and `package` phase generates the snippets.
When `-DskipTests` is present the normal test phase is skipped. When `-Pdocs` is not provided then the snippets are not generated.


.github/workflows/ci-pr.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       shell: bash
       timeout-minutes: 75
       run: |
-        ./mvnw -B -s .github/settings.xml -Pdocs clean install --no-transfer-progress
+        ./mvnw -B -s .github/settings.xml -Pfull,docs clean install --no-transfer-progress
```
onobc, line 20: What is `full` vs. `docs`? I guess I don't understand how `full` fulfills the PR description of `executes rest-docs generation as well`.

corneil: I didn't choose the name. We can change to `rest-docs` to ensure it is unambiguous.

corneil: Having checked:
* `docs` means javadocs
* `full` means all rest-docs and asciidocs.

How about we rather use `javadocs` and `fulldocs`?

cppwfs: Makes sense to me .   @onobc what do you think?

cppwfs (changes requested):
@corneil, Love the renames.  
But, when we make these changes we will have to update the following files:

- ci.yml
- build-snapshot-worker.yml
- build-docs.yml
- spring-cloud-skipper/README.adoc
- appendix-building.adoc

spring-cloud-skipper/spring-cloud-skipper-server-core/pom.xml
```diff
@@ -279,28 +283,6 @@
...
-						<artifactId>maven-surefire-plugin</artifactId>
-						<configuration>
-							<argLine>--add-opens java.base/java.util=ALL-UNNAMED</argLine>
-							<threadCount>1</threadCount>
-							<forkCount>1</forkCount>
-							<failIfNoTests>true</failIfNoTests>
-							<skipTests>false</skipTests>
-							<includes>
-								<include>**/*Documentation.java</include>
-								<include>**/*Tests.java</include>
```
cppwfs, line 297: We probably want to include the `**/*Tests.java` to the docs plugin below.   WDYT?

cppwfs (changes requested):
Looks Great!   One small question below.

spring-cloud-skipper/spring-cloud-skipper-server-core/pom.xml
corneil: I've changed to the surefire in `docs` profile to use the `package` phase which means `test` phase executes normal unit tests and `package` phase generates the snippets. 
When `-DskipTests` is present the normal test phase is skipped. When -Pdocs is not provides then the snippets are not generated.

corneil:
The latests commit fixes tests for the actual resources present that do work when doing local deployments.

corneil:
We have a pass locally fails in CI-PR situation. 😢

cppwfs (approved):
LGTM.  @corneil Thanks for all your hard work on this!

cppwfs:
Rebased, Squashed, Merged.

@corneil Thanks for the hard work on this.

PR #6009, opened 2024-10-23
Deprecate search field for taskName search in TaskDefinitionController

### LINKED ISSUE #5993: TaskDefinitionController Deprecation Cleanup
TaskDefinitionController has the `search`  parameter marked as depricated.  Verify that it is not used before removing.

cppwfs:
resolves #5993

* Remove search as a parameter from the javadoc
* Fix docs such that queryParameter descriptions appear in reference docs.
* Rename Request Parameters header  to Query Parameters

### DIFF
--- a/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/TaskControllerTests.java
+++ b/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/TaskControllerTests.java
@@ -334,28 +334,28 @@ class TaskControllerTests {
 		assertThat(myTask2.getName()).isEqualTo("myTask-t2");
 	}
 
-	@ParameterizedTest
-	@ValueSource(strings = {"search", "taskName"})
-	void findTaskNameContainsSubstring(String taskNameRequestParamName) throws Exception {
+	@Test
+	void findTaskNameContainsSubstring() throws Exception {
+		final String TASK_NAME_REQUEST_PARAMETER = "taskName";
 		repository.save(new TaskDefinition("foo", "task"));
 		repository.save(new TaskDefinition("foz", "task"));
 		repository.save(new TaskDefinition("ooz", "task"));
 
-		mockMvc.perform(get("/tasks/definitions").param(taskNameRequestParamName, "f")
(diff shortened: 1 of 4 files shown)


spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/TaskControllerTests.java
```diff
@@ -334,28 +334,28 @@ void saveCompositeTaskWithParameters() throws Exception {
 		assertThat(myTask2.getName()).isEqualTo("myTask-t2");
 	}
 
-	@ParameterizedTest
-	@ValueSource(strings = {"search", "taskName"})
-	void findTaskNameContainsSubstring(String taskNameRequestParamName) throws Exception {
+	@Test
+	void findTaskNameContainsSubstring() throws Exception {
+		final String TASK_NAME_REQUEST_PARAMETER = "taskName";
```
onobc, line 339: **[TinyNit]** Please remove the `final` keyword as we only use them where it adds value (like immutable fields etc..)

spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/TaskControllerTests.java
```diff
@@ -411,7 +411,7 @@ void findDslTextContainsSubstring() throws Exception {
 
 	@Test
 	void findByDslTextAndNameBadRequest() throws Exception {
-		mockMvc.perform(get("/tasks/definitions").param("dslText", "fo").param("search", "f")
+		mockMvc.perform(get("/tasks/definitions").param("dslText", "fo").param("taskName", "f")
```
onobc, line 414: In this case we reference the param by string name rather than constant. I would be in favor of moving the `TASK_NAME_REQUEST_PARAMETER` to a static final constant or just using `"taskName"` everywhere.

onobc (approved):
Thanks for another deprecation removal @cppwfs 🥳 
My comments are all of the **nit** variety.

cppwfs:
@onobc Thank you for the review!!!

Code review comments have been applied.    Rebased, Squashed, Merged.

PR #6010, opened 2024-10-23
Add what's new document for SCDF

cppwfs:
This document discusses the changes that affect our users in one location.

For now it is in the project root directory, but if you know of a better place please leave a comment and I'll move it to that directory


onobc (approved):
Great start @cppwfs , thanks for adding this. The location is fine for now and we will move it where it makes sense when that time comes.

cppwfs:
Thanks @onobc for the idea and the review!

Rebased Merged.

PR #6029, opened 2024-10-28
Initial security migration

cppwfs:
Separate dataflow and skipper configs instead of trying to "reuse" common configs. Skipper config did extend classes from comman but then override everything.

Remove unused CF configs related to tweaking things for "space developer" as this was never used.

Essentially minimal changes to get new security
configuration model working. There's still some
deprecations left to handle but it's better to
get better integration tests before those
are handled.


onobc (approved):
Co-reviewed w/ @cppwfs - ready for merge.

PR #6035, opened 2024-10-29
Remove DisposableBean implements from SimpleJobService

### LINKED ISSUE #6034: Remove @DisposableBean implementation from SimpleJobSevice
Currently SimpleJobService utilizes a `Destoy` method to stop all active jobs.   However the list of `activeJobs` it utilizes is always empty.   It has a `Scheduled` annotated method that once an hour executes method to remove jobs that are not running.   This is dead code, since dataflow does not control jobs.

cppwfs:
This can be removed because SCDF does not manage internal jobs, but rather launches boot apps that use Spring Batch

resolves #6034

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
@@ -15,13 +15,8 @@
  */
 package org.springframework.cloud.dataflow.server.batch;
 
-import java.time.LocalDateTime;
-import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.Date;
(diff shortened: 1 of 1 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/SimpleJobService.java
```diff
@@ -339,62 +325,4 @@ private void checkJobExists(String jobName) throws NoSuchJobException {
 			throw new NoSuchJobException("No Job with that name either current or historic: [" + jobName + "]");
 		}
 	}
-
```
onobc, line 342: Yeh this code was ancient. I guess it was also not tested otherwise this should have made some test fail.

onobc (approved):
Thanks for the cleanup on this one @cppwfs . Nice catch!

corneil (approved):
:100:

PR #6037, opened 2024-10-30
Added a negative test to DataflowOAuthIT

corneil:
Added a negative test to DataflowOAuthIT to ensure /about cannot be accessed without credentials.


cppwfs (approved):
LGTM.
Thanks for the contribution.

But curious as to what facilitated this change.   Was it an issue or something you noticed missing?

corneil:
> LGTM. Thanks for the contribution.
> 
> But curious as to what facilitated this change. Was it an issue or something you noticed missing?

I had seen cases where a security implementation seems to work until you actively probe it and discover it doesn't block the request

onobc (approved):
LGTM - nice improvement. Thanks @corneil

PR #6038, opened 2024-10-30
Use Batch's StringToLocalDateTimeConverter instead of SCDF's Deprecated converters

cppwfs:


onobc (changes requested):
Thanks for the cleanup @cppwfs (❤️ deletions). Do you mind adding these removals to the WhatsNew doc as part of this PR?

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/batch/JdbcSearchableJobExecutionDao.java
```diff
@@ -148,7 +148,7 @@ public class JdbcSearchableJobExecutionDao extends JdbcJobExecutionDao implement
 
 	public JdbcSearchableJobExecutionDao() {
 		conversionService = new DefaultConversionService();
-		conversionService.addConverter(new StringToDateConverter());
+		conversionService.addConverter(new StringToLocalDateTimeConverter());
```
onobc, line 151: I verified the Batch converter is functionally equivalent to the removed SCDF converter.

cppwfs:
Thank you @onobc for the review.   Thanks for catching my omission to the whats-new.adoc.

Review comments added to PR, Squashed, Merged

PR #6039, opened 2024-10-30
Remove deprecations from AppRegistryController

### LINKED ISSUE #6011: AppRegistryController Deprecation cleanup
AppRegistryController currently there are 2 `Deprecation`s that need to be removed.  
* [register](https://github.com/spring-cloud/spring-cloud-dataflow/commit/44fddefa172d63e16bd5f089ec41227aea9ac8b0) 
* [info](https://github.com/spring-cloud/spring-cloud-dataflow/commit/f80369fbab34b4cf02db9615bc1153fb247e23c9)
These deprecations were added in 2017 when we added versions for registration.

Conversely we need to add 2 more `Deprecation`s for the current API's as they have an optional `bootVersion` that is no longer supported.

AppRegistryOperations:
* register - Move the `Deprecated` annotations to the methods that have the bootversion enumerated.   But also investigate to see if the method at line [87](https://github.com/spring-cloud/spring-cloud-dataflow/blame/main/spring-cloud-dataflow-rest-client/src/main/java/org/springframework/cloud/dataflow/rest/client/AppRegistryOperations.java#L87C26-L87C34) needs to be removed.

cppwfs:
The original issue states to remove deprecated functions. However, this would remove some of SCDF's default behavior when registering applications. In that it can look the suffix of the URL to obtain the version. The original Issues for the PR that put in the deprecations just wanted to give the user the ability to override the version.

Also deprecated the use of bootVersion when registering an app.

Resolves #6011

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppRegistryController.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppRegistryController.java
@@ -223,7 +223,7 @@ public class AppRegistryController {
 	 * @param type module type
 	 * @param name module name
 	 * @param version module version
-	 * @param bootVersion module boot version or {@code null} to use the default.  Deprecated: bootVersion is no longer needed when registering an app.
+	 * @param bootVersion module boot version or {@code null} to use the default.  Deprecated: bootVersion parameter is ignored.
 	 * @param uri URI for the module artifact (e.g. {@literal maven://group:artifact:version})
 	 * @param metadataUri URI for the metadata artifact
 	 * @param force if {@code true}, overwrites a pre-existing registration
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/controller/AppRegistryController.java
```diff
@@ -224,7 +223,7 @@ else if (entry.getKey().equals("outbound")) {
 	 * @param type module type
 	 * @param name module name
 	 * @param version module version
-	 * @param bootVersion module boot version or {@code null} to use the default
+	 * @param bootVersion module boot version or {@code null} to use the default.  Deprecated: bootVersion is no longer needed when registering an app.
```
onobc, line 226: I have mixed feelings about leaving this in since there is no Boot2/3 in SCDF 3.x. However, since it is a controller it is the upgrade empathy thing to do to leave it in and mark `@Deprecated`. Do you mind adusting the desc to say that the parameter is ignored completely?

onobc (changes requested):
Thanks for the PR @cppwfs . One small suggestion and then also do you mind adding this to the What's New doc as part of this PR? I think we should add these as we go (in the deprecation PRs)

onobc (approved):
Thx for the updates @cppwfs

cppwfs:
Updated per code review request.
Rebased, squashed, merged.

Thanks @onobc for the review!!!!!!!

PR #6040, opened 2024-10-30
Remove Deprecated method from TaskDefinitionRepository

### LINKED ISSUE #6018: TaskDefinitionRepository Deprecation Cleanup
Method findByTaskNameContainsAndDslTextContains can be removed.

cppwfs:
Add deprecation removals from whats-new.doc

Resolves #6018

### DIFF
--- a/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/TaskDefinitionRepository.java
+++ b/spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/TaskDefinitionRepository.java
@@ -36,9 +36,6 @@ public interface TaskDefinitionRepository extends KeyValueRepository<TaskDefinit
 
 	Page<TaskDefinition> findByDslTextContains(String dslText, Pageable pageable);
 
-	@Deprecated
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/TaskDefinitionRepository.java
```diff
@@ -36,9 +36,6 @@ public interface TaskDefinitionRepository extends KeyValueRepository<TaskDefinit
 
 	Page<TaskDefinition> findByDslTextContains(String dslText, Pageable pageable);
 
-	@Deprecated
```
onobc, line 39: Sad that this removal did not cause any test failures 😢

onobc (approved):
Thanks for the cleanup @cppwfs - 🍏

spring-cloud-dataflow-server-core/src/main/java/org/springframework/cloud/dataflow/server/repository/TaskDefinitionRepository.java
cppwfs: So True 😢 .  But it was dead code.

PR #6043, opened 2024-11-01
Remove schedule. deprecated  property key prefix

### LINKED ISSUE #6019: DefaultSchedulerService Deprecation Cleanup
`extractAndQualifySchedulerProperties` is no longer needed as we should no longer support `scheduler.` prefixed properties.

cppwfs:
Use deployer. property prefix when setting schedule properties

Originally scheduled properties uses `schedule.` as a prefix for all the settings. However it was determined they were just a subset of deployer properties and thus the prefix could be deprecated.

resolves #6019


corneil:
Are the users going to receive an error when they use `schedule.`?

cppwfs:
Great Question.  In both implementations they check for a empty/null cron expression and will throw an exception.

corneil:
> Great Question. In both implementations they check for a empty/null cron expression and will throw an exception.

Will users experience failure if they don't recreate their schedules with modified properties?

corneil (approved):
:100:

PR #6054, opened 2024-11-07
Remove deprecation annotation from StepExecutionHistory.java

### LINKED ISSUE #6015: StepExecutionHistory Deprecation Cleanup
StepExecutionHistory deprecated fields need to be removed and those classes that report this information.  This will include  changes to  shell and UI).   
This line `NOTE: The following fields in the stepExecutionHistory are deprecated and will be removed in a future release: rollbackCount, readCount, writeCount, filterCount, readSkipCount, writeSkipCount, processSkipCount, durationPerRead.
` from the api-guide.adoc will need to be removed.

More can be read about it [here](https://github.com/spring-cloud/spring-cloud-dataflow/commit/32799bd5e3893c1baa20aeecfa26ca718b0e1b37).

cppwfs:
Update api-guide to remove deprecation notice

Also polish documents to remove errors.

The original intent was to remove these calls from the Rest request as it was not actively used.
But after discussions, we decided to retain this feature.

Resolves #6015

### DIFF
--- a/spring-cloud-dataflow-docs/src/main/asciidoc/api-guide.adoc
+++ b/spring-cloud-dataflow-docs/src/main/asciidoc/api-guide.adoc
@@ -3072,9 +3072,6 @@ include::{snippets}/job-step-executions-documentation/step-progress/curl-request
 
 include::{snippets}/job-step-executions-documentation/step-progress/http-response.adoc[]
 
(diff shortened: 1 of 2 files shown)


onobc (approved):
Changes look straightforward, we are reversing our decision on actually deprecating these fields to begin with. It may be helpful to add the reason why we are removing the deprecation.

Other than that - LGTM. Thanks @cppwfs

cppwfs:
Updated the commit message based on @onobc request.  Thanks @onobc

PR #6064, opened 2024-11-12
Replace deprecated asList method with asInstanceOf.

cppwfs:


onobc (approved):
LGTM - thanks @cppwfs

Issue #6069, opened 2024-11-13
Update to Prometheus Proxy M2


onobc:
**Problem description:**
Is your feature request related to a problem? Please provide a clear and concise description of what the problem is. 

**Solution description:**
Describe the solution you would like. We need a clear and concise description of what you want to happen. Also, have you considered cases outside the expected flow (edge cases)? What should happen in those cases?

**Description of alternatives:**
Describe alternatives you have considered: We need a clear and concise description of any alternative solutions or features you have considered.

**Additional context:**
Add any other context or explanation about the feature request here.

corneil:
2.0.0-M3 is the latest.
Do we know if 2.0.x is supposed to work with old clients using 1.5.x?

onobc:
> 2.0.0-M3 is the latest. Do we know if 2.0.x is supposed to work with old clients using 1.5.x?

Great question @corneil . No. Recommendation is to update to the latest client (we are doing this in #6068 )

onobc:
Closing via https://github.com/spring-cloud/spring-cloud-dataflow/pull/6090/commits/74a2ce57e5aebeab873d3290180a28392b7f6506

PR #6070, opened 2024-11-13
Update to Prometheus RSocket Proxy 2.0.0-M2

onobc:
This commit updates the Prometheus Rsocket Proxy dependency and Docker images used in the installers to 2.0.0-M2.

See #6069

### DIFF
--- a/spring-cloud-dataflow-build/spring-cloud-dataflow-build-dependencies/pom.xml
+++ b/spring-cloud-dataflow-build/spring-cloud-dataflow-build-dependencies/pom.xml
@@ -32,7 +32,7 @@
 		<commons-compress.version>1.26.2</commons-compress.version>
 		<!-- Specific version overrides to deal w/ CVEs -->
 		<nimbus-jose-jwt.version>9.39.3</nimbus-jose-jwt.version>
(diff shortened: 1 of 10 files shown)


cppwfs:
Looks great!
Only one nitpick:
Do we want to update the `configuration-options.adoc` `configuration-options-scdf.feature.monitoring.prometheusRsocketProxy.image.tag` to 2.0.0-M2 from 1.0.0

onobc:
> Looks great! Only one nitpick: Do we want to update the configuration-options.adoc configuration-options-scdf.feature.monitoring.prometheusRsocketProxy.image.tag to 2.0.0-M2 from 1.0.0

Thanks @cppwfs !
I would not call that a nitpick but rather a great find since I missed it. Updated.

That being said, is this doc manually managed (i.e. not auto-generated)? 
1. I did not even know it existed
2. I wonder what else is out of date?

cppwfs (approved):
LGTM
Merge when you have finished the tests.
Thanks for working on this!

This is not a generated file according to the history.   
But looks like we need to add it our release document for update or automate the generation of it.

PR #6072, opened 2024-11-14
Initialize useUserAccessToken to false.

cppwfs:
With the effort to remove deprecated methods from TaskConfigurationProperties with this [commit](https://github.com/spring-cloud/spring-cloud-dataflow/commit/b27f0c22094c20ad63fe7f4c7e0bc6d41ad7c79d), we introduced a NPE when DefaultTaskExecutionService.handleAccessToken tried to use useUserAccessToken. The original code would use TaskServiceUtils to return a false if it was null.

This behavior was a shim to maintain backwards compatibility which is no longer needed.
So the solution is to set the default for this field to false.

### DIFF
--- a/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/service/impl/ComposedTaskRunnerConfigurationPropertiesTests.java
+++ b/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/service/impl/ComposedTaskRunnerConfigurationPropertiesTests.java
@@ -18,6 +18,7 @@ package org.springframework.cloud.dataflow.server.service.impl;
 
 import org.junit.jupiter.api.Test;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
 
 public class ComposedTaskRunnerConfigurationPropertiesTests {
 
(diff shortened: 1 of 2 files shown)


spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/service/impl/ComposedTaskRunnerConfigurationPropertiesTests.java
```diff
@@ -43,7 +44,16 @@ void useUserAccessTokenFromCTRPropFalse() {
 	void useUserAccessTokenFromCTRPropNotSet() {
 		ComposedTaskRunnerConfigurationProperties composedTaskRunnerConfigurationProperties =
 			new ComposedTaskRunnerConfigurationProperties();
-		assertThat(composedTaskRunnerConfigurationProperties.isUseUserAccessToken()).as("Use user access token should be false").isNull();
+		assertThat(composedTaskRunnerConfigurationProperties.isUseUserAccessToken()).as("Use user access token should be false").isFalse();
```
onobc, line 47: I have found the `as` and also Junit `describe` end up being another maintenance point that comes out of date very quickly (i.e. they are more hassle than they are worth IMO). It was here already, just saying this out loud though.

onobc (approved):
Nice work @cppwfs . I have 2 tiny nits that you can leave (or change - no biggie either way).

cppwfs:
Implemented the requested changes.

Thanks @onobc for the review!!!

PR #6074, opened 2024-11-14
Disable Trivy by default

corneil:
Adds an input flag `enableSecurityScan` that is false by default and will only include the scan job if set to true.


onobc (approved):
Thanks for removing this @corneil

PR #6075, opened 2024-11-15
Disable test for backwards compatibilty in DB IT

corneil:
Adds `@Disabled` to migration testing backwards compatibility
The backwards compatibility between 2.10 and 2.11 can be tested easily, but between 2.11 and 3.0 will not work because the dataflow client of 3.0 cannot interact with 2.11
So we can only start doing these types of tests when we have 3.0 and 3.1

### DIFF
--- a/spring-cloud-dataflow-server/src/test/java/org/springframework/cloud/dataflow/integration/test/db/AbstractPostgresDatabaseTests.java
+++ b/spring-cloud-dataflow-server/src/test/java/org/springframework/cloud/dataflow/integration/test/db/AbstractPostgresDatabaseTests.java
@@ -20,6 +20,7 @@ import java.net.URI;
 import java.net.URISyntaxException;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import org.junit.jupiter.api.Disabled;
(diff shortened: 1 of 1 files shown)


spring-cloud-dataflow-server/src/test/java/org/springframework/cloud/dataflow/integration/test/db/AbstractPostgresDatabaseTests.java
```diff
@@ -79,15 +80,16 @@ public void migrationError() {
 	@SuppressWarnings("deprecation")
 	@Test
 	@DataflowMain
-	public void migration210211() throws URISyntaxException {
+	@Disabled("Reactivate when 3.1 is released")
```
onobc, line 83: Do you mind explaining in the PR description why we are disabling it?

onobc: Thanks for updating the PR description.

onobc (approved):

PR #6076, opened 2024-11-15
Disable security scan in ci-pr.yml

corneil:
Added env to control the scan.


.github/workflows/ci-pr.yml
```diff
@@ -3,6 +3,9 @@ name: CI PRs
 on:
   pull_request:
 
+env:
```
onobc, line 6: I don't think there is a way to set an ENV on a PR is there?

PR #6077, opened 2024-11-15
Disable security scan in ci-pr.yml in 2.11.x

corneil:
Added env to control the scan.

### DIFF
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -4,7 +4,9 @@ on:
   pull_request:
 
 env:
-  ENABLE_SECURITY_SCAN: 'false'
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-pr.yml
```diff
@@ -3,6 +3,9 @@ name: CI PRs
 on:
   pull_request:
 
+env:
+  ENABLE_SECURITY_SCAN: 'false'
```
onobc, line 7: I don't think there is a way to set an ENV on a PR is there?

corneil: I think we can use a repo variable to set the default value for Trivy Scans.

corneil:
Changed to different defaults.

PR #6081, opened 2024-11-18
⚠️ Do not merge: Resolve Postgres IT test failure

cppwfs:
We are currently testing 2.10 to 2.11 migrations in the 3.x project

This is causing a failure in the IT test: migration210211

The resolution is to test it the 2.11 -> 3.0 migration.


corneil:
I believe we need to use plain Rest to check the first leg when using 2.11.5 because it is not going to satisfy the 3.x DataFlowClient.

cppwfs:

Issue #6068, opened 2024-11-13
Update to Prometheus 1.x client


onobc:
We are currently using the 0.x Prometheus client (simple client). We want to move up to the 1.x client as Boot will remove the support for simple client auto-configuration in SB 3.5.x

> [!NOTE]
> The update to 1.x client changes output of metrics and is likely to break our dashboard.
> Pay attention to: https://github.com/micrometer-metrics/micrometer/wiki/1.13-Migration-Guide

PR #6090, opened 2024-12-13
Use new Prometheus RSocket client

onobc:
Replaces the `micrometer.io:micrometer-registry-prometheus-simpleclient` with `micrometer.io:micrometer-registry-prometheus`.

Also updates the Prometheus RSocket version from `2.0.0-M3` to `2.0.0-M4`.

See #6068
See #6069

### DIFF
--- a/spring-cloud-dataflow-build/spring-cloud-dataflow-build-dependencies/pom.xml
+++ b/spring-cloud-dataflow-build/spring-cloud-dataflow-build-dependencies/pom.xml
@@ -32,7 +32,7 @@
 		<commons-compress.version>1.26.2</commons-compress.version>
 		<!-- Specific version overrides to deal w/ CVEs -->
 		<nimbus-jose-jwt.version>9.39.3</nimbus-jose-jwt.version>
-		<prometheus-rsocket.version>2.0.0-M2</prometheus-rsocket.version>
+		<prometheus-rsocket.version>2.0.0-M4</prometheus-rsocket.version>
 		<java-cfenv.version>2.3.0</java-cfenv.version>
 		<spring-cloud-services-starter-config-client.version>3.5.4</spring-cloud-services-starter-config-client.version>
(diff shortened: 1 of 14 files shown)


spring-cloud-dataflow-build/spring-cloud-dataflow-build-dependencies/pom.xml
```diff
@@ -32,7 +32,7 @@
 		<commons-compress.version>1.26.2</commons-compress.version>
 		<!-- Specific version overrides to deal w/ CVEs -->
 		<nimbus-jose-jwt.version>9.39.3</nimbus-jose-jwt.version>
-		<prometheus-rsocket.version>2.0.0-M2</prometheus-rsocket.version>
+		<prometheus-rsocket.version>2.0.0-SNAPSHOT</prometheus-rsocket.version>
```
onobc, line 35: > [!NOTE]
> As stated in the description... this will move to `2.0.0-M4` next then finally to `2.0.0`.

src/carvel/config/values/values.yml
```diff
@@ -112,5 +112,5 @@ scdf:
         enabled: false
         image:
           repository: micrometermetrics/prometheus-rsocket-proxy
-          tag: 2.0.0-M2
```
onobc, line 115: It would be nice to find a way to templatize this variable so that we set it in one place (even in a few places would be better than in every yml file in the installers).

cppwfs:
In the description it says we are migrating `M3` to `M4`,  but in the files changed we went from `M2` to `M4`.   Probably want to reflect that in the description.

cppwfs (approved):
LGTM 

One little nit: In the description it says we are migrating M3 to M4, but in the files changed we went from M2 to M4. Probably want to reflect that in the description.

corneil (approved):
:100:

PR #6098, opened 2024-12-31
Added metaDataUri to AppRegistrationResource

### LINKED ISSUE #6092: AppRegistryController does not return metadataUri
When user issues a get on the `apps` endpoint the list contains the `URI` but not the `metadataURI`.   SCDF-UI has a column to display the `metaDataURI` but it is currently empty.   This is the same for the `detail page` which uses the `/apps/<TYPE>/<APP-NAME>` endpoint.

corneil:
Added metaDataUri to AppRegistrationResource
Added test for presence of metaDataUri

Fixes #6092

### DIFF
--- a/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/AppRegistryControllerTests.java
+++ b/spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/AppRegistryControllerTests.java
@@ -129,12 +129,15 @@ class AppRegistryControllerTests {
 		mockMvc.perform(
 				post("/apps/sink/log1/3.0.0")
 						.queryParam("bootVersion", "3")
-						.param("uri", "maven://org.springframework.cloud.stream.app:log-sink-rabbit:3.0.0").accept(MediaType.APPLICATION_JSON)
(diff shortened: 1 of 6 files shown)


spring-cloud-dataflow-server-core/src/test/java/org/springframework/cloud/dataflow/server/controller/AppRegistryControllerTests.java
```diff
@@ -129,12 +129,15 @@ void findRegisteredApp() throws Exception {
 		mockMvc.perform(
 				post("/apps/sink/log1/3.0.0")
 						.queryParam("bootVersion", "3")
-						.param("uri", "maven://org.springframework.cloud.stream.app:log-sink-rabbit:3.0.0").accept(MediaType.APPLICATION_JSON)
+						.param("uri", "maven://org.springframework.cloud.stream.app:log-sink-rabbit:3.0.0")
```
cppwfs, line 132: We still need a test here for when there is no metadata sent on registration.  I'll add it on merge.

cppwfs (approved):
Thank you @corneil for the PR!!!!

cppwfs:
LGTM Thanks for the contribution @corneil !!!!

PR #6106, opened 2025-02-10
Remove the 'version: 3' to match the main docker compose file.

afsg77:
This avoids the warning:
> the attribute 'version' is obsolete, it will be ignored, please remove it to avoid potential confusion


onobc (approved):
Thanks for the contribution @afsg77

